Issue #154, opened 2014-10-08
Support SSL for v0.8.3


joestein:
Kafka is adding SSL support => https://issues.apache.org/jira/browse/KAFKA-1477 and would be great to have this available in go libraries.

uovobw:
any news on this? i am planning on exposing a kafka cluster to the net securing it with stunnel and client side certificates. if there already is some code i can contribute to that...

eapache:
Sarama's support is ready to go... we're waiting on upstream kafka to release this in the broker

uovobw:
i heard that is scheduled to happen in version 0.9, that might be a long way to come yet. 
is there any chance that, with external ssl-termination as i wrote above, this could be pushed earlier?

eapache:
#156 needs to be rebased and have any merge conflicts resolved, but it should be all you need.

uovobw:
awesome, i'll get around to it if i get the time. thanks

uovobw:
@eapache i have submitted a new pr https://github.com/Shopify/sarama/pull/493 that rebases this change to master (banch support-ssl). tests pass, let me know how i can help

eapache:
Implemented (for now) by #493

PR #156, opened 2014-10-10
Add configuration to support broker SSL

eapache:
Possibly implements #154, if my assumptions about the implementation are
correct.

@joestein @wvanbergen

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,6 +1,9 @@
 package sarama
 
-import "time"
+import (
+	"crypto/tls"
+	"time"
+)
 
 // Config is used to pass multiple configuration options to Sarama's constructors.
 type Config struct {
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -12,6 +15,11 @@ type Config struct {
 		DialTimeout  time.Duration // How long to wait for the initial connection to succeed before timing out and returning an error (default 30s).
 		ReadTimeout  time.Duration // How long to wait for a response before timing out and returning an error (default 30s).
 		WriteTimeout time.Duration // How long to wait for a transmit to succeed before timing out and returning an error (default 30s).
+
+		TLS struct {
+			Enable bool        // Whether or not to use TLS when connecting to the broker (defaults to false).
+			Config *tls.Config // The TLS configuration to use for secure connections if specified by UseTLS (defaults to nil).
```
wvanbergen: godoc: `UseTLS` is probably an old name, and should be `Enable`?

eapache: Good catch, this was left over from a rebase before unified config.

eapache:
@wvanbergen @uovobw Rebased this branch correctly and added a note about compat guarantees. If this looks good to everyone I'll merge it.

wvanbergen:
:tada: Does it use a default config if you leave the config set to `nil`?

I'd like to wait for a confirmation that this works as expected for @uovobw before merging.

eapache:
>  DialWithDialer interprets a nil configuration as equivalent to the zero configuration; see the documentation of Config for the defaults.

uovobw:
@wvanbergen thanks for the trust but i no longer have the infrastructure in place to test this completely (since broker-originating connections do not support ssl) and currently do not have time to setup a local broker and test against this. if this merge is not urgent i think i can on this in a couple of days. thanks for the feedback guys

wvanbergen:
@uovobw Given that you are the only person we know of that needs this feature, you can take your time :)

PR #172, opened 2014-11-04
Concurrent metadata refresh

crxpandion:
Changes broker metadata to refresh concurrently rather than sequentially. If all brokers error out the last call to `Broker.GetMetadata(...)` that returned an error is exposed and the client enters the resurrection retry loop.

This also make users of `sarama.Client` (e.g. consumer/producer) be solely responsible for disconnecting brokers who timeout or misbehave. 

Related to https://github.com/Shopify/sarama/issues/15

### DIFF
--- a/client.go
+++ b/client.go
@@ -311,16 +311,14 @@ func (client *Client) refreshMetadata(topics []string, retries int) error {
 		return err
 	default:
 		// means all brokers are timedout... At what level should we disconnect a broker?
-
-	}
-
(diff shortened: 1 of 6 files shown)


eapache:
The client does need to be worked on before this can be tackled properly, but once that is done this will probably have to be rewritten anyways, so there's not much point in leaving the PR open. It can/should be mined for ideas at some point.

PR #194, opened 2014-11-16
Address some golint suggestions

wvanbergen:
- Error strings should not end with a dot.
- No need for `else` when if clause has return statement.
- Exported constants should be documented.
- Some minor changes about variables.

The following glint suggestions have not been implemented.
- Document all methods, or do not export them.
- Error variables should start with Err.

@eapache

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -463,7 +463,7 @@ func (w *brokerConsumer) subscriptionConsumer() {
 			return
 		}
 
-		for child, _ := range w.subscriptions {
+		for child := range w.subscriptions {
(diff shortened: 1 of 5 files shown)


wvanbergen:
Revived this PR:

- Don't use unnecessary `_` variables.
- Fixed some error message punctuation.
- Renamed CoordinatorId to CoordinatorID. While this technically is a API breaking change, we don't even use this field internally so I doubt this will cause any problems.

The error variable names I will tackle in a separate PR, because that will be more breaking.

@Shopify/kafka

eapache:
LGTM

PR #222, opened 2014-12-08
Rewrite consumer to support multiple partitions

### LINKED ISSUE #140: MultiConsumer
@wvanbergen @snormore @mkobetic (and anyone else who cares)

The current consumer works just fine, but is limited to fetching just a single partition of a single topic at once. You can run multiple consumers, but this has several issues:
 - They can't use the same client, or else the broker connection has contention problems (brokers handle requests on a single connection serially, so if it has to wait for `MaxWaitTime` before responding, this introduces severe latency when multiplexing busy and non-busy topics). Using separate clients works, but generates a *lot* of extra connections and metadata traffic.
 - You don't get messages from multiple partitions batched into a single response, when that makes sense for efficiency.

The ideal solution is to have a consumer capable of fetching from multiple topic/partitions at once. This has been at the back of my mind for a while now, so I already have a design ready; if the design makes sense to y'all, then somebody just needs to write it :)

While the user currently specifies a single topic/partition to the constructor, they should now be able to specify a set of topic/partitions. Since some of the configuration that is currently a singleton actually needs to be per-partition (`OffsetMethod` and `OffsetValue` at least), this would probably be of type `map[string]map[int32]*ConsumerPartitionConfig`. I considered permitting the dynamic adding/removing of partitions to the set, but I don't see a strong use case and it complicated a bunch of things.

Events are returned to the user exactly the same way they are now, over a single channel. I considered a separate channel per topic/partition but it complicated the base case, and the events already contain information on topic/partition so it's not hard for the user to dispatch appropriately if they really want to.

The constructor starts up one "controller" goroutine, which starts up and manages one goroutine per broker-that-has-a-partition-we-care-about and is responsible for (initially) dispatching each topic/partition to the appropriate broker's goroutine. The broker goroutine looks a lot like the current `fetchMessages` method with a few tweaks:
 - Some minor work needed to handle multiple blocks in the requests and responses.
 - When a topic/partition is reassigned to a new broker, that topic/partition gets returned to the controller via a channel; the goroutine tracks how many it is "responsible" for and exits if that reaches 0.
 - Similarly when a broker goes down hard, all topic/partitions are returned to the controller for re-dispatching and the goroutine exits.

I expect the success case to be fairly straightforward - as always, the complexity will come when reasoning through the failure cases and ensuring that topic/partitions are redispatched correctly, messages are not accidentally skipped in that case, etc. etc.

When the consumer is closed, it signals the controller which cleans up its children before exiting.

Thoughts?

### LINKED ISSUE #200: Consumer tests take forever
This has been bugging me forever, filing a ticket so I don't forget. These tests take anywhere from 0.5-10 seconds apparently at random, when they are entirely local and with appropriate configuration should basically be instantaneous. Something is fishy here.

cc @wvanbergen

### LINKED ISSUE #225: Handling of compacted topics
It appears that Sarama currently has problems on compacted topics. I have the follow test case(error handling omitted for brevity):
```go
package main

import (
  "github.com/Shopify/sarama"
  "log"
)

func main() {
  cfg := sarama.NewClientConfig()
  cli, _:= sarama.NewClient("", []string{"localhost:9092"}, cfg)
  ccfg := sarama.NewConsumerConfig()
  ccfg.OffsetMethod = sarama.OffsetMethodOldest
  consumer, _ := sarama.NewConsumer(cli, "compacted_topic", 0, "", ccfg)
  for m := range consumer.Events() {
    log.Printf("%s", m)
  }
}
```
I run this against a Kafka 0.8.1 broker, with the compacted_topic being a compacted topic with options:
```
segment.bytes=100000000
cleanup.policy=compact
segment.ms=3600000
```
The topic has a single partition, which currently starts at offset 169314.

When I run the above code, I receive a stream of:
```
2014/12/15 01:23:18 &{  expert.alerts %!s(int32=0) %!s(int64=0) kafka: Response did not contain all the expected topic/partition blocks.}
```
Which I believe is the IncompleteResponse error. After some digging around, it appears that Kafka reports the initial offset(using GetOffset) as 8, but when the actual request is made the offset is 169314. Kafkacat(via librdkafka) handles this gracefully, but sarama simply keeps giving that same error. Looking at the code, I believe the error is produced at:
https://github.com/Shopify/sarama/blob/master/consumer.go#L277-L285

I'm not sure if this a Sarama oversight, or a Kafka bug, but I believe that compaction can introduce gaps in the offset indexes received, so it seems like an oversight in Sarama. My gut tells me that the code needs to check that the offset is monotonically increasing, but can't assume that there are no gaps.

Best regards,
Lasse Dalegaard

eapache:
- `Consumer` is now a much lighter "manager" that spawns `PartitionConsumer`s
  instead. `PartitionConsumer`s behave and have an API almost exactly like the
  old `Consumer`.
- Automatically groups and redispatches PartitionConsumers when leadership moves
  in the cluster.

As with the producer-ng, this is effectively a rewrite and (with the exception of the changes to functional_test.go) should be reviewed wholesale rather than by diff.

@mkobetic @snormore @wvanbergen @yagnik @drdee and anyone else who cares, this is ready to be looked at whenever is convenient. Thoughts on the API? Thoughts on the TODOs still sprinkled about?

Fixes #140 
Fixes #200 
Fixes #225

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -1,6 +1,8 @@
 package sarama
 
 import (
+	"fmt"
+	"sync"
 	"time"
 )
 
@@ -8,41 +10,95 @@ import (
 type OffsetMethod int
 
 const (
-	// OffsetMethodManual causes the consumer to interpret the OffsetValue in the ConsumerConfig as the
-	// offset at which to start, allowing the user to manually specify their desired starting offset.
-	OffsetMethodManual OffsetMethod = iota
 	// OffsetMethodNewest causes the consumer to start at the most recent available offset, as
 	// determined by querying the broker.
-	OffsetMethodNewest
+	OffsetMethodNewest OffsetMethod = iota
 	// OffsetMethodOldest causes the consumer to start at the oldest available offset, as
 	// determined by querying the broker.
 	OffsetMethodOldest
+	// OffsetMethodManual causes the consumer to interpret the OffsetValue in the ConsumerConfig as the
+	// offset at which to start, allowing the user to manually specify their desired starting offset.
+	OffsetMethodManual
 )
 
 // ConsumerConfig is used to pass multiple configuration options to NewConsumer.
 type ConsumerConfig struct {
-	// The default (maximum) amount of data to fetch from the broker in each request. The default is 32768 bytes.
-	DefaultFetchSize int32
 	// The minimum amount of data to fetch in a request - the broker will wait until at least this many bytes are available.
 	// The default is 1, as 0 causes the consumer to spin when no messages are available.
 	MinFetchSize int32
-	// The maximum permittable message size - messages larger than this will return MessageTooLarge. The default of 0 is
-	// treated as no limit.
-	MaxMessageSize int32
 	// The maximum amount of time the broker will wait for MinFetchSize bytes to become available before it
 	// returns fewer than that anyways. The default is 250ms, since 0 causes the consumer to spin when no events are available.
 	// 100-500ms is a reasonable range for most cases. Kafka only supports precision up to milliseconds; nanoseconds will be truncated.
 	MaxWaitTime time.Duration
+}
 
-	// The method used to determine at which offset to begin consuming messages.
+// NewConsumerConfig creates a ConsumerConfig instance with sane defaults.
+func NewConsumerConfig() *ConsumerConfig {
+	return &ConsumerConfig{
+		MinFetchSize: 1,
+		MaxWaitTime:  250 * time.Millisecond,
+	}
+}
+
+// Validate checks a ConsumerConfig instance. It will return a
+// ConfigurationError if the specified value doesn't make sense.
+func (config *ConsumerConfig) Validate() error {
+	if config.MinFetchSize <= 0 {
+		return ConfigurationError("Invalid MinFetchSize")
+	}
+
+	if config.MaxWaitTime < 1*time.Millisecond {
+		return ConfigurationError("Invalid MaxWaitTime, it needs to be at least 1ms")
+	} else if config.MaxWaitTime < 100*time.Millisecond {
+		Logger.Println("ConsumerConfig.MaxWaitTime is very low, which can cause high CPU and network usage. See documentation for details.")
+	} else if config.MaxWaitTime%time.Millisecond != 0 {
+		Logger.Println("ConsumerConfig.MaxWaitTime only supports millisecond precision; nanoseconds will be truncated.")
+	}
+
+	return nil
+}
+
+// PartitionConsumerConfig is used to pass multiple configuration options to AddPartition
+type PartitionConsumerConfig struct {
+	// The default (maximum) amount of data to fetch from the broker in each request. The default is 32768 bytes.
+	DefaultFetchSize int32
+	// The maximum permittable message size - messages larger than this will return MessageTooLarge. The default of 0 is
+	// treated as no limit.
+	MaxMessageSize int32
+	// The method used to determine at which offset to begin consuming messages. The default is to start at the most recent message.
 	OffsetMethod OffsetMethod
 	// Interpreted differently according to the value of OffsetMethod.
 	OffsetValue int64
-
 	// The number of events to buffer in the Events channel. Having this non-zero permits the
 	// consumer to continue fetching messages in the background while client code consumes events,
-	// greatly improving throughput. The default is 16.
+	// greatly improving throughput. The default is 64.
 	EventBufferSize int
+}
+
+// NewPartitionConsumerConfig creates a PartitionConsumerConfig with sane defaults.
+func NewPartitionConsumerConfig() *PartitionConsumerConfig {
+	return &PartitionConsumerConfig{
+		DefaultFetchSize: 32768,
+		EventBufferSize:  64,
+	}
+}
+
+// Validate checks a PartitionConsumerConfig instance. It will return a
+// ConfigurationError if the specified value doesn't make sense.
+func (config *PartitionConsumerConfig) Validate() error {
+	if config.DefaultFetchSize <= 0 {
+		return ConfigurationError("Invalid DefaultFetchSize")
+	}
+
+	if config.MaxMessageSize < 0 {
+		return ConfigurationError("Invalid MaxMessageSize")
+	}
+
+	if config.EventBufferSize < 0 {
+		return ConfigurationError("Invalid EventBufferSize")
+	}
+
+	return nil
 }
 
 // ConsumerEvent is what is provided to the user when an event occurs. It is either an error (in which case Err is non-nil) or
@@ -55,28 +111,28 @@ type ConsumerEvent struct {
 	Err        error
 }
 
-// Consumer processes Kafka messages from a given topic and partition.
-// You MUST call Close() on a consumer to avoid leaks, it will not be garbage-collected automatically when
-// it passes out of scope (this is in addition to calling Close on the underlying client, which is still necessary).
-type Consumer struct {
-	client *Client
+// ConsumeErrors is a type that wraps a batch of "ConsumerEvent"s and implements the Error interface.
+// It can be returned from the PartitionConsumer's Close methods to avoid the need to manually drain errors
+// when stopping.
+type ConsumeErrors []*ConsumerEvent
 
-	topic     string
-	partition int32
-	group     string
-	config    ConsumerConfig
-
-	offset        int64
-	broker        *Broker
-	stopper, done chan bool
-	events        chan *ConsumerEvent
+func (ce ConsumeErrors) Error() string {
+	return fmt.Sprintf("kafka: %d errors when consuming", len(ce))
 }
 
-// NewConsumer creates a new consumer attached to the given client. It will read messages from the given topic and partition, as
-// part of the named consumer group.
-func NewConsumer(client *Client, topic string, partition int32, group string, config *ConsumerConfig) (*Consumer, error) {
-	// Check that we are not dealing with a closed Client before processing
-	// any other arguments
+// Consumer manages PartitionConsumers which process Kafka messages from brokers.
+type Consumer struct {
+	client *Client
+	config ConsumerConfig
+
+	lock     sync.Mutex
+	children map[string]map[int32]*PartitionConsumer
+	workers  map[*Broker]*consumerWorker
+}
+
+// NewConsumer creates a new consumer attached to the given client.
+func NewConsumer(client *Client, config *ConsumerConfig) (*Consumer, error) {
+	// Check that we are not dealing with a closed Client before processing any other arguments
 	if client.Closed() {
 		return nil, ClosedClient
 	}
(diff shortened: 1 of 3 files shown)


eapache:
As with the producer, I wrote a brief architectural wiki page which should help with review: https://github.com/Shopify/sarama/wiki/Consumer

mkobetic:
In light of the discussions in https://github.com/Shopify/reportify/issues/1190, we're seriously considering bumping up the partition counts on the analytics topics. But for that we will definitely need this feature. Given that the stack is not in production yet, we may be a good guinea pigs for this. Any concerns about that @eapache?

eapache:
Go for it - if you want an overview/architecture/review discussion we can schedule that whenever you want.

wvanbergen:
:balloon:

snormore:
:metal:

consumer.go
```diff
@@ -89,298 +145,453 @@ func NewConsumer(client *Client, topic string, partition int32, group string, co
...
+	child.consumer.removeChild(child)
+	close(child.events)
+}
+
+func (child *PartitionConsumer) dispatch() error {
+	if err := child.consumer.client.RefreshTopicMetadata(child.topic); err != nil {
+		return err
+	}
+
+	if leader, err := child.consumer.client.Leader(child.topic, child.partition); err != nil {
```
mkobetic: Same chunk of code is in ConsumePartition, could it just call dispatch() ?

eapache: The problem is the `addChild()` in the middle of this code in `ConsumerPartition` - if I just call `dispatch()` then that call gets moved and things can happen in the wrong order. It's not obvious to me how to fix that without just duplicating the code.

mkobetic: Ah, I missed that. :+1:

consumer.go
```diff
@@ -89,298 +145,453 @@ func NewConsumer(client *Client, topic string, partition int32, group string, co
...
+			w.abort(err)
+			return
+		}
+
+		for child, _ := range w.work {
+			block := response.GetBlock(child.topic, child.partition)
+			if block == nil {
+				// TODO should we be doing anything else here? The fact that we didn't get a block at all
+				// (not even one with an error) suggests that the broker is misbehaving, or perhaps something in the
+				// request/response pipeline is ending up malformed, so we could be better off aborting entirely...
```
mkobetic: Can we check whether the response has any valid blocks above?

eapache: What would that win us?

consumer.go
```diff
@@ -89,298 +145,453 @@ func NewConsumer(client *Client, topic string, partition int32, group string, co
...
 
-	if config.MaxMessageSize < 0 {
-		return ConfigurationError("Invalid MaxMessageSize")
+	return w.broker.Fetch(w.consumer.client.id, request)
+}
+
+func (w *consumerWorker) handleResponse(child *PartitionConsumer, block *FetchResponseBlock) {
+	switch block.Err {
+	case NoError:
+		break
```
mkobetic: Why do we need the break here? Because there's nothing else in the clause?

eapache: I guess it's not strictly necessary, but it makes it clearer (to me, coming from C code) that this case doesn't fall through into the one below :)

consumer.go
```diff
@@ -89,298 +145,453 @@ func NewConsumer(client *Client, topic string, partition int32, group string, co
...
+
+	atLeastOne := false
+	prelude := true
+	for _, msgBlock := range block.MsgSet.Messages {
+
+		for _, msg := range msgBlock.Messages() {
+			if prelude && msg.Offset < child.offset {
+				continue
+			}
+			prelude = false
```
mkobetic: prelude means that we've already seen this message?

consumer.go
mkobetic: I was just thinking about the comment there, that maybe we should abort, but what if there are other valid blocks in the response. But it seems that right now we won't detect the case when there isn't any valid block in the response. Would we want to abort in that case?

consumer.go
eapache: If we ask for offset X, kafka is allowed to send us a few messages <X if it already has e.g. a buffer in memory from X-5 to X+5, so we have to just ignore those messages. This happens in particular when using compression, since messages get compressed in batches and kafka is just going to send us the entire batch containing whichever offset we ask for.

consumer.go
mkobetic: I agree it is clearer this way. I just wanted to make sure I'm not missing something important.

consumer.go
eapache: Hmm, that makes sense... there are similar comments in `handleResponse`, I wonder if it will play nicely with that

mkobetic:
So I went through the code, unsurprisingly it's pretty tight and well written. I admit not having entirely clear picture of how the various channels interact, so that would be something I'd love to cover in our review session on Friday. BTW, I recommend running this through godoc -analyze, I find it helps especially navigating the channels. Here's a sample illustration:

screen shot 2015-02-11 at 1 33 22 pm

consumer.go
eapache: I just pushed https://github.com/Shopify/sarama/commit/9fdbfb71360174abee96d44bb19efec7d7e945f8 which should address this. The commit message explains my logic.

mkobetic: The rationale sounds good to me.

consumer.go
```diff
@@ -8,43 +10,97 @@ import (
...
+}
+
+// PartitionConsumerConfig is used to pass multiple configuration options to AddPartition
+type PartitionConsumerConfig struct {
+	// The default (maximum) amount of data to fetch from the broker in each request. The default is 32768 bytes.
+	DefaultFetchSize int32
+	// The maximum permittable message size - messages larger than this will return MessageTooLarge. The default of 0 is
+	// treated as no limit.
+	MaxMessageSize int32
+	// The method used to determine at which offset to begin consuming messages. The default is to start at the most recent message.
```
drdee: Kafka docs warn that `fetch.message.max.bytes` should be at least as large as `message.max.bytes` else you can produce messages that cannot be consumed. 

Assuming that `DefaultFetchSize` maps to `fetch.message.max.bytes` and `MaxMessageSize` on the producer size is set to 1M bytes (https://github.com/Shopify/sarama/blob/afe4e78f8b33036dac4576e0a7f58ff9bcbedd9a/producer.go#L25)  then I think we should bump `DefaultFetchSize` from 32k to 1M bytes as well.

eapache: `MaxMessageSize` on the consumer maps to `fetch.message.max.bytes` and it defaults to unlimited.

wvanbergen:
Updated the wiki pages to use the new names.

wvanbergen:
One last naming issue: I am annoyed by the fact we have MessageToSend and ConsumerEvent. there's only a minor difference between them. Can they be merged? If not, can we at least both call them Message?

wvanbergen:
Also, small inconsistency related to this: MessageToSend has no Err field. In case of an error, we wrap it in a ProducerError struct. For ConsumerEvent, we simply have a separate field in the struct that can be nil.

Maybe we should have a dual channel approach, one for events, and one for errors, like the Producer?

consumer.go
drdee: What does `DefaultFetchSize` map to? And can't we have property names that map 1:1 to Kafka names?

eapache: > What does DefaultFetchSize map to?

Nothing, it's a feature kafka doesn't support.

> And can't we have property names that map 1:1 to Kafka names?

Our implementation is different (I intentionally ignored their consumer and didn't use it as a guide) because goroutines and channels lend themselves to a substantially different architecture, so using them to write a scala design would probably have gone poorly.

The producer has a similar "issue" - the scala implementation is much different under the hood, so not all the properties map exactly in sometimes subtle ways.

eapache:
> One last naming issue: I am annoyed by the fact we have MessageToSend and ConsumerEvent. there's only a minor difference between them. 

The important difference (which isn't obviously surmountable) is that `MessageToSend` accepts encoders for the key/value, whereas `ConsumerEvent` can only return bytes. We could in theory have `MessageToSend` take bytes directly, but that requires the caller to serialize their payloads themselves, and it prevents them from using the actual golang values in e.g. partitioners if they so desire.

eapache:
> If not, can we at least both call them Message?

They can't have the same name, but we could call them `ProducerMessage` and `ConsumerMessage` I guess...?

wvanbergen:
+1 - it's a sucky backwards incompatible API change but it always bugs me :)

consumer.go
drdee: Fair enough but this could be an ongoing source of confusion for people -- just like what I did they would look to the Kafka documentation and try to map Sarama's config options to the documented Kafka config  options. An idea to add as comments the Kafka counterpart config option (if it exists)?

eapache:
Actually I kind of want to make the producer take bytes and really merge them:
- it avoids the need for the stupid `StringEncoder` (and most payloads seem to be strings or bytes already anyways)
- it deduplicates these two types
- smart partitioners can just use the fancy new `Metadata` field

eapache:
But I'll do that in a separate PR

consumer.go
eapache: I'd like to think that our documentation is good enough that they won't need to look at the kafka counterparts to understand - that is the first-order win imho

eapache: This can be a separate PR though

eapache: And to be clear, I'm not against adding links to counterparts, I just wish it weren't necessary :)

wvanbergen:
All my ideas can be addressed in a separate PR. This can be merged IMHO.

eapache:
I have opened tickets to track the larger changes coming out of this review. :rocket: time!

consumer.go
wvanbergen: Any reason why `EventBufferSize`, `DefaultFetchSize` and `MaxMessageSize` are part of the Partition configuration, and `MinFetchSize` and `MaxWaitTime` of the Consumer configuration?

Personally I'd like them to be all part of the `ConsumerConfig`, so the `PartitionConsumerConfig` can just focus on offset management. Or use no `PartitionConsumerConfig` at all: `NewPartitionConsumer(consumer, topic, partition, startOffset)`, where startOffset can be -1 or -2 for the 2 special values.

eapache: Options available per-partition are because they are set per-partition in the kafka fetch request. Options available globally are because they are set globally in the kafka fetch request.

I agree it would be slightly nice to unify them, but I'm not sure how to do that without removing some of the flexibility that kafka provides.

wvanbergen: I wouldn't mind moving this to ConsumerConfig and lose some flexibility, but I guess it makes sense to stay close to the protocol.

I do think that separating out the offset parameter makes sense though, so in most cases you can just set the PartitionConsumerConfig parameter to `nil`.

``` go
// compare
consumer.ConsumePartition("topic", 1, sarama.OffsetOldest, nil)
consumer.ConsumePartition("topic", 1, 12345, nil)
// vs
config := sarama.NewPartitionConsumerConfig()
config.OffsetMethod = sarama.OffsetMethodManual
config.OffsetValue = 12345
consumer.ConsumePartition("topic", 1, config)
```

eapache: My concern is when we add new offset methods (offset-from-zookeeper or offset-from-kafka or whatever) then we will have to move it back to the config anyways in order to pass in the relevant information...

PR #234, opened 2014-12-18
Retry more than once

### LINKED ISSUE #232: Support multiple in-order retries
We already support as many out-of-order retries as you want via just feeding failed messages back to `Input()`. Supporting multiple in-order retries is quite difficult, as it involves an exponential growth in the number of states in the `leaderDispatcher`. Currently we have a three-state machine; supporting multiple retries would involve duplicating this machine for every level of retries we support (up to five retries? five copies) then handling every cross-machine combination of states. Handling `n` retries would require a switch with `3^n` cases.

eapache:
Implement multiple in-order retries in the producer.

The mental model here is a fairly straightforward expansion of the existing single-retry logic (where the messages go through the pipeline again, and the `leaderDispatcher` pauses un-retried messages until all the retries have happened in order to preserve ordering).

The new version just keeps track of the high-watermark retry in progress (aka how many times the most-retried in-flight message has been retried) and makes sure that all messages with a lower retry-count are paused until the chaser has been seen. 

Fixes #232

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -96,7 +96,7 @@ func (config *ProducerConfig) Validate() error {
 		return ConfigurationError("Invalid RetryBackoff")
 	}
 
-	if config.MaxRetries < 0 || config.MaxRetries > 1 {
+	if config.MaxRetries < 0 {
 		return ConfigurationError("Invalid MaxRetries")
 	}
 
--- a/producer.go
+++ b/producer.go
@@ -387,12 +387,6 @@ func (p *Producer) leaderDispatcher(topic string, partition int32, input chan *M
 		return nil
 	}
 
-	highWatermark := 0
-	retryState := make([]struct {
-		buf          []*MessageToSend
-		expectChaser bool
-	}, p.config.MaxRetries+1)
-
 	// try to prefetch the leader; if this doesn't work, we'll do a proper breaker-protected refresh-and-fetch
 	// on the first message
 	leader, _ = p.client.Leader(topic, partition)
@@ -400,9 +394,19 @@ func (p *Producer) leaderDispatcher(topic string, partition int32, input chan *M
 		output = p.getBrokerWorker(leader)
 	}
 
+	// highWatermark tracks the "current" retry level, which is the only one where we actually let messages through,
+	// all other messages get buffered in retryState[msg.retries].buf to preserve ordering
+	// retryState[msg.retries].expectChaser simply tracks whether we've seen a chaser message for a given level (and
+	// therefore whether our buffer is complete and safe to flush)
+	highWatermark := 0
+	retryState := make([]struct {
+		buf          []*MessageToSend
+		expectChaser bool
+	}, p.config.MaxRetries+1)
+
 	for msg := range input {
 		if msg.retries > highWatermark {
-			// on the very first retried message we send off a chaser so that we know when everything "in between" has made it
+			// new, higher, retry level; send off a chaser so that we know when everything "in between" has made it
 			// back to us and we can safely flush the backlog (otherwise we risk re-ordering messages)
 			highWatermark = msg.retries
 			Logger.Printf("producer/leader state change to [retrying-%d] on %s/%d\n", highWatermark, topic, partition)
@@ -413,8 +417,9 @@ func (p *Producer) leaderDispatcher(topic string, partition int32, input chan *M
 			output = nil
 			time.Sleep(p.config.RetryBackoff)
 		} else if highWatermark > 0 {
+			// we are retrying something (else highWatermark would be 0) but this message is not a *new* retry level
 			if msg.retries < highWatermark {
-				// we are retrying but not ready yet, buffer it until we can flush things in the right order (unless it's a chaser)
+				// in fact this message is not even the current retry level, so buffer it for now (unless it's a just a chaser)
 				if msg.flags&chaser == chaser {
 					retryState[msg.retries].expectChaser = false
 				} else {
@@ -422,7 +427,8 @@ func (p *Producer) leaderDispatcher(topic string, partition int32, input chan *M
 				}
 				continue
 			} else if msg.flags&chaser == chaser {
-				// msg.retries == highWatermark && chaser flag set, we can flush/reset at least some of our buffers
+				// this message is of the current retry level (msg.retries == highWatermark) and the chaser flag is set,
+				// meaning this retry level is done and we can go down (at least) one level and flush that
 				retryState[highWatermark].expectChaser = false
 				Logger.Printf("producer/leader state change to [normal-%d] on %s/%d\n", highWatermark, topic, partition)
 				for {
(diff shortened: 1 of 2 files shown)


eapache:
There are simple tests, and I am fairly happy with the theoretical aspects of the state machine.

@wvanbergen @drdee @yagnik @mkobetic @snormore feel free to take a look

eapache:
@Shopify/kafka this is ready for review. Let me know if you want a walkthrough meeting (it's actually a fairly small number of lines-of-code changed, but the logic can be fairly subtle).

producer.go
```diff
@@ -395,45 +401,60 @@ func (p *Producer) leaderDispatcher(topic string, partition int32, input chan *M
...
-				if err := breaker.Run(doUpdate); err != nil {
-					Logger.Printf("producer/leader state change to [normal] after \"%s\" on %s/%d\n", err, topic, partition)
-					p.returnErrors(backlog, err)
-					backlog = nil
-					continue
+		if msg.retries > highWatermark {
+			// on the very first retried message we send off a chaser so that we know when everything "in between" has made it
+			// back to us and we can safely flush the backlog (otherwise we risk re-ordering messages)
+			highWatermark = msg.retries
+			Logger.Printf("producer/leader state change to [retrying-%d] on %s/%d\n", highWatermark, topic, partition)
```
wvanbergen: I've been using `%s:%d` for topic/partition elsewhere. Not sure what is more common.

eapache: I've been using `%s/%d` because `/` is one of the few chars that is not valid in a topic (I *think* `:` is although I know we don't use it)

producer.go
```diff
@@ -395,45 +401,60 @@ func (p *Producer) leaderDispatcher(topic string, partition int32, input chan *M
...
+			retryState[msg.retries].expectChaser = true
+			output <- &MessageToSend{Topic: topic, partition: partition, flags: chaser, retries: msg.retries - 1}
+			Logger.Printf("producer/leader abandoning broker %d on %s/%d\n", leader.ID(), topic, partition)
+			p.unrefBrokerWorker(leader)
+			output = nil
+			time.Sleep(p.config.RetryBackoff)
+		} else if highWatermark > 0 {
+			if msg.retries < highWatermark {
+				// we are retrying but not ready yet, buffer it until we can flush things in the right order (unless it's a chaser)
```
wvanbergen: "but not ready yet" ?

eapache: but we're still waiting on messages in flight to come back or whatever; flushing now would break ordering

eapache:
It's probably easiest to review the code for this by just reading https://github.com/Shopify/sarama/blob/multiple-retries/producer.go#L397-L479 which is the main loop of the `leaderDispatcher` goroutine.

producer.go
```diff
@@ -394,48 +394,78 @@ func (p *Producer) leaderDispatcher(topic string, partition int32, input chan *M
 		output = p.getBrokerWorker(leader)
 	}
 
+	// highWatermark tracks the "current" retry level, which is the only one where we actually let messages through,
+	// all other messages get buffered in retryState[msg.retries].buf to preserve ordering
+	// retryState[msg.retries].expectChaser simply tracks whether we've seen a chaser message for a given level (and
+	// therefore whether our buffer is complete and safe to flush)
+	highWatermark := 0
```
wvanbergen: Do we need this variable, or can we just query `len(retryState)`?

producer_test.go
```diff
@@ -394,6 +394,91 @@ func TestProducerBrokerBounce(t *testing.T) {
...
+	response3 := new(MetadataResponse)
+	response3.AddBroker(broker3.Addr(), broker3.BrokerID())
+	response3.AddTopicPartition("my_topic", 0, broker3.BrokerID(), nil, nil, NoError)
+	broker1.Returns(response3)
+	broker3.Returns(response2)
+	broker1.Returns(response1)
+	broker2.Returns(response2)
+	broker1.Returns(response1)
+	broker2.Returns(response2)
+	broker1.Returns(response3)
```
wvanbergen: Can we give these responses more descriptive names? That would make this test case a lot easier to parse.

producer.go
eapache: I suppose we could manage it by making `p.config.MaxRetries+1` the capacity instead of the length, and using `append` as appropriate. The annoying part is if we are no state at all (normal operation) and we get a message with `retries==9` or something then we have to loop adding empty elements to the array.

Not an unreasonable suggestion, but I think I like the current method better unless you have additional thoughts.

producer_test.go
eapache: good idea, I will fix them all in a separate PR since there are a lot of other tests that look like this

wvanbergen: So, this tests the following sequence?

- broker 1 (metadata broker) tells us the partition lives on broker 2.
- broker 2 (produce attempt 1) tells us that is not actually the case
- broker 1 (metadata broker) tells us the partition now lives on broker 3.
- broker 3 (produce attempt 2) tells us this is not the case either.
- broker 1 (metadata broker) tells us the partition now lives on broker 2 again.
- broker 2 (produce attempt 3) tells us this is not the case
- broker 1 (metadata broker) tells us the partition now lives on broker 2 again.
- broker 2 (produce attempt 4) tells us this is still not the case
- broker 1 (metadata broker) tells us the partition now lives on broker 3.
- broker 4 (produce attempt 5, retry 4) produces successfully.

So this flow exactly fits in 4 retries.

wvanbergen: I guess this only tests one retry level to be active at the same time. Maybe we want to test the scenario of having different batches on different retry levels as well, and see if the ordering is still OK?

eapache: exactly

producer_test.go
```diff
@@ -394,6 +394,91 @@ func TestProducerBrokerBounce(t *testing.T) {
...
+	config.MaxRetries = 4
+	config.RetryBackoff = 0
+	producer, err := NewProducer(client, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for i := 0; i < 10; i++ {
+		producer.Input() <- &MessageToSend{Topic: "my_topic", Key: nil, Value: StringEncoder(TestMessage)}
+	}
```
wvanbergen: If I understand it correctly, these will all end up on one produce request, which will fail.

producer_test.go
```diff
@@ -394,6 +394,91 @@ func TestProducerBrokerBounce(t *testing.T) {
...
+		case msg := <-producer.Successes():
+			if msg.flags != 0 {
+				t.Error("Message had flags set")
+			}
+		}
+	}
+
+	for i := 0; i < 10; i++ {
+		producer.Input() <- &MessageToSend{Topic: "my_topic", Key: nil, Value: StringEncoder(TestMessage)}
+	}
```
wvanbergen: These should now all succeed as a single batch on the first attempt (right?)

producer_test.go
eapache: Yes, because `FlushMsgCount`

producer_test.go
eapache: Yup, just making sure the state is sane for new messages after we recover.

producer_test.go
eapache: The difficulty of testing multiple batches is that there's no easy way to control the concurrency and write reliable tests :(

wvanbergen:
This looks good, but maybe we should add some comments to the test case to make the flow more clear.

producer.go
wvanbergen: yeah, this is nice and clear. The int extra variable will not kill us in memory usage.

PR #242, opened 2015-01-02
make it possible to configure inner Producer in SimpleProducer

mihasya:
This is a fairly simple PR. The underlying motivation is that we occasionally have larger messages, so we need to be able to configure `ProducerConfig.MaxMessageBytes`

The pre-existing behavior is preserved exactly as is, so this is fully backwards compatible.

### DIFF
--- a/simple_producer.go
+++ b/simple_producer.go
@@ -14,15 +14,17 @@ type spExpect struct {
 	result chan error
 }
 
-// NewSimpleProducer creates a new SimpleProducer using the given client, topic and partitioner. If the
-// partitioner is nil, messages are partitioned by the hash of the key
-// (or randomly if there is no key).
-func NewSimpleProducer(client *Client, topic string, partitioner PartitionerConstructor) (*SimpleProducer, error) {
+// NewSimpleProducerWithConfig is the same as NewSimpleProducer, except it
+// includes the option to set certain configuration parameters on the
+// underlying producer
+func NewSimpleProducerWithConfig(client *Client, topic string, partitioner PartitionerConstructor, config *ProducerConfig) (*SimpleProducer, error) {
 	if topic == "" {
 		return nil, ConfigurationError("Empty topic")
 	}
 
-	config := NewProducerConfig()
+	if config == nil {
+		config = NewProducerConfig()
+	}
 	config.AckSuccesses = true
 	if partitioner != nil {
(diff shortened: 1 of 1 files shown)


simple_producer.go
```diff
@@ -14,15 +14,17 @@ type spExpect struct {
...
 }
 
-// NewSimpleProducer creates a new SimpleProducer using the given client, topic and partitioner. If the
-// partitioner is nil, messages are partitioned by the hash of the key
-// (or randomly if there is no key).
-func NewSimpleProducer(client *Client, topic string, partitioner PartitionerConstructor) (*SimpleProducer, error) {
+// NewSimpleProducerWithConfig is the same as NewSimpleProducer, except it
+// includes the option to set certain configuration parameters on the
+// underlying producer
+func NewSimpleProducerWithConfig(client *Client, topic string, partitioner PartitionerConstructor, config *ProducerConfig) (*SimpleProducer, error) {
```
eapache: There's no point in taking a separate `partitioner` argument in this form, since you can just set it directly on the passed-in config instance.

eapache: I'm not a huge fan of the `WithConfig` suffix, but I can't come up with anything better off the top of my head.

eapache:
A few minor notes, but otherwise this is fine.

mihasya:
@eapache re `WithConfig` yeah if we were going from scratch, I might have opted for a `Builder` pattern or something, but it seems overkill to do that since there aren't that many members ultimately. Was just trying to add the know while preserving b/w compatibility.

Good call on the partitioner bit, I'll make the change.

eapache:
LGTM

@wvanbergen ?

mihasya:
Hope this simple PR doesn't die on the vine :worried:

wvanbergen:
I'd prefer without `WithConfig`, if @eapache agrees we are still in "breaking changes mode" for Sarama.

mihasya:
@wvanbergen could you clarify what implementation you would prefer?

wvanbergen:
Instead of adding another method, I just want to replace the existing one to allow for a producer config. That breaks backwards compatibility, but that might still be OK if @eapache agrees.

eapache:
I'd be OK with that I guess, we still have to land the multi-consumer so the API isn't frozen yet.

eapache:
Effectively covered by #248

PR #244, opened 2015-01-15
producer: add warning about RequiredAcks>1

eapache:
Per https://groups.google.com/d/msg/kafka-clients/4dR2FeN9GKE/_I3WRMoF8AsJ

Setting it to a value > 1 will raise an exception in 0.8.2.0 and later, so
update the godoc and log a warning when validating such a configuration.

@wvanbergen @snormore @drdee 

cc @mkobetic @yagnik


wvanbergen:
LGTM.

Issue #245, opened 2015-01-19
panic deep in fetchMessages() b/c c.broker is nil when kafka cluster stops


nsd20463:
I have a reproducible panic in the sarama library.

If we stop the kafka cluster by stopping each broker one by one (in this case our 3 brokers named kafka-000, kafka-001 and kafka-002) the client panics in *Broker.send() because b (the pointer to the Broker) is nil.

This is with the current sarama (74f18a8a1b61b4658b466f7eb07702839e207e0d), go 1.4 on linux/amd64 (tag "go1.4" == 21924:9ef10fde754f), kafka version 0.8.2beta, and rather ordinary client code:

    cl, err := sarama.NewClient("change", []string{*KAFKA_BROKER}, nil)
    ...(error checking elided)...
    partitions, err := cl.Partitions(KAFKA_CHANGE_TOPIC)
    ...
    con_config := sarama.NewConsumerConfig()
    partition := partitions[0]
    con_config.OffsetMethod = sarama.OffsetMethodNewest
    con, err := sarama.NewConsumer(cl, KAFKA_CHANGE_TOPIC, partition, "change_group", con_config)

    for {
        select {
        case ev := <-con.Events():
        ...

The sarama logger out and the panic are:

<start when we stop kafka>
G72 LOG> 19:58:15.883013 consumer.go:191: Unexpected error processing FetchRequest; disconnecting broker kafka-001:6667: EOF
G79 LOG> 19:58:15.883025 consumer.go:191: Unexpected error processing FetchRequest; disconnecting broker kafka-001:6667: EOF
G79 LOG> 19:58:15.883071 client.go:310: Disconnecting Broker 172314977
G79 LOG> 19:58:15.883090 client.go:353: Fetching metadata from broker kafka-000:6667
G79 LOG> 19:58:15.883201 client.go:376: Error from broker while fetching metadata: EOF
G79 LOG> 19:58:15.883224 client.go:310: Disconnecting Broker -1
G79 LOG> 19:58:15.883244 client.go:353: Fetching metadata from broker kafka-002:6667
G86 LOG> 19:58:15.883337 broker.go:134: Closed connection to broker kafka-001:6667
G87 LOG> 19:58:15.883381 broker.go:134: Closed connection to broker kafka-000:6667
G72 LOG> 19:58:15.883403 client.go:310: Disconnecting Broker 172314977
G72 LOG> 19:58:15.883420 client.go:353: Fetching metadata from broker kafka-000:6667
[...]

PR #246, opened 2015-01-20
Don't fully trust kafka metadata errors

eapache:
Verify that we really got a leader back, even in the case where the error is not
LeaderNotAvailable. Might solve #245.

@nsd20463 does this fix your bug? If so I will merge it, but I would also suggest filing a ticket upstream that Kafka should really return `LeaderNotAvailable` if it's not going to send us a broker.


eapache:
r: @drdee @yagnik
cc: @wvanbergen @snormore @mkobetic

wvanbergen:
Kind of shitty that this is necessary, but :shipit:

PR #250, opened 2015-01-30
Metadata interface{} field added to MessageToSend

schleppy:
This simple change adds an additional interface{} field to the MessageToSend struct
named Metadata.  It is the intention of this field to allow clients to
attach arbitrary data of their choosing to the MessaageToSend so they
can retrieve that data in the Successes and Errors channels and not have
to do any sort of bookkeeping themselves.

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -153,6 +153,7 @@ const (
 type MessageToSend struct {
 	Topic      string
 	Key, Value Encoder
+	Metadata   interface{}
 
 	// these are filled in by the producer as the message is processed
 	offset    int64
--- a/producer.go
+++ b/producer.go
@@ -154,7 +154,7 @@ type MessageToSend struct {
 	Topic    string      // The Kafka topic for this message
 	Key      Encoder     // The partitioning key for this message. It must implement the Encoder interface. Pre-existing Encoders include StringEncoder and ByteEncoder
 	Value    Encoder     // The actual message to store in Kafka. It must implement the Encoder interface. Pre-existing Encoders include StringEncoder and ByteEncoder
-	Metadata interface{} // This field is used to hold arbitrary data you wish to include so it will be available when recieving on the Succcesses and Errors channels
+	Metadata interface{} // This field is used to hold arbitrary data you wish to include so it will be available when recieving on the Successes and Errors channels.  Sarma completely ignores this field and is only to be used for pass-through data.
 
 	// these are filled in by the producer as the message is processed
 	offset    int64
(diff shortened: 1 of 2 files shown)


joestein:
Cool, this looks great, looking forward to being able to use it. One nice thing I like about this is it makes being able to test different scenarios and failure cases much more succinct. Much more straight forward to read your meta data field rather than having to deserialize the entire message again or do hashing operations on the bytes, etc.

producer.go
```diff
@@ -153,6 +153,7 @@ const (
 type MessageToSend struct {
 	Topic      string
 	Key, Value Encoder
+	Metadata   interface{}
```
wvanbergen: We should probably add some godoc to this struct.

wvanbergen:
This looks good to me, just missing some documentation :+1:

producer.go
```diff
@@ -151,8 +151,10 @@ const (
 
 // MessageToSend is the collection of elements passed to the Producer in order to send a message.
 type MessageToSend struct {
-	Topic      string
-	Key, Value Encoder
+	Topic    string      // The Kafka topic for this message
+	Key      Encoder     // The partitioning key for this message. It must implement the Encoder interface. Pre-existing Encoders include StringEncoder and ByteEncoder
+	Value    Encoder     // The actual message to store in Kafka. It must implement the Encoder interface. Pre-existing Encoders include StringEncoder and ByteEncoder
+	Metadata interface{} // This field is used to hold arbitrary data you wish to include so it will be available when recieving on the Succcesses and Errors channels
```
eapache: typo in "Succcesses" - also it would probably be good to mention that Sarama completely ignores this field so it's clear it's just for pass-through data

producer.go
```diff
@@ -151,8 +151,10 @@ const (
 
 // MessageToSend is the collection of elements passed to the Producer in order to send a message.
 type MessageToSend struct {
-	Topic      string
-	Key, Value Encoder
+	Topic    string      // The Kafka topic for this message
+	Key      Encoder     // The partitioning key for this message. It must implement the Encoder interface. Pre-existing Encoders include StringEncoder and ByteEncoder
+	Value    Encoder     // The actual message to store in Kafka. It must implement the Encoder interface. Pre-existing Encoders include StringEncoder and ByteEncoder
+	Metadata interface{} // This field is used to hold arbitrary data you wish to include so it will be available when recieving on the Successes and Errors channels.  Sarma completely ignores this field and is only to be used for pass-through data.
```
wvanbergen: Typo: `receiving`

wvanbergen: Also `Sarama` instead of `Sarma`. Sorry for nitpicking :)

schleppy: Haha... not typing on my normal keyboard.  Nitpicking is a good thing.

eapache:
Merged via manual squash/push (so github's interface is confused, but the change *is* in master with the correct authorship)

wvanbergen:
@schleppy Thanks for contributing, much appreciated!

schleppy:
@wvanbergen My pleasure.  Thank you and the whole team for a great project.

PR #251, opened 2015-01-30
simple_producer: Retry publish on failure

abraithwaite:
Add metadata retries to the simple producer.
Reset retry count when a message has been successfully acked
by the broker.

For context: we had an issue where Kafka was configured to auto-create topics when published to.  When we kept getting the broker election error in the simple producer, we figured out it was this.

It also [appears that this is the expected behavior](https://github.com/apache/kafka/blob/7847e9c703f3a0b70519666cdb8a6e4c8e37c3a7/core/src/test/scala/unit/kafka/integration/TopicMetadataTest.scala#L111-L136) for creating topics.  That is to say: Kafka's behavior is to return this error and when the producer tries again, the topic should have been created.

### DIFF
--- a/simple_producer.go
+++ b/simple_producer.go
@@ -1,5 +1,7 @@
 package sarama
 
+import "time"
+
 // SimpleProducer publishes Kafka messages. It routes messages to the correct broker, refreshing metadata as appropriate,
 // and parses responses for errors. You must call Close() on a producer to avoid leaks, it may not be garbage-collected automatically when
 // it passes out of scope (this is in addition to calling Close on the underlying client, which is still necessary).
@@ -42,7 +44,6 @@ func (sp *SimpleProducer) SendMessage(topic string, key, value Encoder) error {
 	expectation := &spExpect{msg: msg, result: make(chan error)}
 	sp.newExpectations <- expectation
 	sp.producer.Input() <- msg
-
 	return <-expectation.result
 }
 
@@ -50,6 +51,8 @@ func (sp *SimpleProducer) matchResponses() {
 	newExpectations := sp.newExpectations
 	unmatched := make(map[*MessageToSend]chan error)
 	unmatched[nil] = nil // prevent it from emptying entirely
+
+	retry := sp.producer.client.config.MetadataRetries
 
 	for len(unmatched) > 0 {
 		select {
@@ -61,9 +64,17 @@ func (sp *SimpleProducer) matchResponses() {
 			}
 			unmatched[expectation.msg] = expectation.result
 		case err := <-sp.producer.Errors():
+			if err.Err == LeaderNotAvailable && retry > 0 {
+				time.AfterFunc(sp.producer.config.RetryBackoff, func() {
+					sp.producer.Input() <- err.Msg
+				})
+				retry--
+				continue
+			}
 			unmatched[err.Msg] <- err.Err
 			delete(unmatched, err.Msg)
 		case msg := <-sp.producer.Successes():
+			retry = sp.producer.client.config.MetadataRetries
 			close(unmatched[msg])
 			delete(unmatched, msg)
 		}


abraithwaite:
Also, another issue I saw (mostly with Kafka, not sarama) was that there's a number of configs which essentially do the same thing:

From Kafka's docs, these are: Producer's `retry.backoff.ms` Consumer's `rebalance.backoff.ms`

In Sarama, there's the client metadata retry timeout and the producer retry timeout.  Please tell me which one you'd prefer to use.

abraithwaite:
Also, if this conflicts with #234, consider it a bug report instead :beetle:

eapache:
I can't speak for kafka's configuration values, but those two sarama configuration values do *very* different things, and should not be conflated or merged.

- in the client, `WaitForElection` and `MetadataRetries` control how it retries *fetching metadata* in the case when the cluster is in the middle of leadership election (such as when a topic is auto-created)
- in the producer, `RetryBackoff` (and once #234 lands, `MaxRetries`) control how it retries *producing messages* in the producer in the case where a broker fails or the network connection is lost or etc. etc.

As far as I understand it, this PR should already be unnecessary (even without #234) to support the case of auto-created topics. The first produced message will fail and automatically retry, and will succeed the second time. All other messages should then succeed. What configuration are you using that this is not the case?

abraithwaite:
What we observed for simple_producer was that it would fail immediately on the first `LeaderNotAvailable` error.  It does not automatically retry.  When I was debugging it, I never saw `producer/flusher state change to [retrying]...` which if I understand right, is where it should be queued for retry.

The root problem though seems to be that the simple producer reads from the `Errors()` channel and if a message matches (error or not), it considers it processed (see `unmatched`).

eapache:
> The root problem though seems to be that the simple producer reads from the Errors() channel and if a message matches (error or not), it considers it processed (see unmatched).

No, once a message comes out the `Errors()` channel that's it - retries happen internally to the producer and do not come out `Errors()` until/unless all retries have failed.

> When I was debugging it, I never saw `producer/flusher state change to [retrying]...` which if I understand right, is where it should be queued for retry.

Not necessarily. `LeaderNotAvailable` is a metadata error which suggests to me (in combination with the fact you're using auto-created partitions) that the retries should be happening in the client instead. The message I would expect is along the lines of `Some partitions are leaderless, waiting 250ms for election... (2 retries remaining)` (from `client.go#refreshMetadata`). Did you see any of those?

If you didn't see those messages either, what configuration are you using? Are you using struct literals for any configuration values that might accidentally be setting `MetadataRetries` to 0 or something?

abraithwaite:
> No, once a message comes out the Errors() channel that's it - retries happen internally to the producer and do not come out Errors() until/unless all retries have failed.

I wasn't aware.  Thanks.

> The message I would expect is along the lines of Some partitions are leaderless, waiting 250ms for election... (2 retries remaining) (from client.go#refreshMetadata). Did you see any of those?

Unfortunately I didn't see any of those messages.  Here is how I'm creating the producer:
```
	brokers := strings.Split(*brokerList, ",")
	client, err := sarama.NewClient("testclient", brokers, sarama.NewClientConfig())
	if err != nil {
		log.Fatalf("NewClient: %v", err)
	}
	producer, err := sarama.NewSimpleProducer(client, topic, pc)
	if err != nil {
		log.Fatalf("NewProducer: %v", err)
	}
```
So the retries are set to 3 (otherwise the patch wouldn't work either).

abraithwaite:
One issue might be that sarama doesn't even know about the partition yet (the metadata hasn't refreshed after the initial send), so can't know that they're leaderless until the next refresh.  Meanwhile, the synchronous send either needs to block or return something and the error it gets back from Kafka is NoLeaderForPartition.

Thoughts?

eapache:
> the error it gets back from Kafka is NotLeaderForPartition.

earlier you said LeaderNotAvailable - which is it?

abraithwaite:
Sorry, that's what I meant.

eapache:
I can't say for sure without seeing the stack trace, but I suspect it's not even getting to the point of sending the message. In order to choose a partition (which must happen before the full message can be constructed), the producer asks the client for the available partitions for a given topic. If the client doesn't know about that topic yet (which it wouldn't) it refreshes metadata for that specific topic (which should also create the topic on the brokers). This triggers a full retry in `refreshMetadata`, which *should* log the message you're not seeing...

eapache:
I took a guess in #252 could you let me know if that fixes the problem? If not, a set of sarama logs would be very helpful.

abraithwaite:
That seemed to fix it.  I now also see the retry message you were referencing.

Thanks for the help!

PR #254, opened 2015-02-05
Require a non-zero timeout when WaitForAll

eapache:
Otherwise requests are almost bound to fail - there's no good reason for this
config.

@wvanbergen @drdee


wvanbergen:
:+1:

wvanbergen:
I guess we don't really have to deal with the case of values RequiredAcks > 1 anymore, given that that is deprecated

eapache:
That was my thought as well.

PR #260, opened 2015-02-11
Run integration tests against 0.8.2

eapache:


wvanbergen:
+1, let's hope the download URL still works :)

PR #262, opened 2015-02-12
Handle ReplicaNotAvailable like LeaderNotAvailable when initializing a client

wvanbergen:
According to the logs, this was the cause of the kafka_sysv_mq_producer was not restarting during our testing. Apparently we already found that this caused issues when using LeaderNotAvailable (#144), and apparently ReplicaNotAvailable can happen to. We should handle them similarly

@Shopify/kafka


eapache:
:+1:

wvanbergen:
@dalehamel FYI

PR #264, opened 2015-02-12
When we get a new leader, check it is connected

eapache:
If we don't do this we'll foist it off on the flusher, which is legit but means
we waste our retry. We can fail fast here and the resulting behaviour will be
better.

@wvanbergen


wvanbergen:
:cool:

PR #265, opened 2015-02-12
Process partition metadata in event of topic-level error

eapache:
As per 1b465e78c274a76c0a522272cac25e39ddaa0d07, "I can't think of a case where
the stale data from the last "complete" update is somehow more useful than the
up-to-date partial data."

@wvanbergen

### DIFF
--- a/client.go
+++ b/client.go
@@ -537,10 +537,8 @@ func (client *Client) update(data *MetadataResponse) ([]string, error) {
 			break
 		case LeaderNotAvailable:
 			toRetry[topic.Name] = true
-			continue
 		default:
 			err = topic.Err
-			continue
 		}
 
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -537,10 +537,8 @@ func (client *Client) update(data *MetadataResponse) ([]string, error) {
 			break
 		case LeaderNotAvailable:
 			toRetry[topic.Name] = true
-			continue
```
wvanbergen: Should we handle ReplicaNotAvailable here as well?

wvanbergen:
Anyway, this makes sense. :+1:

client.go
eapache: there's no election to wait for in that case, the partition is just underreplicated

PR #267, opened 2015-02-13
Fix flaky SimpleProducer test

eapache:
This test failed randomly in an unrelated PR:
https://travis-ci.org/Shopify/sarama/jobs/50586636

Turns out we were relying on the go scheduler to end up batching 100 messages
into exactly two requests, which isn't guaranteed at all. Now that we can pass a
full config into the SimpleProducer, just set the FlushMsgCount and expect a
single request.

cc @wvanbergen @drdee @yagnik


wvanbergen:
Nice find! This is not the flaky heisenbug we saw in the integration test, right?

eapache:
No, sadly not.

Issue #269, opened 2015-02-13
Merge MessageToSend and ConsumerEvent


eapache:
See discussion on https://github.com/Shopify/sarama/pull/222

cc @wvanbergen

eapache:
Closing in favour of #305

PR #273, opened 2015-02-15
[WIP] Unite MessageToSend and ConsumerEvent

wvanbergen:
Start of #269.

1. New names: `ConsumerEvent` becomes `ConsumerMessage` and in the producer, `MessageToSend` becomes `ProducerMessage`. Once we're ready to unite it I'd like to simply call it `Message`, but right now that is in used internally for messages between client and broker. Can we find a new name for this, e.g. `BrokerMessage` or `ProtocolMessage`?

2. Remove encoder interface, and simply use `[]byte` for keys and values.

Still to resolve: `Metadata`, and `Err`.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -101,9 +101,9 @@ func (config *PartitionConsumerConfig) Validate() error {
 	return nil
 }
 
-// ConsumerEvent is what is provided to the user when an event occurs. It is either an error (in which case Err is non-nil) or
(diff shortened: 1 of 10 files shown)


wvanbergen:
@mkobetic @Shopify/kafka 

As discussed in out walkthrough of the new consumer.

wvanbergen:
Going to close this as this is outdated and I am not so sure about the approach anymore, I'll open an issue instead;

PR #275, opened 2015-02-16
Set default producer request timeout to 10 seconds

wvanbergen:
This value should be at least the amount of time it takes for a non-responsive broker to be considered out of sync. The default value for that is 10 seconds. So I set it to 10 seconds, which matches the default setting of the JVM producer. Maybe we should set it a bit higher as a safety buffer?

@Shopify/kafka @mkobetic


eapache:
Include the default in the godoc of the struct field, then :shipit:

wvanbergen:
I'll tackle the godoc as part of #276

Issue #270, opened 2015-02-13
Document configuration equivalents


eapache:
per discussion on https://github.com/Shopify/sarama/pull/222

- [x] producer
- [x] consumer
- [x] client
- [x] broker

cc @drdee

eapache:
Finished by #311

PR #276, opened 2015-02-16
Document setting equivalents with JVM producer

wvanbergen:
Part of #270, this documents the equivalents of the producer. AFAICS the other settings do not have equivalents.
 
Besides documenting, maybe we can also fix some naming inconsistencies and ambiguities ?

@drdee @eapache

### DIFF
--- a/producer.go
+++ b/producer.go
@@ -13,19 +13,22 @@ func forceFlushThreshold() int {
 }
 
 // ProducerConfig is used to pass multiple configuration options to NewProducer.
+//
+// Some of these configuration settings match settings with the JVM producer, but some of
+// these are implementation specific and have no equivalent in the JVM producer.
 type ProducerConfig struct {
-	Partitioner       PartitionerConstructor // Generates partitioners for choosing the partition to send messages to (defaults to hash).
-	RequiredAcks      RequiredAcks           // The level of acknowledgement reliability needed from the broker (defaults to WaitForLocal).
-	Timeout           time.Duration          // The maximum duration the broker will wait the receipt of the number of RequiredAcks. This is only relevant when RequiredAcks is set to WaitForAll or a number > 1. Only supports millisecond resolution, nanoseconds will be truncated.
-	Compression       CompressionCodec       // The type of compression to use on messages (defaults to no compression).
-	FlushMsgCount     int                    // The number of messages needed to trigger a flush. This is a minimum, not an upper limit (use MaxMessagesPerReq for that).
-	FlushFrequency    time.Duration          // If this amount of time elapses without a flush, one will be queued. This is a minimum, not an upper limit.
-	FlushByteCount    int                    // If this many bytes of messages are accumulated, a flush will be triggered. This is a minimum, not an upper limit.
+	Partitioner       PartitionerConstructor // Generates partitioners for choosing the partition to send messages to (defaults to hash). Similar to the `partitioner.class` setting for the JVM producer.
+	RequiredAcks      RequiredAcks           // The level of acknowledgement reliability needed from the broker (defaults to WaitForLocal). Equivalent to the `request.required.acks` setting of the JVM producer.
+	Timeout           time.Duration          // The maximum duration the broker will wait the receipt of the number of RequiredAcks (defaults to 10 seconds). This is only relevant when RequiredAcks is set to WaitForAll or a number > 1. Only supports millisecond resolution, nanoseconds will be truncated. Equivalent to the JVM producer's `request.timeout.ms` setting.
+	Compression       CompressionCodec       // The type of compression to use on messages (defaults to no compression). Similar to `compression.codec` setting of the JVM producer.
+	FlushMsgCount     int                    // The number of messages needed to trigger a flush. This is a best effort; the number of messages may be more or less. Use `MaxMessagesPerReq` to set a hard upper limit.
+	FlushFrequency    time.Duration          // If this amount of time elapses without a flush, one will be queued. The frequency is a best effort, and the actual frequency can be more or less. Equivalent to `queue.buffering.max.ms` setting of JVM producer.
+	FlushByteCount    int                    // If this many bytes of messages are accumulated, a flush will be triggered. This is a best effort; the number of bytes may be more or less. Use the gloabl `sarama.MaxRequestSize` to set a hard upper limit.
 	AckSuccesses      bool                   // If enabled, successfully delivered messages will be returned on the Successes channel.
-	MaxMessageBytes   int                    // The maximum permitted size of a message (defaults to 1000000)
-	MaxMessagesPerReq int                    // The maximum number of messages the producer will send in a single broker request. Defaults to 0 for unlimited. The global setting MaxRequestSize still applies.
-	ChannelBufferSize int                    // The size of the buffers of the channels between the different goroutines. Defaults to 0 (unbuffered).
-	RetryBackoff      time.Duration          // The amount of time to wait for the cluster to elect a new leader before processing retries. Defaults to 250ms.
+	MaxMessageBytes   int                    // The maximum permitted size of a message (defaults to 1000000). Equivalent to the broker's `message.max.bytes`.
+	MaxMessagesPerReq int                    // The maximum number of messages the producer will send in a single broker request. Defaults to 0 for unlimited. The global setting MaxRequestSize still applies. Similar to `queue.buffering.max.messages` in the JVM producer.
+	ChannelBufferSize int                    // The size of the buffers of the channels between the different goroutines (defaults to 0, i.e. unbuffered).
+	RetryBackoff      time.Duration          // The amount of time to wait for the cluster to elect a new leader before processing retries (defaults to 250ms). Similar to the retry.backoff.ms setting of the JVM producer
 }
 
 // NewProducerConfig creates a new ProducerConfig instance with sensible defaults.


producer.go
```diff
@@ -13,19 +13,22 @@ func forceFlushThreshold() int {
...
+	Timeout           time.Duration          // The maximum duration the broker will wait the receipt of the number of RequiredAcks (defaults to 10 seconds). This is only relevant when RequiredAcks is set to WaitForAll or a number > 1. Only supports millisecond resolution, nanoseconds will be truncated.
+	Compression       CompressionCodec       // The type of compression to use on messages (defaults to no compression). Similar to compression.codec setting of the JVM producer.
 	FlushMsgCount     int                    // The number of messages needed to trigger a flush. This is a minimum, not an upper limit (use MaxMessagesPerReq for that).
-	FlushFrequency    time.Duration          // If this amount of time elapses without a flush, one will be queued. This is a minimum, not an upper limit.
+	FlushFrequency    time.Duration          // If this amount of time elapses without a flush, one will be queued. This is a minimum, not an upper limit. Equivalent to queue.buffering.max.ms setting of JVM producer.
```
wvanbergen: "This is a minimum, not an upper limit." What does that mean? When can it be fore than this interval?  It makes more sense as a maximum to me.

producer.go
```diff
@@ -13,19 +13,22 @@ func forceFlushThreshold() int {
...
-	Compression       CompressionCodec       // The type of compression to use on messages (defaults to no compression).
+	Partitioner       PartitionerConstructor // Generates partitioners for choosing the partition to send messages to (defaults to hash). Similar to the partitioner.class setting for the JVM producer.
+	RequiredAcks      RequiredAcks           // The level of acknowledgement reliability needed from the broker (defaults to WaitForLocal). Equavilent to the request.required.acks setting of the JVM producer.
+	Timeout           time.Duration          // The maximum duration the broker will wait the receipt of the number of RequiredAcks (defaults to 10 seconds). This is only relevant when RequiredAcks is set to WaitForAll or a number > 1. Only supports millisecond resolution, nanoseconds will be truncated.
+	Compression       CompressionCodec       // The type of compression to use on messages (defaults to no compression). Similar to compression.codec setting of the JVM producer.
 	FlushMsgCount     int                    // The number of messages needed to trigger a flush. This is a minimum, not an upper limit (use MaxMessagesPerReq for that).
```
wvanbergen: > This is a minimum, not an upper limit (use MaxMessagesPerReq for that).

Should our naming be more consistent here? E.g. `MinMessagesPerReq` ?

producer.go
```diff
@@ -13,19 +13,22 @@ func forceFlushThreshold() int {
...
-	FlushFrequency    time.Duration          // If this amount of time elapses without a flush, one will be queued. This is a minimum, not an upper limit.
+	FlushFrequency    time.Duration          // If this amount of time elapses without a flush, one will be queued. This is a minimum, not an upper limit. Equivalent to queue.buffering.max.ms setting of JVM producer.
 	FlushByteCount    int                    // If this many bytes of messages are accumulated, a flush will be triggered. This is a minimum, not an upper limit.
 	AckSuccesses      bool                   // If enabled, successfully delivered messages will be returned on the Successes channel.
 	MaxMessageBytes   int                    // The maximum permitted size of a message (defaults to 1000000)
```
wvanbergen: `Message` may be ambiguous here? Is this the maximum size of a single message to be produced, or the protocol message to be sent to the server? Maybe `MaxRequestBytes`?

producer.go
```diff
@@ -13,19 +13,22 @@ func forceFlushThreshold() int {
...
+// these are implementation specific and have no equivalent in the JVM producer.
 type ProducerConfig struct {
-	Partitioner       PartitionerConstructor // Generates partitioners for choosing the partition to send messages to (defaults to hash).
-	RequiredAcks      RequiredAcks           // The level of acknowledgement reliability needed from the broker (defaults to WaitForLocal).
-	Timeout           time.Duration          // The maximum duration the broker will wait the receipt of the number of RequiredAcks. This is only relevant when RequiredAcks is set to WaitForAll or a number > 1. Only supports millisecond resolution, nanoseconds will be truncated.
-	Compression       CompressionCodec       // The type of compression to use on messages (defaults to no compression).
+	Partitioner       PartitionerConstructor // Generates partitioners for choosing the partition to send messages to (defaults to hash). Similar to the partitioner.class setting for the JVM producer.
+	RequiredAcks      RequiredAcks           // The level of acknowledgement reliability needed from the broker (defaults to WaitForLocal). Equavilent to the request.required.acks setting of the JVM producer.
```
eapache: typo Equivalent

producer.go
```diff
@@ -13,19 +13,22 @@ func forceFlushThreshold() int {
...
+	Timeout           time.Duration          // The maximum duration the broker will wait the receipt of the number of RequiredAcks (defaults to 10 seconds). This is only relevant when RequiredAcks is set to WaitForAll or a number > 1. Only supports millisecond resolution, nanoseconds will be truncated.
```
eapache: `request.timeout.ms` they also default to 10s btw, so :+1:

producer.go
eapache: But it's not a minimum either, it's a trigger :(

We handle blocking and non-blocking flushes, meaning:
- blocking flushes do not accept new messages until the current batch is on the wire; we use this when the request gets too large (package-global `MaxRequestSize` or the producer-level `MaxMessagesPerReq`) and adding more messages to it risks having the broker reject it
- non-blocking flushes will flush as soon as the network pipe is available, but will continue to add new messages to the request in the meantime: when the request exceeds `FlushMsgCount` or `FlushFrequency` or `FlushByteCount`)

producer.go
eapache: As per my other comment, if another request is still on the wire this will wait. There was a ticket where somebody was confused about that, so I added "this is a minimum, not an upper limit" to all the descriptions, although of course (per my other comment) it's not technically a minimum either if the other `Flush*` values are set since they could trigger first.

producer.go
eapache: This is the max size of a single message - I tried to use message vs request consistently throughout although if you have even better terminology I'm open to it.

producer.go
```diff
@@ -13,19 +13,22 @@ func forceFlushThreshold() int {
...
 	FlushByteCount    int                    // If this many bytes of messages are accumulated, a flush will be triggered. This is a minimum, not an upper limit.
 	AckSuccesses      bool                   // If enabled, successfully delivered messages will be returned on the Successes channel.
 	MaxMessageBytes   int                    // The maximum permitted size of a message (defaults to 1000000)
-	MaxMessagesPerReq int                    // The maximum number of messages the producer will send in a single broker request. Defaults to 0 for unlimited. The global setting MaxRequestSize still applies.
-	ChannelBufferSize int                    // The size of the buffers of the channels between the different goroutines. Defaults to 0 (unbuffered).
-	RetryBackoff      time.Duration          // The amount of time to wait for the cluster to elect a new leader before processing retries. Defaults to 250ms.
+	MaxMessagesPerReq int                    // The maximum number of messages the producer will send in a single broker request. Defaults to 0 for unlimited. The global setting MaxRequestSize still applies. Similar to batch.num.messages in the JVM producer.
```
eapache: this is not `batch.num.messages`, `FlushMsgCount` is

eapache: this is more like `queue.buffering.max.messages`

producer.go
eapache: should be in sync with `message.max.bytes` on the broker

eapache:
Yes, the flushing logic is non-trivial, @burke and tobi and I had a big long discussion about it several years ago (see #41 and #48).

In the current code it's entirely encapsulated in the `messageAggregator` method though, so if you read/understand that function you should grok it.

producer.go
wvanbergen: OK, I guess we can just leave the name as is. Instead of naming, it may make sense to have the flushing behavior documented in the wiki or something.

wvanbergen:
Fixed according to code review.

producer.go
wvanbergen: If this is not a guaranteed maximum, do you risk having messages rejected on the broker if this setting matches the broker's value?

eapache: `MaxMessageBytes` is a guaranteed maximum. It's the `Flush*` values that are best-effort and so guarantee neither maximum nor minimum.

producer.go
```diff
@@ -13,19 +13,22 @@ func forceFlushThreshold() int {
...
+	Compression       CompressionCodec       // The type of compression to use on messages (defaults to no compression). Similar to `compression.codec` setting of the JVM producer.
 	FlushMsgCount     int                    // The number of messages needed to trigger a flush. This is a minimum, not an upper limit (use MaxMessagesPerReq for that).
-	FlushFrequency    time.Duration          // If this amount of time elapses without a flush, one will be queued. This is a minimum, not an upper limit.
-	FlushByteCount    int                    // If this many bytes of messages are accumulated, a flush will be triggered. This is a minimum, not an upper limit.
+	FlushFrequency    time.Duration          // If this amount of time elapses without a flush, one will be queued. This is a minimum, not an upper limit. Equivalent to `queue.buffering.max.ms` setting of JVM producer.
+	FlushByteCount    int                    // If this many bytes of messages are accumulated, a flush will be triggered. This is a minimum, not an upper limit. This should match the broker's `message.max.bytes` setting.
```
eapache: No, this has nothing to do with `message.max.bytes`, this is a best-effort target of bytes-per-request, whereas `message.max.bytes` (our `MaxMessageBytes`) is a hard limit on the size of individual messages

eapache:
Now that I've used the term "best-effort" for the `Flush*` values I'm liking that description. They are simply best-effort targets for request frequency and size, for tuning purposes. They are not limits (either maximum or minimum*).

* technically with the correct configuration you can have `FlushFrequency` act as a hard minimum, but that's an implementation side-effect and not really documentation-worthy

wvanbergen:
Cool, I'll update :+1:

wvanbergen:
Updated.

producer.go
```diff
@@ -13,19 +13,22 @@ func forceFlushThreshold() int {
...
+	Compression       CompressionCodec       // The type of compression to use on messages (defaults to no compression). Similar to `compression.codec` setting of the JVM producer.
+	FlushMsgCount     int                    // The number of messages needed to trigger a flush. This is a best effort; the number of messages may be more or less. Use `MaxMessagesPerReq` to set a hard upper limit.
+	FlushFrequency    time.Duration          // If this amount of time elapses without a flush, one will be queued. The frequency is a best effort, and the actual frequency can be more or less. Equivalent to `queue.buffering.max.ms` setting of JVM producer.
+	FlushByteCount    int                    // If this many bytes of messages are accumulated, a flush will be triggered. This is a best effort; the number of bytes may be more or less. Use `MaxMessageBytes` to set a hard upper limit.
```
eapache: No, `MaxMessageBytes` is per message, `FlushByteCount` is per request.

I suppose the global `MaxRequestSize` is what you'd use to set a hard upper limit, but that is a package-global, so...

producer.go
```diff
@@ -13,19 +13,22 @@ func forceFlushThreshold() int {
...
 	AckSuccesses      bool                   // If enabled, successfully delivered messages will be returned on the Successes channel.
-	MaxMessageBytes   int                    // The maximum permitted size of a message (defaults to 1000000)
-	MaxMessagesPerReq int                    // The maximum number of messages the producer will send in a single broker request. Defaults to 0 for unlimited. The global setting MaxRequestSize still applies.
-	ChannelBufferSize int                    // The size of the buffers of the channels between the different goroutines. Defaults to 0 (unbuffered).
-	RetryBackoff      time.Duration          // The amount of time to wait for the cluster to elect a new leader before processing retries. Defaults to 250ms.
+	MaxMessageBytes   int                    // The maximum permitted size of a message (defaults to 1000000). Equivalent to the JVM's producer `message.max.bytes`.
```
eapache: "JVM broker's" not "JVM's producer"

wvanbergen:
I suck at doing this. Updated again :)

eapache:
:shipit:

PR #279, opened 2015-02-18
Move retries to its own field and make MaxRetries configurable

eapache:
Doesn't ~~work~~ do anything really yet, but this is the first group of superficial changes necessary to support
multiple retries in the producer.

@Shopify/kafka


wvanbergen:
Looks good.

PR #281, opened 2015-02-18
Producer: back off after failing to find a leader

eapache:
Not just on flusher-triggered retries.

cc @Shopify/kafka


wvanbergen:
:sparkles: - this seems to have resolved the issue I was seeing in my test setup

PR #282, opened 2015-02-19
Use Metadata field in SimpleProducer

eapache:
Rather than storing a `map` from messages to expectations, store the
expectations in our handy new `Metadata` field. This is certainly simpler, and
is probably slightly more efficient.

Also call `AsyncClose` instead of `Close` when shutting down the underlying
Producer, to avoid the hypothetical race condition where `Close` swallows a
success or error and that call to `SendMessage` never returns.

@Shopify/kafka

### DIFF
--- a/simple_producer.go
+++ b/simple_producer.go
@@ -1,16 +1,13 @@
 package sarama
 
+import "sync"
+
 // SimpleProducer publishes Kafka messages. It routes messages to the correct broker, refreshing metadata as appropriate,
 // and parses responses for errors. You must call Close() on a producer to avoid leaks, it may not be garbage-collected automatically when
 // it passes out of scope (this is in addition to calling Close on the underlying client, which is still necessary).
 type SimpleProducer struct {
-	producer        *Producer
-	newExpectations chan *spExpect
-}
-
-type spExpect struct {
-	msg    *MessageToSend
-	result chan error
+	producer *Producer
+	wg       sync.WaitGroup
 }
 
 // NewSimpleProducer creates a new SimpleProducer using the given client  and configuration.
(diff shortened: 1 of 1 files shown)


simple_producer.go
```diff
@@ -26,47 +23,36 @@ func NewSimpleProducer(client *Client, config *ProducerConfig) (*SimpleProducer,
...
 
-func (sp *SimpleProducer) matchResponses() {
-	newExpectations := sp.newExpectations
-	unmatched := make(map[*MessageToSend]chan error)
-	unmatched[nil] = nil // prevent it from emptying entirely
+func (sp *SimpleProducer) handleSuccesses() {
+	defer sp.wg.Done()
+	for msg := range sp.producer.Successes() {
+		expectation := msg.Metadata.(chan error)
+		close(expectation)
```
eapache: food for thought: should this be `expectation <- nil` instead? That would be a little clearer and more consistent with the `handleErrors` case...?

simple_producer.go
```diff
@@ -26,47 +23,36 @@ func NewSimpleProducer(client *Client, config *ProducerConfig) (*SimpleProducer,
...
 
 	return sp, nil
 }
 
 // SendMessage produces a message to the given topic with the given key and value. To send strings as either key or value, see the StringEncoder type.
 func (sp *SimpleProducer) SendMessage(topic string, key, value Encoder) error {
-	msg := &MessageToSend{Topic: topic, Key: key, Value: value}
-	expectation := &spExpect{msg: msg, result: make(chan error)}
-	sp.newExpectations <- expectation
+	expectation := make(chan error)
```
eapache: a channel buffer size of one here wouldn't hurt, and would (in theory) be more efficient by letting the `handleSuccesses` and `handleErrors` functions loop a few times without context switching, but I'm wondering if that's too far into the land of micro-optimizations

simple_producer.go
```diff
@@ -26,47 +23,36 @@ func NewSimpleProducer(client *Client, config *ProducerConfig) (*SimpleProducer,
...
-			delete(unmatched, err.Msg)
-		case msg := <-sp.producer.Successes():
-			close(unmatched[msg])
-			delete(unmatched, msg)
-		}
+func (sp *SimpleProducer) handleErrors() {
+	defer sp.wg.Done()
+	for err := range sp.producer.Errors() {
+		expectation := err.Msg.Metadata.(chan error)
+		expectation <- err.Err
```
wvanbergen: should we close the expectation here as well? Or does the garbage collector do that for us eventually? If GC deals with the problem, I say we should do `expectation <- nil` in the successes case as well.

wvanbergen:
Didn't even realize we could use Metadata for this. so much nicer his way :+1:

simple_producer.go
eapache: gc will still collect open channels as long as they are no longer referenced

I'll update to do `<- nil` in the other case

PR #283, opened 2015-02-19
Permit starting a mockbroker on a specific address

eapache:
I've been playing with writing more resiliency tests with mockbrokers and the
only thing missing was the ability to shut down a broker and then bring it back
on the same address - this PR adds that capability.

@Shopify/kafka @Sirupsen


Sirupsen:
That's cool :+1:

wvanbergen:
:ship:

Issue #286, opened 2015-02-20
More efficient cachedPartitions ideas


cep21:
Hello,

I've written a golang kafka client using sarama (it sends a lot of kafka messages) and when I use pprof the following information shows up
```
(pprof) top100
670ms of 670ms total (  100%)
      flat  flat%   sum%        cum   cum%
     220ms 32.84% 32.84%      220ms 32.84%  github.com/Shopify/sarama.(*int32Slice).Less
      90ms 13.43% 46.27%      370ms 55.22%  sort.doPivot
      50ms  7.46% 53.73%       50ms  7.46%  github.com/Shopify/sarama.(*int32Slice).Swap
      40ms  5.97% 59.70%      520ms 77.61%  github.com/Shopify/sarama.(*Client).cachedPartitions
      40ms  5.97% 65.67%       60ms  8.96%  runtime.mapiternext
      40ms  5.97% 71.64%       50ms  7.46%  sort.insertionSort
      30ms  4.48% 76.12%       30ms  4.48%  runtime.writebarrierptr
      20ms  2.99% 79.10%       20ms  2.99%  runtime.MSpan_Sweep
      20ms  2.99% 82.09%       20ms  2.99%  runtime.cas
      10ms  1.49% 83.58%       10ms  1.49%  MCentral_Grow
      10ms  1.49% 85.07%       10ms  1.49%  encoding/json.cachedTypeFields
...
...
...
```

I appear to spend a lot of time sorting the partitions inside [cachedPartitions()](https://github.com/Shopify/sarama/blob/master/client.go#L425)  Does this need to happen so frequently?  Can this be cached per topic/partitionSet combination?  I'm not super familiar with kafka's protocol so I'm not sure if the solution is so simple.

Thoughts?

eapache:
Yes, this can be cached, it's just a little tricky and there are other performance bottlenecks that are more broadly applicable (e.g. #255). You must have an *enormous* number of partitions for this to be showing up so heavily.

cep21:
1024 partitions

eapache:
That's a lot :)

Patches welcome, if you want to take a crack. In short, you'll have to:
- add an additional map to use as a cache that works in tandem with `metadata` at client.go:67
- use and/or lazily fill the cache in `Client.cachedPartitions()` *being careful to respect the `partitionSet` parameter*
- clear the cache around client.go:586

Otherwise one of us will probably pick this up eventually, but it's not a high priority at the moment.

[...]

PR #287, opened 2015-02-20
Cache client partition results to save the repeated sort() calls

cep21:
For more information, see #286

### DIFF
--- a/client.go
+++ b/client.go
@@ -65,7 +65,11 @@ type Client struct {
 
 	brokers  map[int32]*Broker                       // maps broker ids to brokers
 	metadata map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
-	lock     sync.RWMutex                            // protects access to the maps, only one since they're always written together
+
+	// If the number of partitions is large, we can get some churn calling cachedPartitions,
+	// so the result is cached.  It is important to update this value whenever metadata is changed
+	cachedPartitionsResults map[string][maxPartitionIndex][]int32
+	lock                    sync.RWMutex // protects access to the maps, only one since they're always written together
 }
 
 // NewClient creates a new Client with the given client ID. It connects to one of the given broker addresses
@@ -87,14 +91,15 @@ func NewClient(id string, addrs []string, config *ClientConfig) (*Client, error)
 	}
 
 	client := &Client{
-		id:              id,
-		config:          *config,
-		closer:          make(chan none),
-		seedBrokerAddrs: addrs,
-		seedBroker:      NewBroker(addrs[0]),
-		deadBrokerAddrs: make(map[string]none),
-		brokers:         make(map[int32]*Broker),
-		metadata:        make(map[string]map[int32]*PartitionMetadata),
+		id:                      id,
+		config:                  *config,
+		closer:                  make(chan none),
+		seedBrokerAddrs:         addrs,
+		seedBroker:              NewBroker(addrs[0]),
+		deadBrokerAddrs:         make(map[string]none),
+		brokers:                 make(map[int32]*Broker),
+		metadata:                make(map[string]map[int32]*PartitionMetadata),
+		cachedPartitionsResults: make(map[string][maxPartitionIndex][]int32),
 	}
 	_ = client.seedBroker.Open(config.DefaultBrokerConf)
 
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -65,7 +65,11 @@ type Client struct {
 
 	brokers  map[int32]*Broker                       // maps broker ids to brokers
 	metadata map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
-	lock     sync.RWMutex                            // protects access to the maps, only one since they're always written together
+
+	// If the number of partitions is large, we can get some churn calling cachedPartitions,
+	// so the result is cached.  It is important to update this value whenever metadata is changed
+	cachedPartitionsResults map[string]map[int][]int32
```
eapache: Given there are only two possible values I'm not sure using a nested map is really necessary

client.go
```diff
@@ -425,11 +430,22 @@ func (client *Client) cachedMetadata(topic string, partitionID int32) *Partition
...
 	partitions := client.metadata[topic]
+
 	if partitions == nil {
 		return nil
 	}
+	// Important:  If refactoring this, please keep the client.lock.RLock() above
+	previousResultMap, resMapExists := client.cachedPartitionsResults[topic]
+	if resMapExists {
+		if previousResult, exists := previousResultMap[partitionSet]; exists {
+			return previousResult
```
eapache: This subtlety changes the behavior of this method by not returning a copy - if anyone has been relying on being able to edit the result their code will break.

I'm not sure if we need to care or not, but it's something to think about.

cep21:
I pushed a new change that uses an array rather than a map.  I think not returning a copy does not let people modify the result, but it also saves on memory allocations.  I think this behavior should be ok(?)  Not sure why someone would depend upon that.

client.go
```diff
@@ -65,7 +65,11 @@ type Client struct {
 
 	brokers  map[int32]*Broker                       // maps broker ids to brokers
 	metadata map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
-	lock     sync.RWMutex                            // protects access to the maps, only one since they're always written together
+
+	// If the number of partitions is large, we can get some churn calling cachedPartitions,
+	// so the result is cached.  It is important to update this value whenever metadata is changed
+	cachedPartitionsResults map[string][][]int32
```
eapache: I was thinking two separate maps (or even a fixed array `[2]int32`) but I guess a slice is just as good

client.go
```diff
@@ -422,11 +433,21 @@ func (client *Client) cachedMetadata(topic string, partitionID int32) *Partition
...
 }
 
-func (client *Client) cachedPartitions(topic string, partitionSet int) []int32 {
+func (client *Client) cachedPartitions(topic string, partitionSet partitionType) []int32 {
 	client.lock.RLock()
 	defer client.lock.RUnlock()
 
 	partitions := client.metadata[topic]
+
+	if partitions == nil {
```
eapache: It's a bit weird to check `metadata[topic]` for nil, and then unconditionally access `cachedPartitionsResults` (I realize it works because they are always set together, but it reads a little odd)

client.go
```diff
@@ -441,6 +462,7 @@ func (client *Client) cachedPartitions(topic string, partitionSet int) []int32 {
 
 	sort.Sort(int32Slice(ret))
 	return ret
+
```
eapache: nit: unnecessary blank line added

eapache:
> also saves on memory allocations

Yes it does, which is a good point.

> Not sure why someone would depend upon that.

Neither am I (we don't, as far as I am aware) but I've learned to expect to be surprised :)

---

I've mentioned a few style nits, but I don't think there's anything really wrong with this as it stands.

@Shopify/kafka for an additional review please?

client.go
wvanbergen: I'd prefer to use a static array here, because it makes it more clear that this is not a list but a tuple.

client.go
```diff
@@ -422,11 +433,21 @@ func (client *Client) cachedMetadata(topic string, partitionID int32) *Partition
...
 
 	partitions := client.metadata[topic]
+
+	if partitions == nil {
+		return nil
+	}
+	return client.cachedPartitionsResults[topic][partitionSet]
+}
+
+func (client *Client) cachedPartitionsNoLock(topic string, partitionSet partitionType) []int32 {
```
wvanbergen: I find this function name slightly confusing; based on the name I expect `cachedPartitions` to simply call grab the lock and call this function.

Maybe `setPartitionCache` or `updatePartitionCache`?

wvanbergen:
This looks good to me; just some comments about naming to make the intent clearer.

cep21:
I've pushed one more version based upon the comments.  If there are more thoughts on names, it's ok if you want to pull my branch and make them before pushing upstream.

wvanbergen:
This looks great now. Thanks for your contribution!

PR #288, opened 2015-02-21
A handful of cleanups to the broker configuration

eapache:
- drop default network timeouts from 60 to 30 seconds to match JVM-land
- document the default, and the JVM-land equivalent (part of #270)
- validate `DialTimeout`
- change the semantics of `MaxOpenRequests` to actually be the maximum number of
  open requests, as opposed to the size of the buffer directly (since an
  unbuffered channel still permits one open request at a time). This involves
  bumping the default by one, validating for >0 (not just >=0) and subtracting
  one to get the actual channel buffer size.

@Shopify/kafka

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -11,27 +11,33 @@ import (
 
 // BrokerConfig is used to pass multiple configuration options to Broker.Open.
 type BrokerConfig struct {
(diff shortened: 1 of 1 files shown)


wvanbergen:
:shipit:

PR #291, opened 2015-02-23
Update producer's MaxRetries and RetryBackoff default values

wvanbergen:
Matches the default values of the JVM producer. Also updates godoc.

@Shopify/kafka


eapache:
Frankly I think 100ms is too short, but I have no problem with blaming upstream if people complain :)

:shipit:

wvanbergen:
Let's make sure we override the default in our apps.

PR #292, opened 2015-02-23
Only log metadata retry refreshes when using a list of topics

wvanbergen:
Right now, our logs are full of "starting metadata refresh" lines due to the 10 minute background refresh interval we have in place. This makes the logs noisy.

I think we only request the metadata for all topics:
- during startup
- during background refresh.

I think we can live without the logs in these cases.

Not sure if this is a good idea. @Shopify/kafka, what do you think?


eapache:
I don't consider one log every ten minutes to be "noisy".

wvanbergen:
Closing this.

PR #297, opened 2015-02-23
Assert there are no outstanding successes in the producer tests

eapache:
@Shopify/kafka


wvanbergen:
LGTM. Are we still using `safeClose` elsewhere?

eapache:
`safeClose` takes an `io.Closer` and asserts that the result of `Close()` is nil, so it is still used for the consumer and the client tests.

PR #298, opened 2015-02-24
Improve expectation API to allow for testing of more intricate scenarios

wvanbergen:
This makes the API for mock broker expectations a bit more powerful. 

- Allow setting a latency for the response (this replaces the previous broker-wide latency setting)
- Allow setting a `Before` and `After` callback around the response.

Use these capabilities to add some tests:

- The first client tests asserts that if one broker times out, it will continue to use the second broker to discover metadata.
- The second client test will make sure that eventually `NewClient` will return an error if all brokers have high latency.
- Test the producer with multiple batches in flight at different retry levels.

@Shopify/kafka

### DIFF
--- a/mockbroker.go
+++ b/mockbroker.go
@@ -40,12 +40,15 @@ type MockCluster map[int32]*MockBroker
 
 type callback func()
 
+// BrokerExpectation allos you to specify how the respond to a request the MockBroker will receive.
+// See MockBroker's Expects method to add expectation to a mockbroker.
 type BrokerExpectation struct {
-	Response           encoder
-	Latency            time.Duration
-	BeforeCallback     callback
-	AfterCallback      callback
-	IgnoreServerErrors bool
+	Before   callback      // Before will be called before sending the response after a request has been received
+	Latency  time.Duration // Latency before the response will be sent
+	Response encoder       // Response holds what will be sent back to the client
+	After    callback      // After will be called after the response has been sent to the client.
+
+	IgnoreConnectionErrors bool // IgnoreConnectionErrors should be set to true if connectivity issues while receiving the request or sending the response are to be expected.
 }
 
 func (b *MockBroker) BrokerID() int32 {
@@ -83,18 +86,18 @@ func (b *MockBroker) serverLoop() (ok bool) {
 	for expectation := range b.expectations {
 		_, err = io.ReadFull(conn, reqHeader)
 		if err != nil {
-			return b.serverError(err, conn, expectation.IgnoreServerErrors)
+			return b.serverError(err, conn, expectation.IgnoreConnectionErrors)
 		}
 		body := make([]byte, binary.BigEndian.Uint32(reqHeader))
 		if len(body) < 10 {
 			return b.serverError(errors.New("Kafka request too short."), conn, false)
 		}
 		if _, err = io.ReadFull(conn, body); err != nil {
-			return b.serverError(err, conn, expectation.IgnoreServerErrors)
+			return b.serverError(err, conn, expectation.IgnoreConnectionErrors)
 		}
 
-		if expectation.BeforeCallback != nil {
-			expectation.BeforeCallback()
+		if expectation.Before != nil {
+			expectation.Before()
 		}
 
 		if expectation.Latency > 0 {
@@ -112,14 +115,14 @@ func (b *MockBroker) serverLoop() (ok bool) {
 		binary.BigEndian.PutUint32(resHeader, uint32(len(response)+4))
 		binary.BigEndian.PutUint32(resHeader[4:], binary.BigEndian.Uint32(body[4:]))
 		if _, err = conn.Write(resHeader); err != nil {
-			return b.serverError(err, conn, expectation.IgnoreServerErrors)
+			return b.serverError(err, conn, expectation.IgnoreConnectionErrors)
 		}
 		if _, err = conn.Write(response); err != nil {
-			return b.serverError(err, conn, expectation.IgnoreServerErrors)
+			return b.serverError(err, conn, expectation.IgnoreConnectionErrors)
 		}
 
-		if expectation.AfterCallback != nil {
-			expectation.AfterCallback()
+		if expectation.After != nil {
+			expectation.After()
 		}
 	}
 	if err = conn.Close(); err != nil {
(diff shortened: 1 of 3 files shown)


mockbroker.go
```diff
@@ -31,14 +31,18 @@ type MockBroker struct {
...
 	stopper      chan bool
-	expectations chan encoder
+	expectations chan *MockBrokerExpectation
 	listener     net.Listener
 	t            TestState
-	latency      time.Duration
 }
 
-func (b *MockBroker) SetLatency(latency time.Duration) {
-	b.latency = latency
```
wvanbergen: This can now be set separately for every expectation/response.

mockbroker.go
```diff
@@ -168,6 +180,23 @@ func NewMockBrokerAddr(t TestState, brokerID int32, addr string) *MockBroker {
...
+}
+
+func (e *MockBrokerExpectation) Before(callback callback) *MockBrokerExpectation {
+	e.beforeCallback = callback
+	return e
+}
+
+func (e *MockBrokerExpectation) After(callback callback) *MockBrokerExpectation {
+	e.afterCallback = callback
+	return e
```
wvanbergen: Not sure about these method names; suggestions welcome.

producer_test.go
```diff
@@ -559,6 +559,120 @@ func TestProducerMultipleRetries(t *testing.T) {
...
+	seedBroker.Returns(metadataLeader2).After(func() {
+		for i := 0; i < 10; i++ {
+			producer.Input() <- &MessageToSend{Topic: "my_topic", Key: nil, Value: StringEncoder(TestMessage), Metadata: "Batch 3"}
+		}
+	})
+
+	prodSuccess := new(ProduceResponse)
+	prodSuccess.AddTopicPartition("my_topic", 0, NoError)
+	leader2.Returns(prodSuccess) // batch 1
+	leader2.Returns(prodSuccess) // batch 2 & batch 3
```
wvanbergen: I was expecting to have to do a separate ack for batch 2 and 3, but I guess the retry logic combines those messages in to a single flush?

producer_test.go
```diff
@@ -559,6 +559,120 @@ func TestProducerMultipleRetries(t *testing.T) {
...
+			}
+		case msg := <-producer.Successes():
+			if msg.flags != 0 {
+				t.Error("Message had flags set")
+			}
+			if msg.Metadata.(string) != "Batch 3" {
+				t.Error("Expected batch 3")
+			}
+		}
+	}
```
wvanbergen: I should probably extract this out to a helper method :)

mockbroker.go
```diff
@@ -168,6 +180,23 @@ func NewMockBrokerAddr(t TestState, brokerID int32, addr string) *MockBroker {
 	return broker
 }
 
-func (b *MockBroker) Returns(e encoder) {
-	b.expectations <- e
+func (b *MockBroker) Returns(response encoder) *MockBrokerExpectation {
+	expectation := &MockBrokerExpectation{response: response}
+	b.expectations <- expectation
+	return expectation
```
wvanbergen: This now returns itself so the methods below can be chained.

Technically this is a bit shitty, because we immediately add it to the expectations channel, but allow it to be altered later. I guess this could cause concurrency issues, but because of the synchronous way our tests are set up I don't think this will cause issues.

wvanbergen: Blegh, the race detector is on to me. I guess this needs fixing.

wvanbergen:
CI is failing due to https://github.com/kisielk/errcheck/issues/62

wvanbergen:
This is failing on Travis, even though this works fine locally:

```
--- FAIL: TestSlowCluster (0.12 seconds)
	allocs.go:1: write tcp 127.0.0.1:46446: broken pipe
	allocs.go:1: write tcp 127.0.0.1:47507: broken pipe
	allocs.go:1: write tcp 127.0.0.1:55751: broken pipe
```

any ideas?

wvanbergen:
I assume this is caused by the mock broker trying to write a response during/after the reader timeout of the client.

wvanbergen:
Miraculously, it works by adding a parameter on whether to expect errors to the expectation.

wvanbergen:
Hmmm, in some cases is the read timeout appears to be triggered before the mock broker has figured out a request is incoming, and so it won't resolve the expectation. Any ideas of making this more robust, compared to simply raising the read timeout value?

mockbroker.go
```diff
@@ -31,14 +31,21 @@ type MockBroker struct {
...
-	expectations chan encoder
+	expectations chan *BrokerExpectation
 	listener     net.Listener
 	t            TestState
-	latency      time.Duration
 }
 
-func (b *MockBroker) SetLatency(latency time.Duration) {
-	b.latency = latency
+type MockCluster map[int32]*MockBroker
```
eapache: why not just a slice?

wvanbergen: To make sure the index matches the BrokerID. Broker IDs are 1-indexed, while slice indices start at 0.

producer_test.go
```diff
@@ -559,6 +559,127 @@ func TestProducerMultipleRetries(t *testing.T) {
...
+
+	metadataLeader2 := new(MetadataResponse)
+	metadataLeader2.AddBroker(leader2.Addr(), leader2.BrokerID())
+	metadataLeader2.AddTopicPartition("my_topic", 0, leader2.BrokerID(), nil, nil, NoError)
+	seedBroker.Returns(metadataLeader2)
+	leader2.Returns(prodNotLeader)
+	seedBroker.Returns(metadataLeader1)
+	leader1.Expects(&BrokerExpectation{
+		Response: prodNotLeader,
+		BeforeCallback: func() {
```
eapache: I'd put the "before" before the "response" for readability (since that's the order they actually happen in)

eapache: also, I believe this is so you get messages going into the producer while the flusher is stuck, which is cool, but not obvious

producer_test.go
eapache: this doesn't do what you think it does. specifically, I don't think the before/after hooks give you enough control over the internal concurrency of the producer to actually test what you want to test

producer_test.go
eapache: DRYing up the tests in general (there are other things too) is added to the testing TODO list for diederik and yagnik to look at

eapache:
This will not be merged in its current form - #312 should be fixed first.

PR #299, opened 2015-02-24
Error variables

wvanbergen:
- Prefix error variables (including `KError` instances) with `Err` (golint suggestion)
- Change `(Packet)EncodingError` to be a type like `(Packet)DecodingError`.

@Shopify/kafka

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -5,38 +5,39 @@ import (
 	"fmt"
 )
 
-// OutOfBrokers is the error returned when the client has run out of brokers to talk to because all of them errored
+// ErrOutOfBrokers is the error returned when the client has run out of brokers to talk to because all of them errored
 // or otherwise failed to respond.
-var OutOfBrokers = errors.New("kafka: Client has run out of available brokers to talk to. Is your cluster reachable?")
+var ErrOutOfBrokers = errors.New("kafka: Client has run out of available brokers to talk to. Is your cluster reachable?")
 
-// ClosedClient is the error returned when a method is called on a client that has been closed.
-var ClosedClient = errors.New("kafka: Tried to use a client that was closed")
+// ErrClosedClient is the error returned when a method is called on a client that has been closed.
+var ErrClosedClient = errors.New("kafka: Tried to use a client that was closed")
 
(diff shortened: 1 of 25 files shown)


errors.go
```diff
@@ -65,62 +71,62 @@ type KError int16
...
-	RequestTimedOut                 KError = 7
-	BrokerNotAvailable              KError = 8
-	ReplicaNotAvailable             KError = 9
-	MessageSizeTooLarge             KError = 10
-	StaleControllerEpochCode        KError = 11
-	OffsetMetadataTooLarge          KError = 12
-	OffsetsLoadInProgress           KError = 14
-	ConsumerCoordinatorNotAvailable KError = 15
-	NotCoordinatorForConsumer       KError = 16
+	ErrNoError                         KError = 0
```
eapache: ErrNoError makes me sad :(

sarama.go
```diff
@@ -28,13 +28,13 @@ type StdLogger interface {
 var PanicHandler func(interface{})
 
 // MaxRequestSize is the maximum size (in bytes) of any request that Sarama will attempt to send. Trying
-// to send a request larger than this will result in an EncodingError. The default of 100 MiB is aligned
+// to send a request larger than this will result in an ErrPacketEncodingFailure. The default of 100 MiB is aligned
```
eapache: s/Failure/Error/ ?

errors.go
wvanbergen: Is this one ever exposed outside of Sarama?

eapache: It's part of the kafka spec so anyone using the low-level request/response code will need it.

eapache: I think it's just one of those things where kafka's protocol decision is inherently non-idiomatic in go just because kafka was written in scala and not go. In go you'd use a nil `error` interface for no error, but since java is not go you end up using a typed int of 0. Since in Go you have to type the int, you end up with the 0 case still technically being an error value even though it represents no error.

eapache:
One comment typo, otherwise LGTM

eapache:
:+1:

PR #303, opened 2015-02-24
Split consumer’s Events() channel into Messages() and Errors()

wvanbergen:
This is more in line with the producer. Also, it is really easy to forget checking for the `Err` field for every `ConsumerEvent`, which would lead to bugs.

@Shopify/kafka

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -278,6 +278,7 @@ func (c *Consumer) unrefBrokerConsumer(broker *Broker) {
 // PartitionConsumer processes Kafka messages from a given topic and partition. You MUST call Close()
 // on a consumer to avoid leaks, it will not be garbage-collected automatically when it passes out of
 // scope (this is in addition to calling Close on the underlying consumer's client, which is still necessary).
+// You have to read from both the Messages and Errors channels to prevent the consumer from locking eventually.
 type PartitionConsumer struct {
 	consumer  *Consumer
 	config    PartitionConsumerConfig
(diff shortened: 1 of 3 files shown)


consumer.go
```diff
@@ -94,30 +94,39 @@ func (config *PartitionConsumerConfig) Validate() error {
...
-		return ConfigurationError("Invalid EventBufferSize")
+	if config.ChannelBufferSize < 0 {
+		return ConfigurationError("Invalid ChannelBufferSize")
 	}
 
 	return nil
 }
 
-// ConsumerEvent is what is provided to the user when an event occurs. It is either an error (in which case Err is non-nil) or
+// ConsumerMessage is what is provided to the user when an event occurs. It is either an error (in which case Err is non-nil) or
```
eapache: the remainder of this godoc is now out-of-date

consumer.go
```diff
@@ -94,30 +94,39 @@ func (config *PartitionConsumerConfig) Validate() error {
...
+
+func (ce ConsumerError) Error() string {
+	return fmt.Sprintf("kafka: error while consuming %s/%d: %s", ce.Topic, ce.Partition, ce.Err)
+}
+
+// ConsumeErrors is a type that wraps a batch of errors and implements the Error interface.
 // It can be returned from the PartitionConsumer's Close methods to avoid the need to manually drain errors
 // when stopping.
-type ConsumeErrors []*ConsumerEvent
+type ConsumeErrors []error
```
eapache: why `error` and not `ConsumerError`?

consumer.go
```diff
@@ -376,11 +393,13 @@ func (child *PartitionConsumer) Close() error {
 	// close itself)
 	close(child.dying)
 
+	for _ = range child.messages {
```
eapache: as in the producer's shutdown draining of successes, I believe this must be in a goroutine to prevent deadlock

consumer.go
```diff
@@ -376,11 +393,13 @@ func (child *PartitionConsumer) Close() error {
 	// close itself)
 	close(child.dying)
```
eapache: for more parallels with the producer, we should move this to an `AsyncClose` method

eapache:
a few little things (and one suggestion of an additional enhancement) but nothing major

consumer.go
wvanbergen: Do you remember why that was?

wvanbergen:
Should I guard multiple calls to `AsyncClose()` with a `sync.Once` to prevent panics?

consumer.go
eapache: If the user calls close, then the internal goroutines produce a flood of errors, this will hang waiting for success channel to close, and internal will hang trying to write errors

eapache:
Nope, I think that's overly-defensive

consumer.go
```diff
@@ -94,30 +94,40 @@ func (config *PartitionConsumerConfig) Validate() error {
...
-// ConsumeErrors is a type that wraps a batch of "ConsumerEvent"s and implements the Error interface.
+// ConsumerError is what is provided to the user when an error occurs.
+// It wraps an error and includes the topic and partition.
+type ConsumerError struct {
+	Topic     string
+	Partition int32
+	Err       error
+}
+
+func (ce ConsumerError) Error() string {
```
eapache: I'm not 100% convinced this needs to implement `error` itself, given we never (ourselves) actually stuff it into an `error` type, but I guess it doesn't hurt.

wvanbergen:
e.g., to ensure something like this doesn't cause issues:

```
	wg.Add(1)
	go func() {
		defer wg.Done()
		for err := range consumer.Errors() {
			consumer.AsyncClose()
		}
	}()
```

eapache:
I'd consider that just bad code o.O

consumer.go
wvanbergen: The nice part is that it will add some context to the error message, e.g. when printing the errors that you consume from the `Errors()` channel.

        for err := range consumer.Errors() {
            fmt.Printf(err)
        }

eapache:
Pretty sure it's not legit to call `Close` more than once on any of the standard library IO constructs.

wvanbergen:
I guess you can just break the for loop, and call `(Async)Close` outside the loop. It may leave some errors undrained but that is not really an issue.

eapache:
You can just call `Close` outside the loop and it will drain/return any remaining errors.

wvanbergen:
OK, I think I addressed all the outstanding issues.

eapache:
LGTM

PR #306, opened 2015-02-25
Make Broker.Open truly non-blocking

eapache:
The intent behind the design was so that the method was always safe to call e.g.
when a lock was held, by making it async, however there were still certain cases
where it could have blocked:
 - if another thread had already called Open, subsequent calls would block until
   that network io completed
 - if the connection was already open *and* the pipeline was full
   (MaxOpenRequests were in flight) then calling Open would block until at least
   one response was received

This adds a simple atomic flag we can use to short-circuit appropriately, such
that Open should now have the following properties
 - it *never* blocks on network IO
 - it guarantees that the broker is connected or connecting when it returns
   (modulo another thread calling Close in the interim)

AFAICT these are the properties it needs for lazy broker connection to be
possible in the client.

@Shopify/kafka

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -61,6 +62,7 @@ type Broker struct {
 	conn          net.Conn
 	connErr       error
 	lock          sync.Mutex
+	opened        int32
 
 	responses chan responsePromise
 	done      chan bool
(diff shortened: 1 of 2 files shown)


broker.go
```diff
@@ -156,6 +169,8 @@ func (b *Broker) Close() (err error) {
 	b.done = nil
 	b.responses = nil
 
+	atomic.StoreInt32(&b.opened, 0)
+
```
wvanbergen: Can we short-circuit `Close()` as well using this flag, instead of doing the `if b.conn == nil return ErrNotConnected` check (see above)? Not sure if this is useful.

wvanbergen:
Is the change from "connected" to "connecting" significant? E.g., would we ever check for that error value and start using the broker if it says it's already connected?

broker.go
eapache: We can't *instead of*, but we could do *as well as*, if we wanted to avoid taking the mutex here - I'm not sure that it's all that useful though?

eapache:
I'm not sure what you mean, the change where?

wvanbergen:
> `// ErrAlreadyConnected is the error returned when calling Open() on a Broker that is already connected or connecting.`

wvanbergen:
I guess previously this already was the case, but now the godoc actually reflects that. Should we update the string of the error value as well?

eapache:
Yes, it has always been the case that `Open` only guaranteed connecting, not connected. I've updated the error string to match

wvanbergen:
:checkered_flag:

eapache:
I just vastly simplified the atomic op in `Open`, because the previous version was rather stupid.

@wvanbergen take another look at that logic please?

wvanbergen:
Ha, so much better :+1:

PR #307, opened 2015-02-25
Unexport mockbroker.

wvanbergen:
Don't export MockBroker; it's internal only. This allows us to break its API without having to do a major release.

@Shopify/kafka

### DIFF
--- a/broker_test.go
+++ b/broker_test.go
@@ -51,7 +51,7 @@ func TestBrokerAccessors(t *testing.T) {
 }
 
 func TestSimpleBrokerCommunication(t *testing.T) {
-	mb := NewMockBroker(t, 0)
+	mb := newMockBroker(t, 0)
 	defer mb.Close()
 
 	broker := NewBroker(mb.Addr())
--- a/client_test.go
+++ b/client_test.go
@@ -20,7 +20,7 @@ func TestDefaultClientConfigValidates(t *testing.T) {
 }
 
 func TestSimpleClient(t *testing.T) {
-	seedBroker := NewMockBroker(t, 1)
+	seedBroker := newMockBroker(t, 1)
 
 	seedBroker.Returns(new(MetadataResponse))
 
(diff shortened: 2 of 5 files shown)


eapache:
So for reference, I added this in its own package during my initial aborted attempt to split Sarama into subpackages.

**Commit:** 46618879c6606202b910c6bded7fbbcd2be1e091
**Commit Message:** "Apparently even public test functions can't be imported by other packages, so I have to move this to its own one to be able to use it from both protocol and kafka."
**Comment from that commit:** "It exists solely for testing other parts of the Sarama stack. It is in its own package so that it can be imported by tests in multiple different packages."

eapache:
Which is to say: old me was silly. :shipit:

schleppy:
For what it's worth, the exported MockBroker was serving a purpose in our testing.  We have a library, when used in conjunction with the `Metadata` field on `ProducerMessage`, allows us to attempt a message N times * `MaxRetries` before deciding the message isn't going to get stored.  When we make that decision we have a configurable "Plan" for what to do with the message (store it elsewhere, log to disk, etc.)  The exported MockBroker allowed us to test that functionality, but with this change, the only option was to change it back to exported.

Any chance of changing this back on your end?

eapache:
I'd be ok to export it in a separate package, but it really shouldn't be a part of Sarama's "proper" API since it is testing-only. It only makes use of a single other private API at the moment (the `encoder` interface) so it will have to be tweaked slightly to remove that dependency since it will no longer be able to use it from within a separate package.

eapache:
Opened #312 to track that.

Tangentially: why do you need to retry a message N times * `MaxRetries`? Now that `MaxRetries` is arbitrarily configurable, you should just be able to set it to the desired value in the first place.

wvanbergen:
In one of our projects, we have a Mock producer that makes it really easy
to test scenarios like this. When I get back from vacation, I will work on
publishing that piece.

schleppy:
@eapache The reason we don't just set `MaxRetries` to our target retry number is because we implement a back-off strategy based on the number of times we have attempted to send the message.  Additionally we want to track how many times we have sent a message through, attempted `MaxRetries` times, and still failed - so the stats are collected at that time.

eapache:
Ah ok. A configurable back-off strategy is probably something that could be usefully included in Sarama proper with a bit of work. I suppose we could also expose how many times a message was retried internally, I'm not sure if that makes sense or not.

PR #309, opened 2015-02-26
Lazily connect to brokers in the client

eapache:
Instead of opening a connection to all brokers immediately upon receiving their
information in metadata, wait until we are asked for them either via a call to
`Leader` or a call to `any`.

It seems simple enough (simpler than I was expecting actually) but I'm a bit wary, since it changes the internal semantics so much. Strong opinions and careful review please.

@Shopify/kafka

### DIFF
--- a/client.go
+++ b/client.go
@@ -337,10 +337,12 @@ func (client *Client) any() *Broker {
 	defer client.lock.RUnlock()
 
 	if client.seedBroker != nil {
+		_ = client.seedBroker.Open(client.conf)
 		return client.seedBroker
 	}
 
 	for _, broker := range client.brokers {
+		_ = broker.Open(client.conf)
 		return broker
 	}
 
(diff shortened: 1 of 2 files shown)


wvanbergen:
This looks innocent enough :+1: 

Should we do some manual testing?

eapache:
Ya, I'm really suspicious mostly *because* it looks so innocent. Throw whatever cases you can think of at it :)

yagnik:
This code looks good, what do you guys think about genghis like stress testing tool for kafka, we could possibly just leverage genghis ?

wvanbergen:
King of unrelated to this PR, but we need:

- testing tool for just sarama (see https://github.com/wvanbergen/kafka/tree/master/tools/stressproducer for an initial attempt).
- testing tool for our complete pipeline.

Let's discuss this offline.

eapache:
I ran this code (in combination with #338) through a bunch of failure scenarios using the vagrant cluster and willem's stressproducer and didn't have any problems.

I'll merge this and rebase #338 and then it can be reviewed separately.

PR #310, opened 2015-02-26
Rename SimpleProducer to SyncProducer

eapache:
Return the partition and offset of successful messages to the user as well.

As suggested by Willem in
https://github.com/Shopify/sarama/issues/302#issuecomment-76159600

@Shopify/kafka


wvanbergen:
LGTM

Issue #308, opened 2015-02-26
Single configuration type?


wvanbergen:
After a Twitter conversation: instead of having several different configuration types, we could consolidate them in one.

If we would have a single config struct, The broker/producer/consumer can just grab the relevant settings from it  and ignore the rest. This will make configuration a thing you only have to do once as user, and be done with it. 

``` go
config := sarama.NewClientConfig() // NewConfig? NewKafkaConfig?
config.MaxProduceRetryAttempts = 3
// ...

client := sarama.NewClient(..., config)

// By default, it will grab the config from the client
producer := sarama.NewProducer(client) 

// If you really want to create different producers on the same client with different configs, 
// we could provide an extra method to make that possible:
alternateProducer := sarama.NewProducerWithConfig(client, alternateConfig) 
```

The notable exception here is the offset value setting for the PartitionConsumer, but I feel that should be a function argument instead of a config setting anyway.

@Shopify/kafka Thoughts?

wvanbergen:
@martink @snormore as consumers of the API your input is welcome too!

eapache:
Alternatively, if we do this, in the common case there's no need to expose the client at all:
```go
config := sarama.NewConfig()

// creates an internal client instance, we could expose
// NewProducerFromClient(client, config) if we really wanted to
producer := sarama.NewProducer(config)
```

eapache:
I do like a single `NewConfig()` with a master type, but I also would like to keep the namespacing of the various values, so

```go
config := sarama.NewConfig()
config.Broker.ReadTimeout = 10 * time.Second
config.Producer.RequiredAcks = sarama.WaitForAll
```

etc

wvanbergen:
:+1: to namespacing.

eapache:
The thing I am fighting with now is whether we do

```go
struct Config {
    Client struct {
        // fields
    }
}
```

and make all the subtypes anonymous, or if we leave them as separate named types (like now) and just wrap them in a single master struct. Anonymous feels cleaner, but:
- it means that e.g. the Broker has to carry a *full* config type, and must reference *all* its variables with the full namespace (`b.config.Broker.Foo` instead of `b.config.Foo`)
- it means that we can't construct our default using a literal, and have to assign all the fields manually

karlseguin:
I think this is headed in the right direction. To document my original view:

[...]

PR #311, opened 2015-02-26
Unify config

eapache:
Unify configuration structures into `config.go` per discussion on #308.

Also take the opportunity to rename and reorganize some values, and to document
some equivalents with the JVM version, finishing up #270.

Open questions:
- I don't know what to do about the `addrs` parameter to `NewClient`. Karl suggested it should be moved to the constructor of the config object, but in hindsight that doesn't make sense because if you're constructing a config object *for* a broker, then you should not have to specify multiple addresses at all. It really should be a parameter of the constructor I think.
- Given that there are non-config parameters to the client constructor, how do we go about "hiding" the client from users of the producer/consumer? Just pass those parameters through?

I'm thinking currently we want to end up with:

```go
func NewProducer(addrs []string, conf *Config) {}
func NewConsumer(addrs []string, conf *Config) {}
func NewClient(addrs []string, conf *Config) {}
func NewProducerFromClient(client *Client) {}
func NewConsumerFromClient(client *Client) {}
```

Thoughts?

@Shopify/kafka @karlseguin

### DIFF
--- a/config.go
+++ b/config.go
@@ -85,6 +85,9 @@ type Config struct {
 		MaxWaitTime time.Duration
 	}
 
+	// A user-provided string sent with every request to the brokers for logging, debugging, and auditing purposes.
+	// Defaults to "sarama", but you should probably set it to something specific to your application.
+	ClientID string
 	// The number of events to buffer in internal and external channels. This permits the producer and consumer to
 	// continue processing some messages in the background while user code is working, greatly improving throughput.
 	// Defaults to 256.
@@ -141,6 +144,9 @@ func (c *Config) Validate() error {
 	}
 	if c.Consumer.MaxWaitTime%time.Millisecond != 0 {
 		Logger.Println("Consumer.MaxWaitTime only supports millisecond precision; nanoseconds will be truncated.")
+	}
+	if c.ClientID == "sarama" {
+		Logger.Println("ClientID is the default of 'sarama', you should consider setting it to something application-specific.")
 	}
 
 	// validate Net values
--- a/config.go
+++ b/config.go
@@ -1,0 +1,213 @@
+package sarama
+
+import "time"
+
+// Config is used to pass multiple configuration options to Sarama's constructors.
+type Config struct {
+	// Net is the namespace for network-level properties used by the Broker, and shared by the Client/Producer/Consumer.
+	Net struct {
+		MaxOpenRequests int // How many outstanding requests a connection is allowed to have before sending on it blocks (default 5).
+
+		// All three of the below configurations are similar to the `socket.timeout.ms` setting in JVM kafka.
+		DialTimeout  time.Duration // How long to wait for the initial connection to succeed before timing out and returning an error (default 30s).
+		ReadTimeout  time.Duration // How long to wait for a response before timing out and returning an error (default 30s).
+		WriteTimeout time.Duration // How long to wait for a transmit to succeed before timing out and returning an error (default 30s).
+	}
+
+	// Metadata is the namespace for metadata management properties used by the Client, and shared by the Producer/Consumer.
+	Metadata struct {
+		Retries         int           // How many times to retry a metadata request when a partition is in the middle of leader election (default 3).
+		WaitForElection time.Duration // How long to wait for leader election to finish between retries (default 250ms). Similar to `retry.backoff.ms` in the JVM version.
+		// How frequently to refresh the cluster metadata in the background. Defaults to 10 minutes.
+		// Set to 0 to disable. Similar to `topic.metadata.refresh.interval.ms` in the JVM version.
+		RefreshFrequency time.Duration
+	}
+
+	// Producer is the namespace for configuration related to producing messages, used by the Producer.
+	Producer struct {
+		// The maximum permitted size of a message (defaults to 1000000). Equivalent to the broker's `message.max.bytes`.
+		MaxMessageBytes int
+		// The level of acknowledgement reliability needed from the broker (defaults to WaitForLocal).
+		// Equivalent to the `request.required.acks` setting of the JVM producer.
+		RequiredAcks RequiredAcks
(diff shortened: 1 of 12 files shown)


eapache:
Alternatively, we could move addrs to the config struct, leave it out of the constructor, and simply require users to create a config struct and set that value - but I think that's not very obvious from a first-use perspective.

wvanbergen:
For ConsumePartition, I still feel we should combine offset method and value into one parameter. The only other method I can imagine supporting is to load it from Kafka's offset management system, and we can easily create a constant with a magic value for that. (Similar to what oldest and newest already are. )

wvanbergen:
With regards to the addrs parameter, I have mixed feelings. On the one hand it's a required parameter, but I can imagine supporting a zookeeper connection string as well, and discover the cluster from Zk.

wvanbergen:
The method names to instantiate objects look good to me.

wvanbergen:
One final naming nitpick you are free to ignore: I prefer 'Consumer' over 'Consume'

wvanbergen:
Agreed. We can always add separate functions if we want to support
zookeeper.

eapache:
I think I've addressed all the comments now - a surprisingly large diff, but mostly trivial renames.

Ready for review.

karlseguin:
From an API point of view, it resolves the original concern. I realize it might seem superficial once you're used to it, especially since it really only deals with the first few lines needed to initialize stuff, but I think anyone new to the library will appreciate the streamlining.

:+1:

wvanbergen:
Final note: should we move the maxrequestsize global to the config struct as well?

eapache:
> should we move the maxrequestsize global to the config struct as well?

No, it is used in the protocol-level encoding/decoding which doesn't necessarily have access to the config struct at all.

eapache:
@schleppy it occurs to me that you might have an opinion on this as well.

config.go
```diff
@@ -0,0 +1,219 @@
...
+			Min int32
+			// The normal maximum amount of data to fetch from the broker in each request (default 32768 bytes). This should be larger than
+			// most messages, or else the consumer will spend a lot of time negotiating sizes and not actually consuming. Similar to the JVM's
+			// `fetch.message.max.bytes`.
+			Default int32
+			// The maximum amount of data to fetch from the broker in a single request. Messages larger than this will return ErrMessageTooLarge.
+			// Defaults to 0 (no limit). Similar to the JVM's `fetch.message.max.bytes`. The global `sarama.MaxResponseSize` still applies.
+			Max int32
+		}
+		// The maximum amount of time the broker will wait for MinFetchSize bytes to become available before it
```
wvanbergen: Outdated: `MinFetchSize`

config.go
```diff
@@ -0,0 +1,219 @@
...
+		// All three of the below configurations are similar to the `socket.timeout.ms` setting in JVM kafka.
+		DialTimeout  time.Duration // How long to wait for the initial connection to succeed before timing out and returning an error (default 30s).
+		ReadTimeout  time.Duration // How long to wait for a response before timing out and returning an error (default 30s).
+		WriteTimeout time.Duration // How long to wait for a transmit to succeed before timing out and returning an error (default 30s).
+	}
+
+	// Metadata is the namespace for metadata management properties used by the Client, and shared by the Producer/Consumer.
+	Metadata struct {
+		Retries         int           // How many times to retry a metadata request when a partition is in the middle of leader election (default 3).
+		WaitForElection time.Duration // How long to wait for leader election to finish between retries (default 250ms). Similar to `retry.backoff.ms` in the JVM version.
```
wvanbergen: Naming consistency: `Retry.Max` & `Retry.Backoff`, similar to producer?

config.go
```diff
@@ -0,0 +1,219 @@
...
+			// The default is 1, as 0 causes the consumer to spin when no messages are available. Equivalent to the JVM's `fetch.min.bytes`.
+			Min int32
+			// The normal maximum amount of data to fetch from the broker in each request (default 32768 bytes). This should be larger than
+			// most messages, or else the consumer will spend a lot of time negotiating sizes and not actually consuming. Similar to the JVM's
+			// `fetch.message.max.bytes`.
+			Default int32
+			// The maximum amount of data to fetch from the broker in a single request. Messages larger than this will return ErrMessageTooLarge.
+			// Defaults to 0 (no limit). Similar to the JVM's `fetch.message.max.bytes`. The global `sarama.MaxResponseSize` still applies.
+			Max int32
+		}
```
wvanbergen: I am on the fence on this one, but: `Bytes` is absent from these names. Might be clearer to include this in the name to tell people its bytes; not messages.

config.go
```diff
@@ -0,0 +1,219 @@
...
+		Retries         int           // How many times to retry a metadata request when a partition is in the middle of leader election (default 3).
+		WaitForElection time.Duration // How long to wait for leader election to finish between retries (default 250ms). Similar to `retry.backoff.ms` in the JVM version.
+		// How frequently to refresh the cluster metadata in the background. Defaults to 10 minutes.
+		// Set to 0 to disable. Similar to `topic.metadata.refresh.interval.ms` in the JVM version.
+		RefreshFrequency time.Duration
+	}
+
+	// Producer is the namespace for configuration related to producing messages, used by the Producer.
+	Producer struct {
+		// The maximum permitted size of a message (defaults to 1000000). Equivalent to the broker's `message.max.bytes`.
```
wvanbergen: "Equivalent to" -> "Should be set equal or smaller than"

wvanbergen:
Minor naming/doc nitpicks; I think we are getting close to :checkered_flag:

config.go
eapache: I am also on the fence about this one, I've considering `Consumer.FetchBytes` as well but that also doesn't sit 100% nicely.

eapache: I think I will leave the names as-is but rework the doc to make it super-obvious they are bytes.

eapache:
I believe I've addressed the comments

wvanbergen:
:ship::it:

PR #313, opened 2015-03-04
Update the KError enum for Kafka 0.8.2.0

vrischmann:
Hi,

the KError enum is not complete for Kafka 0.8.2.0. This PR adds the missing errors.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -89,6 +89,10 @@ const (
 	ErrOffsetsLoadInProgress           KError = 14
 	ErrConsumerCoordinatorNotAvailable KError = 15
 	ErrNotCoordinatorForConsumer       KError = 16
+	ErrInvalidTopic                    KError = 17
+	ErrMessageSetSizeTooLarge          KError = 18
+	ErrNotEnoughReplicas               KError = 19
+	ErrNotEnoughReplicasAfterAppend    KError = 20
 )
 
 func (err KError) Error() string {
@@ -129,6 +133,14 @@ func (err KError) Error() string {
 		return "kafka server: Offset's topic has not yet been created."
 	case ErrNotCoordinatorForConsumer:
 		return "kafka server: Request was for a consumer group that is not coordinated by this broker."
+	case ErrInvalidTopic:
+		return "kafka server: The request attempted to perform an operation on an invalid topic."
+	case ErrMessageSetSizeTooLarge:
+		return "kafka server: The request included message batch larger than the configured segment size on the server."
+	case ErrNotEnoughReplicas:
+		return "kafka server: Messages are rejected since there are fewer in-sync replicas than required."
+	case ErrNotEnoughReplicasAfterAppend:
+		return "kafka server: Messages are written to the log, but to fewer in-sync replicas than required."
 	}
 
 	return fmt.Sprintf("Unknown error, how did this happen? Error code = %d", err)


wvanbergen:
Do we need to hands some of these errors in the producer, e.g to trigger a retry?

wvanbergen:
@eapache

eapache:
- Please rebase on the latest master to fix the tests, it appears `errcheck` has recently become more strict.
- It is currently offline due to technical problems, but my usual source for these errors has been the specification wiki at [1]. Last I checked, these errors were not listed, do you know why?

[1] https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol

vrischmann:
Just rebased from master.

You're right, I also checked that page to see why we got a error code 17, and it's not listed there, I guess it's not up to date.

I got the errors from [here](https://github.com/apache/kafka/blob/0.8.2/core/src/main/scala/kafka/common/ErrorMapping.scala) and the messages from [here](https://github.com/apache/kafka/blob/0.8.2/clients/src/main/java/org/apache/kafka/common/protocol/Errors.java)

eapache:
I did some diving into the other kafka docs/code to figure out these new errors, and sent a message to the kafka-users list for additional clarification: https://mail-archives.apache.org/mod_mbox/kafka-users/201503.mbox/%3CCAA4pprA%2BCPnn2%2B3%2B18vX1-6JpAzpCOPZZETgkquH0%2B4DfcKtmw%40mail.gmail.com%3E

I will comment inline on this PR as I figure out what each error actually means, and how we should handle them.

errors.go
```diff
@@ -89,6 +89,10 @@ const (
 	ErrOffsetsLoadInProgress           KError = 14
 	ErrConsumerCoordinatorNotAvailable KError = 15
 	ErrNotCoordinatorForConsumer       KError = 16
+	ErrInvalidTopic                    KError = 17
```
eapache: Invalid topics are e.g. blank, or contain invalid characters (presumably `/` and others). Looks like this might finally have fixed https://github.com/Shopify/sarama/blob/351e68b0267f6421505349526c32c8cd259acf53/client.go#L518-L525

errors.go
```diff
@@ -89,6 +89,10 @@ const (
...
+	ErrMessageSetSizeTooLarge          KError = 18
```
eapache: You can't produce a message-set larger than your segment size. However the (default, rarely-changed) segment size is 1GB and the (default, rarely-changed) max request size is 100MB so this isn't ever a problem unless you've set up your cluster very oddly. I don't feel the need to add specific handling of this unless or until somebody complains that the configurable `MaxRequestSize` we already provide is not enough.

errors.go
```diff
@@ -89,6 +89,10 @@ const (
...
+	ErrNotEnoughReplicas               KError = 19
+	ErrNotEnoughReplicasAfterAppend    KError = 20
```
eapache: These two both occur if you configure `min.insync.replicas` on the topic (or globally) to be >1 and not enough replicas are remaining in-sync. They should be retried in the producer.

wvanbergen: Any idea what the difference between the two is?

eapache: The first is raised if there are not enough ISR when the request is received, so it fails immediately. The second is raised if there are enough initially, but by the time the local append has happened and all remaining ISRs have acked, the number of ISRs is not enough.

Basically: retrying the second could give you duplicates, retrying the first definitely won't.

PR #314, opened 2015-03-04
Check for errors in test code

eapache:
Recent versions of errcheck have started checking in _test.go files as well,
revealing a handful more errors that need handling.

@Shopify/kafka


eapache:
The vast majority of these were in the example code where we were simply doing e.g. `defer producer.Close()` which of course doesn't handle the error it could return.

PR #317, opened 2015-03-05
Make partitioner implementations private

eapache:
The only part the user needs to have access to is the set of constructors, so
make the types themselves private, which somewhat reduces our API surface.

Also remove the ConstantPartitioner. It is trivial, not particularly useful in
practice, and was only added for use in a long-since-defunct functional test.

@Shopify/kafka


yagnik:
I love having a really small api :100:

wvanbergen:
:ship::it:

Issue #312, opened 2015-03-04
Export MockBroker in a separate package


eapache:
Per https://github.com/Shopify/sarama/pull/307#issuecomment-77089589 and https://github.com/Shopify/sarama/pull/307#issuecomment-77091338 this functionality *is* useful to other people, it just shouldn't live as part of Sarama's API.

I know @wvanbergen had plans to add a bunch of functionality to it, so splitting it out will also make that easier.

Currently the only road-block is it's use of the private `encoder` type, but that can just be replaced with `[]byte` and the tests can encode the packets ahead of time. This will actually make it more flexible too, by permitting us to test what happens when the broker gives us back garbage packets.

@yagnik  this is a nice not-entirely-trivial but danger-free ticket if you're looking to work on your go.

cc @Shopify/kafka @schleppy

eapache:
Related to the last point (DRY up the tests) of #271 since as-is removing the use of `encoder` will add a bit more duplication to the test suite.

yagnik:
I can try to take a stab at this over the weekend

eapache:
Current open sarama tickets/PRs that will have to be migrated in some form:
 - #298 
 - #323

wvanbergen:
I finally have an idea for `MockBroker` and `MockClient`. 

1. We implement in a separate package that avoids circular dependencies e.g. `sarama/internal/mocks`
2. We expose them to the outside world by simply importing them into `sarama/mocks`.

``` go
package mocks

import internal_mocks "sarama/internal/mocks"
type Consumer internal_mocks.Consumer
```

Would this work?

eapache:
I believe that would work, :+1: clever

sybrandy:
Do you plan on having the MockBroker and MockClient available?  I have the 1.0.0 release and I don't see them in there.  This is important to me as I have code that I want to test to ensure that I'm properly setting up connections, checking broker health, etc.

sybrandy:
FYI: I'm writing tests right now for what I need to test my interfaces and I'm following the pattern described here: [Creating Fakes in Go With Channels](http://blog.codeship.com/creating-fakes-in-go-with-channels/).  So far, it looks very clean and very easy to integrate into tests.  I just wanted to pass this along as I think this pattern would make for very nice mocks without too much effort.

eapache:
[...]

PR #324, opened 2015-03-06
Add basic MockProducer

wvanbergen:
Part of #312. 

- This adds a `kafkamocks` subpackage under sarama. Name :+1:/:-1: ?
- It adds a simple `KafkaProducer` interface modelled after `sarama.Producer`.
- MockProducer implements this interface and is mostly lifted from KafkaBridge. It also adds a very basic expectations API. Looking for feedback on this part of the API design.

@Shopify/kafka @benbjohnson


kafkamocks/mock_producer.go
```diff
@@ -0,0 +1,113 @@
...
+	closed       chan struct{}
+	input        chan *sarama.ProducerMessage
+	successes    chan *sarama.ProducerMessage
+	errors       chan *sarama.ProducerError
+}
+
+func NewMockProducer(t TestReporter, config *sarama.Config) *MockProducer {
+	if config == nil {
+		config = sarama.NewConfig()
+	}
```
wvanbergen: Config is used to enable `AckSuccesses` (and for `ChannelBufferSize`)

eapache: this prevents Sarama from being able to import this package though (would cause cyclic import) so we couldn't put mockbroker in this package

eapache:
Making this a sarama subpackage means it gets versioned with sarama whether we like it or not, I was kind of thinking we'd make a whole other repo for our mocking stuff.

Re. naming we don't need 'mock' in both the package and type names, `kafkamocks.MockProducer` is silly. Just `mocks.Producer` is probably fine, as long as `kafka` or `sarama` or something is in the import path.

kafkamocks/mock_producer.go
wvanbergen: We need to import sarama anyway for `sarama.ProducerMessage` and `sarama.ProducerError`, because we cannot efficiently capture these with interfaces. Also, we will never use the MockProducer (or a MockConsumer) inside Sarama itself; these are purely provided to make the lives of sarama users easier when it comes to testing.

Maybe the best solution is to keep mockBroker as a private type inside sarama, and if theres a need for it, add a public MockBroker type to this package as a copy?

** This reminds me: should we rename `Broker` in sarama to `BrokerConnection` or `Connection`?

wvanbergen:
I prefer to keep it in the sarama repository. 

- It makes it clear that these mocks are for sarama, and not for anything else.
- This will make sure the mock an actual implementation are always in sync. 

The amount of newly exported identifiers is not that big: only `MockProducerExpectation` and the methods that create them. The rest is covered by the `KafkaProducer` interface that will fall under the backwards compatibility guarantee anyway.

So, we can do two things:
- Give the new types/functions some extra scrutiny, and place them under the same guarantee.
- Make it clear in the docs that this subpackage does not fall under the guarantee. This is [not unprecedented](http://golang.org/pkg/testing/#Cover)

kafkamocks/mock_producer.go
eapache: OK, you're right that we do need to import it for various types. Code duplication is icky though. Perhaps further sub-packaging is the answer: `sarama/mocks/producer` and `sarama/mocks/broker` etc. Then the mockbroker package *doesn't* need to import sarama proper.

cc @yagnik this discussion potentially impacts #312

eapache: > This reminds me: should we rename Broker in sarama to BrokerConnection or Connection?

... why? I'm not necessarily opposed to `BrokerConnection` on any grounds except it being longer. Just `Connection` seems to lose fairly important context (especially when lazy-connections land) because `Connection.ID()` seems to imply something connection-specific, not broker-specific.

eapache:
> Make it clear in the docs that this subpackage does not fall under the guarantee. This is not unprecedented

:+1: I did not realize even the stdlib did this.

eapache:
> this again comes to show that go needs a real solution for dependency management

Looks like they're finally working on it: https://groups.google.com/forum/#!topic/golang-dev/nMWoEAG55v8[1-25]

wvanbergen:
To conclude:

- Use sarama subpackage, called `mocks`
- Make it clear in godoc it's not covered by API stability guarantee
- Name of the mock producer: `mocks.Producer`

Question:

- Should the interface live in the mocks package, in Sarama itself, or a separate interaces subpackage?

eapache:
> Should the KafkaProducer interface live in the mocks package, in Sarama itself, or a separate interaces subpackage?

I'm not sure we should be providing it at all - users of the package who want to test their code can create their own interface that has only the methods they need.

wvanbergen:
That would just mean more work for the user. What is the disadvantage of providing it?

eapache:
If we provide it, it has to live in the proper sarama package (making the user import a second package into their prod. code just to make it testable is silly). But given that not all users will use the entire API, the pattern in go seems to be to make the user define their own interface (see e.g. http://relistan.com/writing-testable-apps-in-go/ which explicitly recommends that over using provided interfaces) with only the methods they need.

wvanbergen:
I don't think the "no unnecessary methods" reason applies here. We already provide a mock implementation that implements all the methods of the interface, so the user doesn't have to implement any methods. She will be forced to define the interface herself if she wants to use the mocks, which is just unnecessary work.

It's always possible to define your own interface if you want to write your own mock.

eapache:
The only other point I have is to make the API smaller, but I guess then we don't even have to expose the real producer struct, just make it private as an implementation of the interface.

wvanbergen:
OK I will make the necessary changes.

wvanbergen:
I moved the interface into Sarama itself, and did the renames. I did the same for SyncProducer and added a mock implementation for it as well.

wvanbergen:
If the consensus is that this is the right approach, I will add docs and examples.

eapache:
LGTM

wvanbergen:
Now with godoc :+1:

mocks/expectations.go
```diff
@@ -0,0 +1,20 @@
+package mocks
+
+import (
+	"errors"
+)
+
+// A simple interafce that includes the testing.T methods we use to report
```
eapache: typo

mocks/producer.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/Shopify/sarama"
+)
+
+// Producer implements sarama's Producer interface for testing purposes.
+// Before you can send messages to it's Input channel, you have to set expectations
+// so it knows how to handle the input. This way you can easily test success and
+// failure scenarios.
+//
+// NOTE: the SyncProducer type currently does not fall under the API stability
+// guarantee of Sarama as it is stiull considered experimental.
```
eapache: typo

eapache: also syncproducer?

eapache:
Ah, I forgot after the last round, but we haven't solved the circular import problem yet. I *do* want to see the mock broker in the mocks package... using even more subpackages doesn't necessarily seem like a bad idea? (e.g. `sarama/mocks/producer/` and `sarama/mocks/broker/` etc)

mocks/expectations.go
```diff
@@ -0,0 +1,20 @@
...
+// expectation violations when using the mock objects.
+type ExpectationViolationReporter interface {
```
eapache: just `ErrorReporter`? not all errors are necessarily going to be violations of expecations

mocks/producer.go
eapache: I would put the note about api stability at the package level, not on individual types

mocks/expectations.go
```diff
@@ -0,0 +1,20 @@
+package mocks
```
eapache: whatever package(s) we end up with will need package-level godoc too

wvanbergen:
My preference would be to put all mock types in the `mocks` package directly, except `MockBroker`. The MockBroker is an exception for several reasons, but primarily it is not really a mock that you can use for dependency injection to replace our `Broker` type . Not sure what package name I'd like for it though; maybe `mocks/broker`.

eapache:
I guess we can always move things around later if we want (because no api guarantees), I just find the inconsistency of having mockbroker on its own would bother me :)

Issue #323, opened 2015-03-06
Add examples to godocs.


sybrandy:
Hello,

As someone who has worked with Kafka before, the way the API works is different from the Java/Scala API.  I'm not saying it's wrong: I actually like it.  However, you have to think about things differently.  I'd like to see examples of how to create consumers/producers, configuring partitioners, etc. in the Godocs to not only help with the transition, but to also make sure that the API is being used how it is meant to be used.  I know I can look through unit tests, but that's really a terrible way to do things when you want to get started.

eapache:
Hi @sybrandy there are already five examples at https://godoc.org/github.com/Shopify/sarama#pkg-examples including two for the consumer and two for the producer. What else are you looking for?

sybrandy:
Dammit...I think I spent so much time looking at the individual methods, I completely missed them.  Sorry.

wvanbergen:
I feel we should make http://shopify.github.io/sarama/ more useful though.

eapache:
ya, I didn't write that, that was part of Shopify's push to give all our open-source projects nice homepages... I'm not sure it really stuck

sybrandy:
Dammit, I just remember what was bothering me: examples for using the MockBroker.  I can't recall the one case off the top of my head, but there was one thing I wanted to test but couldn't figure out how.

wvanbergen:
More reason to move the mock broker/producer etc to a separate package where it will be a first class citizen. See #312 .

wvanbergen:
I added an example for testing using dependency injection to the wiki: https://github.com/Shopify/sarama/wiki/Testing-using-mocks

wvanbergen:
@sybrandy What other areas of Sarama do you think would be helped by adding examples?

sybrandy:
The copy of Sarama I downloaded does not have the mocks, so I can't comment on them.  I don't want to pull down a new version yet as what I have is working for us and I'd rather wait til 1.0 before upgrading.  However, most of my issues were with trying to get the tests working correctly, so documentation on how to do that would be great.  For example, I had to stare at code for a while before I figured out how to properly have the MockBroker respond with multiple results.  Ideally, I'd like to see something like a cookbook where each recipe shows you how to test a certain situation.  Such as "Testing a failure when retrieving a message from the broker." or "Testing a timeout when there are no messages to be retrieved."

wvanbergen:
[...]

PR #326, opened 2015-03-09
Add an example for using the async producer with goroutines

wvanbergen:
@Shopify/kafka

### DIFF
--- a/producer_test.go
+++ b/producer_test.go
@@ -1,7 +1,9 @@
 package sarama
 
 import (
-	"fmt"
+	"log"
+	"os"
+	"os/signal"
 	"sync"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


wvanbergen:
Partially fixes #323

eapache:
LGTM

producer_test.go
```diff
@@ -507,25 +509,92 @@ func TestProducerMultipleRetries(t *testing.T) {
...
 		select {
 		case producer.Input() <- &ProducerMessage{Topic: "my_topic", Key: nil, Value: StringEncoder("testing 123")}:
-			fmt.Println("> message queued")
+			enqueued++
 		case err := <-producer.Errors():
-			panic(err.Err)
+			log.Println("Failed to produce message", err)
+			errors++
+		case <-signals:
+			break
```
eapache: as  the failing test (go vet) caught: you need to break the outer loop, this will just break the select statement by default

PR #331, opened 2015-03-10
Consumer mock

wvanbergen:
@Shopify/kafka


mocks/consumer.go
```diff
@@ -0,0 +1,169 @@
...
+	c.l.Lock()
+	defer c.l.Unlock()
+
+	if c.partitionConsumers[topic] == nil || c.partitionConsumers[topic][partition] == nil {
+		c.t.Errorf("No expectations set for %s/%d", topic, partition)
+		return nil, errOutOfExpectations
+	}
+
+	pc := c.partitionConsumers[topic][partition]
+	pc.consumed = true
```
eapache: if it's already been consumed should we error?

mocks/consumer.go
```diff
@@ -0,0 +1,169 @@
...
+	for _, partitions := range c.partitionConsumers {
+		for _, partitionConsumer := range partitions {
+			_ = partitionConsumer.Close()
+		}
+	}
+
+	return nil
+}
+
+func (c *Consumer) OnPartition(topic string, partition int32) *PartitionConsumer {
```
eapache: what does this method do?

eapache:
The sarama proper changes are fine, and the rest LGTM. :ship: when ready.

mocks/consumer.go
wvanbergen: Does sarama allow consuming the same partition multiple times from the same Consumer?

eapache: Not presently https://github.com/Shopify/sarama/blob/f1a262ce9a6151ef8e3d90452c275c6ccd06727a/consumer.go#L149

mocks/consumer.go
wvanbergen: It registers a mock PartitionConsumer, and returns it, so you can use method chaining to set expectations on it. See `mocks/consumer_test.go` for an example.

This means that the consumer that this topic/partition will be consumed. If it is not consumed, it will report an error. If you consume a partition without registering it, an error is reported as well.

mocks/consumer.go
wvanbergen: Cool, in that case I will return an error as well.

PR #333, opened 2015-03-10
Partitioners operate on the entire ProducerMessage, not just the key

wvanbergen:
This allows you to create custom partitioning schemes, e.g. by using the Metadata field.

Example:

``` go
type PartitionMetadata interface {
  Partition() int32
}

type myPartitioner struct {}

func (mp *myPartitioner) Partition(message *ProducerMessage, numPartitions int32) (int32, error) {
  if pm, ok := message.Metadata.(PartitionMetadata); ok {
    return pm.Partition(), nil
  } else {
    return 0, errors.New("You have to set the partition in the message's Metadata")
  }
}

func (mp *myPartitioner) RequiresConsistency() {
  return true 
}

myPartitionerConstructor := func()  {
  return &myPartitioner{}
}



```

@Shopify/kafka

### DIFF
--- a/partitioner.go
+++ b/partitioner.go
@@ -7,11 +7,11 @@ import (
 	"time"
 )
 
-// Partitioner is anything that, given a Kafka message key and a number of partitions indexed [0...numPartitions-1],
+// Partitioner is anything that, given a Kafka message and a number of partitions indexed [0...numPartitions-1],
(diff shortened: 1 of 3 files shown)


eapache:
godoc for `NewHashPartitioner` could maybe be tweaked, it's not as clear anymore with this change, but otherwise :+1:

wvanbergen:
Tweaked to godoc; fixed the tests. If only Travis wouldn't have such a large lag today...

PR #335, opened 2015-03-11
Make mockbroker part of the _test code

### LINKED ISSUE #334: Using sarama pulls test code into non-test binaries
When using sarama, it pulls a bunch of test options into production binaries. This appears to be a problem caused by mockbroker.go. Addressing #312 seems like it would likely address it.

Sample code:

    package main

    import (
	    "flag"
	    "fmt"

	    "github.com/Shopify/sarama"
    )

    func main() {
	    flag.Parse()

	    saramaConfig := sarama.NewConfig()
	    saramaConfig.ClientID = "id"
	    kafkaClient, err := sarama.NewClient([]string{"localhost:9092"}, saramaConfig)
	    fmt.Printf("kafkaClient: %v, err: %v\n", kafkaClient, err)
    }

For me, this creates a 'sample' binary. The output of running ./sample --help:

    $ ./sample --help
    Usage of ./sample:
      -test.bench="": regular expression to select benchmarks to run
      -test.benchmem=false: print memory allocations for benchmarks
      -test.benchtime=1s: approximate run time for each benchmark
      -test.blockprofile="": write a goroutine blocking profile to the named file after execution
      -test.blockprofilerate=1: if >= 0, calls runtime.SetBlockProfileRate()
      -test.coverprofile="": write a coverage profile to the named file after execution
      -test.cpu="": comma-separated list of number of CPUs to use for each test
      -test.cpuprofile="": write a cpu profile to the named file during execution
      -test.memprofile="": write a memory profile to the named file after execution
      -test.memprofilerate=0: if >=0, sets runtime.MemProfileRate
      -test.outputdir="": directory in which to write profiles
      -test.parallel=1: maximum test parallelism
      -test.run="": regular expression to select tests and examples to run
      -test.short=false: run smaller test suite to save time
      -test.timeout=0: if positive, sets an aggregate time limit for all tests
      -test.v=false: verbose: print additional output

I renamed mockbroker.go => mockbroker_test.go, and the problem went away.

eapache:
Until #312 lands it in its own package, we should avoid pulling go's testing lib
into the production library. Fixes #334.

cc @Shopify/kafka


wvanbergen:
Smart. The alternative is using an interface instead of testing.T like I did in sarama/mocks, but I guess this is OK as a stopgap measure.

eapache:
It did use an interface when it was public and part of the core API, but I removed it when we made it private, not thinking that without the rename it would still get *built* into the core package :P

PR #336, opened 2015-03-12
Configuration tweaks

wvanbergen:
- Make consumer retry backoff configurable, use 10s by default.
- Allow opting out of reading the Errors channel, by setting Consumer.AckErrors and/or Producer.AckErrors to false.
- Fix some documentation.

 #### Discussion

- The retry backoff for the JVM producer is 200ms, we use 10,000ms. :+1:/:-1:?
- I am kind of inclined to default `AckErrors` to `false`, at least for the consumer. The Errors channel is not very useful, because the consumer will retry anyway and keep on going. Figuring out in what scenario you should stop the consumer, because not all errors are created equal. Ideally we would have a way to detect whether somebody is consuming the channel, but I haven't figured out a way.
- I'd rather add a `Consumer.Timeout` setting, so you can make the consumer stop if no messages have been received for an `x` amount of time.

@Shopify/kafka

### DIFF
--- a/config.go
+++ b/config.go
@@ -42,9 +42,9 @@ type Config struct {
 		// Similar to the `partitioner.class` setting for the JVM producer.
 		Partitioner PartitionerConstructor
 		// If enabled, successfully delivered messages will be returned on the Successes channel (default disabled).
-		AckSuccesses bool
+		ReturnSuccesses bool
 		// If enabled, messages that failed to deliver will be returned on the Errors channel, including error (default enabled).
-		AckErrors bool
+		ReturnErrors bool
 
 		// The following config options control how often messages are batched up and sent to the broker. By default,
 		// messages are sent as fast as possible, and all messages received while the current batch is in-flight are placed
@@ -95,8 +95,8 @@ type Config struct {
 		// Equivalent to the JVM's `fetch.wait.max.ms`.
 		MaxWaitTime time.Duration
 
-		// If enabled, any errors that occured while consuming are returned on the Errors channel (default enabled).
-		AckErrors bool
+		// If enabled, any errors that occured while consuming are returned on the Errors channel (default disabled).
+		ReturnErrors bool
 	}
 
 	// A user-provided string sent with every request to the brokers for logging, debugging, and auditing purposes.
@@ -127,13 +127,13 @@ func NewConfig() *Config {
 	c.Producer.Partitioner = NewHashPartitioner
 	c.Producer.Retry.Max = 3
 	c.Producer.Retry.Backoff = 100 * time.Millisecond
-	c.Producer.AckErrors = true
+	c.Producer.ReturnErrors = true
 
 	c.Consumer.Fetch.Min = 1
 	c.Consumer.Fetch.Default = 32768
 	c.Consumer.Retry.Backoff = 2 * time.Second
 	c.Consumer.MaxWaitTime = 250 * time.Millisecond
-	c.Consumer.AckErrors = true
+	c.Consumer.ReturnErrors = false
 
 	c.ChannelBufferSize = 256
 
--- a/config.go
+++ b/config.go
@@ -43,6 +43,8 @@ type Config struct {
 		Partitioner PartitionerConstructor
 		// If enabled, successfully delivered messages will be returned on the Successes channel (default disabled).
 		AckSuccesses bool
+		// If enabled, messages that failed to deliver will be returned on the Errors channel, including error (default enabled).
+		AckErrors bool
 
 		// The following config options control how often messages are batched up and sent to the broker. By default,
 		// messages are sent as fast as possible, and all messages received while the current batch is in-flight are placed
@@ -68,6 +70,11 @@ type Config struct {
 
 	// Consumer is the namespace for configuration related to consuming messages, used by the Consumer.
 	Consumer struct {
+		Retry struct {
+			// How long to wait after a failing to read from a partition before trying again (default 2s).
+			Backoff time.Duration
+		}
+
 		// Fetch is the namespace for controlling how many bytes are retrieved by any given request.
 		Fetch struct {
 			// The minimum number of message bytes to fetch in a request - the broker will wait until at least this many are available.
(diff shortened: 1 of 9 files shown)


config.go
```diff
@@ -68,6 +70,12 @@ type Config struct {
 
 	// Consumer is the namespace for configuration related to consuming messages, used by the Consumer.
 	Consumer struct {
+		Retry struct {
+			// How long to wait after a partition has lost its leader before trying again (default 10s).
+			// Similar to the `refresh.leader.backoff.ms` setting of the JVM producer.
```
eapache: not sure comparing a consumer config to the jvm producer is necessarily useful

config.go
```diff
@@ -211,6 +225,8 @@ func (c *Config) Validate() error {
 		return ConfigurationError("Invalid Consumer.Fetch.Max, must be >= 0")
 	case c.Consumer.MaxWaitTime < 1*time.Millisecond:
 		return ConfigurationError("Invalid Consumer.MaxWaitTime, must be > 1ms")
+	case c.Producer.Retry.Backoff < 0:
```
eapache: this should be Consumer?

consumer.go
```diff
@@ -209,7 +209,9 @@ func (c *consumer) unrefBrokerConsumer(broker *Broker) {
 // PartitionConsumer processes Kafka messages from a given topic and partition. You MUST call Close()
 // on a consumer to avoid leaks, it will not be garbage-collected automatically when it passes out of
 // scope (this is in addition to calling Close on the underlying consumer's client, which is still necessary).
-// You have to read from both the Messages and Errors channels to prevent the consumer from locking eventually.
+// Besides the Messages channel, you have to read from the Errors channels. Otherwisethe consumer will
```
eapache: missing space

consumer.go
```diff
@@ -277,7 +282,7 @@ func (child *partitionConsumer) dispatcher() {
 				select {
 				case <-child.dying:
 					close(child.trigger)
-				case <-time.After(10 * time.Second):
+				case <-time.After(child.conf.Consumer.Retry.Backoff):
```
eapache: In hindsight I think we can actually simplify this loop quite a bit - simply arriving here (via a message on `child.trigger`) indicates that some worker received an error or NotLeader or whatever, so we should back off right away, as dispatching right away probably won't work anyways.

Given that, I think the `<-time.After` case should replace the main `default` case, and this inner select can go away entirely. Makes sense?

config.go
wvanbergen: :whoops:

config.go
eapache: kafka consumer seems to have both `refresh.leader.backoff.ms` and `rebalance.backoff.ms`

wvanbergen:
I don't really like having to tell how you are going to use the API by setting some config flags. So one idea that I had was setting all the `Ack*` values to `false` by default, and then turning them on as soon as you call `Errors()`, or `Successes()`. Unfortunately because we start producing/consuming right away after creating the `Producer` or `PartitionConsumer`, this could lead to messages/errors not being reported in certain concurrency scenarios. Not sure if there's a way around that.

eapache:
I don't understand the desire to be able to blindly ignore errors, that's very non-idiomatic go.

consumer.go
eapache: With this loop fixed I think it makes much more sense to drop the default to 200ms or something more reasonable - the high value was my subconscious trying to point out that the delay was in the wrong place anyways.

wvanbergen:
Because it's basically not an error, it's a warning - at least in the `Consumer`. In the async producer I also see it as valid to ignore errors in AP scenarios.

eapache:
> I'd rather add a Consumer.Timeout setting, so you can make the consumer stop if no messages have been received for an x amount of time.

I have no technical objections, I just don't see the use case. In the vast majority of cases I can think of you just want to keep consuming forever or until an external signal is received?

config.go
wvanbergen: I'm inclined to change the current default to 2s.

eapache:
> Because it's basically not an error, it's a warning - at least in the Consumer.

Hmm, that's fair I guess, although there are some errors that indicate something more serious is wrong.

> In the async producer I also see it as valid to ignore errors in AP scenarios.

That's also true, but I think that should generally be an explicit choice for the user, as opposed to just forgetting or not realizing, and ending up with different guarantees than they want. In that cause, launching a goroutine to drain it is not much more work than setting the flag, so I'm not sure the flag adds much value.

config.go
eapache: :+1:

wvanbergen:
> Hmm, that's fair I guess, although there are some errors that indicate something more serious is wrong.

Exactly. However, because right now all errors are treated equally and the burden is placed on the user to figure out whether something more serious is wrong. This is why I like a timeout value. If you haven't seen any events for `x` seconds, you know something more serious is wrong.

eapache:
> If you haven't seen any events for x seconds, you know something more serious is wrong.

Or it's just a really quiet topic.

eapache:
I think, even in the cases that something more serious is wrong, the consumer should be able to recover once the cluster recovers. I cannot think of a case where e.g. shutting down and restarting the consumer adds any value to just letting it recover on its own.

config.go
```diff
@@ -43,6 +43,8 @@ type Config struct {
 		Partitioner PartitionerConstructor
 		// If enabled, successfully delivered messages will be returned on the Successes channel (default disabled).
 		AckSuccesses bool
+		// If enabled, messages that failed to deliver will be returned on the Errors channel, including error (default enabled).
+		AckErrors bool
```
eapache: I dislike the name `AckErrors` but I don't have anything better :(

Maybe `Return.Errors` and `Return.Successes`?

wvanbergen:
Updated.

consumer_test.go
```diff
@@ -290,10 +292,51 @@ func TestConsumerInterleavedClose(t *testing.T) {
...
+	}
+	log.Println("Processed", msgCount, "messages.")
+}
+
 // This example shows how to use a consumer with a select statement
 // dealing with the different channels.
 func ExampleConsumer_select() {
-	master, err := NewConsumer([]string{"localhost:9092"}, nil)
+	config := NewConfig()
+	config.Consumer.ReturnErrors = true // Handle errors manually instead of logging them.
```
eapache: all we do is log them, so maybe "instead of letting the consumer log them"

consumer_test.go
```diff
@@ -336,7 +379,10 @@ consumerLoop:
 // This example shows how to use a consumer with different goroutines
 // to read from the Messages and Errors channels.
 func ExampleConsumer_goroutines() {
-	master, err := NewConsumer([]string{"localhost:9092"}, nil)
+	config := NewConfig()
+	config.Consumer.ReturnErrors = true // Handle errors manually instead of logging them.
```
eapache: similar

eapache:
One nit, otherwise LGTM I think

PR #337, opened 2015-03-12
Retry UnknownTopicOrPartition on fetching metadata

eapache:
@wvanbergen


wvanbergen:
I guess this makes sense. :+1:

Do we want to ask about this on the mailing list?

eapache:
I just unsubscribed again because it's pretty noisy - feel free :)

This doesn't retry on a different broker, if we want that we need to do something different.

Issue #15, opened 2013-08-15
Better Broker Connection Management


eapache:
Following up to #7, #9, #10, #13 our connection management still has problems.

First some things to know about Kafka that complicate the issue:
 * Broker metadata requests do not seem required to return all the brokers they know about (it is not documented which ones they do return, but I suspect it is the only the subset of brokers that are leading a partition whose data is also in the response).
 * The user only specifies (host, port) pairs while Kafka metadata returns (host, port, brokerId) triples, with no guarantee that the host matches the user-specified hostname, and with no apparent way to query a broker's own brokerId.

Now some of the issues with our current implementation:
 * If the user provides three addresses, only the last of which is valid, we wait for the first two connections to completely fail when fetching metadata before we try the third. Since the TCP connection timeout can be very long (many minutes) this can take a long time. Trying them all in parallel and using the first valid one would be better.
 * Once we've had an address fail, we discard it forever. Over the lifetime of a long-running cluster, all of the nodes may be down at one point or another, meaning that eventually we may 'run out' of nodes and abort.

Thoughts on potential future design:
* We should connect to brokers lazily, ie only when we actually want to talk to them, not just when we know they exist.
* Since there is only ever one leader for a partition, if the connection fails it doesn't make sense to just try another broker. Therefore the caller should decide to proceed, which means that the client's `leader` function shouldn't guarantee that the broker it returns is connected, only connecting.
* However, when fetching metadata, any broker will do, so the `any` function should guarantee that the broker it returns is actually connected. Walking through bad brokers to find a good one could be slow, but connecting to *all* the brokers just to pick one and drop the others is gross. We could try to connect to one, and if it hasn't succeed in 5 seconds, start another, etc. That should cover the common case, since if it hasn't responded after 5 seconds it *probably* won't, and if it does then we'll drop it right away, oh well. More importantly, we won't spawn hundreds of connections if the first broker does respond normally.
* User-provided addresses should be managed separately (which they already are) since they don't have brokerIds. They're only useful for metadata requests, so they should be disconnected entirely 99% of the time. If one of them fails, we should probably just mark it as recently-failed, and retry it after some time has passed. Only if all user-supplied nodes have recently failed (and we have no metadata from the cluster itself) should we abort somehow.

Open Questions:
[...]

PR #338, opened 2015-03-12
Remove external calls to disconnectBroker

eapache:
It is now only called from one place in the client. This looks simple but is
actually super-subtle, and depends on lazy broker connections (PR #309).

disconnectBroker does a whole bunch of different things:
- calls `Close` on the broker connection
- adds the address to the internal `deadBrokerAddrs` map even if the broker is
  not a seed, which I think is wrong, since resurrectDeadBrokers will use it to
  repopulate the seedBrokerAddrs list
- rotates seedBrokers (if the broker was a seed broker)
- removes it from the brokers map (otherwise)

In the producer and consumer where we used to call disconnectBroker:
- We now call `Close` directly on the broker.
- The broker we are dealing with is not a seed broker, so the seedBrokers do not
  need rotating, and I don't think it's a problem that it no longer gets added
  to `deadBrokerAddrs`.
- The reason we removed it from the broker map was so that the next request for
  that broker would trigger a metadata request and reopen the connection. The
  producer and consumer both manually trigger metadata requests when necessary,
  and the fact that we now have lazy connection opening means simply closing it
  (which we do, see first bullet) is enough to cause the connection to reopen
  the next time it is requested, even if no metadata refresh is requested.

In a subsequent PR, the last remaining call will be inlined and refactored. This will effectively fix the last non-trivial issue spawned from #15.

@Shopify/kafka

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -438,7 +438,7 @@ func (w *brokerConsumer) subscriptionConsumer() {
 		response, err := w.fetchNewMessages()
 
 		if err != nil {
-			Logger.Printf("Unexpected error processing FetchRequest; disconnecting broker %s: %s\n", w.broker.addr, err)
+			Logger.Printf("Unexpected error processing FetchRequest; disconnecting from broker %s: %s\n", w.broker.addr, err)
 			w.abort(err)
 			return
 		}
@@ -475,7 +475,6 @@ func (w *brokerConsumer) updateSubscriptionCache(newSubscriptions []*partitionCo
 
 func (w *brokerConsumer) abort(err error) {
 	_ = w.broker.Close() // we don't care about the error this might return, we already have one
-	w.consumer.client.disconnectBroker(w.broker)
 
 	for child := range w.subscriptions {
 		child.sendError(err)
--- a/producer.go
+++ b/producer.go
@@ -532,9 +532,9 @@ func (p *producer) flusher(broker *Broker, input chan []*ProducerMessage) {
 			p.returnErrors(batch, err)
 			continue
 		default:
-			p.client.disconnectBroker(broker)
 			Logger.Printf("producer/flusher/%d state change to [closing] because %s\n", broker.ID(), err)
 			closing = err
+			_ = broker.Close()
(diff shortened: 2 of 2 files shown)


eapache:
The lazy connections have been merged so this is ready for review. The code is trivial (not much to review there), I'm more interested in somebody verifying that my logic is correct around `deadBrokerAddrs`, connection bouncing, etc.

As part of #309 this has already undergone some decent practical testing using the vagrant cluster and willem's stressproducer.

wvanbergen:
Did we run any consumer tests by any chance?

eapache:
Just ran some of those using your new console-consumer, but I'm not too concerned. The changes are symmetric so if the producer works fine then the consumer should (and does) also.

wvanbergen:
I think the logic here is sound.

I don't think it's necessary to `Close()` the broker connection for every error we see in the producer or consumer, but this one of these optimizations that are easy to fuck up.

What is the number of (producer) retries we potentially waste under relatively normal operations?
1. Produce requests fails (e.g. not leader for partition), so it asks for new metadata.
2. Metadata update request fails. This is already unlikely, because it happens only if it completely drains the Metadata retries as well, right?

eapache:
> I don't think it's necessary to Close() the broker connection for every error we see in the producer or consumer

You don't? If a broker gives us a network error or an unexpected protocol error then I don't see what choice we have besides disconnecting from that one and asking metdata for another one?

> What is the number of (producer) retries we potentially waste under relatively normal operations?

No more than we ever have?

wvanbergen:
I meant, not for every error. E.g. if we get a `NotLeaderForPartition`, we don;t technically have to disconnect. Or does that error never end up in `abort`?

wvanbergen:
:shipit:

eapache:
> I meant, not for every error. E.g. if we get a NotLeaderForPartition, we don;t technically have to disconnect. Or does that error never end up in abort?

`abort` cleans up the entire worker - the only errors that end up there are ones returned from `broker.Fetch()`, nothing at the protocol layer.

PR #339, opened 2015-03-12
High water mark offset support to the consumer

wvanbergen:
Expose high water marker offset in consumer for lag monitoring purposes.

@Shopify/kafka

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -3,6 +3,7 @@ package sarama
 import (
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -255,6 +256,11 @@ type PartitionConsumer interface {
 	// errors are logged and not returned over this channel. If you want to implement any custom errpr
 	// handling, set your config's Consumer.Return.Errors setting to true, and read from this channel.
 	Errors() <-chan *ConsumerError
+
+	// HighWaterMarkOffset returns the high water mark offset of the partition, i.e. the offset that will
+	// be used for the next message that will be produced. You can use this to determine how far behind
+	// the processing is.
+	HighWaterMarkOffset() int64
 }
 
 type partitionConsumer struct {
@@ -268,8 +274,9 @@ type partitionConsumer struct {
 	errors         chan *ConsumerError
 	trigger, dying chan none
 
-	fetchSize int32
-	offset    int64
+	fetchSize           int32
+	offset              int64
+	highWaterMarkOffset int64
 }
 
 func (child *partitionConsumer) sendError(err error) {
@@ -391,6 +398,10 @@ func (child *partitionConsumer) Close() error {
 	return nil
 }
 
+func (child *partitionConsumer) HighWaterMarkOffset() int64 {
+	return atomic.LoadInt64(&child.highWaterMarkOffset)
+}
+
 func (child *partitionConsumer) handleResponse(response *FetchResponse) error {
 	block := response.GetBlock(child.topic, child.partition)
 	if block == nil {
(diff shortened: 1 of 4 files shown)


eapache:
need rebase

consumer.go
```diff
@@ -238,6 +239,11 @@ type PartitionConsumer interface {
 	// errors are logged and not returned over this channel. If you want to implement any custom errpr
 	// handling, set your config's Consumer.ReturnErrors setting to true, and read from this channel.
 	Errors() <-chan *ConsumerError
+
+	// HighWaterMarkOffset returns the high water mark offset of the partition, i.e. the offset that will
+	// be used for the next message that will be produced. You can use this to determine high far behind
```
eapache: s/high/how

wvanbergen:
I think we either need to expose the high water mark on the `ConsumerMessage`, or expose the latest consumed offset on the `PartitionConsumer`. With just this, you need to have both a message and the related PartitionConsumer in scope in order to do monitoring. I rather change that to be a single object.

eapache:
With the buffer on the return I don't think there's a reliable non-confusing way to expose that information though. I think in the common case you will have both in scope anyways:

```go
for msg := range pc.Messages() {
    // do whatever, including msg.Offset and pc.HighWaterMark()
}
```

wvanbergen:
As a counter example, in my consumergroup library, you will get messages from all partitions on the channel, and you don't have access to the partition consumers. I guess I could somehow build some code around that to make it work, but it feels painful.

eapache:
What about exposing a `GetChild(topic, partition)` method (or some other name) on the master consumer which returns the PartitionConsumer but doesn't try to create it. Then your library's `HighWaterMark(topic, partition)` implementation can be
```go
consumer.GetChild(topic, partition).HighWaterMark()
```

I think this is potentially a useful method to have anyways, and is very easy to write.

eapache:
I still like this PR as it stands (needs a rebase, but). Maybe we need to expose this information in other places as well, but the current place on the `PartitionConsumer` is the simplest place to start.

wvanbergen:
Rebased.

consumer.go
```diff
@@ -255,6 +256,11 @@ type PartitionConsumer interface {
 	// errors are logged and not returned over this channel. If you want to implement any custom errpr
 	// handling, set your config's Consumer.Return.Errors setting to true, and read from this channel.
 	Errors() <-chan *ConsumerError
+
+	// HighWaterMarkOffset returns the high water mark offset of the partition, i.e. the offset that will
+	// be used for the next message that will be produced. You can use this to determine how far behind
+	// the processing is.
+	HighWaterMarkOffset() int64
```
eapache: FWIW, the kafka docs have "watermark" as one word, so I guess `HighWatermarkOffset` would be more consistent, but YMMV

eapache: oh, I guess that's my fault for misnaming the variable in `FetchResponseBlock` :(

Add it to the list of breaking changes I guess...

wvanbergen: Hmm, grammatically I feel my capitalization closer to `high-water mark` instead of the incorrect `high watermark`.

wvanbergen: Personally, I have a problem with it being **high-water mark offset**; I'd have preferred **high-water mark** or **high-water offset** instead of this Frankensteinian composition. But that would be deviating from the protocol spec too much. Oh well :P

eapache: And for a different opinion, wikipedia doesn't even hyphenate, it just uses three distinct words ("high water mark") everywhere. So I guess this issue is contentious :)

This is fine as-is (and is consistent with `FetchResponseBlock`). We can revisit if anybody complaints

eapache:
This LGTM. :shipit: if you're comfortable with the API.

wvanbergen:
I am ok with this API as a low level building block, but for the high-level consumer we should definitely give this some more thought.

PR #340, opened 2015-03-12
Reduce logging on broker close

eapache:
Don't log "Error closing broker" when the only error is that it's already
disconnected. Don't defer, there's nothing else in the method that can return.

@Shopify/kafka

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -96,16 +96,9 @@ func (b *Broker) Connected() (bool, error) {
 	return b.conn != nil, b.connErr
 }
 
-func (b *Broker) Close() (err error) {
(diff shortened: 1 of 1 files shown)


wvanbergen:
:shipit:

broker.go
```diff
@@ -99,13 +99,6 @@ func (b *Broker) Connected() (bool, error) {
 func (b *Broker) Close() (err error) {
```
wvanbergen: I guess you don't really need to assign a name to the return value anymore.

eapache: good point

PR #341, opened 2015-03-12
Return{Errors,Successes} -> Return.{Errors,Successes}

wvanbergen:
@eapache as discussed.

### DIFF
--- a/config.go
+++ b/config.go
@@ -41,10 +41,16 @@ type Config struct {
 		// Generates partitioners for choosing the partition to send messages to (defaults to hashing the message key).
 		// Similar to the `partitioner.class` setting for the JVM producer.
 		Partitioner PartitionerConstructor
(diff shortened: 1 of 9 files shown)


config.go
```diff
@@ -96,7 +102,13 @@ type Config struct {
 		MaxWaitTime time.Duration
 
 		// If enabled, any errors that occured while consuming are returned on the Errors channel (default disabled).
-		ReturnErrors bool
+		// Return struct specifies what channels will be populated. If they are set to true, you must read from
+		// them to prevent deadlock.
+		Return struct {
+			// Return struct specifies what channels will be populated. If they are set to true, you must read from
```
eapache: godoc should be the "If enabled" line, not this, looks like a bad copy/paste :)

eapache:
:shipit:

PR #344, opened 2015-03-13
Make Client an interface.

wvanbergen:
This allows people to mock a Client more easily for testing purposes. Maybe later, we will provide a mock implementation ourselves.

@Shopify/kafka


eapache:
Providing a mock implementation ourselves might actually make producer/consumer testing easier.

wvanbergen:
Yeah, I think it would be useful for ourselves as well. The only sarama reference in the interface is `OffsetTime` - should we change that to prevent circular dependencies?

PR #345, opened 2015-03-13
Drop ReplicasInSync from Client interface

wvanbergen:
- We don't need it internally.
- The data can be stale on the broker we are requesting this from; you should ask Zookeeper instead.

@Shopify/kafka


eapache:
LGTM

PR #348, opened 2015-03-13
Remove OffsetTime type, just use int64.

wvanbergen:
@Shopify/kafka

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -314,13 +314,11 @@ func (child *partitionConsumer) dispatch() error {
 }
 
 func (child *partitionConsumer) chooseStartingOffset(offset int64) (err error) {
-	var where OffsetTime
+	var time int64
 
 	switch offset {
-	case OffsetNewest:
-		where = LatestOffsets
-	case OffsetOldest:
-		where = EarliestOffset
+	case OffsetNewest, OffsetOldest:
+		time = offset
 	default:
 		if offset < 0 {
 			return ConfigurationError("Invalid offset")
@@ -329,7 +327,7 @@ func (child *partitionConsumer) chooseStartingOffset(offset int64) (err error) {
 		return nil
 	}
 
-	child.offset, err = child.consumer.client.GetOffset(child.topic, child.partition, where)
+	child.offset, err = child.consumer.client.GetOffset(child.topic, child.partition, time)
 	return err
 }
 
--- a/consumer.go
+++ b/consumer.go
@@ -97,15 +97,6 @@ func (c *consumer) Close() error {
 	return nil
 }
 
-const (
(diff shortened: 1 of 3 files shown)


offset_request.go
```diff
@@ -1,19 +1,7 @@
 package sarama
 
-// OffsetTime is used in Offset Requests to ask for all messages before a certain time. Any positive int64
-// value will be interpreted as milliseconds, or use the special constants defined here.
-type OffsetTime int64
-
-const (
-	// LatestOffsets askes for the latest offsets.
-	LatestOffsets OffsetTime = -1
```
eapache: I'd prefer to leave these as documented consts, even if they're of type `int64` instead of type `OffsetTime`

consumer.go
```diff
@@ -119,9 +119,10 @@ func (c *consumer) ConsumePartition(topic string, partition int32, offset int64)
 		fetchSize: c.conf.Consumer.Fetch.Default,
 	}
 
-	if err := child.chooseStartingOffset(offset); err != nil {
-		return nil, err
+	if offset < -2 {
+		return nil, ConfigurationError("Invalid offset")
 	}
+	child.offset = offset
```
eapache: This won't work, we need to call `GetOffset` to convert e.g. -1 to a real offset

offset_request.go
wvanbergen: We already have OffsetOldest and OffsetNewest.

eapache: Which are different things

eapache: Which maybe isn't clear.

consumer.go
wvanbergen: :'(. It was working locally but I was skipping the functional tests because of stupid.

offset_request.go
wvanbergen: I don't really see the difference. It's used in two different context, but in both contexts the mean the same thing: "I want the {oldest,newest} offset". I think having both `OffsetOldest` and `EarliestOffset` is way more confusing.

wvanbergen:
It's working now :)

offset_request.go
eapache: it means different things for the `offsetRequestBlock` because it's used for multiple offsets, so it's a starting point, not a single offset (thus the old comment "Note that because offsets are pulled in descending order, asking for the earliest offset will always return you a single element." because the other choices would return multiple).

eapache: I'm ok with removing these - the only burden is on the users of the low-level API, which there are very few of (if any). However, we need to update the docs on the ones that are left in order to make them not consumer-specific anymore

wvanbergen: godoc has been updated.

eapache:
LGTM

PR #349, opened 2015-03-16
Cleanup some client internals

eapache:
- Inline `getMetadata` it only had one caller left.
- now that `RefreshMetadata` is the only non-recursive caller of
  `tryRefreshMetadata`, move the one-time checks for a closed client and invalid
  topic names outside the recursion, we only need to run them once.

Should be no functional changes.

@Shopify/kafka


wvanbergen:
Belated :+1:

PR #354, opened 2015-03-16
Close the consumer master in tests

eapache:
Since it now owns the client, we leave goroutines hanging around if we don't.

cc @Shopify/kafka


wvanbergen:
:shipit:

PR #355, opened 2015-03-17
SyncProducer’s SendMessage accepts a sarama.ProducerMessage

wvanbergen:
Rationale:

- This is in line with the AsyncProducer.
- This makes it easier to switch between SyncProducer and AsyncProducer if the need arises.
- This allows you to use to `ManualPartioner`, because now you can set the Partition field.

@Shopify/kafka Last API change I promise :)

### DIFF
--- a/sync_producer.go
+++ b/sync_producer.go
@@ -7,9 +7,10 @@ import "sync"
 // it passes out of scope (this is in addition to calling Close on the underlying client, which is still necessary).
 type SyncProducer interface {
 
-	// SendMessage produces a given message, and returns when it has succeeded or failed.
-	// It will set the OPartition and Offset fields for a successfully produced message.
-	SendMessage(*ProducerMessage) error
+	// SendMessage produces a given message, and returns only when it either has succeeded or failed to produce.
+	// It will return the partition and the offset of the produced message, or an error if the message
+	// failed to produce.
+	SendMessage(msg *ProducerMessage) (partition int32, offset int64, err error)
 
 	// Close shuts down the producer and flushes any messages it may have buffered. You must call this function before
 	// a producer object passes out of scope, as it may otherwise leak memory. You must call this before calling Close
@@ -52,7 +53,7 @@ func newSyncProducerFromAsyncProducer(p *asyncProducer) *syncProducer {
 	return sp
 }
 
-func (sp *syncProducer) SendMessage(msg *ProducerMessage) error {
+func (sp *syncProducer) SendMessage(msg *ProducerMessage) (partition int32, offset int64, err error) {
 	oldMetadata := msg.Metadata
 	defer func() {
 		msg.Metadata = oldMetadata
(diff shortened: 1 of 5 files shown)


sync_producer.go
```diff
@@ -6,9 +6,10 @@ import "sync"
 // and parses responses for errors. You must call Close() on a producer to avoid leaks, it may not be garbage-collected automatically when
 // it passes out of scope (this is in addition to calling Close on the underlying client, which is still necessary).
 type SyncProducer interface {
-	// SendMessage produces a message to the given topic with the given key and value. To send strings as either key or value, see the StringEncoder type.
-	// It returns the partition and offset of the successfully-produced message, or the error (if any).
-	SendMessage(topic string, key, value Encoder) (partition int32, offset int64, err error)
+
+	// SendMessage produces a given message, and returns when it has succeeded or failed.
+	// It will set the OPartition and Offset fields for a successfully produced message.
```
eapache: typo

also, I'd specify "on the input ProducerMessage" or something, to make it clear

this is the only part of the change I don't like: we're now using an input pointer as a return value of sorts, which is one of those things go "fixed" from C via multiple return values

eapache: I'd almost be tempted to leave the return types the way they were for this reason

wvanbergen: This is in line what we do with the AsyncProducer. However, we could continue to also return the Offset and Partition besides modifying the ProducerMessage. That felt a bit double to me.

eapache: In the AsyncProducer we only set those values if you get the message back on `Successes`, at which point it feels like a return value (even if it's actually the same struct).

wvanbergen: OK, I will move back the return values :+1:

eapache:
it may or may not be worth noting that the `Metadata` field gets overridden in this api?

wvanbergen:
It stores the Metadata before overriding it, and sets the old metadata back afterwards using a `defer`, so I am not sure it's relevant. The Metadata object is basically only useful for the AsyncProducer, but I decided to mimic the async's producer behaviour in the sync API so people don't have any nasty surprises if they switch between sync and async.

sync_producer.go
wvanbergen: Updated.

sync_producer.go
```diff
@@ -51,14 +53,21 @@ func newSyncProducerFromAsyncProducer(p *asyncProducer) *syncProducer {
 	return sp
 }
 
-func (sp *syncProducer) SendMessage(topic string, key, value Encoder) (partition int32, offset int64, err error) {
+func (sp *syncProducer) SendMessage(msg *ProducerMessage) (partition int32, offset int64, err error) {
+	oldMetadata := msg.Metadata
+	defer func() {
+		msg.Metadata = oldMetadata
```
eapache: ah, I didn't even notice the save+restore, good idea

eapache:
LGTM

PR #356, opened 2015-03-17
Pass the topic to the partitioner constructor

eapache:
This permits custom constructors that return different types of partitioners for
different topics.

Not 100% sure this is necessary/useful, but I figured I'd throw it in the ring before 1.0.

@Shopify/kafka


wvanbergen:
:cool:

Issue #302, opened 2015-02-24
Semantic Versioning


eapache:
We only have a few remaining PRs and/or ideas that require breaking API changes, several of which are almost ready to go. When I started this project, https://gopkg.in was not a thing, so I stuck the current note about versioning in the README to give me some flexibility as we developed:

> A word of warning: the API is not 100% stable. It won't change much (in particular the low-level Broker and Request/Response objects could probably be considered frozen) but there may be the occasional parameter added or function renamed. As far as semantic versioning is concerned, we haven't quite hit 1.0.0 yet. It is absolutely stable enough to use, just expect that you might have to tweak things when you update to a newer version.

Now that we have a sane way of semantically versioning the library and we've settled on a relatively sane API for both the consumer and the producer, we should think about tagging a v1 and following a more proper release process than just "use master".

TODO:
- [x] fix or officially defer all other tickets in https://github.com/Shopify/sarama/milestones/1.0.0
- [x] add a `HISTORY.md` or `CHANGELOG.md` file - #358
- [x] tag `v1.0.0`
- [x] update README to mention API stability via gopkg.in #361  

@Shopify/kafka

wvanbergen:
API breaking changes that I would still like to tackle:

- [ ] Consumer: split `Events()` channel into `Messages()` and `Errors()`
- [ ] See if we can make the `KError` types instead of values so we can add context

wvanbergen:
Also, I feel we export way too many values to be able to effective maintain backwards compatibility while refactoring. E.g., all the messages, and the Broker type are mostly for internal use and greatly reduce our flexibility for refactoring if they all under our semantic versioning policy. 

I'd like to keep the API compatibility to be just the Client, Consumer, Producer, SimpleProducer, and the error types those return. How we achieve this I don't know. We can just write this down, un-export all the other stuff, or create subpackages. Pros and cons for all of these.

wvanbergen:
Ideally, we would specify interfaces for the types we promise backwards compatibility for. Side benefit is that we can easily use those interfaces for dependency injection as well.

eapache:
> Also, I feel we export way too many values to be able to effective maintain backwards compatibility while refactoring. E.g., all the messages, and the Broker type are mostly for internal use and greatly reduce our flexibility for refactoring if they all under our semantic versioning policy. 

I'm not sure what you mean - these types map to the protocol, so the only time they'd change is if kafka upstream changed the protocol, which would dictate a new major version on our end anyways.

[...]

PR #359, opened 2015-03-17
Add examples for random and manual partitioning

wvanbergen:
Towards addressing #323 and #302. @Shopify/kafka

### DIFF
--- a/partitioner_test.go
+++ b/partitioner_test.go
@@ -2,6 +2,7 @@ package sarama
 
 import (
 	"crypto/rand"
+	"log"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


partitioner_test.go
```diff
@@ -120,3 +121,57 @@ func TestManualPartitioner(t *testing.T) {
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer func() {
+		if err := producer.Close(); err != nil {
+			log.Println("Failed to close producer:", err)
+		}
+	}()
+
+	msg := &ProducerMessage{Topic: "test", Partition: 6, Value: StringEncoder("test")}
```
eapache: we should perhaps comment this line

partitioner_test.go
```diff
@@ -120,3 +121,57 @@ func TestManualPartitioner(t *testing.T) {
 		}
 	}
 }
+
+// By default, Sarama uses the message's key to consistently assign a partition to
+// a message using hashing. If no key is not, a random partition will be chosen.
+// This example shows how you can partition messages randomly, even when a key is set
+func ExamplePartitioner_random() {
```
eapache: not sure how useful this example is, it's literally just the one line config change...

eapache:
just nits, no major objections

eapache:
A more useful example might be writing your own partitioner, but I guess there's real live code (the existing implementations) for that

wvanbergen:
It's only a one line change, but it doesn't necessarily make natural sense that this is part of the Config, instead of part of the ProducerMessage struct or SendMessage call

eapache:
¯\_(ツ)_/¯

partitioner_test.go
```diff
@@ -120,3 +121,62 @@ func TestManualPartitioner(t *testing.T) {
 		}
 	}
 }
+
+// By default, Sarama uses the message's key to consistently assign a partition to
+// a message using hashing. If no key is not, a random partition will be chosen.
```
eapache: s/not/set

eapache:
one typo, then :shipit:

PR #360, opened 2015-03-17
Adjust producer flushing semantics

eapache:
If none of the three are set, flush as fast as possible, per normal. If *any* of
the three are set, *only* respect the ones that are set.

@Shopify/kafka


wvanbergen:
This makes much more sense :+1:

PR #368, opened 2015-03-19
producer: bugfix for aggregators getting stuck

eapache:
In circumstances where Flush.Messages and/or Flush.Bytes were set but
Flush.Frequency was not, the producer's aggregator could get stuck on a retry
because a metadata-only chaser message would not be enough on its own to trigger
a flush, and so it would sit in limbo forever.

Always trigger a flush in the aggregator when the message is a chaser. This has
the additional benefit of reducing retry latency when Flush.Frequency *is* set.

Add a test for this case.

@Shopify/kafka

### DIFF
--- a/async_producer_test.go
+++ b/async_producer_test.go
@@ -30,6 +30,22 @@ func closeProducer(t *testing.T, p AsyncProducer) {
 	wg.Wait()
 }
 
+func expectSuccesses(t *testing.T, p AsyncProducer, successes int) {
+	for i := 0; i < successes; i++ {
+		select {
+		case msg := <-p.Errors():
+			t.Error(msg.Err)
+			if msg.Msg.flags != 0 {
+				t.Error("Message had flags set")
+			}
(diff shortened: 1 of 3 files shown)


eapache:
This PR also DRYs up a pattern in the tests that was annoying me

async_producer_test.go
```diff
@@ -30,6 +30,22 @@ func closeProducer(t *testing.T, p AsyncProducer) {
...
+				t.Error("Message had flags set")
+			}
+		case msg := <-p.Successes():
+			if msg.flags != 0 {
+				t.Error("Message had flags set")
+			}
+		}
+	}
+}
+
```
wvanbergen: :+1:

wvanbergen:
This makes sense to me; if the test case went from red to green I am happy.

Would writing these tests be easier if we would have a mock Client?

eapache:
This test in particular was straightforward - I'd have to think about the other two.

PR #369, opened 2015-03-19
consumer: bugfix for broker workers getting stuck

eapache:
Very similar to #367 (on the producer): the consumer could have added new
references to a brokerConsumer that was shutting down, leading it to spin with
an error unnecessarily.

Add a test for this case.

@Shopify/kafka

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -115,10 +115,10 @@ func (c *consumer) ConsumePartition(topic string, partition int32, offset int64)
 		return nil, err
 	}
 
-	if leader, err := c.client.Leader(child.topic, child.partition); err != nil {
+	var leader *Broker
+	var err error
+	if leader, err = c.client.Leader(child.topic, child.partition); err != nil {
 		return nil, err
-	} else {
-		child.broker = leader
 	}
 
 	if err := c.addChild(child); err != nil {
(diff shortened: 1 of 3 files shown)


consumer.go
```diff
@@ -171,31 +171,39 @@ func (c *consumer) refBrokerConsumer(broker *Broker) *brokerConsumer {
...
 
-	brokerWorker := c.brokerConsumers[broker]
 	brokerWorker.refs--
 
 	if brokerWorker.refs == 0 {
 		close(brokerWorker.input)
-		delete(c.brokerConsumers, broker)
+		if c.brokerConsumers[brokerWorker.broker] == brokerWorker {
+			delete(c.brokerConsumers, brokerWorker.broker)
+		}
```
wvanbergen: So if I understand it correctly, this `if` case only triggers during a normal consumer shutdown? When we shut down because of a broker error, this `delete` will already have happened?

consumer.go
```diff
@@ -463,6 +471,7 @@ func (w *brokerConsumer) updateSubscriptionCache(newSubscriptions []*partitionCo
 }
 
 func (w *brokerConsumer) abort(err error) {
+	w.consumer.abandonBrokerConsumer(w)
```
wvanbergen: Another clarification: because we only have one goroutine dealing with this broker, there is no chance of us using the broker's connection between the occurrence of the error, and this call where we abandon it?

consumer.go
eapache: It triggers when all PartitionConsumers move off of the brokerConsumer "safely", either via being closed or via a `NotLeaderForPartition` response. When the broker blows up we abort, and the delete happens immediately to prevent new partitions from taking a reference.

consumer.go
eapache: Technically the subscriptionManager is a separate goroutine, so there are two, but effectively yes.

wvanbergen:
:cool: just was convincing myself what was going on here. :shipit:

PR #370, opened 2015-03-19
Fix go fmt makefile target

eapache:
This wasn't failing CI, but it was effectively not running the expected step so misformatted code might have snuck in accidentally.

@Shopify/kafka


wvanbergen:
:cool:

PR #371, opened 2015-03-19
Add a kaka-console-producer application

wvanbergen:
Nice as an example, and could be nice for shell scripting without the need of installing the JVM to use the one that ships with Kafka. E.g. this could be nice to ship a Kafka event on a deploy of an application.

@Shopify/kafka /cc @graemej

### DIFF
--- a/tools/kafka-console-producer/README.md
+++ b/tools/kafka-console-producer/README.md
@@ -1,0 +1,34 @@
+# kafka-console-producer
+
+A simple command line tool to produce a single message to Kafka.
+
+### Installation
+
+    go install github.com/Shopify/sarama/tools/kafka-console-producer
+
+
+### Usage
+
+    # Minimum invocation
+    kafka-console-producer -topic=test -value=value -brokers=kafka1:9092
+
+    # It will pick up a KAFKA_PEERS environment variable
+    export KAFKA_PEERS=kafka1:9092,kafka2:9092,kafka3:9092
+    kafka-console-producer -topic=test -value=value
+
+    # It will read the value from stdin by using pipes
+    echo "hello world" | kafka-console-producer -topic=test
+
+    # Specify a key:
+    echo "hello world" | kafka-console-producer -topic=test -key=key
+
+    # Partitioning: by default, kafka-console-producer will partition as follows:
+    # - manual partitioning if a -partition is provided
+    # - hash partitioning by key if a -key is provided
+    # - random partioning otherwise.
+    #
+    # You can override this using the -partitioner argument:
+    echo "hello world" | kafka-console-producer -topic=test -key=key -partitioner=random
+
+    # Display all command line options
+    kafka-console-producer -help
--- a/tools/kafka-console-producer/kafka-console-producer.go
+++ b/tools/kafka-console-producer/kafka-console-producer.go
@@ -1,0 +1,118 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"strings"
+
+	"github.com/Shopify/sarama"
+)
+
+var (
+	brokerList  = flag.String("brokers", os.Getenv("KAFKA_PEERS"), "The comma separated list of brokers in the Kafka cluster. You can also set the KAFKA_PEERS environment variable")
+	topic       = flag.String("topic", "", "REQUIRED: the topic to produce to")
+	key         = flag.String("key", "", "The key of the message to produce. Can be empty.")
+	value       = flag.String("value", "", "REQUIRED: the value of the message to produce. You can also provide the value on stdin.")
+	partitioner = flag.String("partitioner", "", "The partitioning scheme to use. Can be `hash`, `manual`, or `random`")
+	partition   = flag.Int("partition", -1, "The partition to produce to.")
+	verbose     = flag.Bool("verbose", false, "Turn on sarama logging to stderr")
+	silent      = flag.Bool("silent", false, "Turn off printing the message's topic, partition, and offset to stdout")
+
+	logger = log.New(os.Stderr, "", log.LstdFlags)
+)
+
+func main() {
+	flag.Parse()
+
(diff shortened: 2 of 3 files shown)


wvanbergen:
We could use Github's releases option to make pre-built binaries available as well, so people don't have to install the go toolchain either.

eapache:
I am split between "tools" and "examples", I feel like that distinction is pretty fine... does it make sense to just put this in examples?

tools/kafka-console-producer/kafka-console-producer.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/Shopify/sarama"
+)
+
+var (
+	brokerList  = flag.String("brokers", os.Getenv("KAFKA_PEERS"), "The comma separated list of brokers in the Kafka cluster. You can also set the KAFKA_PEERS environment variable")
+	topic       = flag.String("topic", "", "REQUIRED: the topic to produce to")
+	key         = flag.String("key", "", "The key of the message to produce. Can be empty.")
+	value       = flag.String("value", "", "REQUIRED: the value of the message to produce. You can also provide the value on stdin.")
+	partitioner = flag.String("partitioner", "hash", "The partitioning scheme to use. Can be `hash`, `manual`, or `random`")
+	partition   = flag.Int("partition", -1, "The partition to produce to. Only makes sense when -partitioner is set to `manual`")
```
eapache: if this is set we should assume `manual`

tools/kafka-console-producer/kafka-console-producer.go
```diff
@@ -0,0 +1,124 @@
...
+	verbose     = flag.Bool("verbose", false, "Turn on sarama logging to stderr")
+	silent      = flag.Bool("silent", false, "Turn off printing the message's topic, partition, and offset to stdout")
+
+	logger = log.New(os.Stderr, "", log.LstdFlags)
+)
+
+func init() {
+	topicFlag := flag.Lookup("topic")
+	topicFlag.DefValue = ""
```
eapache: what is this?

tools/kafka-console-producer/kafka-console-producer.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func main() {
+	flag.Parse()
+
+	if *brokerList == "" {
+		printUsageErrorAndExit("no -brokers specified. ALternatively, set the KAFKA_PEERS environment variable")
```
eapache: s/AL/Al

tools/kafka-console-producer/kafka-console-producer.go
```diff
@@ -0,0 +1,124 @@
...
+			printErrorAndExit(66, "Failed to read data from the standard input: %s", err)
+		}
+		valueEncoder = sarama.ByteEncoder(bytes)
+	} else {
+		printUsageErrorAndExit("-value is required, or you have to provide the value on stdin")
+	}
+
+	config := sarama.NewConfig()
+	config.Producer.RequiredAcks = sarama.WaitForAll
+	config.Producer.Partitioner = partitionerConstructor
```
eapache: create the config earlier and there won't be a need for the local `partitionerConstructor

wvanbergen:
For me the distinction is that a tool is actually useful. It has a bunch of command line flags to make it versatile. I actually want to install them on our infrastructure. Because of the versatility, the code is a bit harder to grasp. I would like to do binary releases of the tools so they are easy to install with chef.

An example on the other hand, is not really a useful application, but purely meant for studying to source code. The code is kept as simple as possible so it is easier to follow.

Does this difference warrant different directories?

tools/kafka-console-producer/kafka-console-producer.go
wvanbergen: whoops, experiment I forgot to remove

tools/kafka-console-producer/kafka-console-producer.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		printErrorAndExit(69, "Failed to open Kafka producer: %s", err)
+	}
+	defer func() {
+		if err := producer.Close(); err != nil {
+			logger.Println("Failed to close Kafka producer cleanly:", err)
+		}
+	}()
+
+	partition, offset, err := producer.SendMessage(&sarama.ProducerMessage{
```
eapache: create the ProducerMessage eariler and there won't be a need for the local `keyEncoder` and `valueEncoder`

eapache:
> Does this difference warrant different directories?

I guess, it just seems odd to have two separate subdirectories for programs that make use of the library.

tools/kafka-console-producer/kafka-console-producer.go
wvanbergen: +1. New rules:

```
    # Partitioning: by default, kafka-console-producer will partition as follows:
    # - manual partitioning if a -partition is provided
    # - hash partitioning by key if a -key is provided
    # - random partioning otherwise.
    #
    # You can override this using the -partitioner argument:
```

wvanbergen:
Updated.

tools/kafka-console-producer/README.md
```diff
@@ -0,0 +1,35 @@
...
+    echo "hello world" | kafka-console-producer -topic=test -key=key
+
+    # Partitioning: by default, kafka-console-producer will partition as follows:
+    # - manual partitioning if a -partition is provided
+    # - hash partitioning by key if a -key is provided
+    # - random partioning otherwise.
+    #
+    # You can override this using the -partitioner argument:
+    echo "hello world" | kafka-console-producer -topic=test -key=key -partitioner=random
+    echo "hello world" | kafka-console-producer -topic=test -key=key -partitioner=roundrobin
```
eapache: roundrobin on a single message will always partition to 0, but eh

wvanbergen: d'oh

eapache:
LGTM

PR #372, opened 2015-03-19
Do not use partition cache for unknown topics

wvanbergen:
If we set the partition cache, we end up return an empty list later on, which ends up causing a weird error message later on.

@Shopify/kafka for review

### DIFF
--- a/client.go
+++ b/client.go
@@ -492,7 +492,7 @@ func (client *client) tryRefreshMetadata(topics []string, retriesRemaining int) 
 		switch err.(type) {
 		case nil:
 			// valid response, use it
-			retry, err := client.update(response)
+			retry, err := client.updateMetadata(response)
 
 			if len(retry) > 0 {
 				if retriesRemaining <= 0 {
@@ -531,7 +531,7 @@ func (client *client) tryRefreshMetadata(topics []string, retriesRemaining int) 
 }
 
 // if no fatal error, returns a list of topics that need retrying due to ErrLeaderNotAvailable
-func (client *client) update(data *MetadataResponse) ([]string, error) {
+func (client *client) updateMetadata(data *MetadataResponse) ([]string, error) {
 	client.lock.Lock()
 	defer client.lock.Unlock()
 
@@ -554,23 +554,37 @@ func (client *client) update(data *MetadataResponse) ([]string, error) {
 
 	var err error
 	for _, topic := range data.Topics {
+
+		delete(client.metadata, topic.Name)
+		delete(client.cachedPartitionsResults, topic.Name)
+
 		switch topic.Err {
 		case ErrNoError:
 			break
-		case ErrLeaderNotAvailable, ErrUnknownTopicOrPartition:
-			toRetry[topic.Name] = true
-		default:
+		case ErrInvalidTopic: // don't retry, don't store partial results
 			err = topic.Err
+			continue
+		case ErrUnknownTopicOrPartition: // retry, do not store partial partition results
+			err = topic.Err
+			toRetry[topic.Name] = true
+			continue
+		case ErrLeaderNotAvailable: // retry, but store partiial partition results
+			toRetry[topic.Name] = true
+			break
+		default: // don't retry, don't store partial results
+			Logger.Printf("Unexpected topic-level metadata error: %s", topic.Err)
+			err = topic.Err
+			continue
 		}
 
 		client.metadata[topic.Name] = make(map[int32]*PartitionMetadata, len(topic.Partitions))
-		delete(client.cachedPartitionsResults, topic.Name)
 		for _, partition := range topic.Partitions {
 			client.metadata[topic.Name][partition.ID] = partition
 			if partition.Err == ErrLeaderNotAvailable {
 				toRetry[topic.Name] = true
 			}
 		}
+
 		var partitionCache [maxPartitionIndex][]int32
 		partitionCache[allPartitions] = client.setPartitionCache(topic.Name, allPartitions)
 		partitionCache[writablePartitions] = client.setPartitionCache(topic.Name, writablePartitions)
--- a/client_test.go
+++ b/client_test.go
@@ -65,6 +65,60 @@ func TestCachedPartitions(t *testing.T) {
 	safeClose(t, client)
 }
 
+func TestClientDoesntCachePartitionsForTopicsWithErrors(t *testing.T) {
+	seedBroker := newMockBroker(t, 1)
+
+	replicas := []int32{seedBroker.BrokerID()}
+
+	metadataResponse := new(MetadataResponse)
+	metadataResponse.AddBroker(seedBroker.Addr(), seedBroker.BrokerID())
+	metadataResponse.AddTopicPartition("my_topic", 1, replicas[0], replicas, replicas, ErrNoError)
+	metadataResponse.AddTopicPartition("my_topic", 2, replicas[0], replicas, replicas, ErrNoError)
+	seedBroker.Returns(metadataResponse)
+
+	config := NewConfig()
+	config.Metadata.Retry.Max = 0
+	client, err := NewClient([]string{seedBroker.Addr()}, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	metadataResponse = new(MetadataResponse)
+	metadataResponse.AddTopic("unknown", ErrUnknownTopicOrPartition)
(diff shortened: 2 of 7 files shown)


wvanbergen:
Added a producer test as well.

client.go
```diff
@@ -571,10 +571,13 @@ func (client *client) update(data *MetadataResponse) ([]string, error) {
 				toRetry[topic.Name] = true
 			}
 		}
-		var partitionCache [maxPartitionIndex][]int32
-		partitionCache[allPartitions] = client.setPartitionCache(topic.Name, allPartitions)
-		partitionCache[writablePartitions] = client.setPartitionCache(topic.Name, writablePartitions)
-		client.cachedPartitionsResults[topic.Name] = partitionCache
+
+		if topic.Err != ErrUnknownTopicOrPartition && topic.Err != ErrInvalidTopic {
```
eapache: this doesn't delete the old cache in the case that a topic we already knew about was deleted

client_test.go
```diff
@@ -65,6 +65,62 @@ func TestCachedPartitions(t *testing.T) {
...
+func TestClientDoesntCachePartitionsForTopicsWithErrors(t *testing.T) {
+	seedBroker := newMockBroker(t, 1)
+
+	replicas := []int32{seedBroker.BrokerID()}
+
+	metadataResponse := new(MetadataResponse)
+	metadataResponse.AddBroker(seedBroker.Addr(), seedBroker.BrokerID())
+	metadataResponse.AddTopicPartition("my_topic", 1, replicas[0], replicas, replicas, ErrNoError)
+	metadataResponse.AddTopicPartition("my_topic", 2, replicas[0], replicas, replicas, ErrNoError)
+	seedBroker.Returns(metadataResponse)
```
eapache: this response can just be empty, there's no need for these

client_test.go
```diff
@@ -65,6 +65,62 @@ func TestCachedPartitions(t *testing.T) {
...
+
+	config := NewConfig()
+	config.Metadata.Retry.Max = 0
+	client, err := NewClient([]string{seedBroker.Addr()}, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	metadataResponse = new(MetadataResponse)
+	metadataResponse.AddBroker(seedBroker.Addr(), seedBroker.BrokerID())
```
eapache: broker not necessary in this response

client_test.go
eapache: ah I see you actually test this later, nevermind

eapache: broker is still not necessary in any of these responses

sync_producer_test.go
```diff
@@ -97,6 +97,36 @@ func TestConcurrentSyncProducer(t *testing.T) {
...
+
+	config := NewConfig()
+	config.Metadata.Retry.Max = 0
+	config.Producer.Retry.Max = 0
+
+	producer, err := NewSyncProducer([]string{broker.Addr()}, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer safeClose(t, producer)
```
eapache: I prefer not to defer in tests because it makes it really hard to tell what's hanging if it hangs while running defers

client.go
eapache: I think I prefer to just unconditionally delete the cache at the top of the loop for all partitions, and let it get filled in again for those we care about

eapache: Additional thought is that we still create the `client.metadata` map entry even if the cache is gone, which could be a little confusing (that the "real" data is `[]` but the cache is `nil`)

wvanbergen: The cache is deleted on line 567

eapache: ah whoops, I'm dumb, sorry.

I believe we may want to similarly delete from `client.metadata` and guard that creation though, still trying to figure that out

eapache:
Don't forget to put this down as a bugfix in the changelog :)

client_test.go
wvanbergen: I thouhgt it was a good idea to ensure we are now not *always* busting the cache.

client.go
eapache: So I think something like the following pseudocode:
```go
for topics {
    delete(client.metadata, ...)
    delete(client.cachedPartitionResults, ...)
    switch {
         NoError: break
         UnknownTopic: set retry, continue
         LeaderNotAvailable: set retry, break
         InvalidTopic: set err, continue
         default: set err, break
    }
    // fill in client.metadata and partitionCache unconditionally
}
```

Does this make sense?

wvanbergen: I think the issue is that we don't cache the error value for `cachedPartitions()`, but we do for `cachedLeader()`. `cachedMetadata()` doesn't either, but the struct that it returns contains the `Err` value.

wvanbergen:
Updated.

client.go
```diff
@@ -554,23 +554,36 @@ func (client *client) update(data *MetadataResponse) ([]string, error) {
...
+
+		delete(client.metadata, topic.Name)
+		delete(client.cachedPartitionsResults, topic.Name)
+
 		switch topic.Err {
 		case ErrNoError:
 			break
-		case ErrLeaderNotAvailable, ErrUnknownTopicOrPartition:
+		case ErrInvalidTopic: // don't retry, don't store partial results
+			err = topic.Err
```
wvanbergen: This makes this method return `ErrInvalidTopic`, even if all the other topics are OK.

eapache: which is fine, I think - we only get this if you explicitly pass in "SomeInvalidTopic\\\\\" to `RefreshMetadata`, so returning `ErrInvalidTopic` in that case is correct, even if not all topics were invalid

wvanbergen: In that case we should probably also set the error on `ErrUnknownTopicOrPartition`, so that gets returned to the user as well.

metadata_response.go
```diff
@@ -182,25 +182,31 @@ func (m *MetadataResponse) AddBroker(addr string, id int32) {
 	m.Brokers = append(m.Brokers, &Broker{id: id, addr: addr})
 }
 
-func (m *MetadataResponse) AddTopicPartition(topic string, partition, brokerID int32, replicas, isr []int32, err KError) {
-	var match *TopicMetadata
+func (m *MetadataResponse) AddTopic(topic string, err KError) *TopicMetadata {
```
eapache: Adding an API requires a minor version bump in semver (1.1.0 vs 1.0.1). I'm torn between just making this private or eating the version bump.

client.go
eapache: Ah, I thought we checked the error return before the retry return, but we do not, so this is OK, good catch :+1:

metadata_response.go
wvanbergen: Let's just do the version bump; there's no good reason this method should be hidden if `AddTopicPartition` is exported.

eapache:
:shipit:

client.go
wvanbergen: Fixed.

metadata_response.go
eapache: Should we doc the new API in the changelog then?

functional_test.go
```diff
@@ -124,6 +124,25 @@ func TestFuncMultiPartitionProduce(t *testing.T) {
...
+func TestProducingToInvalidTopic(t *testing.T) {
+	checkKafkaAvailability(t)
+
+	producer, err := NewSyncProducer(kafkaBrokers, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if _, _, err := producer.SendMessage(&ProducerMessage{Topic: "in/valid"}); err != ErrInvalidTopic {
+		t.Log("Expected ErrInvalidTopic, found", err)
```
eapache: should these be `t.Error` instead of `t.Log`?

eapache: This is failing on the 0.8.1 CI.

eapache: Or is that *why* they're logs, I think you're ahead of me :)

wvanbergen:
Interestingly, when using an invalid topic name, Kafka returns `InvalidTopic` **only** if automatic topic creation is enabled. If it is disabled, we get `UnknownTopicOrPartition`

wvanbergen:
Anyway, if CI is happy am I happy now

eapache:
The checks must be in a funny order. And of course pre-0.8.2 it just returns Unknown regardless. CI is green, ship it.

metadata_response.go
wvanbergen: I don't really feel the need to; it will just show up in the godoc.

PR #373, opened 2015-03-19
producer: wrap partitioning in a circuit-breaker

eapache:
This prevents repeatedly broken topics (e.g. ones that don't even exist) from
choking throughput for other topics.

@Shopify/kafka 

No tests because:
- the breaker itself is well-tested in https://github.com/eapache/go-resiliency/blob/master/breaker/breaker_test.go
- the change itself is stupidly simple
- I'm sick of writing complex producer tests

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -265,10 +265,13 @@ func (p *asyncProducer) topicDispatcher() {
 func (p *asyncProducer) partitionDispatcher(topic string, input chan *ProducerMessage) {
 	handlers := make(map[int32]chan *ProducerMessage)
 	partitioner := p.conf.Producer.Partitioner(topic)
+	breaker := breaker.New(3, 1, 10*time.Second)
 
 	for msg := range input {
 		if msg.retries == 0 {
-			err := p.assignPartition(partitioner, msg)
+			err := breaker.Run(func() error {
+				return p.assignPartition(partitioner, msg)
+			})
 			if err != nil {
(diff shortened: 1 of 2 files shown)


async_producer.go
```diff
@@ -265,10 +265,13 @@ func (p *asyncProducer) topicDispatcher() {
 func (p *asyncProducer) partitionDispatcher(topic string, input chan *ProducerMessage) {
 	handlers := make(map[int32]chan *ProducerMessage)
 	partitioner := p.conf.Producer.Partitioner(topic)
+	breaker := breaker.New(3, 1, 10*time.Second)
```
wvanbergen: We have a bunch of circuit breakers with hard-coded settings now. Should we make the circuit breaker options available in the `Config` struct?

wvanbergen: Given the amount of metadata retries we do with backoff, 3 times in 10 seconds may actually be quite unlikely.

wvanbergen: actually nvm. This is me not understanding the breaker parameters.

wvanbergen:
LGTM.

async_producer.go
eapache: "a bunch" is 2, and they currently have the same settings. We might want to expose those parameters, I'm not sure if anybody really actually cares to tweak them.

eapache: In other words, I just don't care to do that work until somebody actually wants or needs it.

PR #374, opened 2015-03-20
[experimental] Optionally batch metadata requests

eapache:
When a busy broker fails, many goroutines in a producer or consumer may
simultaneously try to refresh metadata for what are only realistically a handful
of topics. Permit batching these requests together with a slight delay to reduce
the number of actual network requests needed.

As it currently stands this has limitation in that it returns the same error to all callers, so if your legit request happens to get batched with an invalid topic, you will get `ErrInvalidTopic` back incorrectly.

cc @Shopify/kafka still playing with this, do not merge

### DIFF
--- a/client.go
+++ b/client.go
@@ -4,6 +4,8 @@ import (
 	"sort"
 	"sync"
 	"time"
+
+	"github.com/eapache/go-resiliency/batcher"
 )
 
 // Client is a generic Kafka client. It manages connections to one or more Kafka brokers.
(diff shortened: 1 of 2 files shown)


eapache:
We *might* want to pursue this by batching at the per-topic level, but this does not work well as-is.

wvanbergen:
This could work well for multiple topics too, but we'd need to split up the response.

eapache:
Requests for "refresh all" would still have to be done separately.

PR #376, opened 2015-03-20
client: given both retries and errors, prefer retries

eapache:
Now that some errors (e.g. UnknownTopicOrParition) set *both* a retry flag and
an error, we should handle retries first. This is sort-of non-idiomatic in that
`updateMetadata` now returns a non-nil error with a "valid" result, but I think
that's OK.

@Shopify/kafka follow-up to #372

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -199,6 +199,92 @@ func TestClientMetadata(t *testing.T) {
 	safeClose(t, client)
 }
 
+func TestClientReceivingUnknownTopic(t *testing.T) {
+	seedBroker := newMockBroker(t, 1)
+
+	metadataResponse1 := new(MetadataResponse)
+	seedBroker.Returns(metadataResponse1)
+
+	config := NewConfig()
+	config.Metadata.Retry.Max = 0
+	client, err := NewClient([]string{seedBroker.Addr()}, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	metadataUnknownTopic := new(MetadataResponse)
+	metadataUnknownTopic.AddTopic("new_topic", ErrUnknownTopicOrPartition)
+	seedBroker.Returns(metadataUnknownTopic)
+
+	if err := client.RefreshMetadata("new_topic"); err != ErrUnknownTopicOrPartition {
+		t.Error("ErrUnknownTopicOrPartition expected, got", err)
+	}
+
+	// If we are asking for the leader of a partition of the non-existing topic.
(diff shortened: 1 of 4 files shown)


wvanbergen:
+1 this makes sense

client_test.go
```diff
@@ -199,6 +199,100 @@ func TestClientMetadata(t *testing.T) {
...
+	if err := client.RefreshMetadata("new_topic"); err != ErrUnknownTopicOrPartition {
+		t.Error("ErrUnknownTopicOrPartition expected, got", err)
+	}
+
+	// If we are asking for the leader of a partition of the non-existing topic.
+	// we will request metadata again.
+
+	metadataResponse3 := new(MetadataResponse)
+	metadataResponse3.AddTopic("new_topic", ErrUnknownTopicOrPartition)
+	seedBroker.Returns(metadataResponse3)
```
eapache: you can just re-use `metadataResponse2` here, there's no need to construct a new one that's exactly the same

client_test.go
```diff
@@ -199,6 +199,100 @@ func TestClientMetadata(t *testing.T) {
 	safeClose(t, client)
 }
 
+func TestClientReceivingUnknownTopic(t *testing.T) {
+	seedBroker := newMockBroker(t, 1)
+
+	metadataResponse1 := new(MetadataResponse)
+	metadataResponse1.AddBroker(seedBroker.Addr(), seedBroker.BrokerID())
```
eapache: I don't think we need this line?

client_test.go
```diff
@@ -199,6 +199,100 @@ func TestClientMetadata(t *testing.T) {
...
+	}
+
+	// If we are asking for the leader of a partition that didn't have a leader before,
+	// we will do another metadata request.
+
+	metadataResponse3 := new(MetadataResponse)
+	metadataResponse3.AddTopic("new_topic", ErrLeaderNotAvailable)
+	metadataResponse3.AddTopicPartition("new_topic", 0, leader.BrokerID(), replicas, replicas, ErrNoError)
+	metadataResponse3.AddTopicPartition("new_topic", 1, -1, replicas, []int32{}, ErrLeaderNotAvailable)
+	seedBroker.Returns(metadataResponse3)
```
eapache: as above, you can just re-use the same response here

eapache:
Good tests.

(It would be nice to also have tests that the retry behaviour is sane - somehow I missed writing any of those the first time around - but one thing at a time).

PR #378, opened 2015-03-21
Better consumer offset handling

wvanbergen:
- Add `Client.GetOffsetRange` to get the available range of offsets for a partition in one request. I had to make some changes to `OffsetRequest` and `OffsetResponse` to make it possible to ask for more than one offset per partition, but no changes to the public API were necessary.

- When calling ConsumePartition, always check whether the offset is within the offset range. This means we now have to do an OffsetRequest for every `ConsumePartition` call, even if the offset is provided. The good news is that the method will immediately return an error and never start a goroutine, instead of starting the goroutine and returning an error in the `Errors()` channel which you can easily ignore.

@Shopify/kafka

### DIFF
--- a/client.go
+++ b/client.go
@@ -144,7 +144,6 @@ func (client *client) Config() *Config {
 }
 
 func (client *client) Close() error {
-	// Check to see whether the client is closed
 	if client.Closed() {
 		// Chances are this is being called from a defer() and the error will go unobserved
 		// so we go ahead and log the event in this case.
@@ -177,7 +176,6 @@ func (client *client) Closed() bool {
 }
 
 func (client *client) Topics() ([]string, error) {
-	// Check to see whether the client is closed
 	if client.Closed() {
 		return nil, ErrClosedClient
 	}
@@ -194,7 +192,6 @@ func (client *client) Topics() ([]string, error) {
 }
 
 func (client *client) Partitions(topic string) ([]int32, error) {
-	// Check to see whether the client is closed
 	if client.Closed() {
 		return nil, ErrClosedClient
 	}
@@ -217,7 +214,6 @@ func (client *client) Partitions(topic string) ([]int32, error) {
 }
 
 func (client *client) WritablePartitions(topic string) ([]int32, error) {
-	// Check to see whether the client is closed
 	if client.Closed() {
 		return nil, ErrClosedClient
 	}
@@ -271,6 +267,10 @@ func (client *client) Replicas(topic string, partitionID int32) ([]int32, error)
 }
 
 func (client *client) Leader(topic string, partitionID int32) (*Broker, error) {
+	if client.Closed() {
+		return nil, ErrClosedClient
+	}
+
 	leader, err := client.cachedLeader(topic, partitionID)
 
 	if leader == nil {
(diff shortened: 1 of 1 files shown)


eapache:
> make it possible to ask for more than one offset per partition

Is that even possible? It's never explicitly stated, but I'd always assumed based on the layout of the protocol that it wasn't really supported to include multiple blocks for a single partition.

wvanbergen:
Apparently it works in Kafka 0.8.1 and 0.8.2. The OffsetRequest documentation is especially unclear, but this led me to believe that is was supported:

> The response contains the starting offset of each segment for the requested partition as well as the "log end offset" i.e. the offset of the next message that would be appended to the given partition.

You can even do more than two, as long as you request the timestamps in descending order.

client.go
```diff
@@ -292,6 +297,42 @@ func (client *client) RefreshMetadata(topics ...string) error {
 	return client.tryRefreshMetadata(topics, client.conf.Metadata.Retry.Max)
 }
 
+func (client *client) GetOffsetRange(topic string, partitionID int32) (int64, int64, error) {
```
eapache: Why do all this work when you could have just made two `GetOffset` calls?

wvanbergen: 1 request vs. 2 requests.

offset_request.go
```diff
@@ -12,7 +12,7 @@ func (r *offsetRequestBlock) encode(pe packetEncoder) error {
 }
 
 type OffsetRequest struct {
-	blocks map[string]map[int32]*offsetRequestBlock
+	blocks map[string]map[int32][]*offsetRequestBlock
```
eapache: since there's no need for uniqueness per partition anymore, this can just become `map[string][]*offsetRequestBlock` and you can put a field `partition int32` into `offsetRequestBlock`.

eapache: Since this is apparently legit (who knew) we should probably update all the other request types as well - perhaps that deserves a separate PR before this one.

client.go
eapache: it only happens once on startup, doesn't seem like a big deal (premature optimization and all that)

wvanbergen: In general I find this something that is generally useful. E.g., for monitoring purposes, you'd also want to know the available range. Of course, we could do two requests for this purpose as well.

offset_request.go
```diff
@@ -51,16 +53,20 @@ func (r *OffsetRequest) version() int16 {
...
+		r.blocks = make(map[string]map[int32][]*offsetRequestBlock)
 	}
 
 	if r.blocks[topic] == nil {
-		r.blocks[topic] = make(map[int32]*offsetRequestBlock)
+		r.blocks[topic] = make(map[int32][]*offsetRequestBlock)
+	}
+
+	if r.blocks[topic][partitionID] == nil {
+		r.blocks[topic][partitionID] = make([]*offsetRequestBlock, 0, 1)
```
eapache: this is unnecessary, `nil` is a valid thing to pass to `append`

wvanbergen: TIL

wvanbergen:
WTF :/ Github marked this as merged after I rebased.

wvanbergen:
Closing in favor of #418

PR #380, opened 2015-03-22
client: connect to a random seed broker

eapache:
Rather than walking them in order, walk them randomly using go's map iterator.
This should lighten the load on the first seed in a lot of failover scenarios.

Also rework a bunch of how seeds are stored:
- store both seeds and dead brokers as Broker objects, not addresses, to avoid
  constructing new Brokers all the time
- lazily set `seedBroker` in `any()` rather than eagerly in the constructor and
  in `disconnectBroker()`
- never add non-seed brokers to `deadBrokers`
- substantially simplify `resurrectDeadBrokers`

This doesn't break the predictability of any of our tests, fortunately, because
we only ever pass one seed in anyways.

@Shopify/kafka

### DIFF
--- a/client.go
+++ b/client.go
@@ -69,9 +69,9 @@ type client struct {
 	// the broker addresses given to us through the constructor are not guaranteed to be returned in
 	// the cluster metadata (I *think* it only returns brokers who are currently leading partitions?)
 	// so we store them separately
-	seedBrokerAddrs []string
-	seedBroker      *Broker
-	deadBrokerAddrs map[string]none
+	seedBroker  *Broker
+	seedBrokers map[*Broker]none
+	deadBrokers map[*Broker]none
 
 	brokers  map[int32]*Broker                       // maps broker ids to brokers
 	metadata map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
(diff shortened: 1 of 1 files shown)


wvanbergen:
"Randomly" means: differently on different Client instances, right? One client instance will always iterate through the list the same way, based on the address of the Broker struct?

wvanbergen:
NVM, just found this in the docs: 

> Since Go 1 the runtime randomizes map iteration order, as programmers relied on the stable iteration order of the previous implementation.

wvanbergen:
This is a nice simplification. :+1: 

I do want to test its behaviour under load though before merging this.

eapache:
> Question: if it is wrong to depend on stable iteration, is it also wrong to depend on random iteration?

I thought it was actually part of the language spec, but I guess not:

>  The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next. If map entries that have not yet been reached are removed during iteration, the corresponding iteration values will not be produced. If map entries are created during iteration, that entry may be produced during the iteration or may be skipped. The choice may vary for each entry created and from one iteration to the next. If the map is nil, the number of iterations is 0.

So, yes, it is technically wrong. I guess we should be using https://golang.org/pkg/math/rand/#Perm

eapache:
Ugh, there is no efficient (spec-compliant) way to take a random element from a map at all. I took a crack at writing something with `rand` but it got really messy. I guess there's no point in doing this.

wvanbergen:
I still like the simplification of using Broker structs though. Worth salvaging from this PR?

eapache:
In progress :)

PR #381, opened 2015-03-23
[WIP] HTTP SSE consumer example

wvanbergen:
This example HTTP server allows you to consume a partition as an SSE stream. 

### TODO
- [ ] In order to write decent tests for this, I will need a MockClient.

### DIFF
--- a/examples/http_sse/README.md
+++ b/examples/http_sse/README.md
@@ -1,0 +1,3 @@
+# SSE server example
+
+This example HTTP server provides a Kafka partition as an SSE stream.
--- a/examples/http_sse/http_sse.go
+++ b/examples/http_sse/http_sse.go
@@ -1,0 +1,199 @@
+package main
+
+import (
+	"github.com/Shopify/sarama"
+
+	"flag"
+	"fmt"
+	"html/template"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+const (
+	templateFilename = "sse_template.html"
+)
+
+var (
+	addr    = flag.String("addr", ":8080", "The address to bind to")
+	brokers = flag.String("brokers", os.Getenv("KAFKA_PEERS"), "The Kafka brokers to connect to, as a comma separated list")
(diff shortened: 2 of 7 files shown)


wvanbergen:
For a potential mock `Client`, I noticed that `Leader` still returns a `*Broker` type, so if it lives in another package we will not be able to use it from sarama itself :(

eapache:
I am coming to the conclusion that we're not going to have any choice but to duplicate our mocks for testing of sarama itself.

examples/http_sse/README.md
```diff
@@ -0,0 +1,3 @@
+# SSE server example
+
+This example HTTP server provides a Kafka partition as an SSE stream.
```
eapache: what does SSE stand for?

examples/http_sse/http_sse.go
```diff
@@ -0,0 +1,199 @@
...
+	if *brokers == "" {
+		flag.PrintDefaults()
+		os.Exit(1)
+	}
+
+	brokerList := strings.Split(*brokers, ",")
+	log.Printf("Kafka brokers: %s", strings.Join(brokerList, ", "))
+
+	config := sarama.NewConfig()
+	client, err := sarama.NewClient(brokerList, config)
```
eapache: you can just pass `nil` for all default config

examples/http_sse/http_sse.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+type SSEServer struct {
+	client sarama.Client
+}
+
+func (s *SSEServer) Handler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		matches := pathMatcher.FindStringSubmatch(r.URL.Path)
+		if len(matches) != 4 {
```
eapache: I only see 3 match groups in your regex?

wvanbergen: The first entry is the full match, the rest are the captures groups.

examples/http_sse/README.md
wvanbergen: Server-side events. I should probably add this to the README.

eapache:
Closing as stale.

PR #382, opened 2015-03-23
Simplify how we manage seed brokers

### LINKED ISSUE #375: Don't log "not connected" when calling `Close()`
Since we have lazy connections, not all registered brokers will actually be connected. When we shut down the client, we should not try to disconnect those, or at least not log the error.

eapache:
Create all the Broker objects themselves right up front rather than storing the
addresses. Don't eagerly open the seed connection in the constructor, we open it
lazily in `any()` already. Never add non-seed brokers to the dead set.
Substantially simplify `resurrectDeadBrokers`. Don't try and close a broker that
isn't even connected.

The good bits as extracted from #380. Fixes #375.

@Shopify/kafka

### DIFF
--- a/client.go
+++ b/client.go
@@ -69,9 +69,8 @@ type client struct {
 	// the broker addresses given to us through the constructor are not guaranteed to be returned in
 	// the cluster metadata (I *think* it only returns brokers who are currently leading partitions?)
 	// so we store them separately
-	seedBrokerAddrs []string
-	seedBroker      *Broker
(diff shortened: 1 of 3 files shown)


wvanbergen:
Can we add a test that touches deadBrokers? I don't think we have one yet, and it's a vital piece of making sure the client keeps running.

eapache:
Test added.

wvanbergen:
LGTM.

PR #383, opened 2015-03-23
Totally remove consumer.children

eapache:
There's something fishy about this change, since it seems really odd for the
consumer not to actually have any references to its children, but I can't
actually come up with any problems, it's much simpler, and it permits (in
theory) multiple PartitionConsumers for the same partition to share a single
consumer.

@Shopify/kafka

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
   - KAFKA_PEERS=localhost:6667,localhost:6668,localhost:6669,localhost:6670,localhost:6671
   - KAFKA_INSTALL_ROOT=/home/travis/kafka
   - KAFKA_HOSTNAME=localhost
+  - DEBUG=true
   matrix:
   - KAFKA_VERSION=0.8.1.1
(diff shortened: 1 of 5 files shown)


eapache:
> multiple PartitionConsumers for the same partition to share a single consumer

Making this actually work probably requires changes to the FetchRequest/FetchResponse as well

.travis.yml
```diff
@@ -9,6 +9,7 @@ env:
   - KAFKA_PEERS=localhost:6667,localhost:6668,localhost:6669,localhost:6670,localhost:6671
   - KAFKA_INSTALL_ROOT=/home/travis/kafka
   - KAFKA_HOSTNAME=localhost
+  - DEBUG=true
```
wvanbergen: You may need to change this to LOGGER=true to make it work; I think DEBUG is reserved by Travis or something.

eapache:
Kafka simply doesn't support this, oh well.

wvanbergen:
Too bad!

PR #384, opened 2015-03-23
Refactor partitionConsumer a bit

eapache:
Move all redispatch/error logic out of `handleResponse` and into a single if
statement, based on the new return value of `handleResponse`.

@Shopify/kafka


wvanbergen:
:shipit:

Issue #325, opened 2015-03-07
PartitionConsumer.Close() hangs frequently


dim:
I have a difficult time to recreate the exact issue, since it is intermittent, but it happens frequently enough to be a problem. Calls to PartitionConsumer.Close() tend to hang in https://github.com/Shopify/sarama/blob/6e7ef3c/consumer.go#L341 because the errors channel is never closed by the dispatcher in https://github.com/Shopify/sarama/blob/6e7ef3c/consumer.go#L260. When I dump my goroutines, I can see that it's hanging in https://github.com/Shopify/sarama/blob/6e7ef3c/consumer.go#L231. Does `AsyncClose()` need an additional `client.trigger <- none{}`?

eapache:
Could you provide the complete goroutine trace? The `client.trigger` channel should be closed by `updateSubscriptionCache` when it detects that the `dying` channel has closed.

dim:
Here's the relevant bits:

```
goroutine 44 [chan receive]:
github.com/Shopify/sarama.(*Broker).responseReceiver(0xc2080b40e0)
  /home/me/go/src/github.com/Shopify/sarama/broker.go:332 +0xe3
github.com/Shopify/sarama.*Broker.(github.com/Shopify/sarama.responseReceiver)·fm()
  /home/me/go/src/github.com/Shopify/sarama/broker.go:84 +0x27
github.com/Shopify/sarama.withRecover(0xc2080d61b0)
  /home/me/go/src/github.com/Shopify/sarama/utils.go:42 +0x3a
created by github.com/Shopify/sarama.func·001
  /home/me/go/src/github.com/Shopify/sarama/broker.go:84 +0x535

goroutine 57 [chan receive]:
github.com/Shopify/sarama.(*Broker).responseReceiver(0xc2080b4540)
  /home/me/go/src/github.com/Shopify/sarama/broker.go:332 +0xe3
github.com/Shopify/sarama.*Broker.(github.com/Shopify/sarama.responseReceiver)·fm()
  /home/me/go/src/github.com/Shopify/sarama/broker.go:84 +0x27
github.com/Shopify/sarama.withRecover(0xc2080d6760)
  /home/me/go/src/github.com/Shopify/sarama/utils.go:42 +0x3a
created by github.com/Shopify/sarama.func·001
  /home/me/go/src/github.com/Shopify/sarama/broker.go:84 +0x535

goroutine 54 [select]:
[...]

PR #386, opened 2015-03-23
Process results of a single fetch request in parallel

eapache:
And abandon partitions that take too long (e.g. because the ~~client~~ user-space reading goroutine is stuck).

@wvanbergen this is a draft attempt at fixing #325 but it kind of sucks. The bulk of the diff is just moving the `handleResponse` method from being on the brokerConsumer to being on the partitionConsumer (split out into #387)

### DIFF
--- a/consumer_test.go
+++ b/consumer_test.go
@@ -46,9 +46,9 @@ func TestConsumerOffsetManual(t *testing.T) {
 		}
 	}
 
+	leader.Close()
 	safeClose(t, consumer)
 	safeClose(t, master)
-	leader.Close()
 }
 
 func TestConsumerLatestOffset(t *testing.T) {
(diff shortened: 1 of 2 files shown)


consumer_test.go
```diff
@@ -278,15 +266,15 @@ func TestConsumerInterleavedClose(t *testing.T) {
...
-	fetchResponse := new(FetchResponse)
-	fetchResponse.AddMessage("my_topic", 0, nil, ByteEncoder([]byte{0x00, 0x0E}), int64(0))
-	leader.Returns(fetchResponse)
-
 	c1, err := master.ConsumePartition("my_topic", 1, 0)
 	if err != nil {
 		t.Fatal(err)
 	}
 
+	time.Sleep(50 * time.Millisecond)
```
eapache: :'(

eapache:
I think I have a better idea (we'll see once I've actually written it :P) but this was enlightening regardless.

PR #387, opened 2015-03-23
Put `handleResponse` method on partitionConsumer

eapache:
It didn't actually use the brokerConsumer at all anymore, so it makes more sense
here.

@Shopify/kafka


wvanbergen:
:shipit:

PR #389, opened 2015-03-24
producer: make Flush.Frequency behaviour better

eapache:
Previously the timer would tick every Flush.Frequency regardless of if a message
was actually queued. This meant that if no message arrived for Flush.Frequency
then the next message would be sent immediately, which is unexpected if not
*exactly* wrong.

With this change the timer is only started when the first message arrives, and
is cleared when a flush occurs. This should result in slightly better batching
for low-volume topics at the result of slightly higher latency (although the
delay will still never be more than Flush.Frequency).

@Shopify/kafka

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -424,7 +424,7 @@ func (p *asyncProducer) messageAggregator(broker *Broker, input chan *ProducerMe
 	var (
 		timer            <-chan time.Time
 		buffer           []*ProducerMessage
-		doFlush          chan []*ProducerMessage
+		flushTriggered   chan []*ProducerMessage
 		bytesAccumulated int
 		defaultFlush     bool
 	)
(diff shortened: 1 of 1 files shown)


async_producer.go
```diff
@@ -421,17 +421,13 @@ func (p *asyncProducer) leaderDispatcher(topic string, partition int32, input ch
...
-	}
-
-	var buffer []*ProducerMessage
-	var doFlush chan []*ProducerMessage
-	var bytesAccumulated int
-	var defaultFlush bool
+	var (
+		timer            <-chan time.Time
+		buffer           []*ProducerMessage
+		doFlush          chan []*ProducerMessage
```
wvanbergen: `flushQueue` ?

wvanbergen:
Looks innocent enough to me :)

async_producer.go
eapache: `flushTriggered`?

wvanbergen: `triggerFush`! :P

PR #390, opened 2015-03-24
Fix OffsetCommitRequest

eapache:
Apparently the kafka protocol spec has been inconsistent for a while,
documenting v1 of this message but claiming it was still v0. Jun finally updated
it, so now we can properly implement *both* versions correctly.

@Shopify/kafka

### DIFF
--- a/offset_commit_request.go
+++ b/offset_commit_request.go
@@ -2,6 +2,7 @@ package sarama
 
 // ReceiveTime is a special value for the timestamp field of Offset Commit Requests which
 // tells the broker to set the timestamp to the time at which the request was received.
+// The timestamp is only used if message version 1 is used, which requires kafka 0.8.2.
 const ReceiveTime int64 = -1
 
 type offsetCommitRequestBlock struct {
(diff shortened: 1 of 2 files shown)


wvanbergen:
Do we have to send the right version of the message to the right Kafka version? Or can kafka 0.8.2 handle the old message as well?

eapache:
AFAIK 0.8.2 can handle both, otherwise you wouldn't necessarily be able to upgrade without downtime.

eapache:
The spec now says "v0 (supported in 0.8.1 or later)"

wvanbergen:
Yeah; I think Kafka 0.8.1 can read the message, but it doesn't do anything useful with it. The functional test for offset management was failing pretty hard on kafka 0.8.1.1 when using the v0 message format.

eapache:
We weren't sending the v0 format before, we were sending v1 and saying v0 because the wiki was wrong :(

wvanbergen:
Aha OK. :+1: in that case

PR #392, opened 2015-03-24
Randomly iterate through seed brokers

eapache:
@Shopify/kafka without relying on golang's `map` this time. It's a bit hairy, not sure if it's worth it...

### DIFF
--- a/client.go
+++ b/client.go
@@ -1,6 +1,7 @@
 package sarama
 
 import (
+	"math/rand"
 	"sort"
 	"sync"
 	"time"
@@ -64,11 +65,13 @@ const (
 
 type client struct {
 	conf   *Config
+	rand   *rand.Rand
 	closer chan none
 
 	// the broker addresses given to us through the constructor are not guaranteed to be returned in
 	// the cluster metadata (I *think* it only returns brokers who are currently leading partitions?)
 	// so we store them separately
+	curSeed     *Broker
 	seedBrokers []*Broker
 	deadSeeds   []*Broker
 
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -69,6 +70,8 @@ type client struct {
 	// the broker addresses given to us through the constructor are not guaranteed to be returned in
 	// the cluster metadata (I *think* it only returns brokers who are currently leading partitions?)
 	// so we store them separately
```
wvanbergen: This comment is now out of place

eapache: fixed by moving `rand` up beside `conf`

wvanbergen:
Did I understand this correctly:

- `curSeed`: the current seed broker
- `seedBrokers`: list of brokers that are not the active seed broker, and also not marked as dead.
- `deadBrokers`: list of brokers that are considered dead.

`any()` Algorithm
1. Use `curSeed`
2. If that doesn't work, grab any form the `seedBrokers`
3. If there's no more `seedBrokers`, grab any broker from the broker list. (should that be random too?)

wvanbergen:
I feel `resurrectDeadBrokers()` should be part of `any()` instead of `tryRefreshMetdata`.

eapache:
Yes, your understanding is correct.

> If there's no more seedBrokers, grab any broker from the broker list. (should that be random too?)

I guess so, but unfortunately there's no efficient way to pick a random element from a map, the best I've been able to come up with requires iterating the entire map. I guess we'd have to maintain the ID->broker map *and* a separate slice of the brokers just for the randomness, which really sucks.

We're back to "It's a bit hairy, not sure if it's worth it..."

eapache:
> I feel resurrectDeadBrokers() should be part of any() instead of tryRefreshMetdata.

Ya, that makes some sense but they happen at different times in the retry loop: `any` can be called several times without spending a retry, while `resurrectDeadBrokers.

wvanbergen:
Yeah, it'll have slightly different behavior, but is that bad? The implementation of both `any()` and `tryRefreshMetadata()` will be simpler.

eapache:
I don't understand what you mean, my understanding is that it would totally break the logic, perhaps you'd better write it as a POC to show me.

eapache:
More discussion will happen on this, but it's probably not going in as-is.

PR #394, opened 2015-03-25
Add retry logic to GetOffset() just like Leader()

eapache:
Which is: refresh the metadata and retry once, in case e.g. our connection or
metadata is stale.

Also add a test for it.

@Shopify/kafka

### DIFF
--- a/client.go
+++ b/client.go
@@ -292,31 +292,16 @@ func (client *client) RefreshMetadata(topics ...string) error {
 }
 
 func (client *client) GetOffset(topic string, partitionID int32, time int64) (int64, error) {
-	broker, err := client.Leader(topic, partitionID)
+	offset, err := client.getOffset(topic, partitionID, time)
+
 	if err != nil {
-		return -1, err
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -292,31 +292,16 @@ func (client *client) RefreshMetadata(topics ...string) error {
...
-	if len(block.Offsets) != 1 {
-		return -1, ErrOffsetOutOfRange
+		if err := client.RefreshMetadata(topic); err != nil {
+			return -1, err
+		}
+		return client.getOffset(topic, partitionID, time)
 	}
 
-	return block.Offsets[0], nil
+	return offset, err
```
wvanbergen: So this will basically do endless retries, as long a `RefreshMetadata` doesn't return an error? Is that something we should watch out for?

wvanbergen: Actually nvm, it just does one retry. I imagined a `for` loop that isn't there.

wvanbergen:
This is a nice improvement :+1:

PR #395, opened 2015-03-25
Cleanup TestFuncMultiPartitionProduce

eapache:
Use SyncProducer so the various goroutines get their own success/error back and
not somebody elses. Don't call Fatal in a goroutine, it doesn't work. Don't pass
the wait group as a parameter, it doesn't change so it's unnecessary.

@Shopify/kafka


wvanbergen:
:+1: pretty straightforward cleanup.

yagnik:
:+1:

PR #396, opened 2015-03-25
Improve memory of producer retry queue

eapache:
The old version just used a simple slice as a queue, using append and slice[1:].
This had correct (amortized O(1)) running time, but did not behave well with the
garbage collector. In particular:
- elements removed from the slice were still pointed to by the underlying array,
  preventing them from being collected until the array was resized
- the array would only be resized when its current capacity was "full", even if
  it had a substantial amount of free space it was wasting

Switch instead to my queue implementation (which I wrote a while ago to solve
these exact problems somewhere else). It uses a ringbuffer that is guaranteed to
stay with a factor of 2 size of the actual number of messages stored, it
generates a lot less garbage when the input and output rates are about the same,
and it nils removed elements in the underlying array so the garbage collector
can pick them up at its leisure.

The only possible concern is that it uses `interface{}` but the usage is so
restrained that type-safety is trivial and the performance will not be noticable
given everything else the producer has to do.

@Shopify/kafka

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/eapache/go-resiliency/breaker"
+	"github.com/eapache/queue"
 )
 
 func forceFlushThreshold() int {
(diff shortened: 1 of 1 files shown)


wvanbergen:
How this this compare to [container/heap](http://golang.org/pkg/container/heap/)?

eapache:
A queue is not a heap... and `container/heap` is just an interface, not an implementation.

wvanbergen:
I find it interesting go doesn't have an efficient queue in its standard library. I guess it just part of the general lack of container types due to the lack of generics. Oh well.

Anyway, this is :+1: for me, if the queue implementation is properly tested.

eapache:
The queue is pretty simple: https://github.com/eapache/queue/blob/master/queue.go and has some tests: https://github.com/eapache/queue/blob/master/queue_test.go

eapache:
There is http://golang.org/pkg/container/list/ that we could use, but it allocates a new struct for each element which makes it somewhat less efficient (in terms of both memory and gc)

eapache:
If we'd rather stick with the stdlib though, I can't imagine the overhead would be too terrible (again, at least compared to all the other things that the producer has to do).

wvanbergen:
Given that it's only 89 lines of go, using this lib is OK I think :+1:

yagnik:
This code lgtm, I'm blindly trusting your queue implementation though.

eapache:
Hmm, now I am tempted to switch to container/list just on the principle of "write less code".

eapache:
Eh, the queue implementation is also used (and thus pretty heavily tested implicitly) in https://github.com/eapache/channels and it's not like it's going to change much.

PR #398, opened 2015-03-26
Process consumed partitions asynchronously

### LINKED ISSUE #325: PartitionConsumer.Close() hangs frequently

eapache:
Rather than inline with the brokerConsumer's goroutine, dispatch them out to the
partitionConsumer's goroutine so that slow partitions don't block the others (as
much).

Also included: split out a `newBrokerConsumer` function.

Fixes #325 

@Shopify/kafka careful :eyes: on this please

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -106,7 +106,7 @@ func (c *consumer) ConsumePartition(topic string, partition int32, offset int64)
 		partition: partition,
 		messages:  make(chan *ConsumerMessage, c.conf.ChannelBufferSize),
 		errors:    make(chan *ConsumerError, c.conf.ChannelBufferSize),
-		trigger:   make(chan none, 1),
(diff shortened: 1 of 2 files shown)


eapache:
This needs revisiting, and possibly splitting apart into several PRs.

PR #400, opened 2015-03-27
The wiki of the offset API has been updated again

eapache:
This should bring us up to date.

@Shopify/kafka 

Somebody should verify against https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-OffsetCommitRequest

### DIFF
--- a/offset_commit_request.go
+++ b/offset_commit_request.go
@@ -13,44 +13,71 @@ type offsetCommitRequestBlock struct {
 
 func (r *offsetCommitRequestBlock) encode(pe packetEncoder, version int16) error {
 	pe.putInt64(r.offset)
-	if version >= 1 {
+	if version == 1 {
 		pe.putInt64(r.timestamp)
+	} else if r.timestamp != 0 {
+		Logger.Println("Non-zero timestamp specified for OffsetCommitRequest not v1, it will be ignored")
 	}
+
 	return pe.putString(r.metadata)
 }
 
 type OffsetCommitRequest struct {
-	ConsumerGroup string
-	Version       int16 // 0 (0.8.1 and later) or 1 (0.8.2 and later, includes timestamp field)
-	blocks        map[string]map[int32]*offsetCommitRequestBlock
+	ConsumerGroup           string
+	ConsumerGroupGeneration int32  // v1 or later
+	ConsumerID              string // v1 or later
+	RetentionTime           int64  // v2 or later
+
+	// Version can be:
+	// - 0 (kafka 0.8.1 and later)
+	// - 1 (kafka 0.8.2 and later)
+	// - 2 (kafka 0.8.3 and later)
+	Version int16
+	blocks  map[string]map[int32]*offsetCommitRequestBlock
 }
 
 func (r *OffsetCommitRequest) encode(pe packetEncoder) error {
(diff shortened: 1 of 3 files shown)


offset_commit_request.go
```diff
@@ -13,44 +13,71 @@ type offsetCommitRequestBlock struct {
...
+
 	return pe.putString(r.metadata)
 }
 
 type OffsetCommitRequest struct {
-	ConsumerGroup string
-	Version       int16 // 0 (0.8.1 and later) or 1 (0.8.2 and later, includes timestamp field)
-	blocks        map[string]map[int32]*offsetCommitRequestBlock
+	ConsumerGroup           string
+	ConsumerGroupGeneration int32  // v1 or later
```
wvanbergen: Do you have any idea what this used for? Couldn't find anything useful in the wiki.

eapache: https://cwiki.apache.org/confluence/display/KAFKA/Kafka+0.9+Consumer+Rewrite+Design

offset_fetch_request.go
```diff
@@ -2,10 +2,15 @@ package sarama
 
 type OffsetFetchRequest struct {
 	ConsumerGroup string
+	Version       int16
 	partitions    map[string][]int32
 }
 
 func (r *OffsetFetchRequest) encode(pe packetEncoder) (err error) {
+	if r.Version < 0 || r.Version > 1 {
```
yagnik: why not r.version != 1, you don't have decimal version do you ?

yagnik: actually it should be > 2

yagnik:
minor comment other lgtm

offset_fetch_request.go
eapache: I'm not sure what you mean? Based on my reading of the protocol guide (and [Jun's comment in JIRA](https://issues.apache.org/jira/browse/KAFKA-1841?focusedCommentId=14270222&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14270222)) the version can be 0 or 1, so if it is `<0` or `>1` we need to bail.

wvanbergen:
This LGTM.

Not sure how long the approach of trying to support all version from one type is going to be sustainable, but for now :shipit:.

offset_fetch_request.go
yagnik: Nevermind then, that looks good

jsvisa:
@eapache @wvanbergen  Seems Kafka 0.8.2 also not support `OffsetManage`. Before you send a `OffsetCommitRequest`, you need to send a `ConsumerMetadataRequest` to get the `Coordinator`, but this request always failed in kafka 0.8.2.0:

```bash
kafka server: Offset's topic has not yet been created. CoordinatorID:-1 CoordinatorHost: CoordinatorPort:-1
```
which is succeed in Kafka 0.8.2-beta.

wvanbergen:
@jsvisa It is working for me on 0.8.2.1 (see #379). The very first time you make this API call, Kafka first has to create the topic which takes some time. The only solution that works is retrying often enough.

jsvisa:
@wvanbergen  Thanks for your advise, I'll try it later in 0.8.2.1

PR #403, opened 2015-03-31
More test prep for toxiproxy

eapache:
@Shopify/kafka so I thought after the last one I was ready to actually write a test, but apparently not :)

This is a grab bag of changes to CI configuration in preparation for toxiproxy tests:
- Always run CI with `DEBUG=true`, it's just useful and it doesn't cost us
  anything.
- Initialize the logger first thing in `init()` when `DEBUG=true` so it can be used
  by other parts of the init.
- Seed the random number generator (and permit overriding the seed via an
  environment variable), this will be used for toxiproxy tests.
- Bump toxiproxy to 1.0.1 to fix a bug resetting proxies.
- Create a topic with 32 partitions for testing multiple partitions per broker.
- Lower the zookeeper timeouts so that we don't have to wait forever for a node
  to die when we use toxiproxy to kill its zookeeper connection.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
   - TOXIPROXY_ADDR=http://localhost:8474
   - KAFKA_INSTALL_ROOT=/home/travis/kafka
   - KAFKA_HOSTNAME=localhost
+  - DEBUG=true
   matrix:
   - KAFKA_VERSION=0.8.1.1
   - KAFKA_VERSION=0.8.2.1
--- a/functional_test.go
+++ b/functional_test.go
@@ -2,6 +2,7 @@ package sarama
 
 import (
 	"log"
+	"math/rand"
 	"net"
 	"os"
 	"strconv"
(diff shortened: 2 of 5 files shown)


wvanbergen:
:shipit: if CI is happy

eapache:
One more fix to toxiproxy's installation (sigh)

wvanbergen:
Let's just keep this branch open until you have your first working toxiproxy test :P

eapache:
I actually have a working toxiproxy test that isn't PRed yet, but I'd like to merge/review this first to keep things simple and separate.

@yagnik or @drdee care to take a look at this? It's mostly more CI config and only a little bit of Go.

eapache:
Working (for certain values) toxiproxy test: https://github.com/Shopify/sarama/commit/cf98aa905881c7ad2e1bd3bb2ff03b8872f7992d

eapache:
This is vagrant and test changes only (completely non-impacting prod code), so shipping with only one review.

yagnik:
Sorry for delayed review on this but this looks good!!

Issue #407, opened 2015-04-03
adding tcp keepalive support for the broker's tcp connection


epsniff:
We've been using the sarama client on Google Compute Engine (GCE) and it works well, expect that once in a while we get `read tcp 10.240.20.4:9093: i/o timeout`.   This occurs often with Go sockets on GCE, because GCE kills idle connections after 10 mins and by default Go's net package turns off tcp keepalives.   

I've created a small patch to support keepalives: https://github.com/lytics/sarama/commit/76aa5518b0b64a801bf534958ed98ce83999d828.   Let me know if you have any feedback?  We're currently running these changes through our staging environment to see this fixes the issue.  

Ref: https://cloud.google.com/compute/docs/troubleshooting#communicatewithinternet

thanks for all your hard work!!!

eapache:
Sounds like a good idea, although based on http://felixge.de/2014/08/26/tcp-keepalive-with-golang.html it looks like keepalives in golang are rather inconsistent across platforms. For style, you may want to take a look at #154 which also adds parameters to the dialer; it may be cleaner to follow that pattern than to wrap it up in an extra function.

epsniff:
Thank you for looking.   Ill look at both and clean up the code accordingly.

epsniff:
Closing the ticket after my PR got merged.

PR #408, opened 2015-04-03
Adding TCP keepalives support for the broker's connection

epsniff:
Addresses issue #407 

Our Google Compute Engine applications have a couple of under utilized topics/partitions that only receive messages once an hour or so.  Google Compute Engine's networking automatically kills any connection thats been idle for more than 10mins.   This causes the Sarama clients on GCE to return this error `read tcp 10.10.10.10:9093: i/o timeout`.  

From Google Compute Engine's Docs:
- https://cloud.google.com/compute/docs/troubleshooting#communicatewithinternet 

>Note that a system-wide "hidden" firewall rule is set to disconnect idle TCP connections after 10
>minutes. If your instance initiates or accepts long-lived connections with an external host, you can
>adjust TCP keep-alive settings to prevent these timeouts from dropping connections. ...

Keepalive Refs: 
- http://felixge.de/2014/08/26/tcp-keepalive-with-golang.html
- http://tldp.org/HOWTO/TCP-Keepalive-HOWTO/overview.html

PS thanks for all the hard work!  You guys rock!

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -4,7 +4,6 @@ import (
 	"io"
 	"sync"
 	"testing"
-	"time"
 )
 
 func safeClose(t *testing.T, c io.Closer) {
@@ -79,7 +78,6 @@ func TestClientDoesntCachePartitionsForTopicsWithErrors(t *testing.T) {
 	seedBroker.Returns(metadataResponse)
 
 	config := NewConfig()
-	config.Net.KeepAlive = 12 * time.Millisecond
 	config.Metadata.Retry.Max = 0
 	client, err := NewClient([]string{seedBroker.Addr()}, config)
 	if err != nil {
--- a/client_test.go
+++ b/client_test.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"sync"
 	"testing"
+	"time"
 )
 
 func safeClose(t *testing.T, c io.Closer) {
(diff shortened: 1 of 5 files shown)


config.go
```diff
@@ -12,6 +12,10 @@ type Config struct {
 		DialTimeout  time.Duration // How long to wait for the initial connection to succeed before timing out and returning an error (default 30s).
 		ReadTimeout  time.Duration // How long to wait for a response before timing out and returning an error (default 30s).
 		WriteTimeout time.Duration // How long to wait for a transmit to succeed before timing out and returning an error (default 30s).
+
+		// KeepAlive specifies the keep-alive period for an active network connection.
+		// If zero, keep-alives are not enabled.
```
eapache: "not enabled" == "disabled"

Also, please mention that the default is 0 (disabled).

config.go
```diff
@@ -126,6 +130,7 @@ func NewConfig() *Config {
 	c.Net.DialTimeout = 30 * time.Second
 	c.Net.ReadTimeout = 30 * time.Second
 	c.Net.WriteTimeout = 30 * time.Second
+	c.Net.KeepAlive = 0 * time.Second
```
eapache: 0 is the default value for uninitialized struct members, so this line is unnecessary

eapache:
A few nits, but generally LGTM. You're welcome to add a line to the changelog if you wish.

client_test.go
```diff
@@ -78,6 +79,7 @@ func TestClientDoesntCachePartitionsForTopicsWithErrors(t *testing.T) {
 	seedBroker.Returns(metadataResponse)
 
 	config := NewConfig()
+	config.Net.KeepAlive = 12 * time.Millisecond
```
wvanbergen: Not sure whether this is a useful addition given that we don't actually test it.

wvanbergen:
LGTM apart from the nitpicks.

epsniff:
@wvanbergen or @eapache - Does the new config setting new a validator ?   https://github.com/Shopify/sarama/blob/master/config.go#L187

eapache:
Assuming that a setting <0 is incoherent then yes, a validator is important - good catch!

PR #409, opened 2015-04-04
Added new client mock.

sybrandy:
I typed this out real quick and I don't have go installed on this computer, so this is untested.  If I get some more time, I'll try to get go installed and test this, but I wanted to get this out there quickly when I found the time.


eapache:
Closing as stale... we're still happy to consider this if you ever get it cleaned up and tested.

PR #411, opened 2015-04-09
Add Client.Coordinator() to retrieve the coordinating broker for a consumer group

wvanbergen:
This is the first step towards consumer group offset management, and later full consumer group management in Kafka 0.8.3.

Careful review of the 2 locks would be appreciated. Do we need to locks or can be just use the single lock for everything?

Extracted from #379.

@Shopify/kafka

### DIFF
--- a/client.go
+++ b/client.go
@@ -42,6 +42,13 @@ type Client interface {
 	// offset, OffsetNewest for the offset of the message that will be produced next, or a time.
 	GetOffset(topic string, partitionID int32, time int64) (int64, error)
 
+	// Coordinator returns the coordinating broker for a consumer group. It will return a locally cached
+	// value if it's available. You can call RefreshCoordinator to update the cached value.
+	Coordinator(consumerGroup string) (*Broker, error)
+
+	// RefreshCoordinator retrieves the coordinator for a consumer group and stores it in local cache.
+	RefreshCoordinator(consumerGroup string) error
+
 	// Close shuts down all broker connections managed by this client. It is required to call this function before
 	// a client object passes out of scope, as it will otherwise leak memory. You must close any Producers or Consumers
 	// using a client before you close the client.
@@ -72,13 +79,15 @@ type client struct {
 	seedBrokers []*Broker
 	deadSeeds   []*Broker
 
-	brokers  map[int32]*Broker                       // maps broker ids to brokers
-	metadata map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
+	brokers      map[int32]*Broker                       // maps broker ids to brokers
+	metadata     map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
+	coordinators map[string]int32                        // Maps consumer group names to coordinating broker IDs
 
 	// If the number of partitions is large, we can get some churn calling cachedPartitions,
 	// so the result is cached.  It is important to update this value whenever metadata is changed
 	cachedPartitionsResults map[string][maxPartitionIndex][]int32
-	lock                    sync.RWMutex // protects access to the maps, only one since they're always written together
+
+	lock sync.RWMutex // protects access to the maps that hold cluster state.
 }
 
 // NewClient creates a new Client. It connects to one of the given broker addresses
@@ -105,6 +114,7 @@ func NewClient(addrs []string, conf *Config) (Client, error) {
 		brokers:                 make(map[int32]*Broker),
 		metadata:                make(map[string]map[int32]*PartitionMetadata),
 		cachedPartitionsResults: make(map[string][maxPartitionIndex][]int32),
+		coordinators:            make(map[string]int32),
 	}
 	for _, addr := range addrs {
 		client.seedBrokers = append(client.seedBrokers, NewBroker(addr))
@@ -304,9 +314,56 @@ func (client *client) GetOffset(topic string, partitionID int32, time int64) (in
 	return offset, err
 }
 
+func (client *client) Coordinator(consumerGroup string) (*Broker, error) {
+	coordinator := client.cachedCoordinator(consumerGroup)
+
+	if coordinator == nil {
+		if err := client.RefreshCoordinator(consumerGroup); err != nil {
+			return nil, err
+		}
+		coordinator = client.cachedCoordinator(consumerGroup)
+	}
+
+	if coordinator == nil {
+		return nil, ErrConsumerCoordinatorNotAvailable
+	}
+
+	_ = coordinator.Open(client.conf)
+	return coordinator, nil
+}
+
+func (client *client) RefreshCoordinator(consumerGroup string) error {
+	response, err := client.getConsumerMetadata(consumerGroup, client.conf.Metadata.Retry.Max)
+	if err != nil {
+		return err
+	}
+
+	client.lock.Lock()
+	defer client.lock.Unlock()
+	client.registerBroker(response.Coordinator)
+	client.coordinators[consumerGroup] = response.Coordinator.ID()
+	return nil
+}
+
 // private broker management helpers
 
-func (client *client) disconnectBroker(broker *Broker) {
+// registerBroker makes sure a broker received by a Metadata or Coordinator request is registered
+// in the brokers map. It returns the broker that is registered, which may be the provided broker,
+// or a previously registered Broker instance. You must hold the write lock before calling this function.
+func (client *client) registerBroker(broker *Broker) {
+	if client.brokers[broker.ID()] == nil {
+		client.brokers[broker.ID()] = broker
+		Logger.Printf("client/brokers Registered new broker #%d at %s", broker.ID(), broker.Addr())
+	} else if broker.Addr() != client.brokers[broker.ID()].Addr() {
+		safeAsyncClose(client.brokers[broker.ID()])
+		client.brokers[broker.ID()] = broker
+		Logger.Printf("client/brokers Replaced registered broker #%d with %s", broker.ID(), broker.Addr())
+	}
+}
+
+// deregisterBroker removes a broker from the seedsBroker list, and if it's
+// not the seedbroker, removes it from brokers map completely.
+func (client *client) deregisterBroker(broker *Broker) {
 	client.lock.Lock()
 	defer client.lock.Unlock()
 
@@ -316,8 +373,9 @@ func (client *client) disconnectBroker(broker *Broker) {
 	} else {
 		// we do this so that our loop in `tryRefreshMetadata` doesn't go on forever,
 		// but we really shouldn't have to; once that loop is made better this case can be
-		// removed, and the function generally can be renamed from `disconnectBroker` to
+		// removed, and the function generally can be renamed from `deregisterBroker` to
 		// `nextSeedBroker` or something
+		Logger.Printf("client/brokers Deregistered broker #%d at %s", broker.ID(), broker.Addr())
 		delete(client.brokers, broker.ID())
 	}
 }
@@ -511,7 +569,7 @@ func (client *client) tryRefreshMetadata(topics []string, retriesRemaining int) 
 			// some other error, remove that broker and try again
 			Logger.Println("Error from broker while fetching metadata:", err)
 			_ = broker.Close()
-			client.disconnectBroker(broker)
+			client.deregisterBroker(broker)
 		}
 	}
 
@@ -538,14 +596,7 @@ func (client *client) updateMetadata(data *MetadataResponse) ([]string, error) {
 	// - if it is an existing ID, but the address we have is stale, discard the old one and save it
 	// - otherwise ignore it, replacing our existing one would just bounce the connection
 	for _, broker := range data.Brokers {
-		if client.brokers[broker.ID()] == nil {
-			client.brokers[broker.ID()] = broker
-			Logger.Printf("Registered new broker #%d at %s", broker.ID(), broker.Addr())
-		} else if broker.Addr() != client.brokers[broker.ID()].Addr() {
-			safeAsyncClose(client.brokers[broker.ID()])
-			client.brokers[broker.ID()] = broker
-			Logger.Printf("Replaced registered broker #%d with %s", broker.ID(), broker.Addr())
-		}
+		client.registerBroker(broker)
 	}
 
 	toRetry := make(map[string]bool)
@@ -594,4 +645,76 @@ func (client *client) updateMetadata(data *MetadataResponse) ([]string, error) {
 		ret = append(ret, topic)
 	}
 	return ret, err
+}
+
+func (client *client) cachedCoordinator(consumerGroup string) *Broker {
+	client.lock.RLock()
+	defer client.lock.RUnlock()
+	if coordinatorID, ok := client.coordinators[consumerGroup]; !ok {
+		return nil
+	} else {
+		return client.brokers[coordinatorID]
+	}
+}
+
+func (client *client) getConsumerMetadata(consumerGroup string, retriesRemaining int) (*ConsumerMetadataResponse, error) {
+	for broker := client.any(); broker != nil; broker = client.any() {
+		Logger.Printf("client/coordinator Requesting coordinator for consumergoup %s from %s.\n", consumerGroup, broker.Addr())
+
+		request := new(ConsumerMetadataRequest)
+		request.ConsumerGroup = consumerGroup
+
+		response, err := broker.GetConsumerMetadata(request)
+
+		if err != nil {
+			Logger.Printf("client/coordinator Request to broker %s failed: %s.\n", broker.Addr(), err)
+
+			switch err.(type) {
+			case PacketEncodingError:
+				return nil, err
+			default:
+				_ = broker.Close()
+				client.deregisterBroker(broker)
+				continue
+			}
+		}
+
+		switch response.Err {
+		case ErrNoError:
+			Logger.Printf("client/coordinator Coordinator for consumergoup %s is #%d (%s:%d).\n", consumerGroup, response.CoordinatorID, response.CoordinatorHost, response.CoordinatorPort)
+			return response, nil
+
+		case ErrConsumerCoordinatorNotAvailable:
+			Logger.Printf("client/coordinator Coordinator for consumer group %s is not available.\n", consumerGroup)
+
+			// This is very ugly, but this scenario will only happen once per cluster.
+			// The __consumer_offsets topic only has to be created one time.
+			// The number of partitions not configurable, but partition 0 should always exist.
+			if _, err := client.Leader("__consumer_offsets", 0); err != nil {
+				Logger.Printf("client/coordinator The __consumer_offsets topic is not initialized completely yet. Waiting 2 seconds...\n")
+				time.Sleep(2 * time.Second)
+			}
+
+			if retriesRemaining > 0 {
+				Logger.Printf("Retrying after %dms... (%d retries remaining)\n", client.conf.Metadata.Retry.Backoff/time.Millisecond, retriesRemaining)
+				time.Sleep(client.conf.Metadata.Retry.Backoff)
+				return client.getConsumerMetadata(consumerGroup, retriesRemaining-1)
+			}
+			return nil, ErrConsumerCoordinatorNotAvailable
+
+		default:
+			return nil, response.Err
+		}
+	}
+
+	Logger.Println("Out of available brokers to request consumer metadata from.")
+
+	if retriesRemaining > 0 {
+		Logger.Printf("Resurrecting dead brokers after %dms... (%d retries remaining)\n", client.conf.Metadata.Retry.Backoff/time.Millisecond, retriesRemaining)
+		time.Sleep(client.conf.Metadata.Retry.Backoff)
+		client.resurrectDeadBrokers()
+		return client.getConsumerMetadata(consumerGroup, retriesRemaining-1)
+	}
+
+	return nil, ErrOutOfBrokers
 }
--- a/consumer_metadata_response.go
+++ b/consumer_metadata_response.go
@@ -41,3 +41,18 @@ func (r *ConsumerMetadataResponse) decode(pd packetDecoder) (err error) {
 
 	return nil
 }
+
+func (r *ConsumerMetadataResponse) encode(pe packetEncoder) error {
+
+	pe.putInt16(int16(r.Err))
+
+	pe.putInt32(r.CoordinatorID)
+
+	if err := pe.putString(r.CoordinatorHost); err != nil {
+		return err
+	}
+
+	pe.putInt32(r.CoordinatorPort)
+
+	return nil
+}
--- a/client_test.go
+++ b/client_test.go
@@ -422,3 +422,135 @@ func TestClientResurrectDeadSeeds(t *testing.T) {
 
 	safeClose(t, c)
 }
+
+func TestClientCoordinatorWithConsumerOffsetsTopic(t *testing.T) {
+	seedBroker := newMockBroker(t, 1)
+	staleCoordinator := newMockBroker(t, 2)
+	freshCoordinator := newMockBroker(t, 3)
+
+	replicas := []int32{staleCoordinator.BrokerID(), freshCoordinator.BrokerID()}
+	metadataResponse1 := new(MetadataResponse)
+	metadataResponse1.AddBroker(staleCoordinator.Addr(), staleCoordinator.BrokerID())
+	metadataResponse1.AddBroker(freshCoordinator.Addr(), freshCoordinator.BrokerID())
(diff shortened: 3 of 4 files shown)


client.go
```diff
@@ -304,11 +312,35 @@ func (client *client) GetOffset(topic string, partitionID int32, time int64) (in
...
+	client.coordinatorLock.RLock()
+	coordinator, ok := client.coordinators[consumerGroup]
+	client.coordinatorLock.RUnlock()
+
+	if ok {
+		_ = coordinator.Open(client.conf)
+		return coordinator, nil
+	}
+
+	coordinator, err := client.refreshCoordinator(consumerGroup, 10)
```
wvanbergen: Number of retries is hardcoded for now.

client.go
```diff
@@ -595,3 +627,61 @@ func (client *client) updateMetadata(data *MetadataResponse) ([]string, error) {
...
+			return nil, response.Err
+		}
+	}
+
+	Logger.Println("Out of available brokers.")
+
+	if attemptsRemaining > 0 {
+		Logger.Printf("client/offsetleader Trying again to find leader for consumer group %s... (%d attempts remaining)\n", consumerGroup, attemptsRemaining)
+
+		time.Sleep(1 * time.Second)
```
wvanbergen: Backoff time is hardcoded for now.

wvanbergen:
Outstanding question: the retry attempts and backoff are hardcoded for now. We could use the standard metadata retry settings for this. Normally that is more than enough, but during the first call to `Coordinator`, Kafka actually creates the `__consumer_offsets` topic, which can take a long time. Any subsequent call to the function will be a lot faster.

wvanbergen:
Note: this only works against Kafka 0.8.2 and higher.

eapache:
- For simplicity, let's stick to just one lock for now. We can add fine-grained locking later if it turns out to be a performance bottleneck, but I doubt it.
- Is there a reason not to follow the same pattern as `Leader` here, with a `cachedCoordinator` function that returns the (opened) broker and a `refreshCoordinator` function that just updates the maps and doesn't return the broker directly?
- For retries, perhaps it is worth detecting the "need to create the topic" case (what error is that?) and adding a one-time separately configurable back-off of 5 seconds for that case, and otherwise using the metadata retries/backoffs. I am alternatively OK with just creating entirely separate retry/backoff configs for this path.

wvanbergen:
1) will do.

2) I think this is simpler and separates concerns better. `refreshCoordinator` just cares about executing the request and has no side effects, and doesn't have to care about cache. `Coordinator` deals with the cache.

3) Let me investigate what happens when you create a second consumergroup. If my understanding is correct, that will return the same error (`ErrConsumerCoordinatorNotAvailable`), but it will be functional much faster because it doesn't need to create the topic. I will report back later.

eapache:
2) It just seems weird to have the broker cache updated in the refresh method but the coordinator cache updated in the Coordinator method - I feel like those updates should live together.

The concern I separated with the Leader design is lazy open - only `cachedLeader` has to worry about opening the broker, since it is the only thing that returns a broker.

Also, nothing uses the `invalidateCoordinator` method?

wvanbergen:
The `invalidateCoordinator` will be used by methods that use the coordinator and figure out that it actually is not the correct one.

eapache:
> The `invalidateCoordinator` will be used by methods that use the coordinator and figure out that it actually is not the correct one.

No no no, this is the same mistake I made with `disconnectBroker` :(

client.go
```diff
@@ -72,13 +76,15 @@ type client struct {
...
+	brokers      map[int32]*Broker                       // maps broker ids to brokers
+	metadata     map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
+	coordinators map[string]*Broker                      // Maps consumer group names to coordinating brokers
 
 	// If the number of partitions is large, we can get some churn calling cachedPartitions,
 	// so the result is cached.  It is important to update this value whenever metadata is changed
 	cachedPartitionsResults map[string][maxPartitionIndex][]int32
-	lock                    sync.RWMutex // protects access to the maps, only one since they're always written together
+
+	lock sync.RWMutex // protects access to the metadata and coordinator maps
```
eapache: and broker map

eapache: and cachedPartitionsResults map

wvanbergen:
Updated `refreshCoordinator` to never open the broker, `Coordinator` now always does that

wvanbergen:
Remind me again why that caused issues for `disconnectBroker`?

wvanbergen:
```
[sarama] 2015/04/09 15:53:45 client/coordinator Requesting coordinator for consumergoup another_new_consumer_group_0 from localhost:9091.
[sarama] 2015/04/09 15:53:45 client/coordinator Coordinator for consumergoup another_new_consumer_group_0 is #9094 (localhost:9094).
[sarama] 2015/04/09 15:53:45 Connected to broker localhost:9094
[sarama] 2015/04/09 15:53:45 client/coordinator Requesting coordinator for consumergoup another_new_consumer_group_1 from localhost:9091.
[sarama] 2015/04/09 15:53:45 client/coordinator Coordinator for consumergoup another_new_consumer_group_1 is #9095 (localhost:9095).
[sarama] 2015/04/09 15:53:45 Connected to broker localhost:9095
[sarama] 2015/04/09 15:53:45 client/coordinator Requesting coordinator for consumergoup another_new_consumer_group_2 from localhost:9091.
[sarama] 2015/04/09 15:53:45 client/coordinator Coordinator for consumergoup another_new_consumer_group_2 is #9091 (localhost:9091).
[sarama] 2015/04/09 15:53:45 Connected to broker localhost:9091
[sarama] 2015/04/09 15:53:45 client/coordinator Requesting coordinator for consumergoup another_new_consumer_group_3 from localhost:9091.
[sarama] 2015/04/09 15:53:45 client/coordinator Coordinator for consumergoup another_new_consumer_group_3 is #9092 (localhost:9092).
[sarama] 2015/04/09 15:53:45 Connected to broker localhost:9092
[sarama] 2015/04/09 15:53:45 client/coordinator Requesting coordinator for consumergoup another_new_consumer_group_4 from localhost:9091.
[sarama] 2015/04/09 15:53:45 client/coordinator Coordinator for consumergoup another_new_consumer_group_4 is #9093 (localhost:9093).
[sarama] 2015/04/09 15:53:45 Connected to broker localhost:9093
[sarama] 2015/04/09 15:53:45 client/coordinator Requesting coordinator for consumergoup another_new_consumer_group_5 from localhost:9091.
[sarama] 2015/04/09 15:53:45 client/coordinator Coordinator for consumergoup another_new_consumer_group_5 is #9094 (localhost:9094).
[sarama] 2015/04/09 15:53:45 Connected to broker localhost:9094
[sarama] 2015/04/09 15:53:45 client/coordinator Requesting coordinator for consumergoup another_new_consumer_group_6 from localhost:9091.
[sarama] 2015/04/09 15:53:45 client/coordinator Coordinator for consumergoup another_new_consumer_group_6 is #9095 (localhost:9095).
[sarama] 2015/04/09 15:53:45 Connected to broker localhost:9095
[sarama] 2015/04/09 15:53:45 client/coordinator Requesting coordinator for consumergoup another_new_consumer_group_7 from localhost:9091.
[sarama] 2015/04/09 15:53:45 client/coordinator Coordinator for consumergoup another_new_consumer_group_7 is #9091 (localhost:9091).
[sarama] 2015/04/09 15:53:45 Connected to broker localhost:9091
[sarama] 2015/04/09 15:53:45 client/coordinator Requesting coordinator for consumergoup another_new_consumer_group_8 from localhost:9091.
[sarama] 2015/04/09 15:53:45 client/coordinator Coordinator for consumergoup another_new_consumer_group_8 is #9092 (localhost:9092).
[sarama] 2015/04/09 15:53:45 Connected to broker localhost:9092
[sarama] 2015/04/09 15:53:45 client/coordinator Requesting coordinator for consumergoup another_new_consumer_group_9 from localhost:9091.
[sarama] 2015/04/09 15:53:45 client/coordinator Coordinator for consumergoup another_new_consumer_group_9 is #9093 (localhost:9093).
[sarama] 2015/04/09 15:53:45 Connected to broker localhost:9093
```

Looks like for all subsequent attempts for new consumergroups, the call succeeds immediately. So `ErrConsumerCoordinatorNotAvailable` is only returned when the topic is being created. A separate longer `sleep` call when this happens sounds like a good idea. :+1:

eapache:
> Remind me again why that caused issues for disconnectBroker?

https://github.com/Shopify/sarama/issues/15
https://github.com/Shopify/sarama/issues/23#issuecomment-28151363

Basically I made it private but used it from outside the client. The semantics were confused. Lots of things.

The solution finally boiled down to two different tasks the user might want to accomplish:
- bounce a broker connection if e.g. it's timed out
- refresh metadata if the leadership has moved

The first can already be accomplished (now that we have lazy connections) by just calling `Close()` on the broker and asking for it again. The second can be accomplished (in the leader case) by calling `RefreshMetadata()`.

So based on all that, I think we should add `RefreshCoordinator(topic) error` as a public client method.

wvanbergen:
Removed `invalidateCoordinator`, added `RefreshCoordinator`.

wvanbergen:
Some final :eyes: ?

client.go
```diff
@@ -42,6 +43,14 @@ type Client interface {
 	// offset, OffsetNewest for the offset of the message that will be produced next, or a time.
 	GetOffset(topic string, partitionID int32, time int64) (int64, error)
 
+	// Coordinator returns the coordinating broker for a consumer group. It will return a locally cached
+	// value if it's available. This may be stale, in which case you should call RefreshCoordinator
```
eapache: I don't like the UX of saying "this may be stale", I think the necessary information is conveyed with just "it's locally cached".

client.go
```diff
@@ -42,6 +43,14 @@ type Client interface {
...
+	// to update the locally cached value.
+	Coordinator(consumerGroup string) (*Broker, error)
+
+	// RefreshCoordinator refreshes the coordinator for a consumer group.
```
eapache: mention that "refreshing the coordinator" involves asking the cluster, not just some local operation

client.go
```diff
@@ -72,13 +81,15 @@ type client struct {
 	seedBrokers []*Broker
 	deadSeeds   []*Broker
 
-	brokers  map[int32]*Broker                       // maps broker ids to brokers
-	metadata map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
+	brokers      map[int32]*Broker                       // maps broker ids to brokers
+	metadata     map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
+	coordinators map[string]*Broker                      // Maps consumer group names to coordinating brokers
```
eapache: I think this should just map to `int32` broker IDs (like the metadata map does), so that we only have one source of truth for "brokers we know about"

client.go
```diff
@@ -595,3 +652,53 @@ func (client *client) updateMetadata(data *MetadataResponse) ([]string, error) {
...
+	for broker := client.any(); broker != nil; broker = client.any() {
+
+		Logger.Printf("client/coordinator Requesting coordinator for consumergoup %s from %s.\n", consumerGroup, broker.Addr())
+
+		request := new(ConsumerMetadataRequest)
+		request.ConsumerGroup = consumerGroup
+
+		response, err := broker.GetConsumerMetadata(request)
+
+		if err != nil {
```
eapache: `PacketEncodingError`s should be returned immediately.

client.go
```diff
@@ -304,9 +316,60 @@ func (client *client) GetOffset(topic string, partitionID int32, time int64) (in
...
+}
+
+func (client *client) RefreshCoordinator(consumerGroup string) error {
+	coordinator, err := client.getCoordinator(consumerGroup, client.conf.Metadata.Retry.Max)
+	if err != nil {
+		return err
+	}
+
+	client.lock.Lock()
+	client.coordinators[consumerGroup] = client.registerBroker(coordinator)
```
eapache: putting this here still feels super-weird to me - why do you prefer this to the `RefreshMetadata` pattern?

client.go
wvanbergen: I think this has better separation of concerns:
- `RefreshCoordinator` only writes cache
- `Coordinator` only reads cache
- `getCoordinator` never touches the cache.

I'd like to refactor metadata similarly, but unfortunately it's a more complicated case because Metadata requests can return partially valid results. That never happens with getCoordinator: it's either good or bad.

client.go
wvanbergen: If the coordinator call returns a new broker, it is always registered in the brokers array as well. What would be the benefit of just storing the ID? AFAICS it's a tradeoff between an extra hash lookup vs. a bit more memory.

client.go
wvanbergen: I think this has better separation of concerns:

- `Coordinator` only reads from cache.
- `RefreshCoordinator` only writes to cache.
- `getCoordinator` never touches the cache.

I'd prefer this approach for metadata as well, except for the fact that it's a more complicated case: Metadata requests can return partially useful responses, while Coordinator responses are either good or bad.

client.go
eapache: OK, I feel like it's just over-complicated to return the result, and *then* write it to the cache in a separate method, when you could simply write it to the cache in the first place. `getCoordinator`'s result is never going to be used for anything but updating the cache, so splitting that apart seems over-engineered to me.

We agree on `Coordinator`/`Leader`/etc should just read from the cache. I'm just not sure why we need two separate methods (fetch from remote, update cache) when that feels like a single operation to me.

eapache: More specifically: we *don't* want anybody to fetch from the cluster without updating the cache. What good would that do? Splitting the two methods gives us that flexibility, yes, but I think that's flexibility we should be avoiding. The cache is the source of truth. There is a single black box method that updates the cache.

client.go
eapache: The tradeoff is the subtleties of updating everything. AFAICT the code is currently fine, but it opens us to cases where a broker ends up in the coordinator map but not in the broker map at all, which is then a leak on shutdown and could lead to weird behaviour. Making `brokers` the source of truth and have everything point into that seems safer.

client.go
eapache: OK, I feel like it's just over-complicated to return the result, and then write it to the cache in a separate method, when you could simply write it to the cache in the first place. getCoordinator's result is never going to be used for anything but updating the cache, so splitting that apart seems over-engineered to me.

We agree on Coordinator/Leader/etc should just read from the cache. I'm just not sure why we need two separate methods (fetch from remote, update cache) when that feels like a single operation to me.

More specifically: we don't want anybody to fetch from the cluster without updating the cache. What good would that do? Splitting the two methods gives us that flexibility, yes, but I think that's flexibility we should be avoiding. The cache is the source of truth. There is a single black box method that updates the cache.

wvanbergen: - Primary reason: the `getCoordinator` method has enough to worry about. It's easier to understand it and refactor it later if we limit the scope.
- Nobody can access `getCoordinator` anyway because it's not exported. Only within sarama itself this is possible, so `anybody` is quite limited in scope. Not sure about the need to protect against ourselves; it feels defensive given that we're all responsible adults here. :P 
- Even if you do end up calling `getCoordinator` without updating the cache, it doesn't cause any issues besides doing potentially unnecessary requests to the cluster.
- It's easier to test stuff in isolation when functions do not have side effects.

eapache: Hmm, maybe our API is wrong then... we are exposing two actions to the user:
- "get", which reads from the cache or reads through to the cluster if it's not in cache
- "refresh", which updates the cache from the cluster if the user thinks the cache is stale

But we are implementing three actions internally.

wvanbergen: > we are implementing three actions internally

No, we're just splitting up functionality in different functions to make it more understandable. We have many more helper functions (e.g. `registerBroker`) that we don't expose.

Maybe our external API can be improved, but that's kind of outside the scope of this PR.

client.go
wvanbergen: Fixed

client.go
eapache: In fact there is a subtle bug in already: if a `RefreshMetadata()` call replaces a broker then the coordinator map will still contain a reference to the old one.

wvanbergen: cool, I'll update.

client.go
eapache: Tangent, but related: I'm an idiot. `ConsumerMetadataResponse` should just have a `*Broker` as a field, it lines up exactly with `MetadataResponse` and the broker already has a `decode` method which does the right thing given an id/host/port tuple. I've added it to the wiki.

Back on topic: I think this change would sit better for me if `getCoordinator` was called `getConsumerMetadata` and just returned the response object on success (letting `RefreshMetadata` construct the broker object).

client.go
wvanbergen: Fixed!

client.go
wvanbergen: Yeah, too bad that would be a breaking change. I'll update

eapache: Actually, we can sort of fix that, one moment for me to PR a change to `ConsumerMetadataResponse`.

wvanbergen: Pushed.

client.go
```diff
@@ -304,9 +316,69 @@ func (client *client) GetOffset(topic string, partitionID int32, time int64) (in
...
+	return coordinator, nil
+}
+
+func (client *client) RefreshCoordinator(consumerGroup string) error {
+	response, err := client.getConsumerMetadata(consumerGroup, client.conf.Metadata.Retry.Max)
+	if err != nil {
+		return err
+	}
+
+	coordinator := &Broker{
```
eapache: once #413 lands this can go away

client.go
```diff
@@ -304,9 +316,69 @@ func (client *client) GetOffset(topic string, partitionID int32, time int64) (in
...
+	return nil
+}
+
 // private broker management helpers
 
-func (client *client) disconnectBroker(broker *Broker) {
+// registerBroker makes sure a broker received by a Metadata or Coordinator request is registered
+// in the brokers map. It returns the broker that is registered, which may be the provided broker,
+// or a previously registered Broker instance.
+func (client *client) registerBroker(broker *Broker) *Broker {
```
eapache: this doesn't need to return anything anymore

client.go
```diff
@@ -42,6 +42,14 @@ type Client interface {
 	// offset, OffsetNewest for the offset of the message that will be produced next, or a time.
 	GetOffset(topic string, partitionID int32, time int64) (int64, error)
 
+	// Coordinator returns the coordinating broker for a consumer group. It will return a locally cached
+	// value if it's available. This value is cached locally; you can call RefreshCoordinator to update
```
eapache: you don't need to says it's cached locally twice :)

client.go
```diff
@@ -304,9 +315,64 @@ func (client *client) GetOffset(topic string, partitionID int32, time int64) (in
...
 
+func (client *client) Coordinator(consumerGroup string) (*Broker, error) {
+	client.lock.RLock()
+	coordinatorID, coordinatorCached := client.coordinators[consumerGroup]
+	client.lock.RUnlock()
+
+	if !coordinatorCached {
+		var err error
+		err = client.RefreshCoordinator(consumerGroup)
+		if err != nil {
```
eapache: the above three lines can be one `if err := client.RefreshCoordinator(consumerGroup); err != nil {`

client.go
```diff
@@ -304,9 +315,64 @@ func (client *client) GetOffset(topic string, partitionID int32, time int64) (in
...
+			return nil, err
+		}
+
+		client.lock.RLock()
+		coordinatorID = client.coordinators[consumerGroup]
+		client.lock.RUnlock()
+	}
+
+	client.lock.RLock()
+	coordinator := client.brokers[coordinatorID]
```
eapache: Because you've released and re-taken the lock this isn't guaranteed to be present. It may be necessary to have a `cachedCoordinator` method like `cachedLeader` that reads the ID and the broker under a single lock, and handles a missing broker by returning the appropriate error instead of `nil, nil`.

client.go
```diff
@@ -304,9 +315,64 @@ func (client *client) GetOffset(topic string, partitionID int32, time int64) (in
...
+func (client *client) RefreshCoordinator(consumerGroup string) error {
+	response, err := client.getConsumerMetadata(consumerGroup, client.conf.Metadata.Retry.Max)
+	if err != nil {
+		return err
+	}
+
+	client.lock.Lock()
+	client.registerBroker(response.Coordinator)
+	client.coordinators[consumerGroup] = response.Coordinator.ID()
+	client.lock.Unlock()
```
eapache: defer this in case something panics?

client.go
```diff
@@ -595,3 +655,54 @@ func (client *client) updateMetadata(data *MetadataResponse) ([]string, error) {
...
+
+		switch response.Err {
+		case ErrNoError:
+			Logger.Printf("client/coordinator Coordinator for consumergoup %s is #%d (%s:%d).\n", consumerGroup, response.CoordinatorID, response.CoordinatorHost, response.CoordinatorPort)
+			return response, nil
+
+		case ErrConsumerCoordinatorNotAvailable:
+			Logger.Printf("client/coordinator Coordinator for consumer group %s not yet available, trying again in %dms...\n", consumerGroup, client.conf.Metadata.Retry.Backoff/time.Millisecond)
+			time.Sleep(2 * time.Second)
+			continue
```
eapache: should this decrement `attemptsRemaining` and check for 0?

wvanbergen: I think that makes sense, but I will bump the Backoff value to 3 seconds because there will be less attempts available.

eapache: We should probably also only do the extra backoff the very first time, if it hasn't been created then it probably won't be.

wvanbergen: Why not? This call should always trigger the creation if it's not there.

eapache: I thought we'd had this conversation already :P

There are two retriable error cases here:
- offsets topic has not yet been created
- offsets topic exists but an election is ongoing so there is no leader right now

The first case returns `ErrConsumerCoordinatorNotAvailable`. It takes some time (~5s?) to create the topic. If the topic fails to create for some reason, waiting another 5 seconds is unlikely to fix anything.

I don't know what the second case returns (`ErrConsumerCoordinatorNotAvailable` or `LeaderNotAvailable` or ...?). This one is worth retrying several times in fairly quick succession (100-500ms) since leadership time depends on how busy the cluster is but is usually fairly fast.

wvanbergen: Not sure what the broker will return. The docs say about `ErrConsumerCoordinatorNotAvailable`:

> The broker returns this error code for consumer metadata requests or offset commit requests if the offsets topic has not yet been created.

That would indicate another error code for the 2nd scenario. Can we trigger this some way?

eapache: Create the topic, find the coordinator, then send it queries while `kill -9` the current coordinator and see what comes out.

eapache: Based on reading the scala it returns `ErrConsumerCoordinatorNotAvailable` for basically all errors.

eapache: So I guess we should have a separate backoff/retry count for this, and default it to 5x1s or something

wvanbergen: Maybe the same backoff, but many more retries?

eapache: I think that might be confusing, but if you have an idea for how to make the godoc make sense go right ahead.

wvanbergen:
@eapache final :eyes:

client.go
```diff
@@ -304,9 +314,56 @@ func (client *client) GetOffset(topic string, partitionID int32, time int64) (in
...
+	return nil
+}
+
 // private broker management helpers
 
-func (client *client) disconnectBroker(broker *Broker) {
+// registerBroker makes sure a broker received by a Metadata or Coordinator request is registered
+// in the brokers map. It returns the broker that is registered, which may be the provided broker,
+// or a previously registered Broker instance.
+func (client *client) registerBroker(broker *Broker) {
```
eapache: maybe worth mentioning this must be called with the lock held in the comment?

consumer_metadata_response.go
```diff
@@ -41,3 +41,19 @@ func (r *ConsumerMetadataResponse) decode(pd packetDecoder) (err error) {
...
 }
+
+func (r *ConsumerMetadataResponse) encode(pe packetEncoder) error {
+
+	pe.putInt16(int16(r.Err))
+
+	pe.putInt32(r.CoordinatorID)
+
+	err := pe.putString(r.CoordinatorHost)
+	if err != nil {
```
eapache: nit: this could be one line

eapache:
Very minor nits, but LGTM. Squash it and :shipit:

zaynetro:
> Note: this only works against Kafka 0.8.2 and higher.

Damn, I spent hours trying to get Coordinator to work on v0.8.1.1. Maybe, it should be added to the docs?

wvanbergen:
@zaynetro https://github.com/Shopify/sarama/pull/453

zaynetro:
Great!

PR #413, opened 2015-04-10
Automatically construct a broker from the ConsumerMetadataResponse

eapache:
@Shopify/kafka 

The three deprecated fields can go away whenever we decide to do 2.0.


wvanbergen:
We should probably add a test

eapache:
Test added.

wvanbergen:
:shipit: once travis is happy.

PR #415, opened 2015-04-10
Refactor metadata retry logic and improve logging

wvanbergen:
This doesn't change the behaviour, but it makes the implementation easier to follow. I also improved logging.

@Shopify/kafka

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -85,7 +85,11 @@ func (b *Broker) Open(conf *Config) error {
 		b.done = make(chan bool)
 		b.responses = make(chan responsePromise, b.conf.Net.MaxOpenRequests-1)
 
-		Logger.Printf("Connected to broker %s\n", b.addr)
+		if b.id >= 0 {
+			Logger.Printf("Connected to broker at %s (registered as #%d)\n", b.addr, b.id)
+		} else {
+			Logger.Printf("Connected to broker at %s (unregistered)\n", b.addr)
+		}
 		go withRecover(b.responseReceiver)
 	})
 
--- a/client.go
+++ b/client.go
@@ -369,11 +369,11 @@ func (client *client) RefreshCoordinator(consumerGroup string) error {
 func (client *client) registerBroker(broker *Broker) {
 	if client.brokers[broker.ID()] == nil {
 		client.brokers[broker.ID()] = broker
-		Logger.Printf("client/brokers Registered new broker #%d at %s", broker.ID(), broker.Addr())
+		Logger.Printf("client/brokers registered new broker #%d at %s", broker.ID(), broker.Addr())
 	} else if broker.Addr() != client.brokers[broker.ID()].Addr() {
 		safeAsyncClose(client.brokers[broker.ID()])
 		client.brokers[broker.ID()] = broker
-		Logger.Printf("client/brokers Replaced registered broker #%d with %s", broker.ID(), broker.Addr())
+		Logger.Printf("client/brokers replaced registered broker #%d with %s", broker.ID(), broker.Addr())
 	}
 }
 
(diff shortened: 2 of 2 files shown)


broker.go
```diff
@@ -85,7 +85,11 @@ func (b *Broker) Open(conf *Config) error {
 		b.done = make(chan bool)
 		b.responses = make(chan responsePromise, b.conf.Net.MaxOpenRequests-1)
 
-		Logger.Printf("Connected to broker %s\n", b.addr)
+		if b.id >= 0 {
+			Logger.Printf("Connected to broker %d at %s\n", b.id, b.addr)
+		} else {
+			Logger.Printf("Connected to seed broker at %s\n", b.addr)
```
eapache: I'm not sure I want the concept of seed-broker to leak into the `Broker` object itself, I feel like it shouldn't really care.

client.go
```diff
@@ -536,29 +536,31 @@ func (client *client) backgroundMetadataUpdater() {
 	}
 }
 
-func (client *client) tryRefreshMetadata(topics []string, retriesRemaining int) error {
+func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int) error {
+	retry := func(err error) error {
+		if attemptsRemaining > 0 {
+			Logger.Printf("client/metadata Retrying after %dms... (%d attempts remaining)\n", client.conf.Metadata.Retry.Backoff/time.Millisecond, attemptsRemaining)
```
eapache: here (and all other log locations) capitalizing the first word after the client/scope seems weird to me, but that's purely aesthetic/stylistic, ymmv

eapache: hmm, more coherently I think this is a grammar thing: I'm trying to read them as single sentences where client/scope is the noun and the verb is being done by the noun, maybe that's crazy of me

client.go
```diff
@@ -536,29 +536,31 @@ func (client *client) backgroundMetadataUpdater() {
...
-					return err
-				}
-				Logger.Printf("Some partitions are leaderless, waiting %dms for election... (%d retries remaining)\n",
-					client.conf.Metadata.Retry.Backoff/time.Millisecond, retriesRemaining)
-				time.Sleep(client.conf.Metadata.Retry.Backoff) // wait for leader election
-				return client.tryRefreshMetadata(retry, retriesRemaining-1)
+			toRetry, err := client.updateMetadata(response)
+			if len(toRetry) > 0 {
+				Logger.Println("client/metadata Some partitions are leaderless.")
+				return retry(err) // note: err can be nil
```
eapache: there's a subtle behaviour change here, in that before we would only retry the topics that were in election, whereas now we will retry them all

client.go
```diff
@@ -567,23 +569,15 @@ func (client *client) tryRefreshMetadata(topics []string, retriesRemaining int)
...
+			Logger.Println("client/metadata Error from broker while fetching metadata:", err)
 			_ = broker.Close()
 			client.deregisterBroker(broker)
 		}
 	}
 
-	Logger.Println("Out of available brokers.")
-
-	if retriesRemaining > 0 {
-		Logger.Printf("Resurrecting dead brokers after %dms... (%d retries remaining)\n",
```
eapache: should we stick a log line in `resurrectDeadBrokers` to make up for this one?

client.go
wvanbergen: This was unintentional, and can be fixed. But is this bad? If we're going to retry again, we can just as well ask for everything again. We will have potentially fresher metadata with simpler code, at the price of slightly more data being transferred.

eapache: eh; if we're going that route, then `updateMetadata` can be simplified since it no longer needs to track a list of topics to retry, it just has to track true/false.

wvanbergen: Yeah, it would be a nice simplification, but if you think keeping the current behaviour is worth it I will replicate it.

broker.go
wvanbergen: The `Broker` doesn't, but the reader of the log does. I can change it to "broker without ID" if that gels better with you.

eapache: That would work, or "manual broker" or "custom broker" or something. The only case where the ID is -1 is when it is manually constructed from an address, and not from a metadata response of some kind.

client.go
eapache: No, I don't have strong feelings for keeping it - let's simplify it then

wvanbergen:
Rebased, and address all comments.

client.go
wvanbergen: Added.

eapache:
LGTM, but it's another thing that will need stress testing before 1.4.0 is released

wvanbergen:
Merging this.

PR #416, opened 2015-04-11
Consistently return ErrClosedClient when calling an exported function on a closed client

wvanbergen:
We have this in place for most of the methods on Client, but not all of them.

@Shopify/kafka


eapache:
:shipit:

PR #418, opened 2015-04-11
Consumer: check offset before returning ConsumePartition.

wvanbergen:
When calling ConsumePartition, always check whether the offset is within the offset range. This means we now have to do two OffsetRequests for every ConsumePartition call, even if the offset is provided. The good news is that the method will immediately return an error and never start a goroutine, instead of starting the goroutine and returning an error in the Errors() channel which you can easily ignore.

This replaces #378. It now uses two OffsetRequests instead of changing OffsetRequests to allow obtaiing multiple offsets for a single partition.

### DIFF
--- a/consumer_test.go
+++ b/consumer_test.go
@@ -350,16 +350,16 @@ func TestConsumerInterleavedClose(t *testing.T) {
 }
 
 func TestConsumerBounceWithReferenceOpen(t *testing.T) {
-	t.Skip("This is not yet working due to concurrency on the mock broker")
-
 	seedBroker := newMockBroker(t, 1)
 	leader := newMockBroker(t, 2)
 	leaderAddr := leader.Addr()
+	tmp := newMockBroker(t, 3)
 
 	metadataResponse := new(MetadataResponse)
 	metadataResponse.AddBroker(leader.Addr(), leader.BrokerID())
+	metadataResponse.AddBroker(tmp.Addr(), tmp.BrokerID())
 	metadataResponse.AddTopicPartition("my_topic", 0, leader.BrokerID(), nil, nil, ErrNoError)
-	metadataResponse.AddTopicPartition("my_topic", 1, leader.BrokerID(), nil, nil, ErrNoError)
+	metadataResponse.AddTopicPartition("my_topic", 1, tmp.BrokerID(), nil, nil, ErrNoError)
 	seedBroker.Returns(metadataResponse)
 
 	config := NewConfig()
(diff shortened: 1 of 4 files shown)


wvanbergen:
@Shopify/kafka I had to disable one consumer test due to concurrency issues on the mock broker. Ideas on how to overcome this are welcome.

consumer_test.go
```diff
@@ -298,6 +350,8 @@ func TestConsumerInterleavedClose(t *testing.T) {
 }
 
 func TestConsumerBounceWithReferenceOpen(t *testing.T) {
+	t.Skip("This is not yet working due to concurrency on the mock broker")
```
eapache: You'll have to add four offset responses (newest/oldest for two partitions) but I didn't think this test was particularly fragile. It should be ok as long as you don't return the second offset pair until after the first `ConsumePartition` call has returned.

consumer_test.go
```diff
@@ -152,10 +172,26 @@ func TestConsumerRebalancingMultiplePartitions(t *testing.T) {
...
+
+	offsetResponseOldest1 := new(OffsetResponse)
+	offsetResponseOldest1.AddTopicPartition("my_topic", 1, 0)
+	leader1.Returns(offsetResponseOldest1)
+
 	// we expect to end up (eventually) consuming exactly ten messages on each partition
 	var wg sync.WaitGroup
-	for i := 0; i < 2; i++ {
-		consumer, err := master.ConsumePartition("my_topic", int32(i), 0)
+	for i := int32(0); i < 2; i++ {
```
eapache: good idea

consumer_test.go
```diff
@@ -152,10 +172,26 @@ func TestConsumerRebalancingMultiplePartitions(t *testing.T) {
...
+	leader0.Returns(offsetResponseOldest0)
+
+	offsetResponseNewest1 := new(OffsetResponse)
+	offsetResponseNewest1.AddTopicPartition("my_topic", 1, 1234)
+	leader1.Returns(offsetResponseNewest1)
+
+	offsetResponseOldest1 := new(OffsetResponse)
+	offsetResponseOldest1.AddTopicPartition("my_topic", 1, 0)
+	leader1.Returns(offsetResponseOldest1)
+
```
eapache: you can dedupe these by moving them inside the loop

it might even be worth writing a helper function that takes mockbroker/topic/partition/newest/oldest and returns the two responses.

consumer_test.go
wvanbergen: The problem is that the two consumers are talking to the same broker. The fetch request/responses that happen right after the offset responses, but we have to ensure we handle the offset requests for both consumers before we start dealing with the fetch requests. AFAICS that's not possible right now.

eapache:
I believe I fixed the test you had skipped. If travis agrees, this LGTM.

wvanbergen:
There's probably some test cleanup we can do, but I'll leave that for a separate PR.

PR #420, opened 2015-04-14
Retry messages on shutdown

### LINKED ISSUE #419: Producer does not retry messages when shutting down
I almost filed this as an enhancement, since this was a design trade-off/simplification I explicitly made when building the current producer, but I guess it qualifies as sufficiently unexpected behaviour (for people other than me) to be considered a bug.

If you call `Close` or `AsyncClose` on a producer instance while messages are inflight, and those messages fail to produce (e.g. because leadership has moved), they will not be retried even if they have retries remaining. They will get fed back to the `topicDispatcher`, which will promptly return them with the confusing error `ErrShuttingDown`.

eapache:
@Shopify/kafka fixes #419.

Rather than use the old complicated system of reference-counting flags to shutdown cleanly, do the much simpler thing: keep a `sync.WaitGroup` counting the number of messages "in flight" (aka owned by the producer). When shutdown is requested, spawn a goroutine that waits for this counter to hit 0, then closes everything in one go.

We add messages to the in-flight set in the `topicDispatcher` (only new messages with retries==0 though). We remove messages from the in-flight set in `returnError` and `returnSuccesses`; even if the `Producer.Return.*` values are false they are still guaranteed to see every message.

We also add/remove chaser messages in `leaderDispatcher`.

~~This still needs tests.~~

I'm not sure what performance impact the waitgroup will have. An alternative might be an atomic counter, and have the shutdown goroutine just poll it every 10ms or something.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -54,6 +54,7 @@ type asyncProducer struct {
 
 	errors                    chan *ProducerError
 	input, successes, retries chan *ProducerMessage
+	inFlight                  sync.WaitGroup
 
 	brokers    map[*Broker]chan *ProducerMessage
 	brokerRefs map[chan *ProducerMessage]int
@@ -105,8 +106,6 @@ type flagSet int8
 
 const (
 	chaser   flagSet = 1 << iota // message is last in a group that failed
-	ref                          // add a reference to a singleton channel
-	unref                        // remove a reference from a singleton channel
 	shutdown                     // start the shutdown process
 )
 
@@ -193,9 +192,7 @@ func (p *asyncProducer) Close() error {
 }
 
 func (p *asyncProducer) AsyncClose() {
-	go withRecover(func() {
-		p.input <- &ProducerMessage{flags: shutdown}
-	})
+	go withRecover(p.shutdown)
 }
 
 ///////////////////////////////////////////
@@ -209,6 +206,7 @@ func (p *asyncProducer) AsyncClose() {
 // dispatches messages by topic
 func (p *asyncProducer) topicDispatcher() {
 	handlers := make(map[string]chan *ProducerMessage)
+	shuttingDown := false
 
 	for msg := range p.input {
 		if msg == nil {
@@ -217,8 +215,14 @@ func (p *asyncProducer) topicDispatcher() {
 		}
 
 		if msg.flags&shutdown != 0 {
-			Logger.Println("Producer shutting down.")
-			break
+			shuttingDown = true
+			continue
+		} else if msg.retries == 0 {
+			p.inFlight.Add(1)
+			if shuttingDown {
+				p.returnError(msg, ErrShuttingDown)
+				continue
+			}
 		}
 
 		if (p.conf.Producer.Compression == CompressionNone && msg.Value != nil && msg.Value.Length() > p.conf.Producer.MaxMessageBytes) ||
@@ -230,7 +234,6 @@ func (p *asyncProducer) topicDispatcher() {
 
 		handler := handlers[msg.Topic]
 		if handler == nil {
-			p.retries <- &ProducerMessage{flags: ref}
 			newHandler := make(chan *ProducerMessage, p.conf.ChannelBufferSize)
 			topic := msg.Topic // block local because go's closure semantics suck
 			go withRecover(func() { p.partitionDispatcher(topic, newHandler) })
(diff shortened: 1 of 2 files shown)


eapache:
> Also, I'm not sure what performance impact the waitgroup will have. An alternative might be an atomic counter, and have the shutdown goroutine just poll it every 10ms or something.

A recent benchmarking and profiling push say: maybe this has a tiny impact on performance, but it's still swamped out by stupid stuff like CRC calculations, so not a concern.

eapache:
CI failing because of https://github.com/kisielk/errcheck/issues/70

eapache:
CI fixed.

@Shopify/kafka this is ready for review.

async_producer.go
```diff
@@ -355,6 +347,7 @@ func (p *asyncProducer) leaderDispatcher(topic string, partition int32, input ch
 				// in fact this message is not even the current retry level, so buffer it for now (unless it's a just a chaser)
 				if msg.flags&chaser == chaser {
 					retryState[msg.retries].expectChaser = false
+					p.inFlight.Done() // this chaser is now useless and will be garbage collected
```
wvanbergen: `s/is now useless/is now handled/` ?

async_producer.go
```diff
@@ -392,6 +385,7 @@ func (p *asyncProducer) leaderDispatcher(topic string, partition int32, input ch
 					}
 
 				}
+				p.inFlight.Done() // this chaser is now useless and will be garbage collected
```
wvanbergen: `/s/useless/handled/` ?

async_producer.go
```diff
@@ -218,7 +218,16 @@ func (p *asyncProducer) topicDispatcher() {
 
 		if msg.flags&shutdown != 0 {
```
wvanbergen: There will only be one message with the `shutdown` message set, right?

wvanbergen:
I think the accounting is correct.

- New (0 retries) messages increment, errors and successes decrement.
- New chaser messages increment, and are decremented when they are handled.
- Only shutdown messages are not accounted for, but there's only one and it doesn't propagate through the goroutines (see my comment above)

:+1:, this is a nice simplification and it's much easier to understand now.

async_producer.go
eapache: > There will only be one message with the shutdown message set, and it doesn't need to propagate further anymore, right?

I guess in theory there could be several if the user calls `AsyncClose` several times, but I don't feel like guarding against that case.

> Instead of sending a message with a flag, could we simply use a closing chan none for this now? Or am I missing something? I'd rather minimize the number of "special" messages that go through the pipeline

It would mean this goroutine would need a `select` instead of a nice simple `range`... not sure if that really ends up being any cleaner, I'll mock it quickly to see.

eapache: > It would mean this goroutine would need a select instead of a nice simple range... not sure if that really ends up being any cleaner, I'll mock it quickly to see.

Really no simpler, because once the `closing` chan is closed then we have to go to a different loop somewhere or we'll spin forever because a closed channel is always ready for reading (and we still have to process things coming in on the input channel for a while longer).

eapache: Working through that did give me an idea for an additional minor simplification though (see the third commit).

eapache:
Your description of the accounting matches my understanding exactly. Once CI is :green_apple: and I've had a quick pair of :eyes: on the third commit, I think this is good to go.

async_producer.go
```diff
@@ -217,8 +215,15 @@ func (p *asyncProducer) topicDispatcher() {
 		}
 
 		if msg.flags&shutdown != 0 {
-			Logger.Println("Producer shutting down.")
-			break
+			shuttingDown = true
+			continue
+		} else if msg.retries == 0 {
+			if shuttingDown {
+				p.returnError(msg, ErrShuttingDown)
```
wvanbergen: This reduces the inflight counter, but it was never incremented for this message. We should probably move the `p.inFlight.Add(1)` up so it always gets executed.

eapache: Oooh, really nice catch. I will fix and see if I can add or adjust a test for this case too.

eapache: OK, this path is fixed and tested now.

wvanbergen:
I think this looks good. We should do some stress testing of this though.

eapache:
I am comfortable enough to push this to master now. I will do some stressing before releasing the next stable version.

PR #423, opened 2015-04-15
client: stop the background thread before closing

### LINKED ISSUE #422: Panic: assignment to entry in nil map
Hi all, 

I'm still digging in (updates to come) but we are seeing this panic from the sarama client: 

```
panic: assignment to entry in nil map

goroutine 3857 [running]:
github.com/Shopify/sarama.(*client).registerBroker(0xc21e3a3f80, 0xc242620230)
	/home/mdmarek/go/root/src/github.com/Shopify/sarama/client.go:367 +0xac
github.com/Shopify/sarama.(*client).updateMetadata(0xc21e3a3f80, 0xc234acde30, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/mdmarek/go/root/src/github.com/Shopify/sarama/client.go:611 +0x127
github.com/Shopify/sarama.(*client).tryRefreshMetadata(0xc21e3a3f80, 0x0, 0x0, 0x0, 0x2, 0x0, 0x0)
	/home/mdmarek/go/root/src/github.com/Shopify/sarama/client.go:563 +0x2d2
github.com/Shopify/sarama.(*client).tryRefreshMetadata(0xc21e3a3f80, 0x0, 0x0, 0x0, 0x3, 0x0, 0x0)
	/home/mdmarek/go/root/src/github.com/Shopify/sarama/client.go:595 +0xb86
github.com/Shopify/sarama.(*client).RefreshMetadata(0xc21e3a3f80, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/mdmarek/go/root/src/github.com/Shopify/sarama/client.go:301 +0x161
github.com/Shopify/sarama.(*client).backgroundMetadataUpdater(0xc21e3a3f80)
	/home/mdmarek/go/root/src/github.com/Shopify/sarama/client.go:541 +0x10a
github.com/Shopify/sarama.*client.(github.com/Shopify/sarama.backgroundMetadataUpdater)·fm()
	/home/mdmarek/go/root/src/github.com/Shopify/sarama/client.go:135 +0x27
github.com/Shopify/sarama.withRecover(0xc21db856d0)
	/home/mdmarek/go/root/src/github.com/Shopify/sarama/utils.go:42 +0x3a
created by github.com/Shopify/sarama.NewClient
	/home/mdmarek/go/root/src/github.com/Shopify/sarama/client.go:135 +0x794
```

eapache:
Otherwise, in the rare case where a background metadata refresh coincided
exactly with a call to `Client.Close()` there was a race where we might have
potentially tried to write to a nil map. Also add a test for this.

Fixes #422.

@Shopify/kafka @epsniff

### DIFF
--- a/client.go
+++ b/client.go
@@ -70,8 +70,8 @@ const (
 )
 
 type client struct {
-	conf   *Config
-	closer chan none
+	conf           *Config
+	closer, closed chan none // for shutting down background metadata updater
 
 	// the broker addresses given to us through the constructor are not guaranteed to be returned in
 	// the cluster metadata (I *think* it only returns brokers who are currently leading partitions?)
@@ -111,6 +111,7 @@ func NewClient(addrs []string, conf *Config) (Client, error) {
 	client := &client{
 		conf:                    conf,
 		closer:                  make(chan none),
+		closed:                  make(chan none),
 		brokers:                 make(map[int32]*Broker),
 		metadata:                make(map[string]map[int32]*PartitionMetadata),
 		cachedPartitionsResults: make(map[string][maxPartitionIndex][]int32),
@@ -129,6 +130,7 @@ func NewClient(addrs []string, conf *Config) (Client, error) {
 		// indicates that maybe part of the cluster is down, but is not fatal to creating the client
 		Logger.Println(err)
 	default:
+		close(client.closed) // we haven't started the background updater yet, so we have to do this manually
 		_ = client.Close()
 		return nil, err
 	}
@@ -151,6 +153,10 @@ func (client *client) Close() error {
 		return ErrClosedClient
 	}
 
+	// shutdown and wait for the background thread before we take the lock, to avoid races
+	close(client.closer)
+	<-client.closed
+
 	client.lock.Lock()
 	defer client.lock.Unlock()
 	Logger.Println("Closing Client")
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -151,6 +152,10 @@ func (client *client) Close() error {
 		return ErrClosedClient
 	}
 
+	// shutdown and wait for the background thread before we take the lock, to avoid races
+	close(client.closer)
+	<-client.closed
```
epsniff: will this deadlock if the [RefreshFrequency](https://github.com/Shopify/sarama/blob/d77967306da5ac9363067b15e6c0f217cfe317e3/client.go#L533) is zero?

eapache: The job still gets started, it just shuts down immediately (see lines 536-540)

epsniff: NM...

epsniff: Just saw that, thanks! deleting my comments.

eapache: It will, however, deadlock if the client calls `Close()` on itself because of an error during initialization. So something still to fix.

eapache: fixed

wvanbergen:
Approach looks good. Too bad the test needs `time.Sleep` like that. :/

eapache:
I'm not as concerned about this test, since the sleeps are mainly to accommodate the timer of the background thread (whose value we control, but which can't just be 0) rather than to ensure things happen in a very particular order. I don't expect this test to be flaky with respect to future changes.

PR #425, opened 2015-04-16
Add some simple producer benchmarks

eapache:
@Shopify/kafka

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -7,7 +7,7 @@ import (
 	"time"
 )
 
-func safeClose(t *testing.T, c io.Closer) {
+func safeClose(t testing.TB, c io.Closer) {
 	err := c.Close()
 	if err != nil {
 		t.Error(err)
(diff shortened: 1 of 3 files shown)


wvanbergen:
Should we store historical results somewhere, e.g. on a wiki page?

eapache:
I've played with these a bit and the results aren't terribly consistent. I'm not sure how useful they will be in practice. If somebody else wants to try running them and see if they can get reliable results that would be useful.

wvanbergen:
:shipit:

wvanbergen:
I guess the performance really depends on Kafka as well - which inside a VM is not very predictable. With KAFKA_PEERS we could point it to a real cluster for potentially more stable results?

eapache:
I am still playing with that, slowly figuring out how to get reliable results. That's all cluster management though, this code can :ship:..

PR #428, opened 2015-04-16
Tweak travis config

eapache:
- only run go tip against latest kafka, there's no reason to add it to the full
  matrix
- let travis finish fast by reporting success as soon as all the required builds
  pass (so the tip build might still be running)

@Shopify/kafka


wvanbergen:
:ship::it:

PR #429, opened 2015-04-17
Producer logging tweaks

eapache:
- in leaderDispatcher, make the topic/partition part of the initial log slug for
  convenience and consistency
- don't log on every switch from normal to flushing in the loop, just the first,
  since 99% of the time we burn right through them all and we only need one line
  to indicate that, not 5 or 10
- similarly, don't log on every switch from flushing to normal, just do so if
  the highWatermark is 0 and it breaks the loop - in combination with the
  previous change we end up with just an entry and exit log for the loop, rather
  than two logs per iteration
- in messageAggregator, make the broker ID part of the slug (it wasn't even
  logged before)

@Shopify/kafka


wvanbergen:
:cool:

PR #431, opened 2015-04-20
Add Topics() and Partitions() to Consumer interface, and mock Consumer implementation.

### LINKED ISSUE #366: Add Consumer.ConsumeTopic(topic, offset)

wvanbergen:
This will allow you to implement most consumer use cases without a `Client` instance, and make them easier to test by adding them to the Consumer mock and allowing to stub values.

As discussed in #430, fixes #366. @Shopify/kafka

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -39,6 +39,15 @@ func (ce ConsumerErrors) Error() string {
 // on a consumer to avoid leaks, it will not be garbage-collected automatically when it passes out of
 // scope.
 type Consumer interface {
+
+	// Topics returns the set of available topics as retrieved from the cluster metadata.
+	// This method is the same as Client.Topics(), and is provided for convenience.
+	Topics() ([]string, error)
+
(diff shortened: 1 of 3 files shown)


consumer.go
```diff
@@ -39,6 +39,15 @@ func (ce ConsumerErrors) Error() string {
...
 // scope.
 type Consumer interface {
+
+	// Topics returns the set of available topics as retrieved from the cluster metadata.
+	// This methods delegates to the internal Client instance.
+	Topics() ([]string, error)
+
+	// Partitions returns the sorted list of all partition IDs for the given topic.
+	// This methods delegates to the internal Client instance.
+	Partitions(topic string) ([]int32, error)
```
eapache: Rather than mention delegation (technical aspect) mention that they are "the same as" and "provided for convenience" or something (UX aspect).

eapache:
Doc string comment, but otherwise LGTM

PR #432, opened 2015-04-24
Fix async producer typo

clutchski:
s/confured/enabled


eapache:
I think it was probably meant to be 'configured' originally, but 'enabled' works fine :)

Thanks!

PR #435, opened 2015-04-27
Clear producer message retry field on success

eapache:
Otherwise a client re-using message structs would run into weird problems.

@Shopify/kafka just occurred to me, we don't reuse messages ourself (and I suspect most people do not) but we already clear flags. Now we do it in one consistent place so this sort of inconsistency doesn't happen again.


wvanbergen:
:shipit:

PR #436, opened 2015-04-28
Refer to brokerConsumers consistently

eapache:
- rename a last few references to "brokerWorker"
- use `bc` instead of `w` as the abbreviation
- extract `newBrokerConsumer` as its own function

@Shopify/kafka no functional changes, just cleanup


wvanbergen:
LGTM.

PR #437, opened 2015-04-28
Consumer async partitions

### LINKED ISSUE #325: PartitionConsumer.Close() hangs frequently

eapache:
Feed consumer responses to the user asynchronously

Give each PartitionConsumer a goroutine whose job is to feed the user, and have
the brokerConsumer feed each partition with the response, then wait for acks.

This makes the consumer much more resilient to weird user consumption ordering,
as the old "one partition at a time" logic was very unforgiving.

Fixes #325 
Replaces #398 

@Shopify/kafka careful :eyes: on this please

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -124,8 +124,9 @@ func (c *consumer) ConsumePartition(topic string, partition int32, offset int64)
 		partition: partition,
 		messages:  make(chan *ConsumerMessage, c.conf.ChannelBufferSize),
 		errors:    make(chan *ConsumerError, c.conf.ChannelBufferSize),
+		feeder:    make(chan *FetchResponse, 1),
 		trigger:   make(chan none, 1),
-		dying:     make(chan none),
+		dying:     make(chan error, 1),
 		fetchSize: c.conf.Consumer.Fetch.Default,
 	}
 
@@ -144,6 +145,7 @@ func (c *consumer) ConsumePartition(topic string, partition int32, offset int64)
 	}
 
 	go withRecover(child.dispatcher)
+	go withRecover(child.responseFeeder)
 
 	child.broker = c.refBrokerConsumer(leader)
 	child.broker.input <- child
@@ -259,10 +261,12 @@ type partitionConsumer struct {
 	topic     string
 	partition int32
 
-	broker         *brokerConsumer
-	messages       chan *ConsumerMessage
-	errors         chan *ConsumerError
-	trigger, dying chan none
+	broker   *brokerConsumer
+	messages chan *ConsumerMessage
+	errors   chan *ConsumerError
+	feeder   chan *FetchResponse
+	trigger  chan none
+	dying    chan error
 
 	fetchSize           int32
 	offset              int64
@@ -306,8 +310,7 @@ func (child *partitionConsumer) dispatcher() {
 		child.consumer.unrefBrokerConsumer(child.broker)
 	}
 	child.consumer.removeChild(child)
-	close(child.messages)
-	close(child.errors)
+	close(child.feeder)
 }
 
 func (child *partitionConsumer) dispatch() error {
@@ -361,11 +364,11 @@ func (child *partitionConsumer) Errors() <-chan *ConsumerError {
 }
 
 func (child *partitionConsumer) AsyncClose() {
-	// this triggers whatever worker owns this child to abandon it and close its trigger channel, which causes
+	// this triggers whatever broker owns this child to abandon it and close its trigger channel, which causes
 	// the dispatcher to exit its loop, which removes it from the consumer then closes its 'messages' and
 	// 'errors' channel (alternatively, if the child is already at the dispatcher for some reason, that will
 	// also just close itself)
-	close(child.dying)
+	child.dying <- nil
 }
 
 func (child *partitionConsumer) Close() error {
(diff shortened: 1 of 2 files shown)


eapache:
After several network errors and one (valid) race condition bug found, CI is finally green.

consumer.go
```diff
@@ -392,6 +394,31 @@ func (child *partitionConsumer) HighWaterMarkOffset() int64 {
...
+			fallthrough
+		default:
+			child.sendError(err)
+			fallthrough
+		case ErrUnknownTopicOrPartition, ErrNotLeaderForPartition, ErrLeaderNotAvailable:
+			// these three are not fatal errors, but do require redispatching
+			Logger.Printf("consumer/%s/%d abandoned broker %d because %s\n", child.topic, child.partition, child.broker.broker.ID(), err)
+			child.broker.acks <- child
+			child.trigger <- none{}
+		}
```
wvanbergen: I personally don't like using `fallthrough` like this, but I suppose it has always worked like this so no need to change.

eapache: The alternative is to duplicate these three lines twice (and the `sendError` call one additional time). Using `fallthrough` explicitly, (as opposed to the stupid implicit C version) seems much more elegant to me.

I don't want to derail the PR (as you mentioned, this code is just moved, not changed), but you've piqued my curiousity - what bothers you about this pattern?

consumer.go
```diff
@@ -551,22 +579,12 @@ func (bc *brokerConsumer) subscriptionConsumer() {
...
-				case ErrUnknownTopicOrPartition, ErrNotLeaderForPartition, ErrLeaderNotAvailable:
-					// these three are not fatal errors, but do require redispatching
-					child.trigger <- none{}
-					delete(bc.subscriptions, child)
-					Logger.Printf("consumer/broker/%d abandoned subscription to %s/%d because %s\n", bc.broker.ID(), child.topic, child.partition, err)
-				}
+			child.feeder <- response
+		}
+
+		for i := len(bc.subscriptions); i > 0; i-- {
```
eapache: Iterating backwards here so we only call `len(bc.subscriptions)` once at the beginning - we may delete elements from it inside the loop, but we want to iterate the total number of times regardless. We also can't use `range` because the order we receive acks in (and thus the order we delete) has no relation to whatever iteration order go would give us

eapache: after writing the above - this is actually rather subtle

can anybody think of a simpler alternative, or should I just put the above as an in-code comment?

consumer.go
wvanbergen: It's just way harder to parse in my mind, especially with the `default` case not be the last one. I'd prefer a bunch of `if` statements.

consumer.go
wvanbergen: Why not a waitgroup? `wg.Add(len(bc.subscriptions))`

consumer.go
```diff
@@ -464,7 +491,7 @@ func (child *partitionConsumer) handleResponse(response *FetchResponse) error {
 type brokerConsumer struct {
 	consumer         *consumer
 	broker           *Broker
-	input            chan *partitionConsumer
+	input, acks      chan *partitionConsumer
```
wvanbergen: An `ack` in here is purely meant to say that a `partitionConsumer` child has handled the response?

consumer.go
eapache: A waitgroup would solve 95% of the problem very elegantly, but it doesn't let us communicate back "there was an error, abandon this subscription". However maybe that can be a totally different thing, let me play with this...

consumer.go
eapache: And to indicate whether the brokerConsumer should continue processing that subscription or if it's been redispatched

eapache:
Thoughts on the third commit? It does make the behaviour less subtle, I think, though perhaps at the cost of another moving part...

wvanbergen:
I like this approach better. Using a WaitGroup for waiting for a bunch of goroutines feels much more natural to me.

I am not a big fan of the extra bool field on PartitionConsumer. Maybe we should make the channel hold a wrapper type with a `partitionConsumer` and a `bool`, instead of adding a member to the `partitionConsumer` type? 

(Rant: Go really should support multi-value channels so you can simply feed multi-value return values to them without having to wrap them in a shitty type.)

eapache:
> Maybe we should make the channel hold a wrapper type with a partitionConsumer and a bool, instead of adding a member to the partitionConsumer type? 

Which channel?

wvanbergen:
the `acks` channel

eapache:
There is no acks channel anymore, it's a wait group o.o

wvanbergen:
Argh, I am an idiot.

eapache:
OK, this version removes the `needsDispatch` flag by switching the `dying` channel to carry actual useful values. Two side-effects:
- we can't close `dying` in the `AsyncClose` function since we might have an internal thread writing to it concurrently now; fortunately writing a `nil` value is just as good
- simplifies the switch which was bothering you, which means it's now just as easy to not use fallthrough, so do that

wvanbergen:
I think this makes sense, but we should stress test this as well.

@shivnagarajan @thegedge, opinions on this?

eapache:
I will merge this now to prep master for stress testing, but feel free to continue commenting @shivnagarajan and @thegedge.

PR #439, opened 2015-04-30
Add a simple topic consumer tool

wvanbergen:
It will consume all partitions for a given topic and write the messages to STDOUT.

@Shopify/kafka

### DIFF
--- a/tools/kafka-console-topicconsumer/kafka-console-topicconsumer.go
+++ b/tools/kafka-console-topicconsumer/kafka-console-topicconsumer.go
@@ -1,0 +1,125 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"os/signal"
+	"strings"
+	"sync"
+
+	"github.com/Shopify/sarama"
+)
+
+var (
+	brokerList = flag.String("brokers", os.Getenv("KAFKA_PEERS"), "The comma separated list of brokers in the Kafka cluster")
+	topic      = flag.String("topic", "", "REQUIRED: the topic to consume")
+	offset     = flag.String("offset", "newest", "The offset to start with. Can be `oldest`, `newest`")
+	verbose    = flag.Bool("verbose", false, "Whether to turn on sarama logging")
+	bufferSize = flag.Int("buffer-size", 256, "The buffer size of the message channel.")
+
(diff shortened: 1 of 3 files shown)


tools/kafka-console-topicconsumer/kafka-console-topicconsumer.go
```diff
@@ -0,0 +1,126 @@
...
+
+	for _, partition := range partitions {
+		pc, err := c.ConsumePartition(*topic, partition, initialOffset)
+		if err != nil {
+			printErrorAndExit(69, "Failed to start consumer for partition %d: %s", partition, err)
+		}
+
+		wg.Add(1)
+		go func(partition int32, pc sarama.PartitionConsumer) {
+			defer wg.Done()
```
eapache: this doesn't have to be part of the wait group, it must terminate before the messages goroutine does (except in the case of OffsetOutOfRange when I don't think it matters)

tools/kafka-console-topicconsumer/kafka-console-topicconsumer.go
```diff
@@ -0,0 +1,126 @@
...
+		printErrorAndExit(69, "Failed to start consumer: %s", err)
+	}
+
+	partitions, err := c.Partitions(*topic)
+	if err != nil {
+		printErrorAndExit(69, "Failed to get the list of partitions: %s", err)
+	}
+
+	var (
+		messages = make(chan *sarama.ConsumerMessage, 1)
```
eapache: I'd give this a bigger buffer to reduce context-switching; sarama's default buffer size for user-facing channels is 256

eapache:
Little things, but LGTM

bouk:
Maybe "kafka-console-consumer" is a better name, not like there's other things to consume in Kafka 😃

eapache:
We actually already have a `kafka-console-partitionconsumer` which only consumes a single partition, but I wouldn't mind just merging them into one `kafka-console-consumer` :)

wvanbergen:
I think `kafka-console-consumer` should be built on top the high-level consumer once we have it.

wvanbergen:
@Shopify/kafka Feel free to still comment on this further; there's definitely ways to improve this tool, but I had an urgent need for it, so I merge this PR.

PR #440, opened 2015-05-01
Kazoo

wvanbergen:
This adds Kazoo, a library to interact with Kafka's metadata in Zookeeper. This allows you to discover the Kafka brokers, and discover with the topic data.

Kazoo also provides most of the functionality to implement high-level consumer groups. The only thing that is missing is wrapper functions for the Zookeeper watches we need. We will need to have at least a watch on this instances of a consumergroup, most likely also on the partitions of the topics you are consuming, and potentially also on the list of topics. I will add these in a separate PR.

@Shopify/kafka @simongui

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ go:
 env:
   global:
   - KAFKA_PEERS=localhost:9091,localhost:9092,localhost:9093,localhost:9094,localhost:9095
+  - ZOOKEEPER_PEERS=localhost:2181,localhost:2182,localhost:2183,localhost:2184,localhost:2185
   - TOXIPROXY_ADDR=http://localhost:8474
   - KAFKA_INSTALL_ROOT=/home/travis/kafka
   - KAFKA_HOSTNAME=localhost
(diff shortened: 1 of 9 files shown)


eapache:
I've almost convinced myself this should have its own repository - thoughts?

wvanbergen:
For Kazoo, I can go either way; there's pros and cons for both.

The more interesting question to me is whether the consumergroup implementation should be a subpackage, or should be inside the sarama package. Adding it to sarama itself will always add the Zookeeper dependency to apps that use sarama, even if you don't need it (e.g. for a producing application). If we don't, I think we will continue to see confusion about the `Consumer` type not being a high level consumer.

eapache:
I think kazoo has a cool enough name (and is distinct enough) to get its own repository :)

> The more interesting question to me is whether the consumergroup implementation should be a subpackage

My plan was to add the offset management stuff first (which doesn't require zookeeper) and by the time we're done that 0.9 has come out and we don't need zookeeper at all. In other words, I would prefer to keep zookeeper out of sarama entirely, especially since talking to it directly will be obsolete soon anyways.

wvanbergen:
OK, closing.

PR #441, opened 2015-05-01
Shuffle seed brokers so we don't always connect to the first one provided

wvanbergen:
I stole this idea from the Zookeeper library because is such a simple yet powerful way to load balance the metadata requests in the cluster over the different brokers. After the initial seed broker shuffle, seed broker handling is deterministic, and will work the exact same way before.

The way I fixed the "resurrect dead brokers" test is not great, but that is primarily because of the limitation of the mock broker implementation. If you have ideas to do this in a better way, tell me :)

@Shopify/kafka

### DIFF
--- a/client.go
+++ b/client.go
@@ -1,6 +1,7 @@
 package sarama
 
 import (
+	"math/rand"
 	"sort"
 	"sync"
 	"time"
@@ -117,8 +118,10 @@ func NewClient(addrs []string, conf *Config) (Client, error) {
 		cachedPartitionsResults: make(map[string][maxPartitionIndex][]int32),
 		coordinators:            make(map[string]int32),
 	}
-	for _, addr := range addrs {
-		client.seedBrokers = append(client.seedBrokers, NewBroker(addr))
+
+	random := rand.New(rand.NewSource(time.Now().UnixNano()))
+	for _, index := range random.Perm(len(addrs)) {
+		client.seedBrokers = append(client.seedBrokers, NewBroker(addrs[index]))
 	}
 
 	// do an initial fetch of all cluster metadata by specifing an empty list of topics
--- a/client_test.go
+++ b/client_test.go
@@ -375,6 +375,21 @@ func TestClientRefreshBehaviour(t *testing.T) {
 }
 
 func TestClientResurrectDeadSeeds(t *testing.T) {
+	initialSeed := newMockBroker(t, 0)
+	emptyMetadata := new(MetadataResponse)
+	initialSeed.Returns(emptyMetadata)
+
+	conf := NewConfig()
+	conf.Metadata.Retry.Backoff = 0
+	conf.Metadata.RefreshFrequency = 0
+	c, err := NewClient([]string{initialSeed.Addr()}, conf)
+	if err != nil {
+		t.Fatal(err)
+	}
+	initialSeed.Close()
+
+	client := c.(*client)
+
 	seed1 := newMockBroker(t, 1)
 	seed2 := newMockBroker(t, 2)
 	seed3 := newMockBroker(t, 3)
@@ -382,17 +397,10 @@ func TestClientResurrectDeadSeeds(t *testing.T) {
 	addr2 := seed2.Addr()
 	addr3 := seed3.Addr()
 
-	emptyMetadata := new(MetadataResponse)
-	seed1.Returns(emptyMetadata)
-
-	conf := NewConfig()
-	conf.Metadata.Retry.Backoff = 0
-	conf.Metadata.RefreshFrequency = 0
-	c, err := NewClient([]string{addr1, addr2, addr3}, conf)
-	if err != nil {
-		t.Fatal(err)
-	}
-	client := c.(*client)
+	// Overwrite the seed brokers with a fixed ordering to make this test deterministic.
+	safeClose(t, client.seedBrokers[0])
+	client.seedBrokers = []*Broker{NewBroker(addr1), NewBroker(addr2), NewBroker(addr3)}
+	client.deadSeeds = []*Broker{}
 
 	wg := sync.WaitGroup{}
 	wg.Add(1)


client.go
```diff
@@ -720,3 +723,11 @@ func (client *client) getConsumerMetadata(consumerGroup string, attemptsRemainin
 	client.resurrectDeadBrokers()
 	return retry(ErrOutOfBrokers)
 }
+
+// shuffleSeedBrokers performs a Fisher-Yates shuffle on the seed broker list
```
eapache: Simpler just to use `rand.Perm()`?

eapache:
Just cast to `*client` briefly in the test and read the order out of the private member, then the rest of the test can proceed exactly as it was.

client.go
wvanbergen: Done

wvanbergen:
Yup, that's what I am doing right now, but I have to do that after the client is initialized.

client.go
```diff
@@ -117,8 +118,9 @@ func NewClient(addrs []string, conf *Config) (Client, error) {
 		cachedPartitionsResults: make(map[string][maxPartitionIndex][]int32),
 		coordinators:            make(map[string]int32),
 	}
-	for _, addr := range addrs {
-		client.seedBrokers = append(client.seedBrokers, NewBroker(addr))
+
+	for _, index := range rand.Perm(len(addrs)) {
```
eapache: we need to seed the random number generator; to avoid destroying our parent program (if it has already seeded with a specific value for some reason) we shouldn't use the global but instantiate a local source instead... not sure if it makes sense to create a package-local in `sarama.go` or just a real local right here

client_test.go
```diff
@@ -375,25 +375,32 @@ func TestClientRefreshBehaviour(t *testing.T) {
...
-	seed1 := newMockBroker(t, 1)
-	seed2 := newMockBroker(t, 2)
-	seed3 := newMockBroker(t, 3)
-	addr1 := seed1.Addr()
-	addr2 := seed2.Addr()
-	addr3 := seed3.Addr()
-
+	initialSeed := newMockBroker(t, 0)
 	emptyMetadata := new(MetadataResponse)
-	seed1.Returns(emptyMetadata)
```
eapache: can't you just create a goroutine that waits until `NewClient` is blocked, then reads off the seed order from the client and returns metadata to the first broker on that list?

eapache:
Did you intend for this to be in 1.4 or not?

client_test.go
wvanbergen: Maybe, but I find this way simpler and easier to understand. The `RefreshMetadata` call is what is being tested, not the `NewClient` call.

wvanbergen:
I think this would be a nice addition to 1.4. It's such a minimal change that it doesn't require an extra batch of stress testing.

client.go
wvanbergen: I've added a package-local `sarama.Random`, which I seed using `time.Now().UnixNano()`.

sarama.go
```diff
@@ -22,6 +24,8 @@ import (
 // but you can set it to redirect wherever you want.
 var Logger StdLogger = log.New(ioutil.Discard, "[Sarama] ", log.LstdFlags)
 
+var Random = rand.New(rand.NewSource(time.Now().UnixNano()))
```
eapache: Should this really be public? I'm not sure why anyone would want to override it. If you want to leave it public, it needs godoc

client_test.go
eapache: the current version leaks a broker goroutine, doesn't it?

client_test.go
```diff
@@ -375,25 +375,32 @@ func TestClientRefreshBehaviour(t *testing.T) {
...
 
+	seed1 := newMockBroker(t, 1)
+	seed2 := newMockBroker(t, 2)
+	seed3 := newMockBroker(t, 3)
+	addr1 := seed1.Addr()
+	addr2 := seed2.Addr()
+	addr3 := seed3.Addr()
+
+	// Overwrite the seed brokers with a fixed ordering to make this test deterministic.
+	client.seedBrokers = []*Broker{NewBroker(addr1), NewBroker(addr2), NewBroker(addr3)}
```
eapache: rather than completely overriding it, just append the two new ones is simpler

sarama.go
```diff
@@ -22,6 +24,10 @@ import (
 // but you can set it to redirect wherever you want.
 var Logger StdLogger = log.New(ioutil.Discard, "[Sarama] ", log.LstdFlags)
 
+// Random is a random number generator instance used throughout sarama for random behavior.
```
eapache: we're not using it for random partitioner instances, should we? it then becomes a single point of contentiont

eapache: I'm starting to think we should have just kept it local to `NewClient` :P

client_test.go
wvanbergen: Fixed that.

client_test.go
wvanbergen: I feel this is simpler to understand, because now you can basically ignore what happened before; the seed broker state is complete defined here. YMMV.

eapache:
This version LGTM.

shivnagarajan:
Looks good. How often is the NewClient called, generating a new permutation? More specifically, how large can the list of brokers be? Generating a permutation often can be expensive, but if the list is small, it should not matter.

Shiv

wvanbergen:
`NewClient` is only called once for an application (usually), so I am not too worried about the cost of this. 

The number of brokers in a cluster may be hundreds for installations that are significantly larger than ours, but especially in those environments I would like to prevent all my clients from connecting to the same broker :)

shivnagarajan:
looks good. :shipit:

eapache:
Also, for what it's worth: building the permutation appears to be `O(n)` in the current implementation: https://golang.org/src/math/rand/rand.go?s=4735:4767#L136

shivnagarajan:
Nice.. I should have looked at that before my claims :-)

PR #442, opened 2015-05-01
Build the partitionConsumer into the topicConsumer

eapache:
@Shopify/kafka this is what I meant by merging them. Adding a flag to the topicConsumer makes the partitionConsumer obsolete.


tools/kafka-console-topicconsumer/kafka-console-topicconsumer.go
```diff
@@ -110,6 +112,24 @@ func main() {
...
+
+	tmp := strings.Split(*partitions, ",")
+	var pList []int32
+	for i := range tmp {
+		val, err := strconv.ParseInt(tmp[i], 10, 32)
+		if err != nil {
+			return nil, err
+		}
+		pList = append(pList, int32(val))
+	}
```
wvanbergen: Should we check whether the partition is the `c.Partitions(*topic)` list, or just let it fail later?

wvanbergen:
Maybe just call it `kafka-console-consumer` now?

tools/kafka-console-topicconsumer/kafka-console-topicconsumer.go
eapache: It fails later pretty nicely with `ERROR: Failed to start consumer for partition -1: kafka server: Request was for a topic or partition that does not exist on this broker.`

wvanbergen: :+1: that should be enough :)

eapache:
Oh, does this whole thing break semantic versioning? Does this need to be a 2.0 change?

wvanbergen:
I don't feel it's needed, but maybe we should state this in the `tools/README`?

eapache:
Added a note to `tools/README` that they are not guaranteed to version like the main library,

eapache:
The alternative would just be to not remove `partitionConsumer`, since `topicConsumer` hadn't been released yet

wvanbergen:
Like I said, I don't feel too strongly about maintaining backwards compatibility for the tools, but if you do that works for me :+1:

eapache:
I'm not 100% sure I want to commit to API stability for the tools for eternity, but in this case it's easy enough. I'll add "remove partitionConsumer" to our 2.0 wiki page.

eapache:
Last :eyes:? I had to do some git dancing to rebase while undeleting the partitionConsumer.

tools/README.md
```diff
@@ -4,7 +4,7 @@ This folder contains applications that are useful for exploration of your Kafka
 Some of these tools mirror tools that ship with Kafka, but these tools won't require installing the JVM to function.
 
 - [kafka-console-producer](./kafka-console-producer): a command line tool to produce a single message to your Kafka custer.
-- [kafka-console-partitionconsumer](./kafka-console-partitionconsumer): a command line tool to consume a single partition of a topic on your Kafka cluster.
-- [kafka-console-topicconsumer](./kafka-console-topicconsumer): a command line tool to consume all partition of a topic on your Kafka cluster.
+- [kafka-console-partitionconsumer](./kafka-console-partitionconsumer): (deprecated) a command line tool to consume a single partition of a topic on your Kafka cluster.
+- [kafka-console-consumer](./kafka-console-consumer): a command line tool to consume arbitrary partitions of a topic on your Kafka cluster.
```
wvanbergen: "all or an arbitrary list of"?

wvanbergen:
minor wording quibble - :shipit:

tools/README.md
eapache: "all of them" still qualifies as an arbitrary list :)

PR #443, opened 2015-05-01
Toxiproxy cleanup

eapache:
Bump toxiproxy to fix one potential issue, and add some framework for using it in functional tests.

@Shopify/kafka only test code changes, so if CI is happy this can go in 1.4 as well.


wvanbergen:
Yeah, this is a nice first step :+1:

eapache:
Updated to cover a few functional tests that I missed in the first pass, merging when :green_apple:

PR #444, opened 2015-05-01
Functional failover test using toxiproxy

eapache:
Write a new test `TestReliableProducing` that uses toxiproxy to take down
zookeeper connections and ensures that everything produces correctly.

Succeeds #404, and probably depends on upstream's https://issues.apache.org/jira/browse/KAFKA-2082 just like its predecessor.


eapache:
Less flaky on 0.9 but still seems flaky...

eapache:
Finally giving up on this one.

PR #447, opened 2015-05-08
trivial: replace x=x+y with x+=y

eapache:
Also organize imports according to the go-import convention used everywhere else
already.

cc @Shopify/kafka


wvanbergen:
belated :+1:

Issue #449, opened 2015-05-14
panic: sync: negative WaitGroup counter


cep21:
Seeing this occasionally in production:

Revision: 23d523386ce0c886e56c9faf1b9c78b07e5b8c90

```
panic: sync: negative WaitGroup counter

goroutine 444 [running]:
sync.(*WaitGroup).Add(0xc208e6c510, 0xffffffffffffffff)
        /usr/src/go/src/sync/waitgroup.go:67 +0x96
sync.(*WaitGroup).Done(0xc208e6c510)
        /usr/src/go/src/sync/waitgroup.go:85 +0x31
github.com/Shopify/sarama.(*asyncProducer).returnSuccesses(0xc208e6c4d0, 0xc29ffe9600, 0x3f, 0x40)
        /opt/goproj/Godeps/_workspace/src/github.com/Shopify/sarama/async_producer.go:753 +0xd3
github.com/Shopify/sarama.(*asyncProducer).flusher(0xc208e6c4d0, 0xc20814c540, 0xc2080b9800)
        /opt/goproj/Godeps/_workspace/src/github.com/Shopify/sarama/async_producer.go:560 +0xafe
github.com/Shopify/sarama.func·006()
        /opt/goproj/Godeps/_workspace/src/github.com/Shopify/sarama/async_producer.go:430 +0x43
github.com/Shopify/sarama.withRecover(0xc20802b220)
        /opt/goproj/Godeps/_workspace/src/github.com/Shopify/sarama/utils.go:42 +0x3a
created by github.com/Shopify/sarama.(*asyncProducer).messageAggregator
        /opt/goproj/Godeps/_workspace/src/github.com/Shopify/sarama/async_producer.go:430 +0x205
```

wvanbergen:
Is this while running normally, or when closing the producer?

eapache:
What configuration are you running with? What do you see in the sarama logs when this occurs? I took a quick look and didn't see any obvious cases where the counter could get out of sync.

cep21:
This is between 3 and 8 minutes of running in production.  We have two tiers of production traffic and only one was seeing this behavior (the one that gets the most traffic).

There is no information in stdout or stderr.  There is nothing in the logs other than messages of the form "client/metadata fetching metadata for all topics from broker kafka17.blah.blah:9092", "Connected to broker at kafka18--bbac.blah:9092 (registered as #10)" and "producer/flusher/12 starting up"

[...]

PR #450, opened 2015-05-14
Ensure we always have called Add() on the inflight counter before we Wait() for it

wvanbergen:
@eapache This is not the cause of #449, but fixes another a potential issue when using the waitgroup.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -221,6 +221,7 @@ func (p *asyncProducer) topicDispatcher() {
 
 		if msg.flags&shutdown != 0 {
 			shuttingDown = true
+			p.inFlight.Done()
 			continue
 		} else if msg.retries == 0 {
 			p.inFlight.Add(1)
@@ -256,7 +257,7 @@ func (p *asyncProducer) topicDispatcher() {
 
 // one per topic
 // partitions messages, then dispatches them by partition
-func (p *asyncProducer) partitionDispatcher(topic string, input chan *ProducerMessage) {
+func (p *asyncProducer) partitionDispatcher(topic string, input <-chan *ProducerMessage) {
 	handlers := make(map[int32]chan *ProducerMessage)
 	partitioner := p.conf.Producer.Partitioner(topic)
 	breaker := breaker.New(3, 1, 10*time.Second)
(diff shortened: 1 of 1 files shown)


async_producer.go
```diff
@@ -610,6 +611,7 @@ func (p *asyncProducer) retryHandler() {
 
 func (p *asyncProducer) shutdown() {
 	Logger.Println("Producer shutting down.")
+	p.inFlight.Add(1)
```
eapache: I don't think this is necessary? If you create a wait group and then immediately Wait on it (without ever calling Add) it just returns right away...

async_producer.go
```diff
@@ -256,7 +257,7 @@ func (p *asyncProducer) topicDispatcher() {
 
 // one per topic
 // partitions messages, then dispatches them by partition
-func (p *asyncProducer) partitionDispatcher(topic string, input chan *ProducerMessage) {
+func (p *asyncProducer) partitionDispatcher(topic string, input <-chan *ProducerMessage) {
```
eapache: these channel restrictions are a good idea though

wvanbergen:
I think, this could potentially be an issue:

``` go
p := NewAsyncProducer(...)
p.Input() <- msg
p.Close()
```

if the `Close` call gets handled before the goroutine has read the first message from the input channel, the inflight counter is still 0.

eapache:
Oh interesting, you're right that's a legit issue. However, I'm not sure your patch fixes all the cases :)
```go
p := NewAsyncProducer(...)
p.Close()
p.Input() <- msg
```
With your patch, if the input message is handled before the waiting goroutine wakes up, then we will increment the counter from 0 before immediately decrementing it... so I think we must stop doing the "increment+decrement" trick for messages we return with the shutting-down error as well.

wvanbergen:
I don't really understand this one? `Close()` doesn't return until the `Wait()` is finished?

eapache:
1. `Close()` calls `Wait()`
2. Go switches contexts
3. the message is sent on `Input()`, is received, causes the wait counter to increment from 0 (the problem case) and decrement again since it is immediately returned as an error

wvanbergen:
(I assume you means `AsyncClose()`, otherwise message on the input channel can never happen until the `Wait()` is finished.)

I don't really see what the problem is: after calling AsyncClose(), no message will ever make it into the processing logic, and we always wait for all the messages being processed to be completed. The worst that can happen is that we wait a bit too long, for `msg` to be returned as error?

eapache:
> I assume you means AsyncClose()

Ya, sorry

> no message will ever make it into the processing logic

But it will still cause the wait counter to be `Add(1)`ed to, which is the bug

eapache:
OK, this is an improvement regardless so I'll merge it; I'll also make a PR which should (I hope) fix and demonstrate the other case I'm thinking of.

PR #451, opened 2015-05-18
Fix another hypothetical wait group issue

eapache:
Don't increment and immediately decrement the wait group on a message which we
just return with ErrShuttingDown. Otherwise:
```
prod.AsyncClose() // starts prod.shutdown() goroutine
prod.Input() <- msg
```
1. shutdown() goroutine calls Wait()
2. while in the call to Wait(), Go switches contexts
3. the message is sent on Input(), is received, and causes the wait counter to
   increment from 0 while being waited on

Where the WaitGroup docs say "calls with a positive delta that occur when the
counter is zero must happen before a Wait".

@Shopify/kafka @wvanbergen hopefully this explanation makes sense, it was the second issue I came up with from #450.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -224,11 +224,18 @@ func (p *asyncProducer) topicDispatcher() {
 			p.inFlight.Done()
 			continue
 		} else if msg.retries == 0 {
-			p.inFlight.Add(1)
 			if shuttingDown {
-				p.returnError(msg, ErrShuttingDown)
+				// we can't just call returnError here because that decrements the wait group,
+				// which hasn't been incremented yet for this message, and shouldn't be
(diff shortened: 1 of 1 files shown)


eapache:
Oh hmm. Willem's fix may have already fixed this case (though more subtly) because once the shutdown message has arrived and caused `Done()` to be called the wait group's internal semaphore has already been triggered, causing the `Wait()` in `shutdown` to unblock, so it *probably* doesn't matter that the increment happens from 0 because the wait isn't really waiting anymore? I'm not sure.

Better safe than sorry on this I guess?

wvanbergen:
Yeah, I don't think this is strictly necessary, but it doesn't hurt either.  Given that the message never really is "in flight", it makes sense not to increment the counter for this.

PR #453, opened 2015-05-22
Update godoc for Client.Coordinator and Client.RefreshCoordinator

wvanbergen:
Mention requirement of kafka 0.8.2. /cc @eapache and @zaynetro


eapache:
belated :+1:

PR #455, opened 2015-05-27
fix code.google.com/p/snappy-go/snappy to github.com/golang/snappy

qaohv:

PR #457, opened 2015-05-27
Fix import location of snappy.

t15k:
Project has been moved from code.google.com/p/snappy-go/snappy to
github.com/golang/snappy. As the code.google.com redirects to the
new location, it breaks a clean build of the current master.


eapache:
I believe #455 beat you to it, thanks though!

PR #459, opened 2015-05-28
Update and simplify doc for MaxResponseSize

### LINKED ISSUE #458: Clarification on sarama.MaxResponseSize limit
I don't think the comment here is necessarily correct: https://github.com/Shopify/sarama/blob/master/sarama.go#L44
as I have observed receiving a 170MB fetch response that is hitting the MaxResponseSize limit. It seems like socket.request.max.bytes is a limit only on the request size and not the response size. Also this: https://issues.apache.org/jira/browse/KAFKA-2063 seems to indicate that you can't actually control the fetch response size.

Also just curious as to whether there is some protocol limitation that absolutely prevents a response from being larger than 2GB?

eapache:
Hopefully sufficient to fix #458.

@4578395263256 does this read better?

cc @Shopify/kafka

### DIFF
--- a/sarama.go
+++ b/sarama.go
@@ -40,7 +40,8 @@ var PanicHandler func(interface{})
 var MaxRequestSize int32 = 100 * 1024 * 1024
 
 // MaxResponseSize is the maximum size (in bytes) of any response that Sarama will attempt to parse. If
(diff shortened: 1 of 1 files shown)


4578395263256:
Looks good, thanks!

wvanbergen:
LGTM.

PR #461, opened 2015-05-29
OffsetManager Implementation

eapache:
There are still some TODOs sprinkled throughout, but not all of them necessarily need to be fixed before we ship this (it's already pretty close to the minimum viable implementation). The only critical missing piece is implementing `AsyncClose` for the `PartitionOffsetManager`.

Regardless, it's at the state where it can be opened up for comments. This should completely supersede #379.

### DIFF
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -384,6 +384,10 @@ func (bom *brokerOffsetManager) mainLoop() {
 
 func (bom *brokerOffsetManager) flushToBroker() {
 	request := bom.constructRequest()
+	if request == nil {
+		return
+	}
+
 	response, err := bom.broker.CommitOffset(request)
 
 	if err != nil {
@@ -428,12 +432,20 @@ func (bom *brokerOffsetManager) constructRequest() *OffsetCommitRequest {
 		Version:       1,
 		ConsumerGroup: bom.parent.group,
 	}
+
 	for s := range bom.subscriptions {
 		s.lock.Lock()
-		r.AddBlock(s.topic, s.partition, s.offset, 0, s.metadata)
+		if s.dirty {
+			r.AddBlock(s.topic, s.partition, s.offset, 0, s.metadata)
+		}
 		s.lock.Unlock()
 	}
-	return r
+
+	if len(r.blocks) > 0 {
+		return r
+	}
+
+	return nil
 }
 
 func (bom *brokerOffsetManager) abort(err error) {
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -243,7 +243,7 @@ func (pom *partitionOffsetManager) fetchInitialOffset(retries int) error {
 		return nil
 	case ErrNotCoordinatorForConsumer:
 		if retries <= 0 {
-			return err
+			return block.Err
 		}
 		if err := pom.selectBroker(); err != nil {
 			return err
@@ -251,12 +251,12 @@ func (pom *partitionOffsetManager) fetchInitialOffset(retries int) error {
 		return pom.fetchInitialOffset(retries - 1)
 	case ErrOffsetsLoadInProgress:
 		if retries <= 0 {
-			return err
+			return block.Err
 		}
 		time.Sleep(pom.parent.conf.Metadata.Retry.Backoff)
 		return pom.fetchInitialOffset(retries - 1)
 	default:
-		return err
+		return block.Err
 	}
 }
 
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -140,6 +140,7 @@ type partitionOffsetManager struct {
 	lock     sync.Mutex
 	offset   int64
 	metadata string
+	dirty    bool
 	broker   *brokerOffsetManager
 
 	errors    chan *ConsumerError
@@ -278,6 +279,16 @@ func (pom *partitionOffsetManager) SetOffset(offset int64, metadata string) {
 
 	pom.offset = offset
 	pom.metadata = metadata
+	pom.dirty = true
+}
+
+func (pom *partitionOffsetManager) updateCommitted(offset int64, metadata string) {
+	pom.lock.Lock()
+	defer pom.lock.Unlock()
+
+	if pom.offset == offset && pom.metadata == metadata {
+		pom.dirty = false
+	}
 }
 
 func (pom *partitionOffsetManager) Offset() (int64, string) {
@@ -376,6 +387,8 @@ func (bom *brokerOffsetManager) flushToBroker() {
 
 		switch err {
 		case ErrNoError:
+			block := request.blocks[s.topic][s.partition]
+			s.updateCommitted(block.offset, block.metadata)
 			break
 		case ErrUnknownTopicOrPartition, ErrNotLeaderForPartition, ErrLeaderNotAvailable:
 			delete(bom.subscriptions, s)
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -1,0 +1,419 @@
+package sarama
+
+import (
+	"sync"
+	"time"
+)
+
+// Offset Manager
+
+// OffsetManager uses Kafka to store and fetch consumed partition offsets.
+type OffsetManager interface {
+	// ManagePartition creates a PartitionOffsetManager on the given topic/partition. It will
+	// return an error if this OffsetManager is already managing the given topic/partition.
+	ManagePartition(topic string, partition int32) (PartitionOffsetManager, error)
+}
+
+type offsetManager struct {
+	client Client
+	conf   *Config
+	group  string
+
+	lock sync.Mutex
+	poms map[string]map[int32]*partitionOffsetManager
+	boms map[*Broker]*brokerOffsetManager
+}
+
+// NewOffsetManagerFromClient creates a new OffsetManager from the given client.
+// It is still necessary to call Close() on the underlying client when finished with the partition manager.
+func NewOffsetManagerFromClient(group string, client Client) (OffsetManager, error) {
+	// Check that we are not dealing with a closed Client before processing any other arguments
+	if client.Closed() {
+		return nil, ErrClosedClient
+	}
+
+	om := &offsetManager{
+		client: client,
+		conf:   client.Config(),
+		group:  group,
+		poms:   make(map[string]map[int32]*partitionOffsetManager),
+		boms:   make(map[*Broker]*brokerOffsetManager),
+	}
+
+	return om, nil
+}
+
+func (om *offsetManager) ManagePartition(topic string, partition int32) (PartitionOffsetManager, error) {
+	pom, err := om.newPartitionOffsetManager(topic, partition)
+	if err != nil {
+		return nil, err
+	}
+
+	om.lock.Lock()
+	defer om.lock.Unlock()
+
+	topicManagers := om.poms[topic]
+	if topicManagers == nil {
+		topicManagers = make(map[int32]*partitionOffsetManager)
+		om.poms[topic] = topicManagers
+	}
+
+	if topicManagers[partition] != nil {
+		return nil, ConfigurationError("That topic/partition is already being managed")
+	}
+
+	topicManagers[partition] = pom
+	return pom, nil
+}
+
+func (om *offsetManager) refBrokerOffsetManager(broker *Broker) *brokerOffsetManager {
+	om.lock.Lock()
+	defer om.lock.Unlock()
+
+	bom := om.boms[broker]
+	if bom == nil {
+		bom = om.newBrokerOffsetManager(broker)
+		om.boms[broker] = bom
+	}
+
+	bom.refs++
+
+	return bom
+}
+
+func (om *offsetManager) unrefBrokerOffsetManager(bom *brokerOffsetManager) {
+	om.lock.Lock()
+	defer om.lock.Unlock()
+
+	bom.refs--
+
+	if bom.refs == 0 {
+		close(bom.updateSubscriptions)
+		if om.boms[bom.broker] == bom {
+			delete(om.boms, bom.broker)
+		}
+	}
+}
+
+func (om *offsetManager) abandonBroker(bom *brokerOffsetManager) {
+	om.lock.Lock()
+	defer om.lock.Unlock()
+
+	delete(om.boms, bom.broker)
+}
+
+// Partition Offset Manager
+
+// PartitionOffsetManager uses Kafka to store and fetch consumed partition offsets. You MUST call Close()
+// on a partition offset manager to avoid leaks, it will not be garbage-collected automatically when it passes
+// out of scope.
+type PartitionOffsetManager interface {
+	// Offset returns the current offset and metadata according to the manager; this value has not necessarily
+	// been flushed to the cluster yet.
+	Offset() (int64, string)
+
+	// SetOffset sets the current offset and metadata according to the manager; this value (or a subsequent update)
+	// will eventually be flushed to the cluster based on configuration.
+	SetOffset(offset int64, metadata string)
+
+	// Errors returns a read channel of errors that occur during offset management, if enabled. By default,
+	// errors are logged and not returned over this channel. If you want to implement any custom error
+	// handling, set your config's Consumer.Return.Errors setting to true, and read from this channel.
+	Errors() <-chan *ConsumerError
+	// AsyncClose initiates a shutdown of the PartitionOffsetManager. This method will return immediately,
+	// after which you should wait until the 'errors' channel has been drained and closed.
+	// It is required to call this function, or Close before a consumer object passes out of scope,
+	// as it will otherwise leak memory.  You must call this before calling Close on the underlying
+	// client.
+	AsyncClose()
+	// Close stops the PartitionOffsetManager from managing offsets. It is required to call this function
+	// (or AsyncClose) before a PartitionOffsetManager object passes out of scope, as it will otherwise
+	// leak memory. You must call this before calling Close on the underlying client.
+	Close() error
+}
+
+type partitionOffsetManager struct {
+	parent    *offsetManager
+	topic     string
+	partition int32
+
+	lock     sync.Mutex
+	offset   int64
+	metadata string
+	broker   *brokerOffsetManager
+
+	errors    chan *ConsumerError
+	rebalance chan none
+	dying     chan none
+}
+
+func (om *offsetManager) newPartitionOffsetManager(topic string, partition int32) (*partitionOffsetManager, error) {
+	pom := &partitionOffsetManager{
+		parent:    om,
+		topic:     topic,
+		partition: partition,
+		errors:    make(chan *ConsumerError, om.conf.ChannelBufferSize),
+		rebalance: make(chan none, 1),
+		dying:     make(chan none),
+	}
+
+	if err := pom.selectBroker(); err != nil {
+		return nil, err
+	}
+
+	if err := pom.fetchInitialOffset(om.conf.Metadata.Retry.Max); err != nil {
+		return nil, err
+	}
+
+	go withRecover(pom.mainLoop)
+
+	return pom, nil
+}
+
+func (pom *partitionOffsetManager) mainLoop() {
+	for {
+		select {
+		case <-pom.rebalance:
+			if err := pom.selectBroker(); err != nil {
+				pom.handleError(err)
+				pom.rebalance <- none{}
+			}
+		case <-pom.dying:
+			if pom.broker != nil {
+				select {
+				case <-pom.rebalance:
+				case pom.broker.updateSubscriptions <- pom:
+				}
+				pom.parent.unrefBrokerOffsetManager(pom.broker)
+			}
+			close(pom.errors)
+			return
+		}
+	}
+}
+
+func (pom *partitionOffsetManager) selectBroker() error {
+	if pom.broker != nil {
+		pom.parent.unrefBrokerOffsetManager(pom.broker)
+		pom.broker = nil
+	}
+
+	var broker *Broker
+	var err error
+
+	if err = pom.parent.client.RefreshCoordinator(pom.parent.group); err != nil {
+		return err
+	}
+
+	if broker, err = pom.parent.client.Coordinator(pom.parent.group); err != nil {
+		return err
+	}
+
+	pom.broker = pom.parent.refBrokerOffsetManager(broker)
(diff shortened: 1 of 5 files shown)


eapache:
@Shopify/kafka

piotrkowalczuk:
What is approximate ETA?

wvanbergen:
We don't have any direct need for this, so finishing this PR does not have high priority for us. If you're willing to work on this, the following things need to be added before we can merge it:

- `AsyncClose`
- unit tests
- functional tests

eapache:
I just added a quick and dirty `AsyncClose` implementation, so this is *theoretically* feature-complete in a minimal state, but I can pretty much guarantee it's broken since I've never actually run it.

eapache:
It's not clear to me if kafka-based offset management is functional in 0.8.2 since v1 of the request requires fields which won't have meaning until the full consumer rewrite lands in 0.9 (https://cwiki.apache.org/confluence/display/KAFKA/Kafka+0.9+Consumer+Rewrite+Design)

wvanbergen:
It's functional; when I was testing this it was successfully storing and retrieving offsets and metadata.

eapache:
What did you put in those fields then? Zeros?

(Specifically: ConsumerGroupGeneration, and ConsumerID)

wvanbergen:
Yup, I just didn't set them: https://github.com/Shopify/sarama/pull/379/files

horkhe:
Hey guys, I was looking at the offset manager interface from the end user perspective and I have a couple of suggestions to make:
* It would be nice if consumer group was not read from config, but rather accepted as a parameter along with topic and partition. That would allow using one offset manager for many consumer groups.
* Why does the user should be concerned with creating a committer for each particular topic/partition? This seems to be unnecessary complication. Why not make the OffsetConsumer take care of that and provide a stupid simple interface. e.g.:
```go
type OffsetResult struct {
    offset int64
    err error
}

type OffsetManger interface {
    FetchOffset(group, topic string, partition int32, replyTo chan<- OffsetResult)
    CommitOffset(group, topic string, partition int32, offset int64, replyTo chan<- error)
    Close() error
}

NewOffsetManagerWithClient(c *Client) (*OffsetManger, error)
```
I understand that you were modeling the API to align with Consumer/PartitionConsumer, but the OffsetManager seems to be a higher level abstraction that asks for a simpler interface. Just think how much code needs to be written to make use of the OffsetManager implemented in this PR comparing to the usage of the suggested interface.

This is certainly just my subjective opinion. In any case thank you very much, guys, for what you are doing.

eapache:
> allow using one offset manager for many consumer groups

That's actually surprisingly difficult. While you can include multiple topics and partitions in a single offset fetch or commit, you can only include a single consumer group (this is a limitation of the kafka protocol) so having an OffsetManager support multiple consumer groups would require a great deal more work.

> Why does the user should be concerned with creating a committer for each particular topic/partition?

Eventually they won't have to create anything at all, it will be baked into the consumer (so creating a PartitionConsumer will automatically create a PartitionOffsetManager behind the scenes that keeps itself up-to-date with the messages you've consumed). I just haven't written that part yet :)

> I understand that you were modeling the API to align with Consumer/PartitionConsumer

Yup, see my previous answer.

horkhe:
@eapache makes sense. Thank you for the clarification.

horkhe:
@eapache Could you please consider making **consumer group** into a parameter of `NewOffsetManager`, for having it in `sarama.Config` forces us to use one `sarama.Client` per offset manager.

aaronkavlie-wf:
We're using Sarama, and looking at our options for offset management.

Is this PR still a low priority for?

wvanbergen:
@aaronkavlie-wf We are not in need of this feature ourselves (we use Zookeeper to manage our offsets). To get this rolling again, it would be great if you can try out this branch to a) see if it works and b) give feedback on the proposed API.

horkhe:
@eapache @wvanbergen I have tried to use this offset manager in our balanced consumer implementation here are my comments (some of them have already been mentioned above by I decided to list them all here anyway):
* It would be nice to have `group` as a `OffsetManager.ManageParitition` parameter or at least as a `NewOffsetManager` parameter rather than a part of `Config`. The later approach forces the user to make a client connection per consumer group;
* `metadata` should probably be a parameter of `SetOffset` rather than settable on its own.
* `PartitionOffsetManager.Close` blocks forever because `errors` channel is never closed in the implementation.
* If a `PartitionOffsetManager` is closed a pending offset may not be committed.

aaronkavlie-wf:
@wvanbergen are you using [your kafka repo](https://github.com/wvanbergen/kafka) for offset management (in production)? Or something else?

wvanbergen:
@aaronkavlie-wf yup that is what we are using.

aaronkavlie-wf:
I tried this out yesterday and did some simple testing. It seems to be storing/retrieving offsets just fine.

I agree with @horkhe that it's cumbersome to manage the partition offset managers; baking it into the consumer as @eapache mentions will be a welcome improvement.

horkhe:
@aaronkavlie-wf beware of the issues that I listed above.

horkhe:
@aaronkavlie-wf I have been thinking about offset manager interface and actually agree with @eapache that it is not a higher level abstraction as I thought before but ruther a lower level building block that along with partition consumer can be used to build a high level balanced consumer.

But the way in our org we have implemented it a little bit differently to ensure that commits are never lost during shutdown and that it is easy to use with multiple consumer groups. I am not sure if it is suitable for the master but maybe later I will submit a PR for you to consider.

An interface that we come up with is a bit different:
```go
// OffsetManager uses Kafka to store consumed partition offsets.
type OffsetManager interface {
	// ManagePartition creates a PartitionOffsetManager on the given
	// group/topic/partition. It returns an error if this OffsetManager is
	// already managing the given group/topic/partition.
	ManagePartition(group, topic string, partition int32) (PartitionOffsetManager, error)

	// Close terminates all spawned PartitionOffsetManager's and waits for them
	// to commit pending offsets. So it is not necessary to call Close methods
	// of spawned PartitionOffsetManagers explicitly.
	Close()
}

// PartitionOffsetManager uses Kafka to fetch consumed partition offsets. You
// MUST call Close() on a partition offset manager to avoid leaks, it will not
// be garbage-collected automatically when it passes out of scope.
type PartitionOffsetManager interface {
	// InitialOffset returns a channel that an initial offset will be sent down
	// to, when retrieved.
	InitialOffset() <-chan FetchedOffset

	// CommitOffset triggers saving of the specified commit in Kafka. Actual
	// commits are performed periodically in a background goroutine. The commit
	// interval is configured by `Config.Consumer.Offsets.CommitInterval`.
	CommitOffset(offset int64, metadata string)

	// Errors returns a read channel of errors that occur during offset
	// management, if enabled. By default errors are not returned. If you want
	// to implement any custom error handling logic then you need to set
	// `Consumer.Return.Errors` to true, and read from this channel.
	Errors() <-chan *OffsetCommitError

	// Close stops the PartitionOffsetManager from managing offsets. It is
	// required to call this function before a PartitionOffsetManager object
	// passes out of scope, as it will otherwise leak memory.
	Close()
}

type FetchedOffset struct {
	Offset   int64
	Metadata string
}

type OffsetCommitError struct {
	Group     string
	Topic     string
	Partition int32
	Err       error
}
```
The implementation is [here](https://github.com/mailgun/sarama/blob/maxim/develop/offset_manager.go).

eapache:
Most recent push makes some of the API adjustments that have been suggested (group is now per-offset-manager, and offset/metadata are set/fetched together). It also fixes the missing `close(error)`.

I haven't figured out the best behaviour yet regarding "If a PartitionOffsetManager is closed a pending offset may not be committed.", since we probably don't want to spin forever trying to commit an offset before we give up, but there's also no harm in that as long as `Close` hasn't been called... do we have the user configure a retry-count that is only valid when shutting down? That seems a bit odd.

aaronkavlie-wf:
@eapache Did you amend to your previous commit? This PR is only showing a single commit, "wip".

aaronkavlie-wf:
I was actually starting on some tests; separate commits would make it easier to merge those in.

eapache:
> Did you amend to your previous commit?

Ya, when I'm working on something incomplete like this I'll typically always just amend a single WIP commit.

> I was actually starting on some tests; separate commits would make it easier to merge those in.

Sure, I can do that in future. I'll do one more amend to give it a useful commit message and then I'll put any other changes as separate commits on top.

eapache:
@horkhe what does your fork do in the event it can't commit offsets on shutdown? Spin forever trying again?

horkhe:
@eapache yes, it spins forever. The idea is that Kafka is highly available so it will get back to service eventually. And any service redeployment/restart will be delayed by the spin until this happens.

eapache:
Latest version here should spin forever; it feels weird, but I can't think of any other better behaviour.

At some point we'll probably want to permit configuring a timeout after which to give up, but that's really easy to do with the way things are set up now.

aaronkavlie-wf:
@eapache I'm looking at this again, and the purpose of the `metadata` argument added to `SetOffset()` is not clear to me.

eapache:
Per https://cwiki.apache.org/confluence/display/KAFKA/Offset+Management

> These fields should be mostly self explanatory, except for metadata. This is meant as a way to attach arbitrary metadata that should be committed with this offset commit. It could be the name of a file that contains state information for the processor, or a small piece of state. Basically it is just a generic string field that will be passed back to the client when the offset is fetched. It will likely have a tight size limit to avoid server impact.

eapache:
Tangentially: I just force-pushed a rebase of this branch in order to resolve a small merge conflict and to pull in @horkhe's test infrastructure.

aaronkavlie-wf:
I am making some progress on tests for this.

At least, for basic functionality such as `Offset()`, `SetOffset()`.

Testing `OffsetCommitResponse` errors is tricky; they trigger chain of logic & server interactions and it deadlocks if you don't have everything just so. I haven't sorted that out yet.

eapache:
> it deadlocks if you don't have everything just so

It's entirely possible that your tests will find actual bugs; I can't say much about the current code right now beyond the fact that it compiles. In other words: don't waste too much time trying to figure out why a test is hanging if it looks like it ought to work :)

aaronkavlie-wf:
Discovered a bug

The broker offset manager `mainLoop` is initialized, and starts committing offsets on a timed interval independent of fetching offsets. If the `fetchInitialOffset` call returns `ErrOffsetsLoadInProgress`, it will start committing offset 0.

Here's the test run, with copious logging statements sprinkled throughout `offset_manager.go`
(run with `config.Consumer.Offsets.CommitInterval = 1 * time.Millisecond`):

```
=== RUN TestFetchInitialInProgress
2015/08/13 12:44:28 selectBroker
2015/08/13 12:44:28 refreshing...
2015/08/13 12:44:28 coordinator
2015/08/13 12:44:28 FETCHINITIAL
2015/08/13 12:44:28 flushing...
2015/08/13 12:44:28 About to commit offset 5 on partition 0
2015/08/13 12:44:28 ErrOffsetLoadInProgress
2015/08/13 12:44:28 Sleeping...
2015/08/13 12:44:28 flushing...
2015/08/13 12:44:28 About to commit offset 0 on partition 0
2015/08/13 12:44:28 CommitOffset response:
2015/08/13 12:44:28 (*sarama.OffsetCommitResponse)(nil)
2015/08/13 12:44:28 Trying again...
2015/08/13 12:44:28 FETCHINITIAL
2015/08/13 12:44:28 err from broker.broker
2015/08/13 12:44:28 returning nil, err
--- FAIL: TestFetchInitialInProgress (0.25s)
	offset_manager_test.go:151: kafka: broker not connected
```

eapache:
Nice find; pushed a fix.

aaronkavlie-wf:
@eapache wow, that was an amazingly quick fix!

:+1: :beer:

eapache:
You explained it super-clearly and I happened to have the code open already while working on a different PR :)

aaronkavlie-wf:
I noticed that some of my tests weren't shutting down cleanly, so I made sure to call `.Close()` on everything. However, I still have a runaway `brokerOffsetManager` infinite loop.

Test case:

```go
// Test recovery from ErrNotCoordinatorForConsumer
// on first fetchInitialOffset call
func TestFetchInitialFail(t *testing.T) {
	om := initOM(t)

	// Error on first fetchInitialOffset call
	responseBlock := OffsetFetchResponseBlock{
		Err:      ErrNotCoordinatorForConsumer,
		Offset:   5,
		Metadata: "test_meta",
	}
	
	coordinator.SetHandler(func(req *request) (res encoder) {
		
	})

	fetchResponse := new(OffsetFetchResponse)
	fetchResponse.AddBlock("my_topic", 0, &responseBlock)
	coordinator.Returns(fetchResponse)

	// Refresh coordinator
	newCoordinator := newMockBroker(t, 3)
	refreshResponse := new(ConsumerMetadataResponse)
	refreshResponse.CoordinatorID = newCoordinator.BrokerID()
	refreshResponse.CoordinatorHost = "127.0.0.1"
	refreshResponse.CoordinatorPort = newCoordinator.Port()
	broker.Returns(refreshResponse)

	// Second fetchInitialOffset call is fine
	fetchResponse2 := new(OffsetFetchResponse)
	responseBlock2 := responseBlock
	responseBlock2.Err = ErrNoError
	fetchResponse2.AddBlock("my_topic", 0, &responseBlock2)
	newCoordinator.Returns(fetchResponse2)

	if _, err := om.ManagePartition("my_topic", 0); err != nil {
		t.Error(err)
	}

	broker.Close()
	coordinator.Close()
	newCoordinator.Close()
	testClient.Close()

	// Wait to see what the brokerOffsetManager does
	time.Sleep(500 * time.Millisecond)
}
```

logging output (again, all my random logging added to `offset_manager.go`):

```
=== RUN TestFetchInitialFail
2015/08/14 11:57:56 selectBroker
2015/08/14 11:57:56 refreshing...
2015/08/14 11:57:56 coordinator
2015/08/14 11:57:56 FETCHINITIAL
2015/08/14 11:57:56 ErrNotCoordinatorForConsumer
2015/08/14 11:57:56 selectBroker
2015/08/14 11:57:56 unref
2015/08/14 11:57:56 refreshing...
2015/08/14 11:57:56 coordinator
2015/08/14 11:57:56 FETCHINITIAL
2015/08/14 11:57:56 ErrNoError
2015/08/14 11:57:56 NOT returning nil, err
2015/08/14 11:57:56 1 subscriptions
2015/08/14 11:57:56 Offset commit request: offset 5 on partition 0
2015/08/14 11:57:56 CommitOffset response:
2015/08/14 11:57:56 (*sarama.OffsetCommitResponse)(nil)
2015/08/14 11:57:56 aborting
2015/08/14 11:57:56 rebalancing...
2015/08/14 11:57:56 selectBroker
2015/08/14 11:57:56 unref
2015/08/14 11:57:56 refreshing...
2015/08/14 11:57:56 kafka: tried to use a client that was closed
2015/08/14 11:57:56 No err
2015/08/14 11:57:56 rebalancing...
2015/08/14 11:57:56 selectBroker
2015/08/14 11:57:56 refreshing...
2015/08/14 11:57:56 kafka: tried to use a client that was closed
2015/08/14 11:57:56 No err
2015/08/14 11:57:56 rebalancing...
2015/08/14 11:57:56 selectBroker
2015/08/14 11:57:56 refreshing...
2015/08/14 11:57:56 kafka: tried to use a client that was closed
2015/08/14 11:57:56 No err
2015/08/14 11:57:56 rebalancing...
2015/08/14 11:57:56 selectBroker
2015/08/14 11:57:56 refreshing...
2015/08/14 11:57:57 kafka: tried to use a client that was closed
2015/08/14 11:57:57 No err
2015/08/14 11:57:57 rebalancing...
2015/08/14 11:57:57 selectBroker
2015/08/14 11:57:57 refreshing...
--- PASS: TestFetchInitialFail (0.71s)
```

@eapache

eapache:
Per @horkhe's request, the manager will not shut down until it successfully commits all outstanding offsets. Based on your logs that is not the case.

aaronkavlie-wf:
I suspected that may be the reason, but it does pose a problem for isolating tests.
It doesn't seem that it should try to commit if the pom is not dirty (indicating uncommitted offsets).

eapache:
I fixed a few minor issues around this area, although none of them critical. I also found the missing return call after `bom.abort()` which, given you'd put that in your PR as well, prompted me to cherry-pick your other fixes to this branch immediately (with proper attribution) just to avoid me having to debug problems you've already solved :)

After more thought, this is not an infinite loop in the `brokerOffsetManager`, it is an infinite loop in the `partitionOffsetManager` because you call `ManagePartition()` but never `Close` the result. Closing the underlying client will *not* close any individual PartitionOffsetManagers.

wvanbergen:
I am using this to implement offset management for my kafka consumer: https://github.com/wvanbergen/kafka/pull/72

One issue I ran into: it commits the last processed offset into Kafka. When resuming a consumer, it will return the exact same value as initial offset. If you feed that to sarama, you end up consuming the message with that offset twice.

This is a simple off by one error. The JVM implementation that uses Zookeeper simply stored the last committed offset + 1 into Zookeeper, so when you read it you can just feed the obtained value directly to `ConsumePartition`.

Should the offset manager take care of this off by one error, or is that the responsibility of the user?

wvanbergen:
Another idea: when you have never committed an offset before, `Offset()` will return -1. We could add a setting `Consumer.Offsets.Initial` that can be set to `OffsetOldest` or `OffsetNewest`. the offsetmanager would then return this value whenever the call to the server returns a negative value.

eapache:
> Should the offset manager take care of this off by one error, or is that the responsibility of the user?

What does the JVM consumer do when you tell it to use kafka? Does it send latest+1 or does it just send latest? Either way, the question is if offset manager should handle it or the high-level consumer; the user shouldn't care.

eapache:
> We could add a setting Consumer.Offsets.Initial that can be set to OffsetOldest or OffsetNewest.

Hmm, not sure how best to handle this case; an alternative would be a parameter to `ManagePartition` like the one we take in `ConsumePartition`.

aaronkavlie-wf:
@eapache 

> After more thought, this is not an infinite loop in the brokerOffsetManager, it is an infinite loop in the partitionOffsetManager because you call ManagePartition() but never Close the result. Closing the underlying client will not close any individual PartitionOffsetManagers.

That makes sense, thanks. Pushed up a commit closing the partition offset managers.

wvanbergen:
Final suggestions: add `Close() error` to the `OffsetManager` interface, and require it to be called during shutdown. This may be a noop for this implementation, but other implementations may need to free up shared resources that are not related to a `PartitionOffsetManager`. This is also more in line with basically any other type we expose in sarama.

aaronkavlie-wf:
Are there any remaining todos for this PR?

offset_manager.go
```diff
@@ -0,0 +1,485 @@
...
+
+	// SetOffset sets the offset and metadata according to the manager; this value (or a subsequent update)
+	// will eventually be flushed to the cluster based on configuration. You should only set the offset of
+	// messages that have been completely processed.
+	SetOffset(offset int64, metadata string)
+
+	// Errors returns a read channel of errors that occur during offset management, if enabled. By default,
+	// errors are logged and not returned over this channel. If you want to implement any custom error
+	// handling, set your config's Consumer.Return.Errors setting to true, and read from this channel.
+	Errors() <-chan *ConsumerError
```
wvanbergen: Should we have an `OffsetManagerError` instead? I think I am OK with `ConsumerError` given that offset management only concerns the consumer, but I wanted top bring it up anyway.

wvanbergen:
@aaronkavlie-wf Just one outstanding PR to finish up: #520, and the question about the error type, but after we resolve those we can merge this I think.

I am mostly concerned about getting the interface right; if we find any bugs we can fix them later, but fixing the interface is harder due toe API stability guarantees.

offset_manager.go
eapache: They'd have exactly the same contents, and your point about offset management only concerning the consumer also came to mind. We're on the same page here to leave it as is.

wvanbergen:
https://github.com/Shopify/sarama/pull/520 proposes some API changes, and fixes a bug.

wvanbergen:
I am happy with this now, but @eapache may want to take a final look.

eapache:
I will do a final walkthrough (mostly for my own benefit) this weekend some time. If I don't find anything, and nobody else raises any last-minute issues, we can merge this as-is on Monday.

eapache:
@horkhe you may be interested in the minor API changes we made:
- `Offset` (the getter) is now `NextOffset` and returns +1 the current offset so it is always the correct offset to start consuming at
- `SetOffset` (the setter) is now `MarkOffset` as that seems more precise

eapache:
A few more questions to consider (none of them block merging this though):
- Should `MarkOffset` validate that the offset being marked is > (or >=) the offset already stored?
- Since producing a message to kafka is much cheaper than writing a zookeeper node, should the commit interval default to a lower value (1 second maybe)?
- Do we need any other "flush triggers" than a simple timer (maybe based on number of uncommitted offsets or something)?
- Should we expose a `Dirty() bool` reader on the POM?

eapache:
OK, I am happy with this now. I would probably like to have a few more of the above questions answered before we put this into a versioned release (1.6?) but it can go into master now and we iterate from there.

wvanbergen:
> Should MarkOffset validate that the offset being marked is > (or >=) the offset already stored?

I see a valid use case of "resetting" a consumer by setting all the offsets to a negative value. However, that might be easier using low level primitives (`Coordinator()` and crafting a custom request). So I am OK with adding this.

> Since producing a message to kafka is much cheaper than writing a zookeeper node, should the commit interval default to a lower value (1 second maybe)?

:+1: to 1 sec interval by default.

> Do we need any other "flush triggers" than a simple timer (maybe based on number of uncommitted offsets or something)?

I have no problem if this gets added, but let's first see if anybody is actually interested in this.

> Should we expose a `Dirty() bool` reader on the POM?
> Would it be handy to expose a Flush() method on the POM which blocks until !dirty (just like the equivalent syscall)?

I don't think `Dirty() bool` is not too useful without a `Flush() error`, so it only makes sense if we add them together. I am not so sure if I can think of good use cases for this. The only time you really want to be sure the last offset is flushed is when you stop consuming from the partition. In that case you should always call `Close()`

Issue #466, opened 2015-06-08
async producer not to circuit beak on partitioner errors


ssahukar:
If Partitioner.Partition() function returns an error, the circuit breaker breaks the circuit and drops some of the subsequent messages. I think Sarama should only drop the message for which partitioner returned an error and send that error to the application through the error channel.

In our application with a custom partitioner, we want to drop the messages that get produced with an invalid key, in which case the expected behavior is to drop only that message but to continue processing the other messages as usual, without a circuit break.

wvanbergen:
The reason the circuit breaker was added is the case of messages being produced to a topic that does not exists. This will fail very slowly, due to the fact that the error (`UnknownTopicOrPartition`) can also legitimately be temporary, so it needs to be retried.

So the circuit breaker was added, but we didn't really think of other use cases. Maybe we can fix the above problem in a better way.

eapache:
@ssahukar this should be fixed in the latest master, if I understood you correctly

ssahukar:
Thanks Evan.

I am following the discussion and it seems it would fix the issue I am
seeing.

Appreciate the quick turn around on this.

helayzhang:
> The reason the circuit breaker was added is the case of messages being produced to a topic that does not exists. This will fail very slowly, due to the fact that the error (`UnknownTopicOrPartition`) can also legitimately be temporary, so it needs to be retried.
> 
> So the circuit breaker was added, but we didn't really think of other use cases. Maybe we can fix the above problem in a better way.

@wvanbergen @eapache 

I have met a problem recently. My kafka has disabled automatic topic creation, once a lot of illegal 
 messages try to write to a not exist topic, `circuit breaker` is soon opened, then other legal messages will be dropped also.

I'm confusing that in which case `UnknownTopicOrPartition` may be temporary, so we need `circuit breaker` to solve this? Hope you to help me understand this better, thx.

PR #467, opened 2015-06-09
Shrink the scope of the partition circuit-breaker

eapache:
It shouldn't wrap the entirety of `assignPartition`, since that also protects
things like the call to actually choose the partition. The only thing it is
supposed to protect is the call to get the list of partitions, since that is
what can block.

This may or may not fix (or be the right way to fix) #466.

@ssahukar @Shopify/kafka

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -271,10 +271,7 @@ func (p *asyncProducer) partitionDispatcher(topic string, input <-chan *Producer
 
 	for msg := range input {
 		if msg.retries == 0 {
-			err := breaker.Run(func() error {
-				return p.assignPartition(partitioner, msg)
-			})
-			if err != nil {
+			if err := p.assignPartition(breaker, partitioner, msg); err != nil {
 				p.returnError(msg, err)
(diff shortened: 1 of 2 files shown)


wvanbergen:
I think this should fix #466.

How hard would it be to construct a test for this?

eapache:
A test that partitioner errors are not part of the breaker, or a test that `Partition()` errors *are* part of the breaker? Neither is particularly tricky I don't think, but I'm also not sure if either are particularly useful.

wvanbergen:
I think a test for a custom partitioner would be good. We explicitly designed to allow this but it's not really tested. The fact that it should be not part of the circuit breaker could be part of it.

A test for having `Partition()` inside a breaker would be nice to have to but is less needed IMHO.

eapache:
test added; ended up touching a lot of lines because I generalized `expectSuccesses` to `expectResults`.

:eyes: when ready

wvanbergen:
LGTM :+1:

PR #468, opened 2015-06-09
Don't retry messages until the broker is closed

eapache:
Otherwise there is a case where the retried messages can have their
newly-selected broker closed out from under them if the remainder of this
goroutine gets heavily delayed by the scheduler.

I believe this may be the cause of the flaky failure in
TestAsyncProducerBrokerBounce (see e.g.
https://travis-ci.org/Shopify/sarama/jobs/66053366)

@Shopify/kafka


wvanbergen:
Nice find. :+1:

PR #470, opened 2015-06-09
Tweak where and how we wait for kafka to boot

eapache:
This should improve an occasional case where functional tests would fail
because a service wasn't ready yet.

cc @Shopify/kafka

### DIFF
--- a/vagrant/boot_cluster.sh
+++ b/vagrant/boot_cluster.sh
@@ -2,19 +2,21 @@
 
 set -ex
 
+# Launch and wait for toxiproxy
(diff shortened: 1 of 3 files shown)


wvanbergen:
Should we wait for all of them, instead of one? The checks are cheap if the service is actually running

eapache:
I thought about that, and was too lazy to see how cheap the checks actually are. Waiting for the last one *should* be sufficient, there isn't much variability in boot time of a fresh instance. No objections if you want to PR it though.

PR #478, opened 2015-06-24
Add decode method to request types

horkhe:
This change makes every request-response pair processed by a mock broker visible in test logs. It helped me to understand the flow of the tests, so I figured it may be beneficial for other users too. It might seems like a lot of changes, but basically I just added `decode` function to every request.

### DIFF
--- a/encoder_decoder.go
+++ b/encoder_decoder.go
@@ -9,15 +9,15 @@ type encoder interface {
 }
 
 // Encode takes an Encoder and turns it into bytes.
-func encode(in encoder) ([]byte, error) {
-	if in == nil {
+func encode(e encoder) ([]byte, error) {
+	if e == nil {
 		return nil, nil
 	}
 
 	var prepEnc prepEncoder
 	var realEnc realEncoder
 
-	err := in.encode(&prepEnc)
+	err := e.encode(&prepEnc)
 	if err != nil {
 		return nil, err
 	}
@@ -27,7 +27,7 @@ func encode(in encoder) ([]byte, error) {
 	}
 
 	realEnc.raw = make([]byte, prepEnc.length)
-	err = in.encode(&realEnc)
+	err = e.encode(&realEnc)
 	if err != nil {
 		return nil, err
 	}
--- a/request.go
+++ b/request.go
@@ -1,15 +1,22 @@
 package sarama
 
-type requestEncoder interface {
+import (
+	"encoding/binary"
+	"fmt"
+	"io"
+)
+
+type requestBody interface {
 	encoder
+	decoder
 	key() int16
 	version() int16
 }
 
 type request struct {
 	correlationID int32
-	id            string
-	body          requestEncoder
+	clientID      string
+	body          requestBody
 }
 
 func (r *request) encode(pe packetEncoder) (err error) {
@@ -17,7 +24,7 @@ func (r *request) encode(pe packetEncoder) (err error) {
 	pe.putInt16(r.body.key())
 	pe.putInt16(r.body.version())
 	pe.putInt32(r.correlationID)
-	err = pe.putString(r.id)
+	err = pe.putString(r.clientID)
 	if err != nil {
 		return err
 	}
@@ -26,4 +33,68 @@ func (r *request) encode(pe packetEncoder) (err error) {
 		return err
 	}
 	return pe.pop()
+}
+
+func (r *request) decode(pd packetDecoder) (err error) {
+	var key int16
+	if key, err = pd.getInt16(); err != nil {
+		return err
+	}
+	var version int16
+	if version, err = pd.getInt16(); err != nil {
+		return err
+	}
+	if r.correlationID, err = pd.getInt32(); err != nil {
+		return err
+	}
+	r.clientID, err = pd.getString()
+
+	r.body = allocateBody(key, version)
+	if r.body == nil {
+		return PacketDecodingError{fmt.Sprintf("Unknown request key: %d", key)}
+	}
+	return r.body.decode(pd)
+}
+
+func decodeRequest(r io.Reader) (req *request, err error) {
+	lengthBytes := make([]byte, 4)
+	if _, err := io.ReadFull(r, lengthBytes); err != nil {
+		return nil, err
+	}
+
+	length := int32(binary.BigEndian.Uint32(lengthBytes))
+	if length <= 4 || length > MaxRequestSize {
+		return nil, PacketDecodingError{fmt.Sprintf("Message of length %d too large or too small", length)}
+	}
+
+	encodedReq := make([]byte, length)
+	if _, err := io.ReadFull(r, encodedReq); err != nil {
+		return nil, err
+	}
+
+	req = &request{}
+	if err := decode(encodedReq, req); err != nil {
+		return nil, err
+	}
+	return req, nil
+}
+
+func allocateBody(key, version int16) requestBody {
+	switch key {
+	case 0:
+		return &ProduceRequest{}
+	case 1:
+		return &FetchRequest{}
+	case 2:
+		return &OffsetRequest{}
+	case 3:
+		return &MetadataRequest{}
+	case 8:
+		return &OffsetCommitRequest{Version: version}
+	case 9:
+		return &OffsetFetchRequest{}
+	case 10:
+		return &ConsumerMetadataRequest{}
+	}
+	return nil
 }
--- a/request_test.go
+++ b/request_test.go
@@ -2,17 +2,8 @@ package sarama
 
 import (
 	"bytes"
+	"reflect"
 	"testing"
-)
-
-var (
-	requestSimple = []byte{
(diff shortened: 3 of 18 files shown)


wvanbergen:
This is great. The mock broker is definitely limited, and it is often quite hard to figure out why things are not working, especially in concurrent scenarios. This will make that a lot better! :clap:

@eapache any comments?

request.go
```diff
@@ -1,11 +1,40 @@
...
+	"fmt"
+)
+
 type requestEncoder interface {
 	encoder
 	key() int16
 	version() int16
 }
 
+type requestHeader struct {
```
eapache: The duplication between `requestHeader` and `request` seems a little odd.

request.go
```diff
@@ -27,3 +56,40 @@ func (r *request) encode(pe packetEncoder) (err error) {
...
+		return &OffsetCommitRequest{Version: version}
+	case 9:
+		return &OffsetFetchRequest{}
+	case 10:
+		return &ConsumerMetadataRequest{}
+	}
+	return nil
+}
+
+type unknownKey error
```
eapache: Can this ever bubble up to the user? Should it be public and/or in `errors.go`?

request_test.go
```diff
@@ -2,6 +2,7 @@ package sarama
 
 import (
 	"bytes"
+	"reflect"
```
eapache: I'm not keen on pulling `reflect` in as a dependency.

eapache:
I am sitting in an airport half-asleep and waiting for a connecting flight, so take this with a grain of salt, but off the top of my head the principle looks good. I'm not convinced the implementation in `request.go` is ideal, I think there's probably a way to do it with less duplication and better conforming to the existing pattern, but I'm too tired to tell you what that would be, and I could just be wrong :)

request.go
horkhe: No it cannot. None of the changes that I made are used in production. This particular error is not needed. We can just do with `fmt.Errorf`. In fact it is never return even in tests. Because all requests implement decoder interface now. If a new request is introduced by a developer who forgets to provide `decode` function then this error will be returned, but only in tests.

request_test.go
horkhe: Hmm, what harm in that since it is a standard library. Besides it is only used in tests to perform deep comparisons?

request.go
horkhe: Well... yea. I did not want to touch the production code, hence some duplication.

horkhe:
@eapache I hope you have managed to catch some sleep :). Any comments now that you have a wide awake look at the changes?

request.go
eapache: I think I'd rather touch production code to get the better structure, our test suite will definitely catch any regressions that might be introduced.

request_test.go
eapache: My understanding was that importing `reflect` at any point required the go compiler to include a bunch more information in the final library/binary, but I'm not actually sure if that is true for testing as well

request_test.go
```diff
@@ -53,3 +54,11 @@ func testDecodable(t *testing.T, name string, out decoder, in []byte) {
 		t.Error("Decoding", name, "failed:", err)
 	}
 }
+
+func testDecoder(t *testing.T, name string, e encoder, d decoder, bytes []byte) {
```
eapache: this tests both encoding and decoding, so a name like `testEncodeDecode` seems more appropriate

eapache: if you're sticking with reflect, you can use reflection to create a new instance of the encoder type for testing decoding as well, saving the extra parameter

eapache:
Sorry for the delay, I'm back and awake now :)

I think I'd prefer to see this as two separate PRs to be honest: one that adds all the `decode` methods and their tests, and a second which does the mockbroker and requestHeader related changes. Hopefully it will be simpler to reason about that way.

request_test.go
horkhe: makes sense

request_test.go
horkhe: We have an HTTP proxy running in production that uses reflection for some things and did not notice any issues with performance whatsoever, so I see no reasons to worry about bringing reflection package into action even in production.

horkhe:
@eapache as you suggested I removed changes in mockbroker so this PR just bring capability to parse requests from encoded bytes.

encoder_decoder.go
```diff
@@ -8,16 +8,18 @@ type encoder interface {
 	encode(pe packetEncoder) error
 }
 
+type encodeFunc func(pe packetEncoder) error
+
 // Encode takes an Encoder and turns it into bytes.
-func encode(in encoder) ([]byte, error) {
-	if in == nil {
+func encode(ef encodeFunc) ([]byte, error) {
```
eapache: I don't understand this change at all, what was wrong with taking an interface here?

request.go
```diff
@@ -1,29 +1,90 @@
...
+	decoder
 	key() int16
 	version() int16
 }
 
-type request struct {
-	correlationID int32
-	id            string
-	body          requestEncoder
+func requestEncoder(req request, correlationID int32, clientID string) encodeFunc {
```
eapache: doesn't it make more sense for `requestHeader` to include a `request`, and then just implement the `encoder` interface directly?

encoder_decoder.go
horkhe: The function can be decorated with another function that is how I take `request.encoder` and decorated it with `encodeRequest` function to add header fields during encoding.

request.go
horkhe: At this point `requestHeader` is used for decoding only, for during encoding the majority of header fields either comes from encoded request or calculated (length). I can make it as you say (in that case change in the `encoder` interface would not be necessary) but then we probably should name this structure some what different, e.g. `requestWithHeader` or something along this lines.

eapache: ya, this is actually turning into a bigger refactor in my head, let me play with this a bit

eapache: does something like https://github.com/Shopify/sarama/compare/sample-request-body-refactor make sense to you?

horkhe: Yes, that what I was thinking about too, when I said that I can do it.

horkhe: I will update my PR with that

horkhe:
@eapache is this better?

request.go
```diff
@@ -1,15 +1,23 @@
...
 	encoder
+	decoder
 	key() int16
 	version() int16
 }
 
 type request struct {
+	length        int32
+	key           int16
+	version       int16
```
eapache: key and version do not need to be members of this struct, they are already implicit in the body

request.go
```diff
@@ -1,15 +1,23 @@
...
+
+type requestBody interface {
 	encoder
+	decoder
 	key() int16
 	version() int16
 }
 
 type request struct {
+	length        int32
```
eapache: length does not need to be a member of this struct; on encode it is implicit and on decode it has to be read in as a separate four-byte value anyways to know how much more to read

horkhe:
@eapache I completely forgot goal of this PR is to be able to parse requests from a stream of bytes. So I added `decodeRequest` function that does that and made sure that it is used in request encode/decode tests.

request.go
```diff
@@ -27,3 +33,67 @@ func (r *request) encode(pe packetEncoder) (err error) {
...
+}
+
+func decodeRequest(r io.Reader) (req *request, err error) {
+	lengthBytes := make([]byte, 4)
+	if _, err := io.ReadFull(r, lengthBytes); err != nil {
+		return nil, err
+	}
+
+	var length int32
+	if length, err = (&realDecoder{raw: lengthBytes}).getInt32(); err != nil {
```
eapache: For a single int32 it's almost certainly simpler to just do `int32(binary.BigEndian.Uint32(lengthBytes))`

request.go
```diff
@@ -27,3 +33,67 @@ func (r *request) encode(pe packetEncoder) (err error) {
...
+		return nil, err
+	}
+
+	encodedReq := make([]byte, length)
```
eapache: you'll want to add a validation the length is reasonable, something like https://github.com/Shopify/sarama/blob/master/response_header.go#L15-L17

request.go
```diff
@@ -27,3 +33,67 @@ func (r *request) encode(pe packetEncoder) (err error) {
...
+		return err
+	}
+	if r.correlationID, err = pd.getInt32(); err != nil {
+		return err
+	}
+	r.clientID, err = pd.getString()
+
+	r.body = allocateBody(key, version)
+	if r.body == nil {
+		return fmt.Errorf("Unknown request key: %d", key)
```
eapache: this should be a `PacketDecodingError`

eapache:
OK, this is very close now. Just a few minor simplifications and validations, and I'll merge it. Thanks for your patience!

horkhe:
@eapache done, let me know if that is ok, and then I will squash my changes to avoid clutter in the main repo.

eapache:
This looks great, squash away.

wvanbergen:
:tada:

Issue #300, opened 2015-02-24
Split producer goroutines into sub-structs


eapache:
When I wrote the producer, I kept each of the five pipeline goroutines as methods on the main `Producer` object for simplicity's sake, however several of them (in particular the `flusher` and `leaderDispatcher`) have now grown too large. The problem with this choice is that it makes it difficult to split logical chunks of that code out into helper functions, because the helpers cannot modify state in the parent function without some complicated parameter/return value passing.

My current thoughts around the fix for #294 are going to make this problem even worse, so before we tackle that we should do a small refactor (internal only, no API changes) of the producer to look more like the current consumer layout: small substructs which contain state variables, thus allowing the large goroutines to be broken up into helper functions and make the code flow much more obvious. Nobody likes reading 150-line functions.

Current thoughts include:
- ~~rename `brokerWorker` and use it~~ Create struct to manage the `messageAggregator` and `flusher` goroutines
- create a struct to manage the `leaderDispatcher` goroutines
- split some helpers out of `flusher` and `leaderDispatcher` to make their main loops more understandable

@drdee @yagnik this is a good project for learning more go and getting your hands dirty with the producer design, with very low risk since there should be no functional changes, just code reorg.

eapache:
At most recent count, this refactor would enable (or at least substantially simplify) #294 and #433.

eapache:
Between #510 #511 and #512 I am considering this done.

PR #481, opened 2015-07-07
Refactor producer goroutines somewhat

eapache:
- rename topicDispatcher to dispatcher
- make a topicProducer struct to handle the old partitionDispatcher
- make a partitionProducer struct to handle the old leaderDispatcher
- move and rename methods for a much more readable structure

Part 1 of #300.

There shouldn't be any functional changes. Thoughts @Shopify/kafka ?

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -96,7 +96,7 @@ func NewAsyncProducerFromClient(client Client) (AsyncProducer, error) {
 	}
 
 	// launch our singleton dispatchers
-	go withRecover(p.topicDispatcher)
+	go withRecover(p.dispatcher)
 	go withRecover(p.retryHandler)
 
 	return p, nil
@@ -200,16 +200,9 @@ func (p *asyncProducer) AsyncClose() {
 	go withRecover(p.shutdown)
 }
 
-///////////////////////////////////////////
-// In normal processing, a message flows through the following functions from top to bottom,
-// starting at topicDispatcher (which reads from Producer.input) and ending in flusher
-// (which sends the message to the broker). In cases where a message must be retried, it goes
-// through retryHandler before being returned to the top of the flow.
-///////////////////////////////////////////
-
 // singleton
 // dispatches messages by topic
-func (p *asyncProducer) topicDispatcher() {
+func (p *asyncProducer) dispatcher() {
 	handlers := make(map[string]chan *ProducerMessage)
 	shuttingDown := false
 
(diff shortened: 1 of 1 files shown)


async_producer.go
```diff
@@ -200,16 +200,9 @@ func (p *asyncProducer) AsyncClose() {
 	go withRecover(p.shutdown)
 }
 
-///////////////////////////////////////////
-// In normal processing, a message flows through the following functions from top to bottom,
-// starting at topicDispatcher (which reads from Producer.input) and ending in flusher
-// (which sends the message to the broker). In cases where a message must be retried, it goes
-// through retryHandler before being returned to the top of the flow.
-///////////////////////////////////////////
-
```
wvanbergen: We should probably replace this with something else.

eapache: Ya. Once this is merged I'll try and find time for "part 2" which will be a similar process for the aggregator and flusher. At that point this comment can come back, excepti "flows through the following structs, via each `dispatch()` method".

async_producer.go
```diff
@@ -264,133 +255,158 @@ func (p *asyncProducer) topicDispatcher() {
...
+	} else if choice < 0 || choice >= numPartitions {
+		return ErrInvalidPartition
 	}
 
+	msg.Partition = partitions[choice]
+
+	return nil
+}
+
+// one per partition per topic
```
wvanbergen: Maybe we should say there's only one goroutine accessing the internals, so there's no need for concurrency guards. The only way to communicate with this struct is to use the `input` and `output` channels; all the other stuff should be considered "private". I guess there's no real way to enforce this, except by using an `internal` subpackage.

eapache: ya, it's the same pattern we already used in the consumer... one goroutine per struct, communicate only via channels

not sure it needs explicit documentation in comments, the pattern is pretty obvious based on usage I think?

wvanbergen: Might be nice to document this somewhere as a general design principle. Doesn't have to be in the code, wiki page would be fine too,

wvanbergen: Actually, more precise would be: communicate between objects only using the constructor and method; never using struct members directly. (right?)

eapache: I've played with a couple of different constructor patterns: here the caller passes in an input channel. For brokerConsumer, it's created as part of constructor then accessed as a member by the caller. For partitionConsumer, the constructor is embedded directly in the calling method...

I'm still not sure which I like best, but the overriding directive is: the struct has an associated goroutine, and only that goroutine may access non-channel members directly

wvanbergen: I think I like the idea of creating a method for everything that should be accessible from outside the struct. e.g. by using a method that returns the internally created channel. That way struct members are always off limits for external objects.

eapache: in all the cases so far, all the methods are off-limits too, they're internal-only helpers

wvanbergen:
This looks fine to me.

Like we were discussing above, the main thing I am looking for is a way to communicate how these objects should be used. Whether that would be with docs, or somehow using a code style I don't know.

We can tackle this after step 2.

eapache:
It's going to be part of my talk at Golang UK if nothing else :)

Issue #315, opened 2015-03-04
Tests misuse `time.Sleep()`


eapache:
Very low-priority, and I knew full-well what I was doing at the time, but using `time.Sleep()` in concurrent tests is a really bad practice. We should get rid of it. If this entails actually thinking about the problem, so be it.

Specifically, `TestConsumerRebalancingMultiplePartitions` includes two instances of the following line:

``` go
time.Sleep(5 * time.Millisecond) // dumbest way to force a particular response ordering
```

:fire: them.

eapache:
Client and producer tests also do this in limited situations.

PR #482, opened 2015-07-15
Enhance mock broker

horkhe:
EDIT: This PR introduces `SetHandler` method to the mock broker that allows setting a function that can generate responses based on actual incoming requests. That allows writing tests that account for concurrent execution non-determinism. By default a handler function that returns responses read from the expectations channel populated by the `Returns` method is used. Besides the broker can accept many connections now.

Old PR description:
Let me explain why I have made these changes first: The thing is that the current consumer implementation can block if a message channel of one of consumed partitions is not being drained. This logic does not play well with our use case, we cannot allow one partition consumer affect others, so we have reimplemented the consumer (our [implementation](https://github.com/mailgun/sarama/blob/maxim/develop/consumer.go) in case you are interested). It would be great if we could reuse the same unit test suite to validate our consumer. Alas the mock broker turned out to be too limited in the sense that it was only capable of generating a particular sequence of responses regardless of actually incoming requests. But requests can come in an undefined order when you test concurrent code. To address this deficiency we introduced the following changes that are included in this PR:
* **mockbrocker_test.go**:
  * `SetHandler` function allows to setup a request handler function that can generate responses based on incoming requests;
  * `defaultRequestHandler` function implements the old behavior - reads responses from the expectation channel and returns them one by one. It is used by default;
  * `SetHandlerByMap` function is a syntactic sugar that along with mock response objects (see below) allows defining mockbroker logic in a concise way [e.g.](https://github.com/mailgun/sarama/blob/maxim/mockbroker/consumer_test.go#L68-L80)
* **mockresponses_test.go** provides a set of mock response helpers that can be used in the request handler function to generate responses based on received requests;
* **consumer_test.go** was refactored to take advantage of the mockbrocker request handler and the mock responses. Note that one of the tests is commented out, this is where the original consumer blocks and our implementation keeps working.

So if you guys are interested in the proposed changes or a part of them, then let's begin a code review, if not then just close this PR and we will live with out own clone.

### DIFF
--- a/async_producer_test.go
+++ b/async_producer_test.go
@@ -285,7 +285,10 @@ func TestAsyncProducerFailureRetry(t *testing.T) {
 	closeProducer(t, producer)
 }
 
+// If a Kafka broker becomes unavailable and then returns back in service, then
+// producer reconnects to it and continues sending messages.
 func TestAsyncProducerBrokerBounce(t *testing.T) {
+	// Given
 	seedBroker := newMockBroker(t, 1)
 	leader := newMockBroker(t, 2)
 	leaderAddr := leader.Addr()
@@ -295,30 +298,34 @@ func TestAsyncProducerBrokerBounce(t *testing.T) {
 	metadataResponse.AddTopicPartition("my_topic", 0, leader.BrokerID(), nil, nil, ErrNoError)
 	seedBroker.Returns(metadataResponse)
 
+	prodSuccess := new(ProduceResponse)
+	prodSuccess.AddTopicPartition("my_topic", 0, ErrNoError)
+
 	config := NewConfig()
-	config.Producer.Flush.Messages = 10
+	config.Producer.Flush.Messages = 1
 	config.Producer.Return.Successes = true
 	config.Producer.Retry.Backoff = 0
 	producer, err := NewAsyncProducer([]string{seedBroker.Addr()}, config)
 	if err != nil {
 		t.Fatal(err)
 	}
+	producer.Input() <- &ProducerMessage{Topic: "my_topic", Key: nil, Value: StringEncoder(TestMessage)}
+	leader.Returns(prodSuccess)
+	expectResults(t, producer, 1, 0)
 
-	for i := 0; i < 10; i++ {
-		producer.Input() <- &ProducerMessage{Topic: "my_topic", Key: nil, Value: StringEncoder(TestMessage)}
-	}
+	// When: a broker connection gets reset by a broker (network glitch, restart, you name it).
 	leader.Close()                               // producer should get EOF
 	leader = newMockBrokerAddr(t, 2, leaderAddr) // start it up again right away for giggles
 	seedBroker.Returns(metadataResponse)         // tell it to go to broker 2 again
 
-	prodSuccess := new(ProduceResponse)
-	prodSuccess.AddTopicPartition("my_topic", 0, ErrNoError)
+	// Then: a produced message goes through the new broker connection.
+	producer.Input() <- &ProducerMessage{Topic: "my_topic", Key: nil, Value: StringEncoder(TestMessage)}
 	leader.Returns(prodSuccess)
-	expectResults(t, producer, 10, 0)
-	seedBroker.Close()
-	leader.Close()
+	expectResults(t, producer, 1, 0)
 
 	closeProducer(t, producer)
+	seedBroker.Close()
+	leader.Close()
 }
 
 func TestAsyncProducerBrokerBounceWithStaleMetadata(t *testing.T) {
--- a/mockbroker_test.go
+++ b/mockbroker_test.go
@@ -1,37 +1,64 @@
 package sarama
 
 import (
+	"bytes"
 	"encoding/binary"
-	"errors"
+	"fmt"
 	"io"
 	"net"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
+
+	"github.com/davecgh/go-spew/spew"
 )
 
-// mockBroker is a mock Kafka broker. It consists of a TCP server on a kernel-selected localhost port that
-// accepts a single connection. It reads Kafka requests from that connection and returns each response
-// from the channel provided at creation-time (if a response has a len of 0, nothing is sent, if a response
-// the server sleeps for 250ms instead of reading a request).
+const (
+	expectationTimeout = 250 * time.Millisecond
+)
+
+type requestHandlerFunc func(req *request) (res encoder)
+
+// mockBroker is a mock Kafka broker. It consists of a TCP server on a
+// kernel-selected localhost port that can accept many connections. It reads
+// Kafka requests from that connection and passes them to the user specified
+// handler function (see SetHandler) that generates respective responses. If
+// the handler has not been explicitly specified then the broker returns
+// responses set by the Returns function in the exact order they were provided.
+// (if a response has a len of 0, nothing is sent, and the client request will
+// timeout in this case).
 //
-// When running tests with one of these, it is strongly recommended to specify a timeout to `go test` so that if the broker hangs
-// waiting for a response, the test panics.
+// When running tests with one of these, it is strongly recommended to specify
+// a timeout to `go test` so that if the broker hangs waiting for a response,
+// the test panics.
 //
-// It is not necessary to prefix message length or correlation ID to your response bytes, the server does that
-// automatically as a convenience.
+// It is not necessary to prefix message length or correlation ID to your
+// response bytes, the server does that automatically as a convenience.
 type mockBroker struct {
 	brokerID     int32
 	port         int32
-	stopper      chan bool
+	closing      chan none
+	stopper      chan none
 	expectations chan encoder
 	listener     net.Listener
 	t            *testing.T
 	latency      time.Duration
+	handler      requestHandlerFunc
+	handlerMux   sync.Mutex
 }
 
 func (b *mockBroker) SetLatency(latency time.Duration) {
 	b.latency = latency
+}
+
+// SetHandler sets the specified function as the request handler. Whenever
+// a mock broker reads a request from the wire it passes the request to the
+// function and sends back whatever the handler function returns.
+func (b *mockBroker) SetHandler(handler requestHandlerFunc) {
+	b.handlerMux.Lock()
+	b.handler = handler
+	b.handlerMux.Unlock()
 }
 
 func (b *mockBroker) BrokerID() int32 {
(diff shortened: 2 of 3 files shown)


eapache:
> The thing is that the current consumer implementation can block if a message channel of one of consumed partitions is not being drained. This logic does not play well with our use case, we cannot allow one partition consumer affect others, so we have reimplemented the consumer.

This is something I thought about when designing the current consumer. It wasn't important to us, so I didn't implement it, but I believe I did leave room for it in the current design. There is probably a ~20-line patch on the current consumer that will do what you want without rewriting the whole thing.

> Alas the mock broker turned out to be too limited in the sense that it was only capable of generating a particular sequence of responses regardless of actually incoming requests. But requests can come in an undefined order when you test concurrent code.

This has been a pain point for us a lot (see e.g. #315). I haven't looked at the code yet, but if that's the problem you're addressing, I already want to land this patch :)

> Note that one of the tests is commented out.

Use `t.Skip` instead so that it can be sure to compile at least.

eapache:
In summary: yes, we are definitely interested, but some of us are on vacation and the rest of us are busy with other projects at the moment, so the review process for this may lag a bit.

Something worth noting: review time is roughly quadratic on the size of the patch, so if you feel like submitting this as several smaller, logically independent changes then that would make our lives easier and would probably land it faster :)

horkhe:
@eapache cool. So let's do this one step at a time. I will make this PR devoted to the mock broker changes only.

horkhe:
@eapache I have updated this PR to just introduce mock broker changes as you suggested.

async_producer_test.go
```diff
@@ -315,10 +315,11 @@ func TestAsyncProducerBrokerBounce(t *testing.T) {
 	prodSuccess.AddTopicPartition("my_topic", 0, ErrNoError)
 	leader.Returns(prodSuccess)
 	expectResults(t, producer, 10, 0)
-	seedBroker.Close()
-	leader.Close()
 
+	time.Sleep(300 * time.Millisecond) // give the broker a chance to send a metadata response
```
eapache: hehe, looking forward to getting rid of these :)

mockbroker_test.go
```diff
@@ -2,38 +2,63 @@ package sarama
...
 	"io"
 	"net"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
+
+	"bytes"
+	"fmt"
+	"github.com/davecgh/go-spew/spew"
```
eapache: is there a particular reason for the import groupings here? typically the 'third-party' imports end up separate from the std lib

mockbroker_test.go
```diff
@@ -47,80 +72,134 @@ func (b *mockBroker) Addr() string {
...
-		}
-		body := make([]byte, binary.BigEndian.Uint32(reqHeader))
-		if len(body) < 10 {
-			return b.serverError(errors.New("Kafka request too short."), conn)
-		}
-		if _, err = io.ReadFull(conn, body); err != nil {
-			return b.serverError(err, conn)
+			Logger.Printf("*** mockbroker/%d/%d: invalid request: err=%v, %v", b.brokerID, idx, err, spew.Sdump(req))
+			b.serverError(err)
+			goto done
```
eapache: you don't need a label for these, you can just `break`

mockbroker_test.go
```diff
@@ -47,80 +72,134 @@ func (b *mockBroker) Addr() string {
...
+
+	go func() {
+		<-b.closing
+		safeClose(b.t, b.listener)
+	}()
+
+	wg := &sync.WaitGroup{}
+	for i := 0; ; i++ {
+		if conn, err = b.listener.Accept(); err != nil {
+			goto done
```
eapache: similarly here, you can just `break`

mockbroker_test.go
```diff
@@ -47,80 +72,134 @@ func (b *mockBroker) Addr() string {
...
+	Logger.Printf("*** mockbroker/%d/%d: connection closed, err=%v", b.BrokerID(), idx, err)
 }
 
-func (b *mockBroker) serverError(err error, conn net.Conn) bool {
-	b.t.Error(err)
-	if conn != nil {
-		if err := conn.Close(); err != nil {
-			b.t.Error(err)
+func (b *mockBroker) requestHandler() func(req *request) (res encoder) {
+	b.handlerLock.Lock()
```
eapache: what are the locks for? to make it safe to update the requestHandler while requests are already being handled?

I'm not sure I see a use case for that, 99% of the time you set the requestHandler at the beginning (before any requests come in) and then don't touch it the rest of the test

mockbroker_test.go
```diff
@@ -47,80 +72,134 @@ func (b *mockBroker) Addr() string {
...
+			goto done
 		}
 
 		if b.latency > 0 {
 			time.Sleep(b.latency)
 		}
 
-		response, err := encode(expectation)
+		res := b.requestHandler()(req)
+		if _, ok := req.body.(*FetchRequest); !ok {
```
eapache: what is specific to fetch requests here?

mockbroker_test.go
```diff
@@ -47,80 +72,134 @@ func (b *mockBroker) Addr() string {
...
+			Logger.Printf("*** mockbroker/%d/%d: served %+v -> %+v", b.brokerID, idx, req, res)
+		}
+		var encodedRes []byte
```
eapache: you can just use `:=` in the next line

mockbroker_test.go
```diff
@@ -47,80 +72,134 @@ func (b *mockBroker) Addr() string {
...
+
+func (b *mockBroker) defaultRequestHandler(req *request) (res encoder) {
+	select {
+	case res, ok := <-b.expectations:
+		if !ok {
+			return nil
 		}
+		return res
+	case <-time.After(expectationTimeout):
+		return nil
```
eapache: if you return nil here what actually happens? does the `encode` call end up panicing?

mockbroker_test.go
horkhe: no reason, just a mistake.

mockbroker_test.go
horkhe: I know, it is just imho labeled goto is easier to read. But sure I can change it to `break` if you prefer.

mockbroker_test.go
horkhe: Since handler is set in one goroutine and used in another, and there is no happens-before relationship between a set and a read actions, it is necessary to use some sort of synchronization to make the updated handler value visible in another goroutine.

mockbroker_test.go
horkhe: There is nothing specific, it is just a leftover from some debugging that I had been doing. Good catch.

mockbroker_test.go
horkhe: `encode` returns `[]byte(nil), error(nil)` if it is passed 'nil'. So mock broker sends no response in this case and the caller times out after `Config.Net.ReadTimeout`.

mockbroker_test.go
horkhe: Although I can use an atomic value here instead.

horkhe:
Urghhh, go 1.3 has no `atomic.Value` :(

mockbroker_test.go
eapache: I think I am comfortable requiring the user to set the request handler *before* making any requests, in which case there is no need for a code-enforced 'happens-before' relation... setting the handler while requests are ongoing is just plain user-error, since there is no use case for it I can think of?

horkhe: In fact there is a use case for it. E.g. check out this particular consumer [test](https://github.com/mailgun/sarama/blob/maxim/develop/consumer_test.go#L197-L271) rewritten using the updated broker. Without resetting the handler during the test the logic of the handler would be very complicated.

And by the way what is wrong with using proper synchronization? :)

eapache: > In fact there is a use case for it. E.g. check out this particular consumer test rewritten using the updated broker. Without resetting the handler during the test the logic of the handler would be very complicated.

That test is still racy though, even with the locks... you'd need some other mechanism to actually make it deterministic. I think it would be better if `SetHandlerByMap` took a `map[string][]MockResponse` instead to avoid the need for updating the handler mid-test?

> And by the way what is wrong with using proper synchronization? :)

Nothing, unless it's unnecessary :)

horkhe: I was not saying that this particular lock can resolve all racing conditions in tests, but it is necessary to make handler updates visible in the `MockBroker` goroutine.

I do not see how passing `map[string][]MockResponse` to `SetHandlerByMap` is going to help here. How would the mock broker decide when to stop using one `MockResponse` from the slice and start using another. Note that `MockResponse` is **not just an object** that should be returned, **but a builder** that constructs responses based on received requests. However in complex tests like this the response to be constructed depends not only on a received request but on the test stage too. So we can either write a complex `MockResponse` that is able to generate different responses depending on the test stage, or set different handlers depending on the test stage. I have chosen the later approach for it seemed to render more readable tests IMHO.

Note that there is still problem how to determine when one test stage completes and other starts. In most cases that can be done examining a PartitionConsumer Messages and Errors channels (although in this test it is mostly done via `time.Sleep`), but unfortunately some errors are not propagated to the Errors channel (e.g. ErrUnknownTopicOrPartition, ErrNotLeaderForPartition, ErrLeaderNotAvailable) and therefore there is no reliable way to determine the test stage boundary  and we are forced to use `time.Sleep`.

That said, it makes sense to leave the `SetHandlerByMap` signature and the respective lock as they are.

eapache: > Note that MockResponse is not just an object that should be returned, but a builder that constructs responses based on received requests.

Oh, missed that part.

> Note that there is still problem how to determine when one test stage completes and other starts.

Yup, that was my confusion. I thought this was the problem your PR was meant to solve.

> That said, it makes sense to leave the SetHandlerByMap signature and the respective lock as they are. 

OK

mockbroker_test.go
eapache: in fact, you can use the `Accept` call in the for loop itself and track the index separately, and then you don't need the additional flow control at all

eapache:
One minor comment left but this otherwise looks pretty good to me. I'd like @wvanbergen to take a look when he's back as well.

horkhe:
@eapache done. Please, let me know if I can squash the changes.

eapache:
squash :+1:

horkhe:
@eapache done

async_producer_test.go
```diff
@@ -295,30 +298,34 @@ func TestAsyncProducerBrokerBounce(t *testing.T) {
...
+	config.Producer.Flush.Messages = 1
 	config.Producer.Return.Successes = true
 	config.Producer.Retry.Backoff = 0
 	producer, err := NewAsyncProducer([]string{seedBroker.Addr()}, config)
 	if err != nil {
 		t.Fatal(err)
 	}
+	producer.Input() <- &ProducerMessage{Topic: "my_topic", Key: nil, Value: StringEncoder(TestMessage)}
+	leader.Returns(prodSuccess)
+	expectResults(t, producer, 1, 0)
```
wvanbergen: much better :+1:

mockbroker_test.go
```diff
@@ -1,39 +1,66 @@
...
-	"errors"
+	"fmt"
 	"io"
 	"net"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
+
+	"github.com/davecgh/go-spew/spew"
```
wvanbergen: I think this dependency is :+1: for testing.

wvanbergen:
I think this is a great improvement. There's definitely some more changes needed to the mock broker to make it more versatile, but this is a good start. Thanks for working on this :+1:

Issue #475, opened 2015-06-17
Potential race/deadlock between broker error and `AsyncClose`


eapache:
I was inspecting the consumer code as part of #474 and I realized that there is a potential deadlock condition in the very rare case where a broker returns an error (such as `ErrOffsetOutOfRange`) at the same time that the user calls `AsyncClose` (or just `Close`).

Both events result in a value being written to `child.dying` which has a buffer for only one event. If the `AsyncClose` write succeeds, then the `responseFeeder` write (triggered by the broker error) will block; since the goroutine responsible for reading from that channel is currently waiting on `responseFeeder` to unblock it, I believe this deadlocks.

cc @Shopify/kafka very unlikely (and entirely theoretical at this point) but I'd appreciate if somebody could verify my logic

wvanbergen:
Maybe things like this will be a lot easier to reason about using a `sync.Once` guard around the shutdown logic?

eapache:
?

the problem is that multiple things can legitimately write to `dying`, not that one of those writes is illegitimate

wvanbergen:
Why would we ever want to write more than once to `dying`? We can only shut down a consumer once.

wvanbergen:
Never mind; we're not just `close`ing the channel, but also writing errors to it.

PR #483, opened 2015-07-15
Fix hypothetical consumer deadlock

eapache:
In the case when the user calls `AsyncClose` at the same instant as the broker
leading the partition returns an error.

Fixes bug #475.

Solution is simpler/cleverer than I'd thought originally. Still 100% hypothetical though.

@Shopify/kafka


eapache:
Nevermind, doesn't actually 100% fix the issue.

PR #485, opened 2015-07-16
Dont require all consumers drained

eapache:
If a partitionConsumer fills up and is not being drained (or is taking a long time) remove its subscription until it can proceed again in order to not block other partitions which may still be making progress.

@Shopify/kafka @horkhe

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -1,6 +1,7 @@
 package sarama
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 	"sync/atomic"
@@ -278,6 +279,8 @@ type partitionConsumer struct {
 	highWaterMarkOffset int64
 }
 
+var errTimedOut = errors.New("timed out feeding messages to the user") // not user-facing
+
 func (child *partitionConsumer) sendError(err error) {
 	cErr := &ConsumerError{
 		Topic:     child.topic,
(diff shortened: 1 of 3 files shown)


consumer.go
```diff
@@ -401,10 +401,22 @@ func (child *partitionConsumer) HighWaterMarkOffset() int64 {
...
+			for i, msg := range msgs {
+				select {
+				case child.messages <- msg:
+				case <-time.After(child.conf.Consumer.MaxProcessingTime):
+					child.broker.done <- child
+					for _, msg = range msgs[i:] {
+						child.messages <- msg
+					}
+					child.broker.input <- child
+					continue feederLoop
```
wvanbergen: Can you explain this section a bit more? Can the second `child.messages <- msg` block as well, or am I understand this incorrectly?

eapache: This is happening in its own goroutine. If the timeout triggers, we send a message back to the brokerConsumer telling it to take us out of the pool, and then we feed the rest of the messages to the user at our leisure (so yes, the second message send probably blocks until the user catches up, but it doesn't block any *other* partitions).

Once we've flushed all our messages, we put ourselves back in the pool

wvanbergen:
I think it's doable to add a test for this: consume 2 partitions, drain only one of them.

horkhe:
@eapache @wvanbergen There is such a [test case](https://github.com/Shopify/sarama/pull/492/files#diff-e85d03976d9766c4d00ac176511d4f06R673). It is about to be introduced in #492. It is skipped there.

eapache:
OK, please re-review this PR as it should now also fix all the race conditions that the previous changes introduced.

If it looks good, I'll merge it and then you can rebase #492 without the skipped test.

consumer.go
```diff
@@ -271,13 +272,15 @@ type partitionConsumer struct {
...
 	trigger, dying chan none
-	dispatchReason error
+	responseResult error
 
 	fetchSize           int32
 	offset              int64
 	highWaterMarkOffset int64
 }
 
+var errTimedOut = errors.New("timed out feeding messages to the user") // not user-facing
```
wvanbergen: Capitalize?

consumer.go
wvanbergen: So, if the messages never get drained by the consumer (because their consumer goroutine shut down for some reason, we will end up with a stuck goroutine?

I think that's OK in that scenario, just want to understand the code.

wvanbergen:
I think this looks OK. @horkhe any final comments?

consumer.go
eapache: Yup, it will still shut down if/when `Close` is called though since that drains any outstanding messages.

consumer.go
eapache: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

wvanbergen: Hmmm, all the configuration errors (see config.go above) are capitalized.

eapache: good point, I've submitted a PR to fix them :)

PR #486, opened 2015-07-17
Update snappy repo path and encoding interface

kostyantyn:


ollieparsley:
Frustrating that the snappy library was changed without warning. Good call on the fix.

eapache:
This is the second time in a few months too...

PR #487, opened 2015-07-17
Fix snappy import and Encode return types

ollieparsley:
Seems like over night the snappy library merged in some breaking changes.

https://github.com/golang/snappy/commits/master


ollieparsley:
Fixed in #486 anyway

PR #492, opened 2015-07-23
Further mockbrocker improvements

horkhe:
This PR brings further improvements into the mockbrocker:

* A set of response builders were introduced to construct responses based on actual request data. Note that builders can respond to the most common request parameters. If something very specific is needed then a response can always be constructed directly;
* `mockbroker.SetHandlerByMap` method was introduced that provides a declarative way of defining a request handler that use mockresponse builders.
* **consumer_test.go** was rewritten using the new test primitives. Note that it is not a direct translation. Some tests were split for better isolation of concerns. Note also that `TestConsumerRebalancingMultiplePartitions` is using time delays for synchronization and therefore race prone. It could have been rewritten in almost non-racy way by removing the goroutine that reads from partition consumer channels and putting those reads among the lines of the main test flow. But I opted to leave it this way, because the goroutine code resembles how PartitionCosnumer is used in real life.


horkhe:
@eapache it builds now that I rebased it on the latest master.

Makefile
```diff
@@ -1,7 +1,7 @@
 default: fmt vet errcheck test
 
 test:
-	go test -v -timeout 60s -race ./...
+	go test -v -timeout 2m -race ./...
```
eapache: why does the test suddenly need so much more time?

consumer_test.go
```diff
@@ -1,38 +1,37 @@
...
-
-	offsetResponseOldest := new(OffsetResponse)
-	offsetResponseOldest.AddTopicPartition("my_topic", 0, 0)
-	leader.Returns(offsetResponseOldest)
+// If a particular offset is provided then messages are consumed starting from
+// that offset.
+func TestConsumerOffsetManual(t *testing.T) {
+	// Given
+	broker0 := newMockBroker(t, 0)
+	defer broker0.Close()
```
eapache: we haven't been using defers in tests because if one of them deadlocks (which is a common "found a bug!" indication) then it's really hard to track down

consumer_test.go
```diff
@@ -56,195 +53,491 @@ func TestConsumerOffsetManual(t *testing.T) {
 
 	safeClose(t, consumer)
 	safeClose(t, master)
-	leader.Close()
 }
 
+// If `OffsetNewest` is passed as the initial offset then the first consumed
+// message is indeed corresponds to the offset that broker claims to be the
+// newest in his metadata response.
```
eapache: brokers are not gendered, use "its"

consumer_test.go
```diff
@@ -56,195 +53,491 @@ func TestConsumerOffsetManual(t *testing.T) {
...
+	config.Consumer.Retry.Backoff = 500 * time.Millisecond
+	config.Metadata.Retry.Max = 0
+	c, err := NewConsumer([]string{broker0.Addr()}, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	pc, err := c.ConsumePartition("my_topic", 0, OffsetOldest)
+	if err != nil {
+		t.Errorf("Failed to create a partition consumer, err=%v", err)
```
eapache: this can be fatal, otherwise the next line is just going to panic with a nil pointer

consumer_test.go
```diff
@@ -56,195 +53,491 @@ func TestConsumerOffsetManual(t *testing.T) {
...
+	// Stage 3: finally the metadata returned by broker0 tells that broker1 is
+	// a new leader for my_topic/0. Consumption resumes.
 
-	offsetResponseNewest := new(OffsetResponse)
-	offsetResponseNewest.AddTopicPartition("my_topic", 0, 1234)
-	leader.Returns(offsetResponseNewest)
+	// Unfortunately consumer does not propagate `ErrNotLeaderForPartition`
+	// error to PartitionConsumer.Errors() channel. So there is no other way to
+	// synchronize here by sleep.
+	time.Sleep(300 * time.Millisecond)
```
eapache: the consumer *does* propagate the network timeouts to the errors channel, so you can synchronize on that right?

consumer_test.go
```diff
@@ -56,195 +53,491 @@ func TestConsumerOffsetManual(t *testing.T) {
...
+			SetLeader("my_topic", 0, broker0.BrokerID()),
+		"FetchRequest": newMockWrapper(fetchResponse2),
+	})
+
+	// When
+
+	// Unfortunately consumer does not propagate `ErrNotLeaderForPartition`
+	// error to PartitionConsumer.Errors() channel. So there is no other way to
+	// synchronize here by sleep.
+	time.Sleep(200 * time.Millisecond)
```
eapache: and here, you can sync on the network timeout error

eapache:
A few small points, but otherwise this looks really excellent, thank you!

consumer_test.go
horkhe: makes sense

consumer_test.go
horkhe: Sorry, in russian they are, I will correct that.

consumer_test.go
horkhe: you are right, we can use that as a synchronization event.

Makefile
horkhe: The consumer tests are rather fast, it is that other tests that use the older mock broker logic (the `Returns()` plus the default handler to be more specific) sometimes take longer to complete. The thing is that in the original mock broker implementation it would delay a response to a request **indefinitely** until a reply is provided via `Returns()` method. Now if a mock broker happens to get a request before `Returns()` is called it waits for `expectationTimeout` for `Returns()` to be called and then ignores the request making it IO timeout on the consumer side. Consumer retries eventually but that adds up to the execution time.

I will try to increase the `expectationTimeout` and bring the overall test suite timeout to the original value.

horkhe:
@eapache done

eapache:
:+1: 

@wvanbergen any comments?

wvanbergen:
This looks good to me! Merging...

PR #493, opened 2015-07-23
Add configuration to support broker SSL

uovobw:
as mentioned in https://github.com/Shopify/sarama/issues/154 this has been rebased on current master and can be merged

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -1,6 +1,7 @@
 package sarama
 
 import (
+	"crypto/tls"
 	"fmt"
 	"io"
 	"net"
@@ -68,7 +69,15 @@ func (b *Broker) Open(conf *Config) error {
 	go withRecover(func() {
 		defer b.lock.Unlock()
 
-		b.conn, b.connErr = net.DialTimeout("tcp", b.addr, conf.Net.DialTimeout)
+		dialer := &net.Dialer{
+			Timeout: conf.Net.DialTimeout,
+		}
+
+		if conf.Net.TLS.Enable {
+			b.conn, b.connErr = tls.DialWithDialer(dialer, "tcp", b.addr, conf.Net.TLS.Config)
+		} else {
+			b.conn, b.connErr = dialer.Dial("tcp", b.addr)
+		}
 		if b.connErr != nil {
 			b.conn = nil
 			atomic.StoreInt32(&b.opened, 0)
--- a/broker.go
+++ b/broker.go
@@ -69,12 +69,13 @@ func (b *Broker) Open(conf *Config) error {
 	go withRecover(func() {
 		defer b.lock.Unlock()
 
-		dialer := &net.Dialer{
-			Timeout: conf.Net.DialTimeout,
+		dialer := net.Dialer{
+			Timeout:   conf.Net.DialTimeout,
+			KeepAlive: conf.Net.KeepAlive,
 		}
 
 		if conf.Net.TLS.Enable {
-			b.conn, b.connErr = tls.DialWithDialer(dialer, "tcp", b.addr, conf.Net.TLS.Config)
+			b.conn, b.connErr = tls.DialWithDialer(&dialer, "tcp", b.addr, conf.Net.TLS.Config)
 		} else {
 			b.conn, b.connErr = dialer.Dial("tcp", b.addr)
 		}
--- a/examples/http_server/http_server.go
+++ b/examples/http_server/http_server.go
@@ -20,9 +20,10 @@ var (
 	addr      = flag.String("addr", ":8080", "The address to bind to")
 	brokers   = flag.String("brokers", os.Getenv("KAFKA_PEERS"), "The Kafka brokers to connect to, as a comma separated list")
 	verbose   = flag.Bool("verbose", false, "Turn on Sarama logging")
-	cert_file = flag.String("certificate", "", "The optional certificate file for client authentication")
-	key_file  = flag.String("key", "", "The optional key file for client authentication")
-	ca_file   = flag.String("ca", "", "The optional certificate authority file for TLS client authentication")
+	certFile  = flag.String("certificate", "", "The optional certificate file for client authentication")
+	keyFile   = flag.String("key", "", "The optional key file for client authentication")
+	caFile    = flag.String("ca", "", "The optional certificate authority file for TLS client authentication")
+	verifySsl = flag.Bool("verify", false, "Optional verify ssl certificates chain")
 )
 
 func main() {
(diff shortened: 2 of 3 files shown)


broker.go
```diff
@@ -68,12 +69,15 @@ func (b *Broker) Open(conf *Config) error {
 	go withRecover(func() {
 		defer b.lock.Unlock()
 
-		dialer := net.Dialer{
-			Timeout:   conf.Net.DialTimeout,
-			KeepAlive: conf.Net.KeepAlive,
+		dialer := &net.Dialer{
```
eapache: this struct should not have changed at all

examples/http_server/http_server.go
```diff
@@ -14,14 +17,39 @@ import (
 )
 
 var (
-	addr    = flag.String("addr", ":8080", "The address to bind to")
-	brokers = flag.String("brokers", os.Getenv("KAFKA_PEERS"), "The Kafka brokers to connect to, as a comma separated list")
-	verbose = flag.Bool("verbose", false, "Turn on Sarama logging")
+	addr       = flag.String("addr", ":8080", "The address to bind to")
+	brokers    = flag.String("brokers", os.Getenv("KAFKA_PEERS"), "The Kafka brokers to connect to, as a comma separated list")
+	verbose    = flag.Bool("verbose", false, "Turn on Sarama logging")
+	cert_file  = flag.String("certificate", "", "The certificate file for client authentication")
```
eapache: could you put a reference to TLS (and the fact that it's optional) in the help please?

examples/http_server/http_server.go
```diff
@@ -14,14 +17,39 @@ import (
...
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
+
+		tls_config = &tls.Config{
+			Certificates: []tls.Certificate{cert},
+			//		RootCAs:            caCertPool,
```
eapache: should this be commented out?

examples/http_server/http_server.go
```diff
@@ -14,14 +17,39 @@ import (
...
-	addr    = flag.String("addr", ":8080", "The address to bind to")
-	brokers = flag.String("brokers", os.Getenv("KAFKA_PEERS"), "The Kafka brokers to connect to, as a comma separated list")
-	verbose = flag.Bool("verbose", false, "Turn on Sarama logging")
+	addr       = flag.String("addr", ":8080", "The address to bind to")
+	brokers    = flag.String("brokers", os.Getenv("KAFKA_PEERS"), "The Kafka brokers to connect to, as a comma separated list")
+	verbose    = flag.Bool("verbose", false, "Turn on Sarama logging")
+	cert_file  = flag.String("certificate", "", "The certificate file for client authentication")
+	key_file   = flag.String("key", "", "The key file for client authentication")
+	ca_file    = flag.String("ca", "", "The certificate authority file for client authentication")
+	tls_config *tls.Config
```
eapache: this doesn't need to be global

In fact if you put the tls flag checking *after* the main config has been created, you don't need a variable at all you can just assign straight to the config value

eapache: oh sorry those are in separate functions; still, breaking that code out into a function which gets called from both cases is still probably cleaner

wvanbergen:
We don't really have the guarantee that SSL will work with this once it lands in Kafka proper. I am OK with merging this, with the caveat that once proper SSL support lands in Kafka, and for some reason it doesn't work with this code, we will not maintain backwards compatibility for this code.

eapache:
> we will not maintain backwards compatibility for this code.

Good point, a comment to that effect on the config values is definitely needed.

uovobw:
i am busy with other pieces of the infra at the moment but i still intend to complete - or at least contribute to - this. i will address the various notes but please note that i did only rebase the old branch for the previous pull request to current master and opened the review .

eapache:
Most of the comments are re. changes to the example program, not the original rebased code. To keep it simple I've updated #154 appropriately and will merge that if it looks good. You are welcome to resubmit the example changes if you wish.

broker.go
uovobw: not sure why it did in the first place, changed it back to have the keepalive and use the reference to it in the tls.DialWithDialer call

examples/http_server/http_server.go
uovobw: changed to "The optional certificate authority file for TLS client authentication"

examples/http_server/http_server.go
uovobw: not sure why it slipped in, uncommented

examples/http_server/http_server.go
uovobw: created a createTlsConfiguration() function that evaluates the command line flags and returns a config if they are set and contain valid data

uovobw:
merged from master and addressed your comments, thanks again

eapache:
Looks mostly OK to me. @wvanbergen? You wrote the http server example.

We'll have to re-add a comment re. compatibility guarantees but I think that can be done post-merge? Or would it be better to merge #156 and then rebase this to pull the http server changes in separately?

examples/http_server/http_server.go
```diff
@@ -164,6 +195,11 @@ func newDataCollector(brokerList []string) sarama.SyncProducer {
 	config := sarama.NewConfig()
 	config.Producer.RequiredAcks = sarama.WaitForAll // Wait for all in-sync replicas to ack the message
 	config.Producer.Retry.Max = 10                   // Retry up to 10 times to produce the message
+	tls_config := createTlsConfiguration()
```
wvanbergen: `tlsConfig`

examples/http_server/http_server.go
```diff
@@ -183,6 +219,11 @@ func newAccessLogProducer(brokerList []string) sarama.AsyncProducer {
 	// For the access log, we are looking for AP semantics, with high throughput.
 	// By creating batches of compressed messages, we reduce network I/O at a cost of more latency.
 	config := sarama.NewConfig()
+	tls_config := createTlsConfiguration()
```
wvanbergen: Naming nitpick: `tlsConfig`

examples/http_server/http_server.go
```diff
@@ -47,6 +53,31 @@ func main() {
...
+			log.Fatal(err)
+		}
+
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
+
+		t = &tls.Config{
+			Certificates:       []tls.Certificate{cert},
+			RootCAs:            caCertPool,
+			InsecureSkipVerify: true,
```
wvanbergen: Not sure if this is great as example code.

wvanbergen:
Couple of comments, mostly looks good :+1:

examples/http_server/http_server.go
uovobw: i've been testing with self signed certs and i hardcoded it and forgot, sorry will fix

uovobw:
@wvanbergen  addressed your comments

uovobw:
that is strange, we have [failing tests](https://travis-ci.org/Shopify/sarama/builds/74301680) but go test on my host succeeds

uovobw:
yes i can confirm that the command

> go test -v -timeout 60s -race ./...

run on my host (as is from the travis CI test) completes successfully on my host

wvanbergen:
Seems to have been a temporary fluke - I just did a rebuild and it passed now.

@eapache merge?

PR #494, opened 2015-07-27
consumer: fix another race pointed out by Maxim

eapache:
Take the previous refactor to its logical conclusion by handling *all* the
error logic in the brokerConsumer, not the responseFeeder. This fixes the race
to close the dying channel (since the brokerConsumer can just close the trigger
instead as it has ownership).

At the same time, refactor `updateSubscriptionCache` into `handleResponses`, and
inline the "new subscriptions" bit into the main loop; otherwise we end up
processing the previous iterations results at the very beginning of the next
iteration, rather than at the very end of the current one.

@horkhe @Shopify/kafka should fix the issue reported https://github.com/Shopify/sarama/pull/484#issuecomment-125328327

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -271,7 +271,7 @@ type partitionConsumer struct {
 	feeder   chan *FetchResponse
 
 	trigger, dying chan none
-	dispatchReason error
+	responseResult error
 
 	fetchSize           int32
 	offset              int64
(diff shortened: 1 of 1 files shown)


eapache:
This temporarily breaks a test which #485 re-fixes, so I've merged this fix in there and we can land them together.

PR #495, opened 2015-07-29
Cleanup error formatting

eapache:
@Shopify/kafka 

As pointed out by Willem (https://github.com/Shopify/sarama/pull/485#discussion_r35786611) we have a bunch of not-quite-idiomatic error formats. Fix up a bunch of obvious ones.


wvanbergen:
:ship:

PR #496, opened 2015-07-29
Put a 60s limit on the test suite

eapache:
@Shopify/kafka 

Nothing should run that long, and when flaky tests hang this is a lot better
than Travis's default 10-minute timeout.


wvanbergen:
:cool:

PR #502, opened 2015-08-07
Add History() method to mockbroker

horkhe:
This PR adds History() method to mockbroker. The method returns RequestResponse pairs in the order it has processed them. The method is supposed to facilitate testing of requests generated by the code. E.g. for offset manager testing. The workflow should be as follows:
* mock responses
* execute test logic
* stop mock broker
* inspect the recorded history

Note that lock on history is needed because a mockbroker can have several connections, each of them running in a goroutine of its own updating the history.

### DIFF
--- a/mockbroker_test.go
+++ b/mockbroker_test.go
@@ -46,7 +46,13 @@ type mockBroker struct {
 	t            *testing.T
 	latency      time.Duration
 	handler      requestHandlerFunc
-	handlerMux   sync.Mutex
+	history      []RequestResponse
+	lock         sync.Mutex
+}
+
+type RequestResponse struct {
+	Request  requestBody
+	Response encoder
 }
 
 func (b *mockBroker) SetLatency(latency time.Duration) {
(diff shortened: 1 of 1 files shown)


eapache:
> Note that lock on history is needed because a mockbroker can have several connections, each of them running in a goroutine of its own updating the history.

Then would it make more sense to have a history array per connection instead?

mockbroker_test.go
```diff
@@ -77,6 +84,10 @@ func (b *mockBroker) BrokerID() int32 {
 	return b.brokerID
 }
 
+func (b *mockBroker) History() []RequestResponse {
+	return b.history
```
eapache: if further request/responses can update the history slice, should this not be returning a copy (and probably taking the lock as well)?

eapache:
A few implementation questions, but definitely a good idea, thanks!

mockbroker_test.go
horkhe: The workflow that I expected and in fact use it in our offset manager implementation to inspect the history at the very end of the test when there are no more modifications. But it does make sense to make this method more universal and indeed return a copy. 

And you are right about the lock, for visibility guarantee if for nothing else.

horkhe:
@eapache yea, I thought about that too but decided not to do that. The reasons are:
* it may be hard for the test to figure out what particular connection related slice to look at;
* typical history use case I see are:
  * check if a particular request with particular parameters has been made;
  * check what parameters the most recent request of a particular type has had;
  * check if requests have been made in a correct order;
  
   for all these purposes one history buffer is quite sufficient, and certainly simpler.

wvanbergen:
I agree we should start with one history because it's simpler. If that turns out to not be enough, we can change it later: the mock broker is not part of the stable API promise.

horkhe:
@eapache @wvanbergen I made sure that the history is copied on return. Besides I decided that one lock would be more than enough in this case.

eapache:
:+1: 

@wvanbergen merge at will

horkhe:
Squashed changes to make them look nice in the trunk.

wvanbergen:
Belated :+1:

PR #503, opened 2015-08-07
Add mock responses for OffsetManager testing

horkhe:
This PR brings mock response builders and respective additions to response structs that are used by offset manager.

### DIFF
--- a/consumer_metadata_response.go
+++ b/consumer_metadata_response.go
@@ -20,10 +20,14 @@ func (r *ConsumerMetadataResponse) decode(pd packetDecoder) (err error) {
 	}
 	r.Err = KError(tmp)
 
-	r.Coordinator = new(Broker)
-	if err := r.Coordinator.decode(pd); err != nil {
+	coordinator := new(Broker)
+	if err := coordinator.decode(pd); err != nil {
 		return err
 	}
+	if coordinator.addr == ":0" {
+		return nil
+	}
+	r.Coordinator = coordinator
 
 	// this can all go away in 2.0, but we have to fill in deprecated fields to maintain
 	// backwards compatibility
@@ -43,16 +47,27 @@ func (r *ConsumerMetadataResponse) decode(pd packetDecoder) (err error) {
 }
 
 func (r *ConsumerMetadataResponse) encode(pe packetEncoder) error {
-
 	pe.putInt16(int16(r.Err))
-
+	if r.Coordinator != nil {
+		host, portstr, err := net.SplitHostPort(r.Coordinator.Addr())
+		if err != nil {
+			return err
+		}
+		port, err := strconv.ParseInt(portstr, 10, 32)
+		if err != nil {
+			return err
+		}
+		pe.putInt32(r.Coordinator.ID())
+		if err := pe.putString(host); err != nil {
+			return err
+		}
+		pe.putInt32(int32(port))
+		return nil
+	}
 	pe.putInt32(r.CoordinatorID)
-
 	if err := pe.putString(r.CoordinatorHost); err != nil {
 		return err
 	}
-
 	pe.putInt32(r.CoordinatorPort)
-
 	return nil
 }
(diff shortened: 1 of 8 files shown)


consumer_metadata_response.go
```diff
@@ -20,10 +20,14 @@ func (r *ConsumerMetadataResponse) decode(pd packetDecoder) (err error) {
 	}
 	r.Err = KError(tmp)
 
-	r.Coordinator = new(Broker)
-	if err := r.Coordinator.decode(pd); err != nil {
+	coordinator := new(Broker)
+	if err := coordinator.decode(pd); err != nil {
 		return err
 	}
+	if coordinator.addr == ":0" {
```
eapache: Why this change? Does Kafka return an address of ":0" in special cases?

mockresponses_test.go
```diff
@@ -230,3 +230,152 @@ func (mfr *mockFetchResponse) getHighWaterMark(topic string, partition int32) in
...
+	res := &OffsetFetchResponse{}
+	for topic, partitions := range mr.offsets[group] {
+		for partition, block := range partitions {
+			res.AddBlock(topic, partition, block)
+		}
+	}
+	return res
+}
+
+//func (mr *mockOffsetFetchResponse) getOffset(group, topic string, partition int32) *OffsetFetchResponseBlock {
```
eapache: why is this commented out?

offset_commit_response.go
```diff
@@ -4,9 +4,40 @@ type OffsetCommitResponse struct {
...
+			pe.putInt16(int16(kerror))
+		}
+	}
+	return nil
+}
+
 func (r *OffsetCommitResponse) decode(pd packetDecoder) (err error) {
 	numTopics, err := pd.getArrayLength()
-	if err != nil {
+	if err != nil || numTopics == 0 {
```
eapache: this extra conditional seems unnecessary, it would just fall through the loop without it?

eapache:
A few minor questions but generally looks good.

cc @aaronkavlie-wf you may want to coordinate with @horkhe since it looks like you're both interested in pushing the offset manager forward

offset_commit_response.go
horkhe: It is necessary to make encode/decode symmetric. When a reply that contains no topics encoded it used to contain an empty slice of blocks, but a response object that contains no topics had `nil` slice of blocks instead, so they did not compare equal.

mockresponses_test.go
horkhe: It is a remnant of the previous iteration of the implementation, I will remove that.

consumer_metadata_response.go
horkhe: It is necessary to make encode/decode symmetric. When a response that contains zero coordinator info is encoded it used to contain a zero instance of broker, but a response object that contains no broker had a `nil` broker reference.

aaronkavlie-wf:
Looks like this should take care of some of the hangups I ran into with testing, at least (missing decode/encode methods).

@horkhe have you done any work on tests for the offset manager PR?

horkhe:
@aaronkavlie-wf sorry, man, but I have written [our own](https://github.com/mailgun/sarama/blob/maxim/develop/offset_manager.go) offset manager, and it is tested. With this PR I am sharing some code from that effort with the upstream repo.

aaronkavlie-wf:
@horkhe hmm, why did you decide to write your own?

(Perhaps we should say "fork" as I see that it's clearly based on [the outstanding PR](https://github.com/Shopify/sarama/pull/461)

horkhe:
@aaronkavlie-wf when we wrote it the OffsetManager in the outstanding PR had different interface and a number of issues. You can find the list in one of my comments to the PR. Besides it our implementation the OffsetManager guarantees to commit pending offsets on close, whereas the OffsetManger in the outstanding PR drops them.

eapache:
code :+1: now

aaronkavlie-wf:
@horkhe right, but why not submit the changes as a PR to Sarama?

We have a need for offset management as well, but I'm looking to get it into Sarama to avoid the need to maintain a separate fork.

PR #504, opened 2015-08-09
simplify goroutine constructor semantics slightly

eapache:
@Shopify/kafka this is functionally no different but I think it's conceptually slightly simpler?


wvanbergen:
:+1:

PR #506, opened 2015-08-11
Improve async_producer tests and prep for more

eapache:
- implement a mockProduceResponse
- fix decoding of produceRequests (bless my foresight in making getSubset a
  method on the packetDecoder)
- collapse three of the simpler existing producer tests into one as a
  proof-of-concept

Heavily influenced by Maxim's recent work on the consumer tests.

@Shopify/kafka cc @horkhe

### DIFF
--- a/mockresponses_test.go
+++ b/mockresponses_test.go
@@ -324,6 +324,52 @@ func (mr *mockOffsetCommitResponse) getError(group, topic string, partition int3
 	return kerror
 }
 
+// mockProduceResponse is a `ProduceResponse` builder.
+type mockProduceResponse struct {
+	errors map[string]map[int32]KError
+	t      *testing.T
+}
+
+func newMockProduceResponse(t *testing.T) *mockProduceResponse {
+	return &mockProduceResponse{t: t}
+}
+
+func (mr *mockProduceResponse) SetError(topic string, partition int32, kerror KError) *mockProduceResponse {
+	if mr.errors == nil {
+		mr.errors = make(map[string]map[int32]KError)
+	}
+	partitions := mr.errors[topic]
+	if partitions == nil {
+		partitions = make(map[int32]KError)
+		mr.errors[topic] = partitions
+	}
+	partitions[partition] = kerror
+	return mr
+}
+
+func (mr *mockProduceResponse) For(reqBody decoder) encoder {
+	req := reqBody.(*ProduceRequest)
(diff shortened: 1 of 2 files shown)


produce_request.go
```diff
@@ -93,8 +93,12 @@ func (p *ProduceRequest) decode(pd packetDecoder) error {
 			if messageSetSize == 0 {
 				continue
 			}
+			msgSetDecoder, err := pd.getSubset(int(messageSetSize))
+			if err != nil {
+				return err
+			}
```
wvanbergen: What is fixed by this? How have we never been burned by this before?

mockresponses_test.go
```diff
@@ -324,6 +324,52 @@ func (mr *mockOffsetCommitResponse) getError(group, topic string, partition int3
 	return kerror
 }
 
+// mockProduceResponse is a `ProduceResponse` builder.
+type mockProduceResponse struct {
```
wvanbergen: I think I prefer calling this one (and the others) `produceResponseBuilder`

produce_request.go
eapache: This fixes decoding of produce requests with multiple message sets. This whole `decode` method was just added by @horkhe a few days ago, and is only used for tests so far, none of which caught this case until I wrote the one above.

wvanbergen: :ok_hand:

mockresponses_test.go
eapache: no preference, that rename can go in a separate PR

wvanbergen:
I have mixed feelings about collapsing the tests into one. It's way less repetitive, but there's usually different behavior around [0, 1, many] instances of the concepts involved, and now we're only testing the "many" scenario.

This is a very generic comment; I don't know if that actually makes any difference in this particular case.

eapache:
> but there's usually different behavior around [0, 1, many] instances of the concepts involved

For brokers and partitions and flushes, 0/1 are just degenerate cases of many, I don't think they really have their own behaviour.

mockresponses_test.go
wvanbergen: :+1:

wvanbergen:
:ship:

eapache:
Upon reflection I removed the test collapsing for now; better to do it right and make sure all cases are covered. Will ship the other components of this on :green_apple:.

PR #508, opened 2015-08-12
[WIP] Offset manager tests

aaronkavlie-wf:
This PR adds tests for the offset manager in PR #461.
The majority of the code now has coverage.

Notes
--------

- I was unable to reach the `bom.updateSubscriptions` loop in the `brokerOffsetManager.abort()` method; pointers for that would be appreciated.

- The tests use a pretty simple setup; adding a second partition offset manager (in an attempt to reach the code in the previous point) led to a a broker unreference/refresh and a need for more mock responses that I didn't get to.

- After spending a while with it, the code is still somewhat difficult to reason about, due to a lot of channels and resultant side effects. Also, 3 different objects whose responsibilities are not totally clear.

@eapache @wvanbergen @horkhe

### DIFF
--- a/offset_manager_test.go
+++ b/offset_manager_test.go
@@ -209,6 +209,7 @@ func TestSetOffset(t *testing.T) {
 func TestCommitErr(t *testing.T) {
 	pom := initPOM(t)
 
+	// Error on one partition
 	ocResponse := new(OffsetCommitResponse)
 	ocResponse.AddError("my_topic", 0, ErrOffsetOutOfRange)
 	ocResponse.AddError("my_topic", 1, ErrNoError)
@@ -223,9 +224,45 @@ func TestCommitErr(t *testing.T) {
 		CoordinatorPort: newCoordinator.Port(),
 	})
 
+	// Nothing in response.Errors at all
 	ocResponse2 := new(OffsetCommitResponse)
-	ocResponse2.AddError("my_topic", 0, ErrNoError)
 	newCoordinator.Returns(ocResponse2)
+
+	// For RefreshCoordinator()
+	broker.Returns(&ConsumerMetadataResponse{
+		CoordinatorID:   newCoordinator.BrokerID(),
+		CoordinatorHost: "127.0.0.1",
+		CoordinatorPort: newCoordinator.Port(),
+	})
+
+	// Error on the wrong partition for this pom
+	ocResponse3 := new(OffsetCommitResponse)
+	ocResponse3.AddError("my_topic", 1, ErrNoError)
+	newCoordinator.Returns(ocResponse3)
+
+	// For RefreshCoordinator()
+	broker.Returns(&ConsumerMetadataResponse{
+		CoordinatorID:   newCoordinator.BrokerID(),
+		CoordinatorHost: "127.0.0.1",
+		CoordinatorPort: newCoordinator.Port(),
+	})
+
+	// ErrUnknownTopicOrPartition/ErrNotLeaderForPartition/ErrLeaderNotAvailable block
+	ocResponse4 := new(OffsetCommitResponse)
+	ocResponse4.AddError("my_topic", 0, ErrUnknownTopicOrPartition)
+	newCoordinator.Returns(ocResponse4)
+
+	// For RefreshCoordinator()
+	broker.Returns(&ConsumerMetadataResponse{
+		CoordinatorID:   newCoordinator.BrokerID(),
+		CoordinatorHost: "127.0.0.1",
+		CoordinatorPort: newCoordinator.Port(),
+	})
+
+	// Normal error response
+	ocResponse5 := new(OffsetCommitResponse)
+	ocResponse5.AddError("my_topic", 0, ErrNoError)
+	newCoordinator.Returns(ocResponse5)
 
 	pom.SetOffset(100, "modified_meta")
 
--- a/offset_manager_test.go
+++ b/offset_manager_test.go
@@ -1,0 +1,132 @@
+package sarama
+
+import (
+	"testing"
+	"time"
+)
+
+var (
+	broker, coordinator *mockBroker
+	testClient          Client
+)
+
+func TestNewOffsetManager(t *testing.T) {
+	seedBroker := newMockBroker(t, 1)
+	seedBroker.Returns(new(MetadataResponse))
+
+	testClient, err := NewClient([]string{seedBroker.Addr()}, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = NewOffsetManagerFromClient("grouop", testClient)
+	if err != nil {
+		t.Error(err)
+	}
+
+	testClient.Close()
+
+	_, err = NewOffsetManagerFromClient("group", testClient)
+	if err != ErrClosedClient {
+		t.Errorf("Error expected for closed client; actual value: %v", err)
+	}
+
+	seedBroker.Close()
+}
+
+func initPOM(t *testing.T) PartitionOffsetManager {
+	config := NewConfig()
+	config.Metadata.Retry.Max = 0
+	config.Consumer.Offsets.CommitInterval = 1 * time.Millisecond
+
+	broker = newMockBroker(t, 1)
+	coordinator = newMockBroker(t, 2)
+
+	seedMeta := new(MetadataResponse)
+	seedMeta.AddBroker(coordinator.Addr(), coordinator.BrokerID())
+	// seedMeta.AddTopicPartition("my_topic", 0, 1, []int32{}, []int32{}, ErrNoError)
+	broker.Returns(seedMeta)
+
+	var err error
+	testClient, err = NewClient([]string{broker.Addr()}, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	broker.Returns(&ConsumerMetadataResponse{
+		CoordinatorID:   coordinator.BrokerID(),
+		CoordinatorHost: "127.0.0.1",
+		CoordinatorPort: coordinator.Port(),
+	})
+
+	fetchResponse := new(OffsetFetchResponse)
+	fetchResponse.AddBlock("my_topic", 0, &OffsetFetchResponseBlock{
+		Err:      ErrNoError,
+		Offset:   5,
+		Metadata: "test_meta",
+	})
+	coordinator.Returns(fetchResponse)
+
+	om, err := NewOffsetManagerFromClient("group", testClient)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	pom, err := om.ManagePartition("my_topic", 0)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return pom
+}
+
+func TestOffset(t *testing.T) {
+	pom := initPOM(t)
+	offset, meta := pom.Offset()
+	if offset != 5 {
+		t.Errorf("Expected offset 5. Actual: %v", offset)
+	}
+	if meta != "test_meta" {
+		t.Errorf("Expected metadata \"test_meta\". Actual: %q", meta)
+	}
+
+	pom.Close()
+	testClient.Close()
+	broker.Close()
+	coordinator.Close()
+}
+
+func TestSetOffset(t *testing.T) {
+	pom := initPOM(t)
+
+	ocResponse := new(OffsetCommitResponse)
+	ocResponse.AddError("my_topic", 0, ErrNoError)
+	coordinator.Returns(ocResponse)
+
+	pom.SetOffset(100, "modified_meta")
+	offset, meta := pom.Offset()
+
+	if offset != 100 {
+		t.Errorf("Expected offset 100. Actual: %v", offset)
+	}
+	if meta != "modified_meta" {
+		t.Errorf("Expected metadata \"modified_meta\". Actual: %q", meta)
+	}
+
+	pom.Close()
+	testClient.Close()
+	broker.Close()
+	coordinator.Close()
+}
+
+func TestCloseErrors(t *testing.T) {
+	pom := initPOM(t)
+	// pom.errors = "foo"
+	err := pom.Close()
+	if err != nil {
+		t.Error(err)
+	}
+
+	testClient.Close()
+	broker.Close()
+	coordinator.Close()
+}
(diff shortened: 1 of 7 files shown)


aaronkavlie-wf:
Latest commit includes a small bug fix in `offset_manager.go`

aaronkavlie-wf:
@eapache this is worth a look now; most of the code is covered. Feedback appreciated.

wvanbergen:
You forgot to handle some errors:

```
1.59s$ make errcheck
errcheck github.com/Shopify/sarama/...
github.com/Shopify/sarama/offset_manager_test.go:81:18	testClient.Close()
github.com/Shopify/sarama/offset_manager_test.go:130:11	pom.Close()
github.com/Shopify/sarama/offset_manager_test.go:131:18	testClient.Close()
github.com/Shopify/sarama/offset_manager_test.go:166:11	pom.Close()
github.com/Shopify/sarama/offset_manager_test.go:167:18	testClient.Close()
github.com/Shopify/sarama/offset_manager_test.go:180:11	pom.Close()
github.com/Shopify/sarama/offset_manager_test.go:183:18	testClient.Close()
github.com/Shopify/sarama/offset_manager_test.go:203:11	pom.Close()
github.com/Shopify/sarama/offset_manager_test.go:204:18	testClient.Close()
github.com/Shopify/sarama/offset_manager_test.go:277:18	testClient.Close()
github.com/Shopify/sarama/offset_manager_test.go:302:11	pom.Close()
github.com/Shopify/sarama/offset_manager_test.go:304:18	testClient.Close()
make: *** [errcheck] Error 1
```

We use errcheck to enforce that every error return value is handled (or at least explicitly ignored)

offset_manager_test.go
```diff
@@ -0,0 +1,305 @@
...
+	"time"
+)
+
+var (
+	broker, coordinator *mockBroker
+	testClient          Client
+	config              *Config
+)
+
+func initOM(t *testing.T) OffsetManager {
```
wvanbergen: I'd prefer to not use abbreviations for function names, so `initOffsetManager`

offset_manager_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+	om, err := NewOffsetManagerFromClient("group", testClient)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return om
+}
+
+func initPOM(t *testing.T) PartitionOffsetManager {
```
wvanbergen: `initPartitionOffsetManager`

offset_manager_test.go
```diff
@@ -0,0 +1,305 @@
...
+import (
+	"testing"
+	"time"
+)
+
+var (
+	broker, coordinator *mockBroker
+	testClient          Client
+	config              *Config
+)
```
wvanbergen: @eapache Do we usually have shared state like this for tests? Or do we always keep all state contained inside the test function? Not sure what to think about this.

eapache: I would prefer to avoid globals on principle. the init functions can just return the additional values

offset_manager_test.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	broker.Close()
+	coordinator.Close()
+	newCoordinator.Close()
+	pom.Close()
+	testClient.Close()
+
+	// Wait to see what the brokerOffsetManager does
+	time.Sleep(500 * time.Millisecond)
```
wvanbergen: What are we waiting on here?

offset_manager_test.go
```diff
@@ -0,0 +1,305 @@
...
+	responseBlock2 := responseBlock
+	responseBlock2.Err = ErrNoError
+	fetchResponse2.AddBlock("my_topic", 0, &responseBlock2)
+	newCoordinator.Returns(fetchResponse2)
+
+	pom, err := om.ManagePartition("my_topic", 0)
+	if err != nil {
+		t.Error(err)
+	}
+
```
wvanbergen: Maybe we should check whether `pom.Offset()` has the expected value here.

offset_manager_test.go
```diff
@@ -0,0 +1,305 @@
...
+	pom.SetOffset(100, "modified_meta")
+	offset, meta := pom.Offset()
+
+	if offset != 100 {
+		t.Errorf("Expected offset 100. Actual: %v", offset)
+	}
+	if meta != "modified_meta" {
+		t.Errorf("Expected metadata \"modified_meta\". Actual: %q", meta)
+	}
```
wvanbergen: Do we need to test whether we end up doing a CommitOffset requests at some point?

wvanbergen: Never mind; this is tested in the next test.

offset_manager_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+	broker.Close()
+	coordinator.Close()
+	newCoordinator.Close()
+	testClient.Close()
+}
+
+// Test of recovery from abort
+func TestBOMAbort(t *testing.T) {
```
wvanbergen: `TestAbortPartitionOffsetManager`

offset_manager_test.go
```diff
@@ -0,0 +1,305 @@
...
+		t.Error(err)
+	}
+
+	broker.Close()
+	coordinator.Close()
+	pom.Close()
+	testClient.Close()
+}
+
+func TestOffset(t *testing.T) {
```
wvanbergen: `TestPartitionOffsetManagerOffset`

offset_manager_test.go
```diff
@@ -0,0 +1,305 @@
...
+		t.Errorf("Expected metadata \"test_meta\". Actual: %q", meta)
+	}
+
+	pom.Close()
+	broker.Close()
+	coordinator.Close()
+	testClient.Close()
+}
+
+func TestSetOffset(t *testing.T) {
```
wvanbergen: `TestPartitionOffsetManagerSetOffset`

offset_manager_test.go
```diff
@@ -0,0 +1,305 @@
...
+	if err != ErrClosedClient {
+		t.Errorf("Error expected for closed client; actual value: %v", err)
+	}
+
+	seedBroker.Close()
+}
+
+// Test recovery from ErrNotCoordinatorForConsumer
+// on first fetchInitialOffset call
+func TestFetchInitialFail(t *testing.T) {
```
wvanbergen: `TestOffsetManagerOffsetInitialFetchFail`

wvanbergen:
A couple of questions and naming nitpicks, but this looks good. Thanks for helping on with this!

offset_manager_test.go
aaronkavlie-wf: Thanks for pointing this out, it was for debugging of an infinite loop as I hadn't closed the pom.

aaronkavlie-wf:
The latest batch of commits addresses all the naming comments and removes the `time.Sleep()` call.

offset_manager_test.go
aaronkavlie-wf: The only reason I can remember that I went this route was to simplify return values for those init functions. Shouldn't be too difficult to fix...

wvanbergen:
Green CI :green_heart:. I am happy once the globals are done.

offset_manager_test.go
```diff
@@ -0,0 +1,302 @@
...
+		t.Errorf("Expected metadata \"modified_meta\". Actual: %q", meta)
+	}
+
+	safeClose(t, pom)
+	safeClose(t, testClient)
+	broker.Close()
+	coordinator.Close()
+}
+
+func TestCommitErr(t *testing.T) {
```
wvanbergen: Forgot this one: `TestPartitionOffsetManagerCommitError`

offset_manager_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	broker.Close()
+	coordinator.Close()
+	newCoordinator.Close()
+	safeClose(t, pom)
+	safeClose(t, testClient)
+}
+
+// Test fetchInitialOffset retry on ErrOffsetsLoadInProgress
+func TestFetchInitialInProgress(t *testing.T) {
```
wvanbergen: `TestOffsetManagerFetchInitialLoadInProgress`

offset_manager_test.go
```diff
@@ -0,0 +1,302 @@
...
+// on first fetchInitialOffset call
+func TestOffsetManagerFetchInitialFail(t *testing.T) {
+	om := initOffsetManager(t)
+
+	// Error on first fetchInitialOffset call
+	responseBlock := OffsetFetchResponseBlock{
+		Err:      ErrNotCoordinatorForConsumer,
+		Offset:   5,
+		Metadata: "test_meta",
+	}
```
wvanbergen: Maybe we should verify that the Offset and Metadata values are ignored for an error response, i.e.  `Offset()` will not return 5 after this.

wvanbergen:
The reason I want the long test function names is that it allows you to do `go test -run OffsetManagaer` to run all tests related to the offset manager.

aaronkavlie-wf:
Pushed up another commit to get rid of globals.

4 return values isn't pretty; I guess the only way out of that, though, would be to define a struct or two.

wvanbergen:
You can always create a `testState` struct or something along those lines if you don't want to return 4 separate values.

aaronkavlie-wf:
> The reason I want the long test function names is that it allows you to do go test -run OffsetManager to run all tests related to the offset manager.

Yeah, I wanted to do that during testing. I wish you could target a specific file without worrying about test names.

offset_manager_test.go
aaronkavlie-wf: In the test code? Sarama retries automatically on error so it appears there's no way to call a `pom.Offset()` if you get an error.

wvanbergen: ah true - didn't think of that.

wvanbergen:
I think I am happy with this now :+1: @eapache final comments?

offset_manager_test.go
```diff
@@ -0,0 +1,300 @@
...
+	broker.Returns(seedMeta)
+
+	var err error
+	testClient, err = NewClient([]string{broker.Addr()}, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	broker.Returns(&ConsumerMetadataResponse{
+		CoordinatorID:   coordinator.BrokerID(),
```
eapache: you should just be able to set `Coordinator: coordinator` and it will pick up the ID/Host/Port automatically

offset_manager_test.go
```diff
@@ -0,0 +1,300 @@
...
+
+	om, err = NewOffsetManagerFromClient("group", testClient)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return om, testClient, broker, coordinator
+}
+
+func initPartitionOffsetManager(t *testing.T) (pom PartitionOffsetManager,
```
eapache: I'd rather this function take the offset manager as an argument (so callers would have to call `initOffsetManager` themselves first) so that all methods have access to the root manager. This will make it much easier if/when we add manager-level methods like a global `Close`

offset_manager_test.go
```diff
@@ -0,0 +1,300 @@
...
+	ocResponse := new(OffsetCommitResponse)
+	ocResponse.AddError("my_topic", 0, ErrOffsetOutOfRange)
+	ocResponse.AddError("my_topic", 1, ErrNoError)
+	coordinator.Returns(ocResponse)
+
+	newCoordinator := newMockBroker(t, 3)
+
+	// For RefreshCoordinator()
+	broker.Returns(&ConsumerMetadataResponse{
+		CoordinatorID:   newCoordinator.BrokerID(),
```
eapache: for all of these `ConsumerMetadataResponse`s you should just be able to set the `Coordinator` field and nothing else

eapache:
A few minor points, but generally :+1:. Please squash away all the intermediate commits, and rebase on the branch to make sure my recent changes haven't introduced any bugs :)

offset_manager_test.go
aaronkavlie-wf: Tried that, I get this:

```
cannot use coordinator (type *mockBroker) as type *Broker in field value
```

nor does type assertion (`coordinator.(*Broker)`) fix it.

aaronkavlie-wf:
@eapache 

> Please squash away all the intermediate commits

squash everything here to a single commit?

offset_manager_test.go
eapache: oh right, mockBroker != Broker... nevermind then

eapache:
> squash everything here to a single commit?

It doesn't have to be a single commit, but 21 is way too many. However many makes sense for each commit to be its own self-contained and consistent step.

aaronkavlie-wf:
I rebased and squashed some of my commits.

I mistakenly rebased on master, then rebased again on the offset-manager branch. Still includes master stuff (rebasing the offset-manager branch ought to clean that out).

It's down to 9 commits from this branch, is that reasonable?

There is a CI failure; I'm not able to reproduce locally.

wvanbergen:
CI seems to be a flaky test; I restarted it.

eapache:
I've rebased the offset-manager branch, you should be able to rebase again to get rid of all the extra commits.

aaronkavlie-wf:
Rebased; it still includes a bunch of your commits in this PR for some reason.

eapache:
Ya, something was weird. I ended up manually grabbing both branches locally, doing all the rebase/resolution here, and then pushing the resulting commits up to the existing `offset-manager` branch.

So this has, in effect, been merged, even though github is still confused. Thanks!

aaronkavlie-wf:
:+1: thanks!

myusuf3:
Any docs on how this is supposed to work! We are looking to not to have to deal with tracking offset anymore.

wvanbergen:
Godoc documentation: https://godoc.org/github.com/Shopify/sarama#OffsetManager

This is a building block that stores offsets in Kafka, but doesn't automatically deal with offsets for you. You still have to use this API yourself in your consumer implementation. Example: https://github.com/wvanbergen/kafka/pull/72

myusuf3:
Yeah do using the `OffsetManager` to get where we are at for each topic and feedback into consumer is what I am looking for! Are you guys working toward automatic handling of offset? in that branch you linked?

wvanbergen:
Yeah, the idea of that branch is that it deals with offsets for you, but it requires Zookeeper to manage state of the instances. Because of this it's not in sarama itself.

myusuf3:
Awesome! I am going to starting using new offset stuff today! Thanks guys!

PR #510, opened 2015-08-13
Structure the aggregator and flusher goroutines

eapache:
Another chunk of #300. Once this lands the final step will be to break out the
existing massive `run()` methods into useful helper functions now that
parameter-passing isn't such a pain.

In theory no functional changes; just slightly nicer code and stricter typing
around channel direction.

@Shopify/kafka


wvanbergen:
Really straightforward :+1:

PR #511, opened 2015-08-13
Clean up aggregator via helper methods

eapache:
Also get rid of forceFlushThreshold it does not need its own function

@wvanbergen

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -9,10 +9,6 @@ import (
 	"github.com/eapache/queue"
 )
 
-func forceFlushThreshold() int {
-	return int(MaxRequestSize - (10 * 1024)) // 10KiB is safety room for misc. overhead, we might want to calculate this more precisely?
-}
-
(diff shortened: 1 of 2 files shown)


async_producer.go
```diff
@@ -544,45 +534,77 @@ func (a *aggregator) run() {
...
+	if len(a.buffer) > 0 {
+		a.output <- a.buffer
 	}
 	close(a.output)
 }
 
+func (a *aggregator) wouldOverflow(msg *ProducerMessage) bool {
+	switch {
+	// Would we overflow our maximum possible size-on-the-wire? 10KiB is arbitrary overhead for safety.
+	case a.bufferBytes+msg.byteSize() >= int(MaxRequestSize-(10*1024)):
```
wvanbergen: Can we trust the compiler to optimize the static calculation into a constant?

eapache: We weren't before and nobody ever complained...

wvanbergen:
:+1:

PR #512, opened 2015-08-13
Clean up the flusher via helper methods

eapache:
@wvanbergen

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -504,9 +504,10 @@ func (p *asyncProducer) newBrokerProducer(broker *Broker) chan<- *ProducerMessag
 	go withRecover(a.run)
 
 	f := &flusher{
-		parent: p,
-		broker: broker,
-		input:  bridge,
+		parent:         p,
+		broker:         broker,
+		input:          bridge,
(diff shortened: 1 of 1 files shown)


async_producer.go
```diff
@@ -653,40 +632,77 @@ func (f *flusher) run() {
...
-					currentRetries[topic][partition] = block.Err
-					f.parent.retryMessages(msgs, block.Err)
-				default:
-					f.parent.returnErrors(msgs, block.Err)
+	for i, msg := range batch {
+
+		if f.currentRetries[msg.Topic] != nil && f.currentRetries[msg.Topic][msg.Partition] != nil {
+			// we're currently retrying this partition so we need to filter out this message
+			f.parent.retryMessages([]*ProducerMessage{msg}, f.currentRetries[msg.Topic][msg.Partition])
+			batch[i] = nil
```
eapache: actually some functional changes here; I re-orged a few of these lines so that we retry here *before* we do the chaser check; I don't think this really makes any difference

wvanbergen: :+1:

wvanbergen:
:cool: :+1: :ship:

PR #515, opened 2015-08-14
Make toxiproxy provisioning optional

horkhe:
Some projects use sarama's vagrant install to run functional tests, but sometimes tests that are supposed to be deterministic are being messed up by toxiproxy. This PR makes it possible to exclude toxiproxy from the setup. If `TOXIPROXY_VERSION` env variable is defined then it is provisioned, otherwise it is not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ go:
 env:
   global:
   - KAFKA_PEERS=localhost:9091,localhost:9092,localhost:9093,localhost:9094,localhost:9095
+  - TOXIPROXY_VERSION=1.0.3
   - TOXIPROXY_ADDR=http://localhost:8474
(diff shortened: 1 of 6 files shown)


eapache:
> but sometimes tests that are supposed to be deterministic are being messed up by toxiproxy

I don't immediately understand this - toxiproxy is a deterministic invisible proxy by default, so if it's causing problems in its default state then that sounds like a bug.

horkhe:
Sometimes I saw connections to zookeeper and/or kafka timeout when running tests on Travis CI, so I wanted to remove and extra connection leg from the picture. But turned out that still happens even without toxiproxy. Well if you do not want to have an option to disable the proxy then just close this PR.

eapache:
Since toxiproxy does nothing by default, I don't really want to carry any extra complexity around enabling/disabling it.

PR #518, opened 2015-08-19
Add Close() OffsetManager interface

wvanbergen:
This way, an OffsetManager that uses an external data store can properly clean up the resources it uses.

- Should we add `AsyncClose()` for consistency?
- I added a dummy implementation for the Kafka offset managers. Should be do something there now the function is present?

@eapache

### DIFF
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -12,6 +12,11 @@ type OffsetManager interface {
 	// ManagePartition creates a PartitionOffsetManager on the given topic/partition. It will
 	// return an error if this OffsetManager is already managing the given topic/partition.
 	ManagePartition(topic string, partition int32) (PartitionOffsetManager, error)
+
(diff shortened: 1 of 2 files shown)


eapache:
> Should we add AsyncClose() for consistency?

Does this make sense as a useful function for the ZK-based manager?

eapache:
> I added a dummy implementation for the Kafka offset managers. Should be do something there now the function is present?

Nope. The consumer is in a similar situation - all it does is (potentially) close the underlying client, but we don't even have that to deal with because the only constructor we provide is `OffsetManagerFromClient`.

eapache:
Add AsyncClose if you think it's needed, then LGTM.

wvanbergen:
I don't think we'll need AsyncClose, because there are no channels to wait for.

PR #519, opened 2015-08-19
Add go 1.5 to build matrix

wvanbergen:
@eapache


eapache:
:+1: we can add it to the README too, and in a few months we can remove 1.3.

PR #520, opened 2015-08-19
Offset manager: make initial offset configurable

wvanbergen:
This allows you to set the initial offset `Offset()` will return if no offset was committed yet for the partition. 

Right now, we depend on the implicit behaviour of the Kafka broker, which is returning -1. This patch makes the behaviour explicit, and allows you to change that to either `OffsetOldest` or OffsetNewest`.

I prefer using a config setting instead of an extra argument to ManagePartition, because this value will always be the same for every partition in a consumer.

@Shopify/kafka

### DIFF
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -117,27 +117,33 @@ func (om *offsetManager) abandonBroker(bom *brokerOffsetManager) {
 // on a partition offset manager to avoid leaks, it will not be garbage-collected automatically when it passes
 // out of scope.
 type PartitionOffsetManager interface {
-	// Offset returns the last offset that was marked as processed and associated metadata according to the manager;
-	// this value has not necessarily been flushed to the cluster yet. If you want to resume a partition consumer
-	// from where it left off, remember that you have to increment the offset by one so the partition consumer will
-	// start at the next message. This prevents the last committed message from being processed twice.
-	Offset() (int64, string)
+	// NextOffset returns the next offset that should be consumed for the managed partition, accompanied
+	// by metadata which can be used to reconstruct the state of the partition consumer when it resumes.
+	// NextOffset() will return `config.Consumer.Offsets.Initial` and an empty metadata string if no
+	// offset was committed for this partition yet.
+	NextOffset() (int64, string)
 
-	// SetOffset sets the offset and metadata according to the manager; this value (or a subsequent update)
-	// will eventually be flushed to the cluster based on configuration. You should only set the offset of
-	// messages that have been completely processed.
-	SetOffset(offset int64, metadata string)
+	// MarkOffset marks the provided offset as processed, alongside a metadata string that represents
+	// the state of the partition consumer at that point in time. The metadata string can be used by
+	// another consumer to restore that state, so it can resume consumption.
+	//
+	// Note: calling MarkOffset does not necessarily commit the offset to the backend store immediately
+	// for efficiency reasons, and it may never be committed if your application crashes. This means that
+	// you may end up processing the same message twice, and your processing should ideally be idempotent.
+	MarkOffset(offset int64, metadata string)
 
 	// Errors returns a read channel of errors that occur during offset management, if enabled. By default,
 	// errors are logged and not returned over this channel. If you want to implement any custom error
 	// handling, set your config's Consumer.Return.Errors setting to true, and read from this channel.
 	Errors() <-chan *ConsumerError
+
 	// AsyncClose initiates a shutdown of the PartitionOffsetManager. This method will return immediately,
 	// after which you should wait until the 'errors' channel has been drained and closed.
 	// It is required to call this function, or Close before a consumer object passes out of scope,
 	// as it will otherwise leak memory.  You must call this before calling Close on the underlying
 	// client.
 	AsyncClose()
+
 	// Close stops the PartitionOffsetManager from managing offsets. It is required to call this function
 	// (or AsyncClose) before a PartitionOffsetManager object passes out of scope, as it will otherwise
 	// leak memory. You must call this before calling Close on the underlying client.
@@ -290,7 +296,7 @@ func (pom *partitionOffsetManager) Errors() <-chan *ConsumerError {
 	return pom.errors
 }
 
-func (pom *partitionOffsetManager) SetOffset(offset int64, metadata string) {
+func (pom *partitionOffsetManager) MarkOffset(offset int64, metadata string) {
 	pom.lock.Lock()
 	defer pom.lock.Unlock()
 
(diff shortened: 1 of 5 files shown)


wvanbergen:
Given that this allows you to send the return value directly to `Consumer.ConsumePartition`, we should probably make sure it always returns the next offset to consume, instead of the last offset that was committed.

offset_manager.go
```diff
@@ -250,8 +251,12 @@ func (pom *partitionOffsetManager) fetchInitialOffset(retries int) error {
 
 	switch block.Err {
 	case ErrNoError:
-		pom.offset = block.Offset
-		pom.metadata = block.Metadata
+		if block.Offset >= 0 {
+			pom.offset = block.Offset
+			pom.metadata = block.Metadata
+		} else {
+			pom.offset = pom.parent.conf.Consumer.Offsets.Initial
```
eapache: you don't actually want to return `OffsetNewest` here, you want to ask the broker what the earliest available offset is and return that.

eapache:
I am not convinced this makes sense when the offset manager is just a subcomponent of the higher-level consumer.

> because this value will always be the same for every partition in a consumer.

Is that true? What about the migration case where you want to pass a distinct constant (manual) offset for each partition?

offset_manager.go
eapache: Oh, or is this intentional... that's really not clear.

wvanbergen: I prefer returning the constant. This way to user of the API can know there's no offset stored, and do something specific in that case (e.g. log). Moreover, in the end you feed to value to `ConsumePartition`, which will always do a `FetchOffset` API call to validate/lookup the actual offset to use. So this saves you one roundtrip to the broker.

wvanbergen:
For the migration case, I was thinking either of the following models:

- Use a separate tool to simply commit the current offsets for Kafka, and restart the consumer afterwards.
- Get the `Offset()`, and if it is negative, get it from to old storage engine, and set it with `SetOffset`.

offset_manager.go
eapache: Right. I would prefer to have this `if` in the actual `Offset()` method I think though; then we can put the `+1` in as well and all the processing logic ends up in one place. Thoughts?

eapache: Also then it should probably be called `NextOffset`?

wvanbergen: :+1:, will fix

wvanbergen:
Renamed `Offset` to `NextOffset`, and moved the logic to that function. Also renamed `SetOffset` to `MarkOffset`: I think it's a little bit more descriptive, but I am open to other suggestions.

Also, the godoc comments probably deserve some :eyes: as well to make sure it's clear how this functions should be used.

wvanbergen:
I added a functional test for the offset manager. I discovered a bug with it: once you have managed a partition, and closed the partition offset manager, you cannot open a new partition offset manager for that partition. 

As a consumer this has to work: it's possible to start consuming a partition, then have another instance manage that partition, but later get the partition assigned back to you if the other consumer instances shuts down.

My attempt at fixing: https://github.com/Shopify/sarama/commit/3d1316a183c15224ea66a126e722287ddb326146

offset_manager.go
```diff
@@ -111,33 +111,48 @@ func (om *offsetManager) abandonBroker(bom *brokerOffsetManager) {
...
-	// this value has not necessarily been flushed to the cluster yet. If you want to resume a partition consumer
-	// from where it left off, remember that you have to increment the offset by one so the partition consumer will
-	// start at the next message. This prevents the last committed message from being processed twice.
-	Offset() (int64, string)
-
-	// SetOffset sets the offset and metadata according to the manager; this value (or a subsequent update)
-	// will eventually be flushed to the cluster based on configuration. You should only set the offset of
-	// messages that have been completely processed.
-	SetOffset(offset int64, metadata string)
+	// NextOffset returns the next offset should be consumed for the managed partition, accompanied by the
```
aaronkavlie-wf: "next offset **that** should be consumed..."

offset_manager.go
```diff
@@ -111,33 +111,48 @@ func (om *offsetManager) abandonBroker(bom *brokerOffsetManager) {
...
+	// state metadata that can be used to reconstruct the state of the consumer.
+	// NextOffset() will return `config.Consumer.Offsets.Initial` and and empty metadata string if no
```
aaronkavlie-wf: "and and" -> "and an"

offset_manager.go
```diff
@@ -111,33 +111,48 @@ func (om *offsetManager) abandonBroker(bom *brokerOffsetManager) {
...
+	// offset was committed for this partition yet.
+	NextOffset() (int64, string)
```
aaronkavlie-wf: This should have named arguments as `MarkOffset` does below

offset_manager.go
```diff
@@ -111,33 +111,48 @@ func (om *offsetManager) abandonBroker(bom *brokerOffsetManager) {
...
-	// from where it left off, remember that you have to increment the offset by one so the partition consumer will
-	// start at the next message. This prevents the last committed message from being processed twice.
-	Offset() (int64, string)
-
-	// SetOffset sets the offset and metadata according to the manager; this value (or a subsequent update)
-	// will eventually be flushed to the cluster based on configuration. You should only set the offset of
-	// messages that have been completely processed.
-	SetOffset(offset int64, metadata string)
+	// NextOffset returns the next offset should be consumed for the managed partition, accompanied by the
+	// state metadata that can be used to reconstruct the state of the consumer.
```
aaronkavlie-wf: The meaning of "state metadata" is a little unclear here; that sounds like a special type of metadata. That phrase isn't used elsewhere in the docs.

offset_manager.go
wvanbergen: These are return values, not arguments

offset_manager.go
```diff
@@ -111,33 +111,48 @@ func (om *offsetManager) abandonBroker(bom *brokerOffsetManager) {
...
+	// NextOffset() will return `config.Consumer.Offsets.Initial` and and empty metadata string if no
+	// offset was committed for this partition yet.
+	NextOffset() (int64, string)
+
+	// MarkOffset marks the provided offset as processed, alongside a metadata string that represents
+	// the state of the partition consumer at that point in time. The metadata string can be used by
+	// another consumer to restore that state, so it can resume consumption.
+	// Note: calling MarkOffset does not necessarily commit the offset to the backend store immediately
```
aaronkavlie-wf: Should put a blank line before the note

offset_manager.go
```diff
@@ -111,33 +111,48 @@ func (om *offsetManager) abandonBroker(bom *brokerOffsetManager) {
...
+	// for efficiency reasons, and it may never be committed if your application crashes. This means that
+	// you may end up processing the same message twice, and your processing should ideally be idempotent.
+	MarkOffset(offset int64, metadata string)
```
aaronkavlie-wf: So the name change is due to the fact that we would resume from the offset *after* this one, right?

One point of confusion here is that the name (and docstring) imply that every consumed offset must be marked -- "mark as done" vs. "save the last offset"

offset_manager.go
aaronkavlie-wf: Oh, right.

You can name return values though; might be a good idea here.

offset_manager.go
wvanbergen: While you don't necessarily have to do that, I think that is the best practice to use this API. We implement batching / commit intervals for you, so you don't have to worry about overloading the broker by calling this function often.

offset_manager.go
wvanbergen: We don't do that for any of our interfaces, so for consistency I'll leave it at just types.

eapache:
:+1:

PR #523, opened 2015-08-26
Fix two decoding bugs found by go-fuzz

eapache:
(https://github.com/dvyukov/go-fuzz)

- handle negative message-set sizes in FetchResponses
- handle IPv6 and/or malformed broker addresses

@Shopify/kafka

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -312,7 +312,10 @@ func (b *Broker) decode(pd packetDecoder) (err error) {
 		return err
 	}
 
-	b.addr = fmt.Sprint(host, ":", port)
+	b.addr = net.JoinHostPort(host, fmt.Sprint(port))
+	if _, _, err := net.SplitHostPort(b.addr); err != nil {
+		return err
+	}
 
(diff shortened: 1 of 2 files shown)


broker.go
```diff
@@ -312,7 +312,10 @@ func (b *Broker) decode(pd packetDecoder) (err error) {
 		return err
 	}
 
-	b.addr = fmt.Sprint(host, ":", port)
+	b.addr = net.JoinHostPort(host, fmt.Sprint(port))
+	if _, _, err := net.SplitHostPort(b.addr); err != nil {
+		return err
+	}
```
wvanbergen: I assume this was because of a hostname with a colon in it?

wvanbergen:
Nice work :clap: :ship:

broker.go
eapache: and hostnames with `[` in them and `\0` and all sorts of other interesting garbage that go-fuzz generated

PR #525, opened 2015-08-27
Centralize and unify getSubset error checking

eapache:
The fix in #523 did not cover all cases, as found by subsequent go-fuzz runs.

@wvanbergen


wvanbergen:
Looks good :+1:

PR #527, opened 2015-08-31
Use an optimized crc32 library which is faster

eapache:
Needs testing to ensure that this is actually faster (and is still correct), but may solve #255.

### DIFF
--- a/crc32_field.go
+++ b/crc32_field.go
@@ -2,7 +2,8 @@ package sarama
 
 import (
 	"encoding/binary"
-	"hash/crc32"
+
+	"github.com/klauspost/crc32"
 )
 
 // crc32Field implements the pushEncoder and pushDecoder interfaces for calculating CRC32s.


wvanbergen:
We may also want to check what architectures this will run on, because it uses CPU instructions that may not be available everywhere, and I don't know if it does fallbacks.

eapache:
On a quick glance it uses `+build` flags and implements the stdlib version as a fallback, so we should be OK.

wvanbergen:
OK, let's run some benchmarks!

eapache:
My benchmark shows no difference; possibly vagrant does not support the necessary instructions. Since golang profiling is broken on native MacOS, we may have to run something on one of our servers.

eapache:
Even on a bare-metal server whose `/proc/cpuinfo` says it supports SSE4.2 I see the same amount of time spent in the fallback method... not sure what else I need to do to actually make the optimizations run.

eapache:
Welp, found a bug in *our* code now when benchmarking. Another PR incoming.

eapache:
Short version is: this basically removes CRC32 from the profile entirely, it drops from 15-20% of the CPU time in my test to a statistically insignificant number of samples. Sold.

Also, the upstream author has submitted this to Go, so there's a very good chance Go 1.6 will have this optimization in the stdlib.

wvanbergen:
:clap:

PR #528, opened 2015-08-31
Fix and simplify message decoding

eapache:
- Don't short-circuit return when inflating compressed messages. There are no
  other fields left, but we should still check the wrapping message's CRC32.
- Do short-circuit return that CRC32 field, it's the last thing we do so there's
  no need for a manual if-not-nil return.

@Shopify/kafka found while testing #527


wvanbergen:
:+1:

PR #531, opened 2015-09-01
Fix offset-manager expiry timestamps to 'now'

### LINKED ISSUE #530: OffsetManager does not set timestamp right
When OffsetManager [commits](https://github.com/Shopify/sarama/blob/master/offset_manager.go#L479) an offset it sets timestamp to be `0`, that is the beginning of the Unix epoch. That effectively makes the commit expired immediately. On my Kafka setup I observed that such commits disappear in 1 minute (that is why all tests pass) and Kafka starts reporting `-1` as the next commit. The right value of timestamp that means **now** is `-1`, which is captured by the [ReceiveTime](https://github.com/Shopify/sarama/blob/master/offset_commit_request.go#L6) constant.

eapache:
Thanks to Maxim for pointing out that 0 doesn't mean now, it means unix epoch 0
(a long time ago).

Fixes #530.

@horkhe @Shopify/kafka

### DIFF
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -476,7 +476,7 @@ func (bom *brokerOffsetManager) constructRequest() *OffsetCommitRequest {
 	for s := range bom.subscriptions {
 		s.lock.Lock()
 		if s.dirty {
-			r.AddBlock(s.topic, s.partition, s.offset, 0, s.metadata)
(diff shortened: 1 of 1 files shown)


horkhe:
Much better now :)

PR #538, opened 2015-09-24
Correctly handle unencodable messages in the producer

eapache:
Quick fix for #449, pending more thought and writing tests etc.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -119,6 +119,8 @@ type ProducerMessage struct {
 
 	retries int
 	flags   flagSet
+
+	keyCache, valueCache []byte
 }
 
 func (m *ProducerMessage) byteSize() int {
@@ -135,6 +137,8 @@ func (m *ProducerMessage) byteSize() int {
 func (m *ProducerMessage) clear() {
 	m.flags = 0
 	m.retries = 0
+	m.keyCache = nil
+	m.valueCache = nil
 }
 
 // ProducerError is the type of error generated when the producer fails to deliver a message.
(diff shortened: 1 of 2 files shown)


cep21:
What if the flag setting happened inside return error or return success?

wvanbergen:
can we simply set the returned flag in the `return` functions themselves?

eapache:
> What if the flag setting happened inside return error or return success?

I suspect I'll end up getting rid of the flag (and the setting of the array to `nil` which it partially replaced) entirely, as that seems very fragile. Yes, setting the flag in the returning functions would have fixed this issue, but it also would have potentially hidden others since there are a lot of other places we could use messages without checking for the flag.

I just threw this up as-is for @cep21's benefit, I won't merge it yet.

eapache:
Ya, setting the flag isn't ideal at all because it results in messages reaching the user with flags set, which breaks our API guarantee.

eapache:
I've just pushed a better (i.e. actually correct) fix to this branch. It's still not ideal architecturally, but I think I'm happy enough with it to release it as v1.5.1 if you both agree.

cep21:
This code works for me.  There's something stinky about the existence of ```keyCache, valueCache```, but I think that's what you're talking about w/ "not ideal" and it is an improvement for sure.  I'll probably merge it into my SafeWaitGroup branch.

eapache:
> but I think that's what you're talking about w/ "not ideal"

Understatement is a bad habit of mine :P

wvanbergen:
:+1: - this fix looks pretty straightforward to me.

Side note: another good reason to remove the Encoder interface for sarama v2. :)

eapache:
> Side note: another good reason to remove the Encoder interface for sarama v2. :)

Yup :)

eapache:
I'll release 1.5.1 with this shortly and then file another ticket to refactor/solve it properly.

eapache:
1.*6*.1 released.

eapache:
The longer-term refactor has been rolled into #433 which has just suddenly gone from "nice to have eventually" to "reasonably important".

PR #539, opened 2015-09-24
Don't return from example functions

eapache:
go-vet recently added a rule for this it seems

cc @wvanbergen


wvanbergen:
Weird - not a big fan of this because I don't like setting an example of using panic, but I guess this is the new go way :/

PR #542, opened 2015-09-25
Wrap all godoc at 80 characters

### LINKED ISSUE #540: Config struct commentlines too long
Please wrap your struct doc comments.

https://godoc.org/github.com/Shopify/sarama#Config

These struct docs are hard to read on godoc.org. They could be improved by wrapping at 80 columns. That's common practice.

eapache:
cc @wvanbergen 

closes #540 

The fact that the go docs explicitly say "Go has no line length limit" and then godoc.org enforces 80 characters anyways makes me sad :(

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -17,10 +17,11 @@ import (
 // scope.
 type AsyncProducer interface {
 
-	// AsyncClose triggers a shutdown of the producer, flushing any messages it may have
-	// buffered. The shutdown has completed when both the Errors and Successes channels
-	// have been closed. When calling AsyncClose, you *must* continue to read from those
-	// channels in order to drain the results of any messages in flight.
+	// AsyncClose triggers a shutdown of the producer, flushing any messages it may
+	// have buffered. The shutdown has completed when both the Errors and Successes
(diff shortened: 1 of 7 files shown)


wvanbergen:
:+1:

I really don't like how the godoc generates documentation for interfaces compared to how it generates it for concrete types. :(

Did you use some kind of tool for this?

eapache:
Hand-rolled with a few vim macros :/

adg:
Thanks for doing it. There's an open issue on the golang/go issue tracker to improve the presentation of interface and struct types. Not sure how to resolve it though.

eapache:
Struct/interface members and their docs should be parsed out and reflowed just like top-level types. The exact display format used after that is just CSS and is almost irrelevant.

adg:
Top level docs show declarations and associated doc comments.
What you suggest would split declarations into multiple pieces.
So at the very least this is not "just like top-level types."

eapache:
Hmm... I have some ideas for that once I get back to a real keyboard.

adg:
My point is that there are plenty of ideas, but there is no obvious
solution.

PR #544, opened 2015-09-27
Producer refactor

### LINKED ISSUE #433: Construct produce requests incrementally

eapache:
This is an experimental refactor of the entire aggregator/flusher pipeline. If successful it would close #433 as well as properly solve the issue papered over by #538.

I did a decent job of making each commit stand-alone and self-explanatory, so you can read commit-at-a-time, or just treat it as entirely new code. The total diff against the old code is impossible to follow.

High-level, this is what I did:
- Introduce a `produceSet` structure which collects messages, rejecting those which fail to encode and incrementally constructing a `ProduceRequest` out of those which succeed.
- Keep a running `produceSet` in the aggregator instead of just a `[]*ProducerMessage`. Move all the "how many bytes/messages in this batch so far?" logic into the `produceSet`.
- Merge the flusher and the aggregator into a single `brokerProducer`. Revive the flusher as a trivial goroutine whose sole purpose is to let the aggregator `select` on the result of its network calls.
- Clean up a bunch of now-dead code.

This new layout has a number of benefits:
- Messages live solely in the `produceSet`, which makes them much easier to track vs. having them spread across a slice, a map, and a long-lived request object.
- Messages only even enter the `produceSet` when they encode properly, which means that a whole raft of logic can go away around handling requests which encoded successfully but contained no successful messages. This also means that we no longer need to nil out array entries for failed messages, getting rid of another bunch of ugly logic.
- Retry state is handled message-at-a-time as the messages enter the `brokerProducer`, rather than batch-at-a-time as they enter the `flusher`. This is much easier to follow, cleans up a few hacks that were necessary for the aggregator to correctly pass along flagged messages, and makes for a sane state machine that clearly terminates in all cases.
- The `produceSet` can track how many bytes it has accumulated in each individual partition. It also has access to the actual encoded length, not just the estimate provided by the `Encoder` interface. This makes it more precise, and lets it accumulate much larger batches safely. In particular, this lets it accumulate requests larger than `MaxMessageSize` (usually 1MB) when using compression.
- It is now substantially easier to add support for request pipelining if that project ever gets off the ground on the server side.

@wvanbergen 
cc @kvs @andremedeiros @bai @cep21 

P.S. I don't want to make any claims about performance without measurements to back them up, but in theory this is a performance win too: we do fewer passes over each message, we generate less garbage for the collector, and we do more work without needing to context-switch.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -542,10 +542,8 @@ type aggregator struct {
 	input  <-chan *ProducerMessage
 	output chan<- *produceSet
 
-	buffer      *produceSet
-	bufferBytes int
-	bufferCount int
-	timer       <-chan time.Time
+	buffer *produceSet
+	timer  <-chan time.Time
 }
 
 func (a *aggregator) run() {
@@ -558,7 +556,7 @@ func (a *aggregator) run() {
 				goto shutdown
 			}
 
-			if a.wouldOverflow(msg) {
+			if a.buffer.wouldOverflow(msg) {
 				Logger.Printf("producer/aggregator/%d maximum request accumulated, forcing blocking flush\n", a.broker.ID())
 				a.output <- a.buffer
 				a.reset()
@@ -566,10 +564,8 @@ func (a *aggregator) run() {
 			}
 
 			a.buffer.add(msg)
-			a.bufferBytes += msg.byteSize()
-			a.bufferCount++
 
-			if a.readyToFlush(msg) {
+			if a.buffer.readyToFlush(msg) {
 				output = a.output
 			} else if a.parent.conf.Producer.Flush.Frequency > 0 && a.timer == nil {
 				a.timer = time.After(a.parent.conf.Producer.Flush.Frequency)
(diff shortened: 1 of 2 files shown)


eapache:
Added the requested comment, cleaned up the (mis)-use of the `pending` variable, and then ran into a much bigger problem (third new commit) that I'm not particularly comfortable with yet. TBD.

eapache:
OK, another refactor and I think I'm happy with this again. This most recent commit ended up solving another problem with the version currently in master: if a response is received which invalidates some of the messages we have buffered to the point where we no longer meet our flush conditions, we now stop trying to flush until the conditions are re-met.

eapache:
Worth noting: this mornings rework has partially lost us benefit point 3 (retry state being handled message-at-a-time) since we have to handle responses arriving when a batch is already collected. I think the only way around that would be to not collect batches while a response is pending, which is of course unacceptably slow in the common success case.

eapache:
I am going to extract this into several stand-alone chunks which should be easier to reason about and review.

PR #545, opened 2015-09-27
producer: simplify initial "too large" check

eapache:
Minor oddity I noticed while working on the aggregator/flusher refactor; this
looks like vestigial code from some attempt at handling the weird size
requirements of compressed message batches, but since `msg.byteSize()`
*includes* `msg.Value.Length()` if appropriate, the entire first condition of
this `||` is superceded by the second.

@wvanbergen Saturday's not-so-scary PR :)

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -249,9 +249,7 @@ func (p *asyncProducer) dispatcher() {
 			p.inFlight.Add(1)
 		}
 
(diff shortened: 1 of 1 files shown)


wvanbergen:
:+1: not sure where this came from but the first check is a subset of the second check.

Issue #546, opened 2015-09-29
Timeout issue in consumer on broker death


4578395263256:
We are currently running release 1.6.1. During testing, we killed a broker by ipmi_rebooting it (so that it wouldn't be able to close the connection). This caused consumers that were reading from partitions for which that was the leader to get stuck for 3-4 minutes, at the end of which there was this message repeated once per partition

read tcp <ip address>:9092: i/o timeout

and after that things recovered.

Based on the fact that the read timeout is only 30 seconds, it seems like it shouldn't take 3-4 minutes for things to get unstuck. Here are our settings:

kafkaConfig.ClientID = config.ClientID
kafkaConfig.Consumer.Fetch.Default = 16 * 1024 * 1024
kafkaConfig.Consumer.Fetch.Min = 8 * 1024 * 1024
kafkaConfig.Consumer.Fetch.Max = 64 * 1024 * 1024
kafkaConfig.ChannelBufferSize = 500
kafkaConfig.Consumer.Return.Errors = true
kafkaConfig.Metadata.Retry.Max = 10
kafkaConfig.Metadata.Retry.Backoff = 5 * time.Second

Thanks for any insight you can provide!

wvanbergen:
Can you set `sarama.Logger` and capture the output? That should give you more insight in what is going on.

My best guess is this:
- Broker goes away
- Sarama tries to consume from broker, but fails after timeout.
- Sarama asks a different broker what broker it should now consume from now.
- The other broker looks this up in Zookeeper. Zookeeper has not yet been updated** so it holds the same broker as before.
- Sarama tries to connect to the broker and fails again after a timeout.
Etc.

So basically, Sarama will continue to retry using the same broker for a partition until the leader of that partition has been updated in Zookeeper. When a kafka broker doesn't shut down properly, it doesn't have a chance to hand off leadership to another one, and the handoff is determined by the connection timeout of the Kafka broker to Zookeeper. You can configure it in the Kafka broker configuration file.

4578395263256:
Note things recovered around 20:03:25. The following is the sarama logging output. This part of the log goes from 19:57:03 to 20:07:01.

[Sarama] 2015/09/28 19:59:43 consumer/broker/1 disconnecting due to error processing FetchRequest: read tcp <DEAD BROKER IP>:9092: i/o timeout
[Sarama] 2015/09/28 20:00:18 client/metadata fetching metadata for all topics from broker <OTHER BROKER>:9092
[...]

PR #548, opened 2015-09-30
Make dead brokers die harder

eapache:
When a broker gets an error trying to receive a response (either from the
network layer, or from failing to parse the minimal global header), it should
just abandon ship and die. Save that error and return it immediately for any
further requests we might have made.

- The vast majority of the time the connection is going to be hosed anyways, if
  nothing else by being out-of-sync on correlation IDs (which we don't handle
  and which doesn't seem particularly urgent).
- All of Sarama's built-in callers (producer/consumer/offset-manager)
  immediately `Close` a broker when they receive one of these errors anyways, so
  all this does is speed up that in the common case.

*If* one of these errors is recoverable, and *if* there is user-space code
somewhere which actually tries to recover in one of those cases, then that code
would break.

This neatly satisfies one of the XXX comments I left in about this issue from
way back in 2013. The TODOs about correlation ID matching are still present.

@wvanbergen @4578395263256 another approach to #546, one I am generally happier with. Thoughts?

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -344,16 +344,24 @@ func (b *Broker) encode(pe packetEncoder) (err error) {
 }
 
 func (b *Broker) responseReceiver() {
+	var dead error
 	header := make([]byte, 8)
 	for response := range b.responses {
+		if dead != nil {
+			response.errors <- dead
+			continue
+		}
+
 		err := b.conn.SetReadDeadline(time.Now().Add(b.conf.Net.ReadTimeout))
 		if err != nil {
+			dead = err
 			response.errors <- err
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


wvanbergen:
This change looks good to me :+1:.

I can't really see any use code out there trying to salvage the connection, especially because the correlation IDs are probably broken after something like this happens.

wvanbergen:
Alternatively, we could put in a circuit breaker. Not sure if it is worth the effort.

eapache:
A circuit-breaker would be neat, but we already have other ones higher in the stack so we'd end up with nested breakers on slightly different conditions which would be odd. Also, this is really not something that can heal without re-starting the network connection so it's not an ideal use.

PR #549, opened 2015-09-30
Refactor the producer, part 1

eapache:
First stand-alone chunk extracted from #544. This introduces and uses a
produceSet structure which takes care of collecting messages, rejecting ones
which don't encode, and turning the rest into a `ProduceRequest`.

This has several knock-on effects:
- we no longer nil out array entries, so nil checks in several places (e.g.
  `returnErrors`) can be removed
- parseResponse gets re-indented since it loops over the partitions in the set
  via a callback now rather than a pair of nested loops
- groupAndFilter is much simpler, a lot of its logic now lives in the produceSet
- the flusher has to use the set to return errors/successes/retries, rather than
  `batch` which may now contain messages not in the eventual request
- keyCache and valueCache can be removed

@wvanbergen cc @kvs

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -137,12 +137,12 @@ type ProducerMessage struct {
 
 	retries int
 	flags   flagSet
-
-	keyCache, valueCache []byte
 }
 
+const producerMessageOverhead = 26 // the metadata overhead of CRC, flags, etc.
+
 func (m *ProducerMessage) byteSize() int {
-	size := 26 // the metadata overhead of CRC, flags, etc.
+	size := producerMessageOverhead
 	if m.Key != nil {
 		size += m.Key.Length()
 	}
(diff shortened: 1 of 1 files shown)


async_producer.go
```diff
@@ -700,68 +702,47 @@ func (f *flusher) groupAndFilter(batch []*ProducerMessage) map[string]map[int32]
...
-			// Other non-retriable errors
-			default:
-				f.parent.returnErrors(msgs, block.Err)
+		switch block.Err {
+		// Success
+		case ErrNoError:
+			for i := range msgs {
+				msgs[i].Offset = block.Offset + int64(i)
+			}
+			f.parent.returnSuccesses(msgs)
```
wvanbergen: Maybe we can return a success immediately after setting the offset? That prevents a second iteration of the message collection. Might be a micro-optimization that is not worth it.

wvanbergen:
This is a much easier to swallow changeset. :+1:

async_producer.go
eapache: Ya, not right now anyways, we'd have to split `returnSuccesses` out into a `returnSuccess` method too

Issue #294, opened 2015-02-23
Producer hangs when `Retry.Max` is 0


eapache:
(This is also poor behaviour in other configurations in that it can "waste" some of the configured retries, but is otherwise not a problem - it doesn't break ordering or anything.)

The `leaderDispatcher` function only knows to check for a new leader when it receives a message with a new retry value, but there are at least two cases (possibly more?) where for various reasons the `flusher` fails but never sends a retriable message to the leaderDispatcher:
- all of the messages are at their retry limit, and get returned to the user instead
- there is a topic/partition led by that broker, but with no messages in the current batch, when the broker connection dies

In both of the above cases, we can end up with a `leaderDispatcher` whose broker is toast, but doesn't know it and so doesn't try and get a new one.

In the normal case, the next message (not at the retry limit) will get spun around anyways and kick things off the way it should; this is less than ideal (we waste a retry for the message(s) in question) but not a very severe problem as by default we have three retries configured and typically only need one.

However, in the case where someone sets `MaxRetries` to 0, every subsequent message is immediately returned to the user and the producer ends up stuck.

eapache:
The following test will pass/fail this issue, it should be copied into `producer_test.go` when this is fixed.
```golang
func TestProducerOutOfRetries(t *testing.T) {
        seedBroker := NewMockBroker(t, 1)
        leader := NewMockBroker(t, 2)

        metadataResponse := new(MetadataResponse)
        metadataResponse.AddBroker(leader.Addr(), leader.BrokerID())
        metadataResponse.AddTopicPartition("my_topic", 0, leader.BrokerID(), nil, nil, NoError)
        seedBroker.Returns(metadataResponse)

        client, err := NewClient("client_id", []string{seedBroker.Addr()}, nil)
        if err != nil {
                t.Fatal(err)
        }

        config := NewProducerConfig()
        config.FlushMsgCount = 10
        config.AckSuccesses = true
        config.RetryBackoff = 0
        config.MaxRetries = 0
        producer, err := NewProducer(client, config)
        if err != nil {
                t.Fatal(err)
        }

        for i := 0; i < 10; i++ {
                producer.Input() <- &MessageToSend{Topic: "my_topic", Key: nil, Value: StringEncoder(TestMessage)}
        }

        prodNotLeader := new(ProduceResponse)
        prodNotLeader.AddTopicPartition("my_topic", 0, NotLeaderForPartition)
        leader.Returns(prodNotLeader)

[...]

PR #551, opened 2015-10-14
Producer refactor 3

eapache:
@wvanbergen depends on #550, I'll rebase once that's merged.

This is the final step, it moves the request/response handling and all state into the "aggregator" (which is now named the `brokerProducer`). This makes what used to be the `flusher` really trivial, so it gets inlined into the constructor.

Most of the logic stays the same, the most confusing change is probably around forcing a blocking flush, because we now have to handle the case where we receive a response while we are waiting, and that changes our state (see the new `waitForSpace` method).

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -505,214 +505,249 @@ func (pp *partitionProducer) updateLeader() error {
 	})
 }
 
-// one per broker, constructs both an aggregator and a flusher
+// one per broker; also constructs an associated flusher
 func (p *asyncProducer) newBrokerProducer(broker *Broker) chan<- *ProducerMessage {
-	input := make(chan *ProducerMessage)
-	bridge := make(chan *produceSet)
+	var (
+		input     = make(chan *ProducerMessage)
+		bridge    = make(chan *produceSet)
+		responses = make(chan *brokerProducerResponse)
+	)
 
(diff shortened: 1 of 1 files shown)


eapache:
Rebased on master now that part 2 is merged. The diff is still pretty yucky, but I'm not sure how to split it any further. It may be easier to just read the `brokerProducer` code in the new version from scratch, without worrying about the old version.

eapache:
Landing this will close #433 and should let me approach #294 without tearing my hair out (finally).

wvanbergen:
:+1:

With regards to #433 - this doesn't really construct a ProduceRequest incrementally, because that is still done from scratch in `produceSet.buildRequest`. Can this be done incrementally as we add messages to the produceset?

eapache:
> because that is still done from scratch in produceSet.buildRequest

Not from scratch - the `MessageSet`s are built incrementally. The request itself *could* be build incrementally for uncompressed requests, but for compressed requests we have to wait until we're ready in order to compress the body all at once, so that would be weird. If snappy lib provided an incremental stream compressor we might be able to do both incrementally, but it's not worth the effort at this point.

eapache:
Still tearing my hair out over #294 - I am learning to hate this state machine

Issue #553, opened 2015-10-16
Mock issues in Go 1.5


aaronkavlie-wf:
Several of our tests using `github.com/Shopify/sarama/mocks` are failing in Go 1.5.1 due to count mismatches between expected/actual offsets and the like. The same test suite passes when run with Go 1.4.2.

Have you seen any difference in test code that uses `mocks` with Go 1.5?

eapache:
@wvanbergen ?

Off the top of my head it sounds like a race condition exposed by 1.5's change to the default `GOMAXPROCS`. Does adding `-race` to the test expose anything? Is this producer or consumer?

aaronkavlie-wf:
I'm seeing the test failures in the file that covers our sarama extraction -- both the producer and consumer are used there.

A race condition is likely. I commented out all but the first test, and saw a panic due to nil pointer at first. After adding a logging statement to the sarama source, then removing it, the test now fails with no panic. And running with `-race` made the test pass. Very odd.

aaronkavlie-wf:
I just saw it fail, and then pass, with the `-race` flag and no changes.

Further evidence of a race condition:

```
	<autogenerated>:31: Unexpected offset when calling ConsumePartition for test/0. Expected 4, got 1.
	<autogenerated>:31: Unexpected offset when calling ConsumePartition for test/1. Expected 1, got 4.
	buffer_test.go:258: Next offset expected to be 4; actual: 1
	buffer_test.go:263: Next offset expected to be 1; actual: 4
```

eapache:
It's not super-clear to me, but based on those logs could it be an actual bug in your code that's been uncovered? The mock code involved is fairly simple and protected by a mutex: you're telling it to expect `test/0` with offset 4, and then actually calling `ConsumePartition` with offset 1... and vice-versa for `test/1`.

aaronkavlie-wf:
@eapache I have found a race condition in my code; I was naively assuming loop index order was preserved.

I have another race condition which is not so simple however. Tests now pass *most* of the time, but sometimes, some of the messages inserted with `mocks.PartitionConsumer.YieldMessage()` are not ready when I check `partitionConsumer.Messages()`. Is there any way to ensure that the test messages have been added to the channel?

eapache:
I don't see one off the top of my head. @wvanbergen why does `YieldMessage` go through `handleExpectations` instead of just writing straight to `messages`?

PR #555, opened 2015-10-27
Send messages and errors directly in consumer mock

aaronkavlie-wf:
The mock `PartitionConsumer` sent messages to an expectations channel
on calls to `YieldMessage` and `YieldError`. This led to race conditions
in tests if `pc.Messages` was called before the `handleExpectations`
goroutine added a message to the channel.

This PR sends message and errors to their respective channels
directly. The expectations channel & goroutine are removed
altogether, simplifying the code.

This resolves issue #553.

@eapache @wvanbergen

### DIFF
--- a/mocks/consumer.go
+++ b/mocks/consumer.go
@@ -60,7 +60,6 @@ func (c *Consumer) ConsumePartition(topic string, partition int32, offset int64)
 	}
 
 	pc.consumed = true
-	go pc.handleExpectations()
 	return pc, nil
 }
 
@@ -141,13 +140,12 @@ func (c *Consumer) ExpectConsumePartition(topic string, partition int32, offset 
 
 	if c.partitionConsumers[topic][partition] == nil {
 		c.partitionConsumers[topic][partition] = &PartitionConsumer{
-			t:            c.t,
-			topic:        topic,
-			partition:    partition,
-			offset:       offset,
-			expectations: make(chan *consumerExpectation, 1000),
-			messages:     make(chan *sarama.ConsumerMessage, c.config.ChannelBufferSize),
-			errors:       make(chan *sarama.ConsumerError, c.config.ChannelBufferSize),
+			t:         c.t,
+			topic:     topic,
+			partition: partition,
+			offset:    offset,
+			messages:  make(chan *sarama.ConsumerMessage, c.config.ChannelBufferSize),
+			errors:    make(chan *sarama.ConsumerError, c.config.ChannelBufferSize),
 		}
 	}
 
@@ -169,7 +167,6 @@ type PartitionConsumer struct {
 	topic                   string
 	partition               int32
 	offset                  int64
-	expectations            chan *consumerExpectation
 	messages                chan *sarama.ConsumerMessage
 	errors                  chan *sarama.ConsumerError
 	singleClose             sync.Once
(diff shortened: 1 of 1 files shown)


mocks/consumer.go
```diff
@@ -289,7 +261,13 @@ func (pc *PartitionConsumer) HighWaterMarkOffset() int64 {
 // reasons forthis not to happen. ou can call ExpectMessagesDrainedOnClose so it will
 // verify that the channel is empty on close.
 func (pc *PartitionConsumer) YieldMessage(msg *sarama.ConsumerMessage) {
-	pc.expectations <- &consumerExpectation{Msg: msg}
+	pc.highWaterMarkOffset += 1
```
eapache: Does this still need to be atomic and/or locked somehow? I guess it depends on if people expect to be able to call this method multiple times concurrently, but I don't know if that's true (it was safe to call the old one concurrently).

eapache:
One thought, but generally LGTM.

mocks/consumer.go
wvanbergen: It's atomic in sarama, so I prefer the mock to be atomic as well to match the sarama API as well as possible.

wvanbergen:
:+1:, just need non-atomic increment needs fixing.

mocks/consumer.go
aaronkavlie-wf: I removed the atomic operations because it's no longer executed in a goroutine... then again, `handleExpectations` uses a for loop anyway, so that alone would not lead to race conditions. So I guess it's not totally clear to me why they were there.

If atomic operations are needed in case someone calls code concurrently, that means that plain assignments are never safe in public APIs.

wvanbergen: In sarama it's atomic, because it's possible to have the processing lag monitoring in a separate goroutine. So it's not so much about internal use inside the mock, but more about how the API is used.

aaronkavlie-wf:
switched back to the atomic operations.

wvanbergen:
Thanks!

mocks/consumer.go
```diff
@@ -289,7 +260,13 @@ func (pc *PartitionConsumer) HighWaterMarkOffset() int64 {
 // reasons forthis not to happen. ou can call ExpectMessagesDrainedOnClose so it will
 // verify that the channel is empty on close.
 func (pc *PartitionConsumer) YieldMessage(msg *sarama.ConsumerMessage) {
-	pc.expectations <- &consumerExpectation{Msg: msg}
+	atomic.AddInt64(&pc.highWaterMarkOffset, 1)
```
eapache: two concurrent calls to this method can still interleave, resulting in incorrect `msg.Offset` values

aaronkavlie-wf: Are you suggesting that lock/unlock should be added here too?

eapache: I think it is necessary? `atomic.AddInt64` returns the new value, so you could collapse the atomic calls and reduce the window, but it would still be possible to interleave method calls such that messages ended up in the channel with the wrong order (e.g. offsets `1 2 3 4` instead of `1 2 3 4`). A lock seems the obvious way to force the `highWaterMarkOffset` change to occur synchronously with the channel write.

Note that the variable will still need to be edited atomically in order to avoid locking `HighWaterMarkOffset()` too.

aaronkavlie-wf: I could add it in, this is already merged... will it still work on this branch, or should I do another?

eapache: another branch is probably simpler

PR #557, opened 2015-10-28
Add an example of using the consumer

### LINKED ISSUE #556: Option: Provide an example of consuming message
In the example chapter of the document,  I can't find an example of consuming message.
And in the test case, there doesn't has an complete example too.

eapache:
Somehow we never added this.

@wvanbergen @i19 

Fixes #556.

### DIFF
--- a/consumer_test.go
+++ b/consumer_test.go
@@ -1,6 +1,9 @@
 package sarama
 
 import (
+	"log"
+	"os"
+	"os/signal"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


wvanbergen:
:+1:, this is a good start. (I could have sworn we already had something like this, but apparently not).

Might be nice to also have a more complete example that consumes all partitions of a topic. Maybe that is better suited for the examples folder though.

i19:
Thx a lot. 
I found that sarama doesn't have connection to ZooKeeper. So I wonder how you guys perform the rebalance and how to scale the consumer?

eapache:
Zookeeper logic is in a separate package: https://github.com/wvanbergen/kafka

i19:
Cool ! Thx again !

PR #559, opened 2015-10-29
Collapse two sync.atomic calls

eapache:
It doesn't matter since we're inside a mutex anyways, but this is cleaner.

cc @wvanbergen @aaronkavlie-wf


wvanbergen:
:+1:

aaronkavlie-wf:
:+1:

PR #562, opened 2015-11-10
producer: explicitly register partitions

eapache:
Introduce a `syn` flag (a la TCP) and send a syn message from partitionProducer
to brokerProducer on setup. This lets brokerProducer know exactly which
partitions it should handle at any time, even if they haven't sent an actual
message yet.

For symmetry, rename `chaser` to `fin`, since it is already the last message
from a partitionProducer to a brokerProducer, and also the last from
brokerProducer to partitionProducer. Tweak logging to be a little more explicit around the brokerPoducer state.

First piece of a refactor my subconscious has been working on without me.

@wvanbergen

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -585,6 +585,8 @@ func (bp *brokerProducer) run() {
 			}
 
 			if msg.flags&syn == syn {
+				Logger.Printf("producer/broker/%d state change to [open] on %s/%d\n",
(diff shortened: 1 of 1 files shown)


wvanbergen:
This looks good to me, and also makes the chaser stuff easier to understand. :+1:

PR #563, opened 2015-11-10
Extract produceSet and add tests

eapache:
Now that `produceSet` is its own stand-alone structure with well-defined
behaviour, extract it into its own file (pure copy-paste, nothing to review) to
make async_producer.go not quite so huge.

This makes it more obvious that produceSet can/should have unit tests, so add a
few basic ones.

@wvanbergen


wvanbergen:
:+1: nice refactor.

PR #565, opened 2015-11-11
Officially drop 1.3 (but it still works for now)

eapache:
It has been much more than 2 months since 1.5 came out so we can drop 1.3 now.

Also use more recent point releases of 1.4 and 1.5.

cc @wvanbergen


wvanbergen:
:+1:

PR #569, opened 2015-11-13
Introduce mockSequence and further simplify tests

horkhe:
This PR introduces `mockSequence` response builder that is created of a sequence of (mock)responses and everytime when `MockBroker` calls its `For` method it returns the next response from the sequence. When the last element in the sequence is reached it is returned for every subsequent call of the `For` method.

This change allows to define `MockBrocker` handler with `SetHandlerByMap` exclusively, that makes tests a bit shorter and easier to read.


eapache:
LGTM. @wvanbergen ?

wvanbergen:
Much nicer :+1:

PR #572, opened 2015-11-26
Add kafka 0.9

eapache:
Fix one new constraint on the broker ID which must now be small unless you
configure otherwise. Fortunately it seems older versions just ignore unrecognized config stanzas.

P.S. I believe this is because 0.9 now optionally auto-generates broker IDs for you within a given range? At some point we need to find/read a good changelog.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ env:
   matrix:
   - KAFKA_VERSION=0.8.1.1
   - KAFKA_VERSION=0.8.2.2
+  - KAFKA_VERSION=0.9.0.0
 
 before_install:
 - export REPOSITORY_ROOT=${TRAVIS_BUILD_DIR}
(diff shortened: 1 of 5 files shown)


wvanbergen:
:+1:

eapache:
@wvanbergen review new extra commit please

wvanbergen:
:+1: again :)

eapache:
OK, failure is that before an `OffsetFetch` of a non-existent topic/partition would return `ErrUnknownTopicOrPartition`, now it just returns an offset of -1. I'll mimic the old behaviour for now so I don't have to think about compatibility too hard.

eapache:
Added a skip, even working around that there are several problems with the offset manager. I'll file a ticket instead.

eapache:
Filed #573 and #574.

jan4984:
:+1:

PR #575, opened 2015-11-28
fix lint & alignment issues

hollow:
Updated Makefile to:

* isolate build with goop
* run gofmt with -s
* run gometalinter to check for various issues, including variable shadowing, alignment issues, etc


hollow:
seems to have failed due to network issues on travis. can someone restart this build?

wvanbergen:
Hey, thanks for contributing. Your PR has a bunch of unrelated things cobbled together, which makes this PR a bit hard to swallow..

- We prefer not using a package manager for sarama, because it is a library, not an app. We want sarama to work with the latest versions of our dependencies, which is exactly what using vanilla `go get` gives us.
- Some of the lint-related changes are definitely food, but some of the other ones I may not necessarily agree with. It would be nice if you could open separate PRs for different classes of lint fixes you have applied so we can look at them individually.

hollow:
Hey, I already figured that this will not be nerfed, so we will simply keep all our improvements in the remerge fork

PR #577, opened 2015-12-07
Implement Kafka 0.9 protocol additions

wvanbergen:
@eapache Am I on the right track here?

### DIFF
--- a/describe_groups_response.go
+++ b/describe_groups_response.go
@@ -1,0 +1,162 @@
+package sarama
+
+type DescribeGroupsResponse struct {
+	Groups []*GroupDescription
+}
+
+func (r *DescribeGroupsResponse) encode(pe packetEncoder) error {
+	if err := pe.putArrayLength(len(r.Groups)); err != nil {
+		return err
+	}
+
+	for _, groupDescription := range r.Groups {
+		if err := groupDescription.encode(pe); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *DescribeGroupsResponse) decode(pd packetDecoder) (err error) {
+	n, err := pd.getArrayLength()
+	if err != nil {
+		return err
+	}
+
+	r.Groups = make([]*GroupDescription, n)
+	for i := 0; i < n; i++ {
+		r.Groups[i] = new(GroupDescription)
+		if err := r.Groups[i].decode(pd); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+type GroupDescription struct {
+	Err          KError
+	GroupId      string
+	State        string
+	ProtocolType string
+	Protocol     string
+	Members      map[string]*GroupMemberDescription
+}
+
+func (gd *GroupDescription) encode(pe packetEncoder) error {
+	pe.putInt16(int16(gd.Err))
+
+	if err := pe.putString(gd.GroupId); err != nil {
+		return err
+	}
+	if err := pe.putString(gd.State); err != nil {
+		return err
+	}
+	if err := pe.putString(gd.ProtocolType); err != nil {
+		return err
+	}
+	if err := pe.putString(gd.Protocol); err != nil {
+		return err
+	}
+
+	if err := pe.putArrayLength(len(gd.Members)); err != nil {
+		return err
+	}
+
+	for memberId, groupMemberDescription := range gd.Members {
+		if err := pe.putString(memberId); err != nil {
+			return err
+		}
+		if err := groupMemberDescription.encode(pe); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (gd *GroupDescription) decode(pd packetDecoder) (err error) {
+	if kerr, err := pd.getInt16(); err != nil {
+		return err
+	} else {
+		gd.Err = KError(kerr)
+	}
+
+	if gd.GroupId, err = pd.getString(); err != nil {
+		return
+	}
+	if gd.State, err = pd.getString(); err != nil {
+		return
+	}
+	if gd.ProtocolType, err = pd.getString(); err != nil {
+		return
+	}
+	if gd.Protocol, err = pd.getString(); err != nil {
+		return
+	}
+
+	n, err := pd.getArrayLength()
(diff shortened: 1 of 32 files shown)


join_group_request.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (r *JoinGroupRequest) version() int16 {
+	return 0
+}
+
+type GroupProtocol interface {
+	encodeGroupProtocol(packetEncoder) error
+	decodeGroupProtocol(packetDecoder) error
+}
```
wvanbergen: This part is a bit tricky; the protocol defines this to be extensible based on the `ProtocolType`. It continues to only define the specifics for the "consumer" protocol.

join_group_request.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	pe.putInt16(cgp.Version)
+	if err := pe.putStringArray(cgp.Subscription); err != nil {
+		return err
+	}
+	return pe.putBytes(cgp.UserData)
+}
+
+func (cgp *ConsumerGroupProtocol) decodeGroupProtocol(pd packetDecoder) error {
+	return nil
```
wvanbergen: TODO

errors.go
```diff
@@ -37,6 +37,9 @@ var ErrShuttingDown = errors.New("kafka: message received by producer in process
 // ErrMessageTooLarge is returned when the next message to consume is larger than the configured Consumer.Fetch.Max
 var ErrMessageTooLarge = errors.New("kafka: message is larger than Consumer.Fetch.Max")
 
+// ErrMessageTooLarge is returned when a JoinGroup request returns a protocol type that is not supported by sarama.
+var ErrUnknownGroupProtocol = errors.New("kafka: encountered an unknown group protocol")
```
edenhill: This can't happen, the broker selects the protocols supported by all members, and if no such set exists throws a InconsistentGroupProtocol

eapache:
You're going to need to add all the new `KError` values to this as well.

join_group_request.go
```diff
@@ -0,0 +1,127 @@
...
+		if err := groupProtocol.encodeGroupProtocol(pe); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *JoinGroupRequest) decode(pd packetDecoder) (err error) {
+	r.GroupId, err = pd.getString()
```
eapache: these can all be collapsed into their `if` lines because `err` is predefined

join_group_request.go
eapache: I am tempted to just make this `[]byte` and handle the consumer-specific implementation via `Encoder` interface? Not sure, but I dislike a public interface with no public methods.

errors.go
wvanbergen: Right now this error is only used when the ProtocolType on the request is set (by the user) to an unknown value (i.e. anything but `consumer`).

wvanbergen:
Added new `KError`s, and Heartbeat request and response pair.

errors.go
```diff
@@ -140,6 +154,28 @@ func (err KError) Error() string {
...
+	case ErrRebalanceInProgress:
+		return "kafka server: A rebalance for the group is in progress. Please re-join the group."
+	case ErrInvalidCommitOffsetSize:
+		return "kafka server: The provided commit metadata was too large"
+	case ErrTopicAuthorizationFailed:
+		return "kafka server: The client is not authorized to access this topic"
+	case ErrGroupAuthorizationFailed:
+		return "kafka server: The client is not authorized to access this group"
+	case ErrClusterAuthorizationFailed:
+		return "kafka server: The client is not authorized to send this request type"
```
eapache: for consistency, the existing errors end in punctuation

wvanbergen:
I've implemented all the new response/request pairs. I have also resumed the ConsumerGroup abstraction - We can later decide how to handle that best

eapache:
LGTM. Collapse all the `decode` if statements and add tests, then :shipit:

join_group_request_test.go
```diff
@@ -0,0 +1,62 @@
...
+	joinGroupRequestTwoProtocols = []byte{
+		0, 9, 'T', 'e', 's', 't', 'G', 'r', 'o', 'u', 'p', // Group ID
+		0, 0, 0, 100, // Session timeout
+		0, 12, 'T', 'w', 'o', 'P', 'r', 'o', 't', 'o', 'c', 'o', 'l', 's', // Member ID
+		0, 8, 'c', 'o', 'n', 's', 'u', 'm', 'e', 'r', // Protocol Type
+		0, 0, 0, 2, // 2 group protocol
+		0, 3, 't', 'w', 'o', // Protocol name
+		0, 0, 0, 3, 0x04, 0x05, 0x06, // protocol metadata
+		0, 3, 'o', 'n', 'e', // Protocol name
+		0, 0, 0, 3, 0x01, 0x02, 0x03, // protocol metadata
```
wvanbergen: This gets assembled from a map, so the order is not deterministic. @eapache how have you gotten around this elsewhere?

eapache: Only test the 0/1 cases :P

wvanbergen:
I changed some slice types into maps to make the types easier to use, and added tests to all the things. This is ready for final review.

wvanbergen:
Note: some messages, fields, and errors are also renamed as part of the 0.9 release. For backwards compatibility reasons, these are not included.

describe_groups_response.go
```diff
@@ -0,0 +1,162 @@
+package sarama
+
+type DescribeGroupsResponse struct {
+	Groups []*GroupDescription
```
eapache: kinda/almost/sorta makes sense for this to be a map keyed by group ID, but not sure if it's worth it at this point

eapache: it would make decode rather more complex

eapache:
:shipit:

describe_groups_response.go
wvanbergen: I also like that this matches the request: you ask for a slice of groups, you get a slice of group descriptions back.

eapache: fair

PR #584, opened 2015-12-11
Add a buffer around the broker's net.Conn for reads

benburkert:
Reduces the number of read syscalls issued by short, sequential reads on the broker's net.Conn. A similar technique is used in the [`net/http`](https://github.com/golang/go/blob/master/src/net/http/server.go#L482) package.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -99,7 +99,7 @@ type bufConn struct {
 	buf *bufio.Reader
 }
 
-func NewBufConn(conn net.Conn) *bufConn {
+func newBufConn(conn net.Conn) *bufConn {
 	return &bufConn{
 		Conn: conn,
 		buf:  bufio.NewReader(conn),
--- a/utils.go
+++ b/utils.go
@@ -1,6 +1,10 @@
 package sarama
 
-import "sort"
+import (
+	"bufio"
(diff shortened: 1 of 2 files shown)


ashishgandhi:
:+1:

eapache:
I am actually incredibly surprised to learn that the Go stdlib doesn't do this for you...

Do you have any benchmarks showing how much of an improvement this is?

benburkert:
I don't have anything scientific, but I did manually profile an app before & after this change and saw a reduction in time spent in `syscall.Syscall` from ~20% to ~10%.

utils.go
```diff
@@ -87,3 +91,21 @@ func (b ByteEncoder) Encode() ([]byte, error) {
...
 }
+
+// bufConn wraps a net.Conn with a buffer for reads to reduce the number of
+// reads that trigger syscalls.
+type bufConn struct {
+	net.Conn
+	buf *bufio.Reader
+}
+
+func NewBufConn(conn net.Conn) *bufConn {
```
eapache: method should not be public

wvanbergen:
:+1: from me if you un-export `NewBufConn`

PR #585, opened 2015-12-12
Fix offset_manager for Kafka 0.9.0.0

horkhe:
[Here](https://github.com/apache/kafka/blob/0.9.0/core/src/main/scala/kafka/coordinator/GroupCoordinator.scala#L402-L405) is why the solution works.

### DIFF
--- a/functional_offset_manager_test.go
+++ b/functional_offset_manager_test.go
@@ -1,16 +1,11 @@
 package sarama
 
 import (
-	"os"
 	"testing"
 )
 
 func TestFuncOffsetManager(t *testing.T) {
 	checkKafkaVersion(t, "0.8.2")
-	if os.Getenv("KAFKA_VERSION") == "0.9.0.0" {
-		t.Skip("Offset manager is broken with kafka 0.9 at the moment.")
-	}
-
 	setupFunctionalTest(t)
 	defer teardownFunctionalTest(t)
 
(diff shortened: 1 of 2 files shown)


functional_offset_manager_test.go
```diff
@@ -24,10 +19,6 @@ func TestFuncOffsetManager(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if _, err := offsetManager.ManagePartition("does_not_exist", 123); err != ErrUnknownTopicOrPartition {
-		t.Fatal("Expected ErrUnknownTopicOrPartition when starting a partition offset manager for a partition that does not exist, got:", err)
-	}
-
```
wvanbergen: Why did this have to be removed? Is it because the new group protocol is abstract, and potentially could work for other sources besides kafka partitions? (e.g. Kafka Connect). Or is there another reason?

wvanbergen:
Nice find :+1:

functional_offset_manager_test.go
eapache: You no longer get back `ErrUnknownTopicOrPartition`, you get back `ErrNoError` with an offset of `-1` (just like you do for topic/partitions that exist but haven't yet had an offset committed). I'm honestly not sure if this is a kafka bug, or an intended behaviour change.

eapache: @edenhill ^ ?

eapache:
Very nice find on setting the generation to -1.

horkhe:
I am having second thoughts about using `-1` directly. I should probably define a constant e.g. `groupGenerationUndefined`. Thoughts guys?

functional_offset_manager_test.go
horkhe: Yea, it is just not an error anymore. Whether it is a bug or behaviour change we cannot think of anything smarter than stop checking that.

wvanbergen: :+1: Given the differences in behavior between Kafka versions, removing this assertion is probably best.

wvanbergen:
:+1: for `groupGenerationUndefined` as constant. Other that than I am happy to merge this.

horkhe:
Added the constant and squashed.

PR #586, opened 2015-12-14
Publish GroupGenerationUndefined constant

horkhe:
In #585 I made a mistake declaring `groupGenerationUndefined` as module internal. It is a part of protocol and therefore should be public. It also belongs to **offset_commit_request.go** file along with `ReceiveTime` constant.


wvanbergen:
:+1:

PR #588, opened 2015-12-23
Consumer groups on Kafka 0.9

dim:
Hey, I am working on implementing consumer groups on Kafka 0.9 and here are the required API additions. I am not particularly sure about metadata, as Kafka doesn't prescribe but only recommend a structure for it. I followed the recommendations on https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-JoinGroupRequest, please let me know what you think.

Cheers,
Dim

### DIFF
--- a/config.go
+++ b/config.go
@@ -54,6 +54,20 @@ type Config struct {
 		RefreshFrequency time.Duration
 	}
 
+	// Group is the namespace for group management properties
+	Group struct {
+		Session struct {
+			// The allowed session timeout for registered consumers (defaults to 30s).
+			// Must be within the allowed server range.
+			Timeout time.Duration
+		}
+		Heartbeat struct {
+			// Interval between each heartbeat (defaults to 3s). It should be no more
+			// than 1/3rd of the Group.Session.Timout setting
+			Interval time.Duration
+		}
+	}
+
 	// Producer is the namespace for configuration related to producing messages,
 	// used by the Producer.
 	Producer struct {
@@ -211,6 +225,9 @@ func NewConfig() *Config {
 	c.Metadata.Retry.Max = 3
 	c.Metadata.Retry.Backoff = 250 * time.Millisecond
 	c.Metadata.RefreshFrequency = 10 * time.Minute
+
+	c.Group.Session.Timeout = 30 * time.Second
+	c.Group.Heartbeat.Interval = 3 * time.Second
 
 	c.Producer.MaxMessageBytes = 1000000
 	c.Producer.RequiredAcks = WaitForLocal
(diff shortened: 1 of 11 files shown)


config.go
```diff
@@ -54,6 +54,20 @@ type Config struct {
...
+	// Group is the namespace for group management properties
+	Group struct {
+		Session struct {
+			// The allowed session timeout for registered consumers (defaults to 30s).
+			// Must be within the allowed server range.
+			Timeout time.Duration
+		}
+		Heartbeat struct {
+			// Interval between each heartbeat (defaults to 3s). It should be no more
+			// than 1/3rd of the Group.Session.Timout setting
```
wvanbergen: We could check this in `Config.Validate()`

group_members.go
```diff
@@ -0,0 +1,98 @@
...
+	if m.UserData, err = pd.getBytes(); err != nil {
+		return
+	}
+
+	return nil
+}
+
+type GroupMemberAssignment struct {
+	Version  int16
+	Topics   []GroupMemberAssignedTopic
```
wvanbergen: `map[string][]int32` maybe?

wvanbergen:
Thanks for working on this :+1: 

As for the approach, I think what you are doing makes sense:

- We should follow the protocol's recommended structure for consumer groups.
- Some of the protocol identifiers have been renamed. Due to backwards compatibility reasons, we cannot really do these renames, unless we declare a 2.0 version (something that we are considering). For now, it's probably best to just build it with the current names; the functionality is not really different.
- Let's start with these building blocks, and tackle the higher level consumer construct in a separate PR when we land this.

dim:
@wvanbergen so yes, I am working on a high-level group consumer as a separate package for now. I have addressed your feedback above. Please let me know if there's anything else.

group_members.go
```diff
@@ -0,0 +1,94 @@
+package sarama
+
+type GroupMemberMetadata struct {
```
wvanbergen: As for naming, we should probably call this `ConsumerGroupMemberMetadata`. This is a specialization of the generic group type.

group_members.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	if m.UserData, err = pd.getBytes(); err != nil {
+		return
+	}
+
+	return nil
+}
+
+type GroupMemberAssignment struct {
```
wvanbergen: Similarly, `ConsumerGroupMemberAssignment`

We should probably also name the file `consumer_group_members.rb`

wvanbergen:
This looks pretty good to me. Let's see what @eapache thinks.

dim:
Hey, I have released a v2 branch of https://github.com/bsm/sarama-cluster/tree/v2 which requires this PR to be merged. I would really appreciate if you could do it next week. Feedback on the cluster consumer is also welcome. Thanks, dim

eapache:
I've been away, sorry. I'll take a look tomorrow.

eapache:
I've been away, sorry. I'll take a look tomorrow.

PR #589, opened 2016-01-02
Upstream has spoken on retriable errors

eapache:
Add InvalidMessage, and re-order the others to match their definition order.


wvanbergen:
When would a retry succeed? A topic that is in process of being created?

eapache:
For `ErrUnknownTopicOrPartition`? When you have auto-create turned on.

wvanbergen:
No, I mean `InvalidMessage`, which IIRC gets returned when you use an invalid topic name (e.g. with a slash in it.

eapache:
That's `InvalidTopic` - `InvalidMessage` indicates a bad CRC

wvanbergen:
D'oh. :+1: in that case

PR #592, opened 2016-01-06
Consumer rebalancing

wannes-ds:
We require the automatic consumer rebalancing throughout a new project using Kafka 0.9.

This PR contains tested code to support the automatic rebalancing in a consumer group using a round-robin partition assignment scheme. 

This PR introduces a `topicConsumer`, providing a `Messages` and `Error` channel which is published to by all the `partitionConsumer` to allow the client to consume from one single stream.

In addition, the client API is extended with a `RebalanceEvent` channel, which informs the client of partition consumer changes, inspired by the [Consumer Rebalance Listener](http://kafka.apache.org/090/javadoc/org/apache/kafka/clients/consumer/ConsumerRebalanceListener.html) in the Java API. The `SubscriptionOperation` constant specifies the type of operation (addition, removal).

Example:

>select {
		case msg := <-topicConsumer.Messages():
			log.Printf("Consumed message offset %d\n", msg.Offset)
			consumed++
		case event := <-rebalanceEvents:
			log.Printf("Performing rebalance with %d changes", len(event.Subscriptions))
		case <-signals:
			break ConsumerLoop
		}

As I'm new to golang and to sarama, I expect that some code will be out of place or already present somewhere else. If so, please point this out to me, I had a hard time figuring out al the different parts of the sarama consumer.

I'd appreciate all feedback towards making this production ready and merge-able.

Thanks!

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -33,7 +33,7 @@ type responsePromise struct {
 	errors        chan error
 }
 
-// NewBroker creates and returns a Broker targetting the given host:port address.
+// NewBroker creates and returns a Broker targeting the given host:port address.
 // This does not attempt to actually connect, you have to call Open() for that.
 func NewBroker(addr string) *Broker {
 	return &Broker{id: -1, addr: addr}
--- a/broker.go
+++ b/broker.go
@@ -191,7 +191,7 @@ func (b *Broker) JoinGroup(request *JoinGroupRequest) (*JoinGroupResponse, error
 
 	err := b.sendAndReceive(request, response)
 
-	if (err != nil) {
+	if err != nil {
 		return nil, err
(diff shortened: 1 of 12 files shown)


eapache:
I believe @dim was also working on this, you may want to pool your resources.

dim:
Hey, I have created it as a separate package with extensive tests, but it can be ultimately merged into Sarama I guess: https://github.com/bsm/sarama-cluster

wannes-ds:
Great, no need for this PR then. The only thing that seems missing in `sarama-cluster` is a `Consumer Rebalance Listener` alternative, I'll open an issue to discuss this.

Thanks!

PR #599, opened 2016-01-28
Update docs around TLS

eapache:
Remove the caveat from the configuration, and add the verification to the
CHANGELOG.

cc @wvanbergen


eapache:
@wvanbergen we need to make `errcheck` only install/run on Go 1.5, do you know how to do that in travis?

wvanbergen:
We can just fix up the command in `.travis.yml` based on the env variable I think.

wvanbergen:
Text changes: :+1:

PR #602, opened 2016-02-10
Add support for custom offset retention durations to offset manager

cevian:


eapache:
Thanks! I think this kind of value makes more sense in the `Consumer.Offsets` section of the configuration struct though.

cevian:
Ok changed. Can you take another look?

cevian:
Test this please

eapache:
I'm a human being who has to eat and sleep just like everyone else, please be patient.

offset_manager_test.go
```diff
@@ -228,6 +228,33 @@ func TestPartitionOffsetManagerMarkOffset(t *testing.T) {
 	coordinator.Close()
 }
 
+func TestPartitionOffsetManagerMarkOffsetWithRetention(t *testing.T) {
```
eapache: This test doesn't actually test that the value in the request is set correctly. It should do that, or it should be removed.

cevian:
Sorry that was meant for Travis-ci actually :). The check failed and I thought I could get travis-ci to re-test it with "Test this please" (That is how you force jenkins to re-check, thought Travis-ci is the same.). Didn't mean that to rush you.

config.go
```diff
@@ -183,6 +183,10 @@ type Config struct {
 			// The initial offset to use if no offset was previously committed.
 			// Should be OffsetNewest or OffsetOldest. Defaults to OffsetNewest.
 			Initial int64
+
+			// The retention duration for committed offsets.
+			// A value of 0 cause the retention time to default to the offsets.retention.minutes broker option.
```
eapache: Please follow the style of other option fields:
- keep lines short enough to display properly on godoc.org
- explicitly list the default value
- mention that Kafka only supports precision up to milliseconds; nanoseconds will be truncated.
- add a warning to the `Validate` method if greater-than-millisecond precision is provided

eapache:
Oh, heh I had no idea Jenkins worked like that, Travis doesn't :)

I think you can trigger a rebuild from the Travis page that github links to.

offset_manager_test.go
cevian: Fixed. Is the sleep in here ok?

eapache: Instead of a sleep and a `coordinator.Returns` you should be able to call `coordinator.setHandler` and validate the request values in the callback.

cevian: Ok changed, that does look a lot better.

eapache:
Thanks for your work!

PR #604, opened 2016-02-14
Adjust broker locking order slightly

eapache:
I was doing an unrelated experiment with a config that generated a warning on
validation, and I ended up flooded with warnings because we now call `Open` on
every broker returned from the Client, and we were revalidating the config every
single time. One thing led to another and I made several other adjustments in
the same vein.

- Move the atomic `AlreadyConnected` check to before the config validation in
  `Broker.Open`. Config validation is expensive and can log, no point in doing
  it if we don't have to.
- Move all the `atomic.StoreInt32` calls to *after* any of their related log
  messages. None of these were real problems, but it removes any possibility of
  out-of-order log messages due to unusual race conditions.
- Remove the double-check for `AlreadyConnected` after we've taken the full
  lock. Initially I just removed the log message since it's not something we
  need to log, but upon closer inspection I'm pretty sure this is dead code;
  anything that sets `opened` to 0 also sets `conn` to nil *first*, and does so
  while holding the lock.

@wvanbergen

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -45,6 +45,10 @@ func NewBroker(addr string) *Broker {
 // follow it by a call to Connected(). The only errors Open will return directly are ConfigurationError or
 // AlreadyConnected. If conf is nil, the result of NewConfig() is used.
 func (b *Broker) Open(conf *Config) error {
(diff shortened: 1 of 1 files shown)


wvanbergen:
:+1:

PR #616, opened 2016-03-03
Add additional error handling based on spec

### LINKED ISSUE #596: Audit error code handling
https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol

Thanks to @gwenshap we now have a documented list of exactly which error codes to expect from which API endpoints. We should audit our client, producer, and consumer code to ensure we haven't missed handling any.

eapache:
Gwen posted to the spec a list of expected error codes for each
request/response. This adds handling for the listed codes for consumer, client,
and offset manager. Some errors (e.g. around group membership) have been ignored
because we don't implement the relevant feature yet anyways.
Possible errors for the producer are not yet listed in the spec.

@wvanbergen closes #596

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -642,7 +642,7 @@ func (bc *brokerConsumer) handleResponses() {
 			Logger.Printf("consumer/%s/%d shutting down because %s\n", child.topic, child.partition, result)
 			close(child.trigger)
 			delete(bc.subscriptions, child)
-		case ErrUnknownTopicOrPartition, ErrNotLeaderForPartition, ErrLeaderNotAvailable:
+		case ErrUnknownTopicOrPartition, ErrNotLeaderForPartition, ErrLeaderNotAvailable, ErrReplicaNotAvailable:
 			// not an error, but does need redispatching
 			Logger.Printf("consumer/broker/%d abandoned subscription to %s/%d because %s\n",
 				bc.broker.ID(), child.topic, child.partition, result)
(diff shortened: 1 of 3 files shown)


consumer.go
```diff
@@ -642,7 +642,7 @@ func (bc *brokerConsumer) handleResponses() {
 			Logger.Printf("consumer/%s/%d shutting down because %s\n", child.topic, child.partition, result)
 			close(child.trigger)
 			delete(bc.subscriptions, child)
-		case ErrUnknownTopicOrPartition, ErrNotLeaderForPartition, ErrLeaderNotAvailable:
+		case ErrUnknownTopicOrPartition, ErrNotLeaderForPartition, ErrLeaderNotAvailable, ErrReplicaNotAvailable:
```
wvanbergen: I remember us removing handling `ErrReplicaNotAvailable` because we usually are only interested in the leader for our purposes. Is that not the case here?

wvanbergen:
Just one question, but :+1: for improving this.

consumer.go
eapache: Ya, that's true I think? It's not super-clear from the protocol spec what to do with `ReplicaNotAvailable`. It says of the error "this can be safely ignored" but also marks it as not-retriable, so... treat it like success? Hard fail? @edenhill could you clarify please?

eapache: The only change here in this PR is whether or not we send it out on the error channel (if enabled) before redispatching. If we really do want to ignore it in some other way (treat it as success?) then we have never done that to date and nobody's complained...

wvanbergen: :+1:

PR #621, opened 2016-03-07
Put "no consumergroups" in the docs more times

eapache:
We get a lot of questions on this still.

@wvanbergen


wvanbergen:
:+1:

PR #625, opened 2016-03-10
fix two un-checked accesses to a map in producer.go (missing lock acquisition)

rsrsps:
there are two unchecked accesses to the brokerProducers map in producers.go:

1. in .Close() the range function relies on being able to iterate the map as a reader even though other operations adding to/deleting from the map may occur while this is happening

2. goroutine spawned in newBrokerProducer() can eventually mutate the brokerproducers map

the gratuitous change to the deliveryLocks is to prove that there are not also unchecked accesses there.


rsrsps:
Oh geeze.  Github UI seems to have lured me to the wrong repo.  Thanks, Evan.

eapache:
No problem, although if you really are still using that ancient version of the producer you'll have a much better time if you just upgrade :)

PR #627, opened 2016-03-10
Don't import `testing` in the main lib

### LINKED ISSUE #626: package sarama imports testing
Some files in package sarama import the go stdlib "testing" package.  The most annoying side effect is that the global flags get polluted with all the -test.X ones from package testing in all binaries that link in github.com/Shopify/sarama.   

Package testing should really only be included by tests, not by the main part of any library.

eapache:
This was a minor regression from #570. Fixes #626.

@lvdlvd @wvanbergen

### DIFF
--- a/mockbroker.go
+++ b/mockbroker.go
@@ -9,7 +9,6 @@ import (
 	"reflect"
 	"strconv"
 	"sync"
-	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


lvdlvd:
Very nice. very fast response. thanks a bundle.

2016-03-10 22:40 GMT+01:00 Evan Huus <notifications@github.com>:

PR #633, opened 2016-03-17
Validate client ID

### LINKED ISSUE #632: Check config string properties for invalid characters
I filed a bug in the [Kafka JIRA repo](https://issues.apache.org/jira/browse/KAFKA-3417), but it appears that the invalid characters are not being checked for and are causing unexpected downstream errors. When I initially found this using sarama, the client just hung even though I saw an error in the Kafka logs. I am not sure if Kafka is not returning a response or the failure is being ignored.

With a fresh Kafka setup, create the topic and publish a message.

```bash
./kafka-topics --create --topic foo --zookeeper 192.168.99.100:2181 --partitions 1 --replication-factor 1
./kafka-console-producer --topic foo --broker-list 192.168.99.100:9092
hello world
```

Run the code to get the message, an error, or timeout.

```go
package main

import (
	"log"
	"time"

	"github.com/Shopify/sarama"
)

func main() {
	conf := sarama.NewConfig()
	conf.ClientID = "foo:bar"

	con, err := sarama.NewConsumer([]string{"192.168.99.100:9092"}, conf)
	if err != nil {
		log.Fatal(err)
	}
	defer con.Close()

	part, err := con.ConsumePartition("foo", 0, sarama.OffsetOldest)
	if err != nil {
		log.Fatal(err)
	}
	defer part.Close()

	select {
	case <-time.After(time.Second * 2):
		log.Println("timeout")

	case <-part.Messages():
		log.Println("ok")

	case err := <-part.Errors():
		log.Println("error: ", err)

	}
}
```

eapache:
Fixes #632.

@wvanbergen @bruth

### DIFF
--- a/config.go
+++ b/config.go
@@ -2,8 +2,11 @@ package sarama
 
 import (
 	"crypto/tls"
+	"regexp"
 	"time"
 )
+
+var validID *regexp.Regexp = regexp.MustCompile(`\A[A-Za-z0-9._-]*\z`)
 
 // Config is used to pass multiple configuration options to Sarama's constructors.
 type Config struct {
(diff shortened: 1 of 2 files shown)


wvanbergen:
:+1:

PR #641, opened 2016-04-13
Drop go 1.4 and update tools/docs appropriately

eapache:
@wvanbergen after fighting to get `go vet` working again on 1.4 I realized we were only about a week away from our 2 month guarantee. I'll merge this next week some time.


wvanbergen:
:+1:

wvanbergen:
🎉

eapache:
Turns out @Sirupsen just released the new toxiproxy though which means merging this broke master because the API changed (at least I think that's the reason). It's a test-only breakage at least.

Sirupsen:
@eapache Is this because you're using the Go API from master?

eapache:
Ya I think so; we're installing 1.0.3 from tarball on CI, then using the master go API to talk to it. Is this as simple as switching to the 2.0 tarball?

Sirupsen:
There may have been some minor changes to the Go client API, but I don't remember (Jake and Jacob did most of the work for 2.0)—it's probably about as simple as that, with maybe a couple of small exceptions.

PR #648, opened 2016-04-28
Add support for SASL plain text authentication

rshriram:
Hi
 This is a simple patch that adds support for SASL plain text authentication to Kafka brokers, if enabled. This feature is useful when people want to leverage hosted Kafka services such as IBM Bluemix's Message Hub service (Kafka 0.9 with SASL plain text authentication). By default, the authentication is disabled.

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -492,18 +492,27 @@ func (b *Broker) doSASLPlainAuth() error {
 	binary.BigEndian.PutUint32(authBytes, uint32(length))
 	copy(authBytes[4:], []byte("\x00"+b.conf.Net.SASL.User+"\x00"+b.conf.Net.SASL.Password))
 
-	b.conn.SetWriteDeadline(time.Now().Add(b.conf.Net.WriteTimeout))
-	b.conn.Write(authBytes)
+	err := b.conn.SetWriteDeadline(time.Now().Add(b.conf.Net.WriteTimeout))
+	if err != nil {
+		Logger.Printf("Failed to set write deadline when doing SASL auth: %s\n", err.Error())
+		return nil
+	}
+
+	_, err = b.conn.Write(authBytes)
+	if err != nil {
+		Logger.Printf("Failed to write SASL auth header to broker: %s\n", err.Error())
+		return nil
+	}
 
 	header := make([]byte, 4)
 	n, err := io.ReadFull(b.conn, header)
 	// If the credentials are valid, we would get a 4 byte response filled with null characters.
 	// Otherwise, the broker closes the connection and we get an EOF
 	if err != nil {
-		Logger.Printf("Failed to read response while authenticating with SASL: %s", err.Error())
+		Logger.Printf("Failed to read response while authenticating with SASL: %s\n", err.Error())
 		return err
 	}
 
-	Logger.Printf("SASL authentication successful:%v - %v", n, header)
+	Logger.Printf("SASL authentication successful:%v - %v\n", n, header)
 	return nil
 }
--- a/broker.go
+++ b/broker.go
@@ -9,6 +9,8 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+	"bytes"
+	"encoding/binary"
 )
 
 // Broker represents a single Kafka broker connection. All operations on this object are entirely concurrency-safe.
@@ -82,6 +84,38 @@ func (b *Broker) Open(conf *Config) error {
 		b.conn = newBufConn(b.conn)
 
 		b.conf = conf
+
+		if conf.Net.SASL.Enable {
+			//
+			// Begin SASL/PLAIN authentication
+			//
+			authBytes := []byte("\x00" + b.conf.Net.SASL.User + "\x00" + b.conf.Net.SASL.Password)
+			buf := new(bytes.Buffer)
+
+			err = binary.Write(buf, binary.BigEndian, int32(len(authBytes)))
+			if err != nil {
+				Logger.Printf("Failed to encode payload size (SASL credentials): %s", err.Error())
+			}
+
+			err = binary.Write(buf, binary.BigEndian, authBytes)
+			if err != nil {
+				Logger.Printf("Failed to encode payload (SASL credentials): %s", err.Error())
+			}
+
+			b.conn.SetWriteDeadline(time.Now().Add(b.conf.Net.WriteTimeout))
+			b.conn.Write(buf.Bytes())
+
+			header := make([]byte, 4)
+			n, err := io.ReadFull(b.conn, header)
+			if err != nil {
+				Logger.Printf("Failed to read response while authenticating with SASL: %s", err.Error())
+			}
+			Logger.Printf("SASL authentication successful:\n%v\n%v\n%v", n, header, string(header))
+			//
+			// End SASL/PLAIN authentication
+			//
+		}
+
 		b.done = make(chan bool)
 		b.responses = make(chan responsePromise, b.conf.Net.MaxOpenRequests-1)
 
--- a/broker.go
+++ b/broker.go
@@ -1,7 +1,6 @@
 package sarama
 
 import (
-	"bytes"
 	"crypto/tls"
 	"encoding/binary"
 	"fmt"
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -222,6 +233,7 @@ func NewConfig() *Config {
 	c.Net.DialTimeout = 30 * time.Second
 	c.Net.ReadTimeout = 30 * time.Second
 	c.Net.WriteTimeout = 30 * time.Second
+	c.Net.SASL.Enable = false
```
eapache: you don't need to set this, `false` is the default for go booleans

config.go
```diff
@@ -256,6 +268,14 @@ func (c *Config) Validate() error {
 	if c.Net.TLS.Enable == false && c.Net.TLS.Config != nil {
 		Logger.Println("Net.TLS is disabled but a non-nil configuration was provided.")
 	}
+	if c.Net.SASL.Enable == false {
+		if c.Net.SASL.User != "" {
+			Logger.Println("Net.SASL is disabled but a non-empty username was provided.")
```
eapache: these are good checks; I don't know a lot about SASL, but is it valid to have empty usernames and passwords? If not you should do the opposite check as well - if it's enabled but the user/pass is blank return a configuration error

broker.go
```diff
@@ -82,6 +84,38 @@ func (b *Broker) Open(conf *Config) error {
 		b.conn = newBufConn(b.conn)
 
 		b.conf = conf
+
+		if conf.Net.SASL.Enable {
+			//
+			// Begin SASL/PLAIN authentication
+			//
+			authBytes := []byte("\x00" + b.conf.Net.SASL.User + "\x00" + b.conf.Net.SASL.Password)
```
eapache: could this block of SASL code be moved to a method please to keep `Open` fairly straightforward?

broker.go
```diff
@@ -82,6 +84,38 @@ func (b *Broker) Open(conf *Config) error {
...
+			buf := new(bytes.Buffer)
```
eapache: you know how many bytes you're going to need to write, there's no need to use a dynamic buffer here?

broker.go
```diff
@@ -82,6 +84,38 @@ func (b *Broker) Open(conf *Config) error {
...
+
+			b.conn.SetWriteDeadline(time.Now().Add(b.conf.Net.WriteTimeout))
+			b.conn.Write(buf.Bytes())
+
+			header := make([]byte, 4)
+			n, err := io.ReadFull(b.conn, header)
+			if err != nil {
+				Logger.Printf("Failed to read response while authenticating with SASL: %s", err.Error())
+			}
+			Logger.Printf("SASL authentication successful:\n%v\n%v\n%v", n, header, string(header))
```
eapache: I don't know a whole lot about SASL but don't you need to actually *check* the response header in some way before saying it succeeded?

config.go
```diff
@@ -33,6 +33,17 @@ type Config struct {
 			Config *tls.Config
 		}
 
+		// SASL based authentication with broker. While there are multiple SASL authentication methods
+		// the current implementation is limited to plaintext (SASL/PLAIN) authentication
```
eapache: since we currently only support plain-text, should we enforce that SASL is only used over HTTPS to avoid leaking credentials on the wire?

rshriram: Thanks for the comments.
With regard to enforcing use of TLS (not https) for SASL, I thought I would
keep the auth separate from the encryption to give people a choice should
the opt to use non tls transports with password authentication (as a
framework, it seemed harsh to enforce policies/preferences on the user).

Will address your other comments and send out a new PR soon.

On Thu, Apr 28, 2016 at 3:46 PM Evan Huus <notifications@github.com> wrote:

> In config.go
> <https://github.com/Shopify/sarama/pull/648#discussion_r61490796>:
>
> > @@ -33,6 +33,17 @@ type Config struct {
> >  			Config *tls.Config
> >  		}
> >
> > +		// SASL based authentication with broker. While there are multiple SASL authentication methods
> > +		// the current implementation is limited to plaintext (SASL/PLAIN) authentication
>
> since we currently only support plain-text, should we enforce that SASL is
> only used over HTTPS to avoid leaking credentials on the wire?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/Shopify/sarama/pull/648/files/3834ba1ce06295021c65e824e3aa9337caf5f03f#r61490796>
>
-- 

~shriram

broker.go
rshriram: I admit that I am a bit new to Golang. Can you clarify what you mean by this comment?

eapache: creating a new `bytes.Buffer` creates a growable dynamic buffer that resizes itself as you add more data. In this case it is unnecessary work, since you already know exactly how many bytes the message will be, so you can allocate the precise size up-front.

rshriram:
Hi, I have addressed all of your comments. See if this version of PR is good enough.

broker.go
```diff
@@ -82,6 +83,18 @@ func (b *Broker) Open(conf *Config) error {
 		b.conn = newBufConn(b.conn)
 
 		b.conf = conf
+
+		if conf.Net.SASL.Enable {
+			err = b.doSASLPlainAuth()
+			if err != nil {
+				Logger.Printf("SASL authentication with broker %s failed\n", b.addr)
```
eapache: you log every case in `doSASLPlainAuth` *and* both cases here, that seems a bit much?

broker.go
```diff
@@ -454,3 +467,52 @@ func (b *Broker) responseReceiver() {
...
+func (b *Broker) doSASLPlainAuth() error {
+	length := 1 + len(b.conf.Net.SASL.User) + 1 + len(b.conf.Net.SASL.Password)
+	authBytes := make([]byte, length+4) //4 byte length header + auth data
+	binary.BigEndian.PutUint32(authBytes, uint32(length))
+	copy(authBytes[4:], []byte("\x00"+b.conf.Net.SASL.User+"\x00"+b.conf.Net.SASL.Password))
+
+	err := b.conn.SetWriteDeadline(time.Now().Add(b.conf.Net.WriteTimeout))
+	if err != nil {
+		Logger.Printf("Failed to set write deadline when doing SASL auth: %s\n", err.Error())
+		return nil
```
eapache: should we really be returning `nil` here (and in the case we failed to write)? the connection will potentially continue unauthenticated in that case?

config.go
```diff
@@ -294,6 +313,10 @@ func (c *Config) Validate() error {
 		return ConfigurationError("Net.WriteTimeout must be > 0")
 	case c.Net.KeepAlive < 0:
 		return ConfigurationError("Net.KeepAlive must be >= 0")
+	case (c.Net.SASL.Enable == true) && (c.Net.SASL.User == ""):
```
eapache: nit: I think generally these parentheses are not needed?

broker.go
```diff
@@ -82,6 +83,18 @@ func (b *Broker) Open(conf *Config) error {
 		b.conn = newBufConn(b.conn)
 
 		b.conf = conf
+
+		if conf.Net.SASL.Enable {
+			err = b.doSASLPlainAuth()
+			if err != nil {
+				Logger.Printf("SASL authentication with broker %s failed\n", b.addr)
+				_ = b.Close()
```
eapache: go's mutexes are not re-entrant, so this will deadlock; to handle the error please use `connErr` and follow the pattern established a few lines up when we fail to connect at all; note you will still need to close `b.conn` since it has now been established

broker.go
rshriram: I am sorry. There seems to be some mess up from my older version of code and the newer version. This was intended to return ```err``` .

rshriram:
Any pointers as to why the build fails with Go 1.6.1 ? The log seems to go on for ever.

eapache:
It's just a test that occasionally gets stuck; I've restarted it and it should pass this time.

eapache:
This looks OK to me now I think; @wvanbergen ?

broker.go
```diff
@@ -454,3 +471,52 @@ func (b *Broker) responseReceiver() {
...
+//   passwd    = 1*SAFE ; MUST accept up to 255 octets
+//   UTF8NUL   = %x00 ; UTF-8 encoded NUL character
+//
+//   SAFE      = UTF1 / UTF2 / UTF3 / UTF4
+//                  ;; any UTF-8 encoded Unicode character except NUL
+//
+// When credentials are valid, Kafka returns a 4 byte array of null characters.
+// When credentials are invalid, Kafka closes the connection. This does not seem to be the ideal way
+// of responding to bad credentials but thats how its being done today.
+func (b *Broker) doSASLPlainAuth() error {
```
wvanbergen: as name, I prefer `sendAndReceiveSASLPlainAuth` which is more in line with naming we do elsewhere

broker.go
```diff
@@ -454,3 +471,52 @@ func (b *Broker) responseReceiver() {
...
+	// If the credentials are valid, we would get a 4 byte response filled with null characters.
+	// Otherwise, the broker closes the connection and we get an EOF
+	if err != nil {
+		Logger.Printf("Failed to read response while authenticating with SASL to broker %s: %s\n", b.addr, err.Error())
+		return err
+	}
+
+	Logger.Printf("SASL authentication successful with broker %s:%v - %v\n", b.addr, n, header)
+	return nil
+}
```
wvanbergen: I suppose this is too different from the normal API packets we send over the wire in order to reuse our main `sendAndReceive` method.

rshriram: yes. This authentication preceeds the normal kafka message exchanges and follows a different format (no correlation IDs, etc.). Besides, cant use the send function, as it would deadlock

broker.go
rshriram: fixed

rshriram:
Hi
 Would you like any more changes before merging this PR?

wvanbergen:
I am ok with merging this :+1:. @eapache I will let you do the honours

eapache:
👍  thanks

Issue #650, opened 2016-04-29
Consumer allocates a lot of memory when Consumer.Offsets.Retention has non-zero value


dimchansky:
Marking offset (calling `MarkOffset` method on partition offset manager) in Kafka 0.8.2 when `Consumer.Offsets.Retention` has non-zero value leads to system halt because application eventually uses all available memory. There are a lot of messages in log like:
```
Non-zero timestamp specified for OffsetCommitRequest not v1, it will be ignored
kafka: error while consuming topic_name/0: EOF
```
sarama version: https://github.com/Shopify/sarama/commit/68920ff1be3bbd31449ab1e2c255c50cedeb755c
Kafka version: 0.8.2
Go version: 1.6.2 windows/amd64

eapache:
> Kafka 0.8.2 when `Consumer.Offsets.Retention` has non-zero value

We should document that this is not a valid configuration. Support for `Consumer.Offsets.Retention` was only added in Kafka 0.9. If you check your broker logs I suspect you will see them complaining about an unrecognized message format or something.

> application eventually uses all available memory

Even given the above misconfiguration, I don't see any reason for it to use all available memory. Can you provide more complete logs when this happens?

> Non-zero timestamp specified for OffsetCommitRequest not v1, it will be ignored

This indicates a very minor (totally harmless) issue which is not related to anything else, I'll fix it too.

eapache:
If you can provide the needed information, I am happy to reopen.

PR #651, opened 2016-04-29
Don't try to send partition timestamps in v2 OCRs

eapache:
Only v1 OffsetCommitRequest messages had the per-partition timestamp, so only
send it in that case and let it be 0 otherwise. The `encode` method was already
smart enough to ignore it and log a warning, so this basically just suppresses
the annoying log message.

@wvanbergen @dimchansky fixes the less interesting piece of #650.


wvanbergen:
:+1:

PR #652, opened 2016-04-29
Doc that Offset.Retention requires kafka 0.9

eapache:
It requires v2 of the OffsetCommitRequest which was only introduced in Kafka
0.9; trying to use it with earlier broker versions will fail.

From #650.

@wvanbergen @dimchansky


wvanbergen:
:+1:

PR #655, opened 2016-05-05
Added multi version kafka support

mrafayaleem:
@eapache @wvanbergen This is the PR I have been working on to address https://github.com/Shopify/sarama/issues/617. Can you guys start reviewing it please. I am originally from Python background so please ignore any coding horrors here.

Also checkout how different response versions are handled in `kafka-python` [here](https://github.com/dpkp/kafka-python/blob/master/kafka/protocol/fetch.py). For this PR, I am passing kafka version from request to response so that response knows how to decode itself. Also, please do leave comments on the way kafka version is assigned in test cases. It doesn't look right IMO.

Description:
- Tries to address https://github.com/Shopify/sarama/issues/617

Not implemented/missing/confused:
- Handling offsets to kafka vs zk.
- Config conflicts.

Issues:
- I am aware of at least five test cases that are failing on this branch. These are mostly `Decoded request does not match the encoded one`. Any pointers would be highly appreciated. Failing tests:
 - TestOffsetCommitRequestV0
 - TestOffsetCommitRequestV1
 - TestOffsetFetchRequest
 - TestPartitionOffsetManagerMarkOffsetWithRetention
 - TestProduceRequest

Gotchas:

I figured out that the BNF for FetchResponse v1 is actually 
```
v1 (supported in 0.9.0 or later) and v2 (supported in 0.10.0 or later)
FetchResponse => ThrottleTime [TopicName [Partition ErrorCode HighwaterMarkOffset MessageSetSize MessageSet]]
  ThrottleTime => int32
  TopicName => string
  Partition => int32
  ErrorCode => int16
  HighwaterMarkOffset => int64
  MessageSetSize => int32
```
instead of
```
v1 (supported in 0.9.0 or later) and v2 (supported in 0.10.0 or later)
FetchResponse => [TopicName [Partition ErrorCode HighwaterMarkOffset MessageSetSize MessageSet]] ThrottleTime
  TopicName => string
  Partition => int32
  ErrorCode => int16
  HighwaterMarkOffset => int64
  MessageSetSize => int32
  ThrottleTime => int32
```
I think there is an error in the kafka protocol wiki [here](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol).

### DIFF
--- a/config.go
+++ b/config.go
@@ -8,8 +8,16 @@ import (
 
 var validID *regexp.Regexp = regexp.MustCompile(`\A[A-Za-z0-9._-]*\z`)
 
+type KafkaVersion struct {
+	Release *kVersion
+}
+
 // Config is used to pass multiple configuration options to Sarama's constructors.
 type Config struct {
+
+	// This defines the Kafka broker version and is used for broker version specific behaviour
+	KafkaVersion KafkaVersion
+
 	// Net is the namespace for network-level properties used by the Broker, and
 	// shared by the Client/Producer/Consumer.
 	Net struct {
@@ -219,6 +227,8 @@ type Config struct {
 func NewConfig() *Config {
 	c := &Config{}
 
+	c.KafkaVersion.Release = V0_8_2_2
+
 	c.Net.MaxOpenRequests = 5
 	c.Net.DialTimeout = 30 * time.Second
 	c.Net.ReadTimeout = 30 * time.Second
@@ -362,4 +372,8 @@ func (c *Config) Validate() error {
 	}
 
 	return nil
+}
+
+func (v *KafkaVersion) AtLeast(ver *kVersion) bool {
+	return v.Release.GE(ver)
 }
--- a/config.go
+++ b/config.go
@@ -8,15 +8,11 @@ import (
 
 var validID *regexp.Regexp = regexp.MustCompile(`\A[A-Za-z0-9._-]*\z`)
 
-type KafkaVersion struct {
-	Release *kVersion
-}
-
 // Config is used to pass multiple configuration options to Sarama's constructors.
 type Config struct {
 
 	// This defines the Kafka broker version and is used for broker version specific behaviour
-	KafkaVersion KafkaVersion
+	KafkaVersion *KafkaVersion
 
 	// Net is the namespace for network-level properties used by the Broker, and
 	// shared by the Client/Producer/Consumer.
@@ -227,7 +223,7 @@ type Config struct {
 func NewConfig() *Config {
 	c := &Config{}
 
-	c.KafkaVersion.Release = V0_8_2_2
+	c.KafkaVersion = V0_8_2_2
 
 	c.Net.MaxOpenRequests = 5
 	c.Net.DialTimeout = 30 * time.Second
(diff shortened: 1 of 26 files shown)


eapache:
> I think there is an error in the kafka protocol wiki

That agrees with https://kafka.apache.org/protocol.html#protocol_messages, so I will update the wiki.

eapache:
The requests and responses should *not* store the actual kafka version (0.8.2 or 0.9 or whatever); they should store the actual API version number specific to them (see e.g. what `OffsetFetchRequest` already does on master). This is because different API versions may have substantially different behaviours (offsets to kafka vs zk as you point out) and so even if I'm using Kafka 0.9 I may still want to specify a different version of the message. We need to support this use case. This will also *massively* simplify the definition of `allocateBody`.

The *kafka* version should be stored either as a global (next to `MaxRequestSize` and `MaxResponseSize`) or in the `Config` struct. I'd prefer the config struct, but that would involve passing the config around in a lot more places. We should really do that anyways at some point, since `MaxRequestSize` and `MaxResponseSize` should not be global either, but that's a much larger chunk of work so I'm ok with another global for now.

config.go
```diff
@@ -8,8 +8,16 @@ import (
 
 var validID *regexp.Regexp = regexp.MustCompile(`\A[A-Za-z0-9._-]*\z`)
 
+type KafkaVersion struct {
```
eapache: having both `kVersion` and `KafkaVersion` as structs seems like an unnecessary complication; can't you just make one type out of them?

kafka_versions.go
```diff
@@ -0,0 +1,10 @@
+package sarama
+
+var LatestStable, _ = NewVersion("0.9.0.1")
+var V0_10_0, _ = NewVersion("0.10.0.0")
+var V0_9_0_1 = LatestStable
```
eapache: I'd prefer defining this with a literal `var V0_9_0_1 = NewVersion("0.9.0.1")` and then define the latest stable at the bottom with `var LatestStable = V0_9_0_1`.

utils.go
```diff
@@ -109,3 +111,33 @@ func newBufConn(conn net.Conn) *bufConn {
 func (bc *bufConn) Read(b []byte) (n int, err error) {
 	return bc.buf.Read(b)
 }
+
+type kVersion struct {
+	versionArr [4]int
+}
+
+func NewVersion(ver string) (*kVersion, error) {
```
eapache: there is no reason for this to take a string or return an error `NewVersion(0, 9, 0, 1)`  reads cleaner

mrafayaleem:
I actually removed version number from `OffsetFetchRequest ` but now I see what you mean here and I agree. So essentially, that means that version number should be implemented for every request/response similar to the one for `OffsetFetchRequest` on master, right? And I don't think that there is anyway we can do away without implementing any version on responses because this is how we would be able to decode them properly.

Right now, Kafka version has been made part of the config in this PR. Since our discussion started with this, why not go ahead with it rather than implementing it as global. Test cases would need to be modified  though which I have already partially  hacked here in an ugly way.

utils.go
mrafayaleem: 👍

eapache:
Superseded mostly by #676 (based on this code but simplified) and other work that is ongoing. Thanks for your input, it was helpful in solving this!

mrafayaleem:
Glad to be of help. Cheers!

PR #657, opened 2016-05-06
Fix race condition on OffsetManager shutdown

### LINKED ISSUE #656: OffsetManager.Close() race condition
When closing OffsetManager, it can hang waiting for state to become "clean". 

The Close method uses AsyncClose internally. At L360, AsyncClose waits for a message on the ```clean``` chan before closing the ```dying``` chan:
```
		pom.lock.Lock()
		dirty := pom.dirty
		pom.lock.Unlock()

		if dirty {
			<-pom.clean
		}

		close(pom.dying)
```

However, ```updateCommitted``` can execute between ```AsyncClose()```'s ```pom.lock.Unlock()``` and receive on the ```clean``` chan:
```
	pom.lock.Lock()
	defer pom.lock.Unlock()
...
		select {
		case pom.clean <- none{}:
		default:
		}
```

The ```Close()``` method drains ```pom.errors``` at L371. This channel does not close until ```pom.dying``` is closed, which is selected in the mainLoop at L215

eapache:
If `updateCommitted` ran between the `Unlock` call and the read from `clean`
then we might miss the transition message and hang. Instead, read
unconditionally from `clean` and launch a second goroutine that *writes* a
message to it if we were already clean.

The additional local `cancel` is necessary to avoid leaking the second (writing)
goroutine in the event that the main shutdown goroutine picks up a real
transitional event before our extra writer even runs; without it the writer
could block writing forever.

@wvanbergen @hasnickl fixes #656

### DIFF
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -351,16 +351,22 @@ func (pom *partitionOffsetManager) NextOffset() (int64, string) {
 }
 
 func (pom *partitionOffsetManager) AsyncClose() {
+	cancel := make(chan none)
 	go func() {
 		pom.lock.Lock()
-		dirty := pom.dirty
-		pom.lock.Unlock()
-
-		if dirty {
-			<-pom.clean
+		defer pom.lock.Unlock()
+		if !pom.dirty {
+			select {
+			case pom.clean <- none{}:
+			case <-cancel:
+			}
 		}
+	}()
 
+	go func() {
+		<-pom.clean
 		close(pom.dying)
+		close(cancel)
 	}()
 }
 


hasnickl:
If the first ```go func``` runs while ```pom``` is dirty, ```updateCommitted``` could run and finish before the second ```go func``` receives on ```clean```

A possible solution is to have an internal flag like ```pom.shutdown```. ```AsyncClose``` can lock pom, set shutdown to true, start the go routine to read from ```clean``` and unlock.

```updateCommitted``` could then change the select with default to conditionally start a go routine:

```
if pom.shutdown {
  go func(){
    pom.clean <- none{}
  }
}
```

perhaps ```MarkOffset``` would also check this flag to prevent state going to dirty again while shutting down.

Any thoughts on this?

eapache:
> If the first go func runs while pom is dirty, updateCommitted could run and finish before the second go func receives on clean

You're right, I must have been asleep when I cooked up this solution.

PR #658, opened 2016-05-06
Fix race condition on OffsetManager shutdown

### LINKED ISSUE #656: OffsetManager.Close() race condition

eapache:
If `updateCommitted` ran between the `Unlock` call and the read from `clean`
then we might miss the transition message and hang. Instead, use a condition
value from the `sync` package which does the right thing.

@wvanbergen @hasnickl fixes #656 

Supersedes #657

### DIFF
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -176,7 +176,7 @@ type partitionOffsetManager struct {
 	offset   int64
 	metadata string
 	dirty    bool
-	clean    chan none
+	clean    sync.Cond
 	broker   *brokerOffsetManager
 
 	errors    chan *ConsumerError
@@ -189,11 +189,11 @@ func (om *offsetManager) newPartitionOffsetManager(topic string, partition int32
 		parent:    om,
 		topic:     topic,
 		partition: partition,
-		clean:     make(chan none),
 		errors:    make(chan *ConsumerError, om.conf.ChannelBufferSize),
 		rebalance: make(chan none, 1),
 		dying:     make(chan none),
 	}
+	pom.clean.L = &pom.lock
 
 	if err := pom.selectBroker(); err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


offset_manager.go
```diff
@@ -353,11 +349,10 @@ func (pom *partitionOffsetManager) NextOffset() (int64, string) {
...
 	go func() {
 		pom.lock.Lock()
-		dirty := pom.dirty
-		pom.lock.Unlock()
+		defer pom.lock.Unlock()
 
-		if dirty {
-			<-pom.clean
+		for pom.dirty {
+			pom.clean.Wait()
```
hasnickl: I haven't seen ```sync.Cond``` used before. This looks like a pretty clean solution!

This seems to solve the issue with hanging, but could ```pom.dirty``` change to true after this goroutine finishes? I suppose this would only happen if ```MarkOffset``` is called after a ```Close```, is it worth worrying about?

hasnickl:
lgtm

offset_manager.go
eapache: calling anything after a `Close` or `AsyncClose` is user-error, not worth worrying about

PR #664, opened 2016-06-01
Forbid empty ClientID and use 'sarama' as default

slaunay:
Documentation states that `"sarama"` is the default ClientID but the empty string actually is used.

This is a not a problem for connecting to the brokers but it creates issues when settings policies against the tools binaries (e.g. `kafka-console-producer`) like:
```
bin/kafka-configs.sh  --zookeeper localhost:2181 --describe --entity-name ""  --entity-type clients
Error while executing topic command Path must not end with / character
java.lang.IllegalArgumentException: Path must not end with / character
	at org.apache.zookeeper.common.PathUtils.validatePath(PathUtils.java:58)
	at org.apache.zookeeper.ZooKeeper.getData(ZooKeeper.java:1137)
	at org.apache.zookeeper.ZooKeeper.getData(ZooKeeper.java:1184)
	at org.I0Itec.zkclient.ZkConnection.readData(ZkConnection.java:119)
	at org.I0Itec.zkclient.ZkClient$12.call(ZkClient.java:1094)
	at org.I0Itec.zkclient.ZkClient$12.call(ZkClient.java:1090)
	at org.I0Itec.zkclient.ZkClient.retryUntilConnected(ZkClient.java:985)
	at org.I0Itec.zkclient.ZkClient.readData(ZkClient.java:1090)
	at org.I0Itec.zkclient.ZkClient.readData(ZkClient.java:1085)
	at org.I0Itec.zkclient.ZkClient.readData(ZkClient.java:1074)
	at kafka.admin.AdminUtils$.fetchEntityConfig(AdminUtils.scala:353)
	at kafka.admin.ConfigCommand$$anonfun$describeConfig$1.apply(ConfigCommand.scala:108)
	at kafka.admin.ConfigCommand$$anonfun$describeConfig$1.apply(ConfigCommand.scala:107)
	at scala.collection.immutable.List.foreach(List.scala:318)
	at kafka.admin.ConfigCommand$.describeConfig(ConfigCommand.scala:107)
	at kafka.admin.ConfigCommand$.main(ConfigCommand.scala:57)
	at kafka.admin.ConfigCommand.main(ConfigCommand.scala)
```

Fix uses `"sarama"` as the default ClientID following the documentation and also forbids the empty string (test case provided).

### DIFF
--- a/client.go
+++ b/client.go
@@ -67,6 +67,8 @@ type Client interface {
 }
 
 const (
+	// Default ClientID
+	DefaultClientID = "sarama"
 	// OffsetNewest stands for the log head offset, i.e. the offset that will be
 	// assigned to the next message that will be produced to the partition. You
 	// can send this to a client's GetOffset method to get this offset, or when
--- a/client.go
+++ b/client.go
@@ -67,8 +67,6 @@ type Client interface {
 }
 
 const (
-	// Default ClientID
-	DefaultClientID = "sarama"
 	// OffsetNewest stands for the log head offset, i.e. the offset that will be
 	// assigned to the next message that will be produced to the partition. You
 	// can send this to a client's GetOffset method to get this offset, or when
--- a/config.go
+++ b/config.go
@@ -6,6 +6,8 @@ import (
 	"time"
 )
 
+const defaultClientID = "sarama"
+
 var validID *regexp.Regexp = regexp.MustCompile(`\A[A-Za-z0-9._-]+\z`)
 
 // Config is used to pass multiple configuration options to Sarama's constructors.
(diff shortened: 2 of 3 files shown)


client.go
```diff
@@ -67,6 +67,8 @@ type Client interface {
 }
 
 const (
+	// Default ClientID
+	DefaultClientID = "sarama"
```
eapache: There's no reason to make this a public constant.

Also, while I understand the intent behind putting it in `client.go` I think it makes more sense in `config.go`.

eapache:
Wow I'm dumb. I had two distinct chances to get this trivial thing right: when I wrote the original PR that said "it defaults to sarama!" *and* a whole year later when I wrote the regex and had a chance to use the correct quantifier, but didn't.

Thanks for finding this :)

slaunay:
I applied the changes.
Let me know if you want me to rebase the changes.

Thanks for this sweet library, much easier to read than the official Java client.

eapache:
Thanks :)

PR #677, opened 2016-06-09
SyncProducer batch support

### LINKED ISSUE #674: Support synchronous sending for batches of messages
Currently, the sync producer doesn't support sending batches of messages, even though the underlying protocol allows this (and the java client library supports this too).

We have a need to send a batch of messages reliably with acknowledgment, but don't want to incur the multiple round trips required with sending each message individually.

eapache:
Provide a `SyncProducer.SendMessages` method which produces a given *batch* of
messages (though does not guarantee they will end up in a single batch on the
wire).

Fixes #674 

@tadbook @JasonRosenberg cc @wvanbergen

### DIFF
--- a/mocks/sync_producer.go
+++ b/mocks/sync_producer.go
@@ -57,6 +57,31 @@ func (sp *SyncProducer) SendMessage(msg *sarama.ProducerMessage) (partition int3
 	}
 }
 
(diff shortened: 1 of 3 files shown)


tadbook:
Thanks!

Issue #642, opened 2016-04-14
Planned support for new Kafka 0.10 message format


makkes:
Do you already have plans on supporting the new message format of the upcoming 0.10 release? We're especially interested in using the new timestamps introduced with https://cwiki.apache.org/confluence/display/KAFKA/KIP-32+-+Add+timestamps+to+Kafka+message.

Thanks!

wvanbergen:
No plans as of yet, but we're always interested in contributions.

This may require #617 to be in place first, so we can continue to support older Kafka versions as well. @eapache what do you think?

eapache:
There are two pieces here:
- on-the-wire protocol message support, which we can add at any time and won't conflict with anything
- ability to actually use the new messages in the producer/consumer which depends on #617 to really do properly

Pull requests welcome. I'll probably get around to it eventually otherwise, but it's very low on our list of priorities right now.

eapache:
Support for targeting various kafka versions properly was done in #673 and #676. Support for producer timestamps was done in #678.

PR #678, opened 2016-06-10
Support and provide producer message timestamps

eapache:
- Support v0/v1/v2 of Produce requests in the protocol layer.
- Enforce the correct message version for Kafka version at the Broker.
- Use v2 messages in the producer if the Kafka version supports it, and pass the
  resulting timestamp back to the user in the ProducerMessage.

Addresses #642

@makkes @wvanbergen

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -135,6 +135,11 @@ type ProducerMessage struct {
 	// Partition is the partition that the message was sent to. This is only
 	// guaranteed to be defined if the message was successfully delivered.
 	Partition int32
+	// Timestamp is the timestamp assigned to the message by the broker. This
+	// is only guaranteed to be defined if the message was successfully
+	// delivered, RequiredAcks is not NoResponse, and the Kafka broker is at
+	// least version 0.10.0.
+	Timestamp time.Time
 
 	retries int
 	flags   flagSet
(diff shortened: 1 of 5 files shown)


kchaliki:
Hello @eapache et al, I have pulled this down and tried to use it, I assume it's enough to set the new version onto my config like so

    config := sarama.NewConfig()
    config.Version = sarama.V0_10_0_0

    syncProducer, err := sarama.NewSyncProducer(connections, config)

However attempts to send a message fail with 

panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x5d773c]

goroutine 771 [running]:
panic(0x878b00, 0xc820018130)
	/opt/go/src/runtime/panic.go:481 +0x3e6
github.com/Shopify/sarama.(*Broker).Produce(0xc82013bc00, 0xc82062e620, 0xc820029770, 0x0, 0x0)
	/home/kostasc/go/src/github.com/Shopify/sarama/broker.go:209 +0x44c
github.com/Shopify/sarama.(*asyncProducer).newBrokerProducer.func1()
	/home/kostasc/go/src/github.com/Shopify/sarama/async_producer.go:543 +0x9a
github.com/Shopify/sarama.withRecover(0xc82062e520)
	/home/kostasc/go/src/github.com/Shopify/sarama/utils.go:46 +0x3a
created by github.com/Shopify/sarama.(*asyncProducer).newBrokerProducer
	/home/kostasc/go/src/github.com/Shopify/sarama/async_producer.go:552 +0x366

Changing the version to previous values works but the timestamp is not set

kchaliki:
Some more details I collected, This could potentially be a race between the goroutines started at broker.go:64 which sets the conf on the broker instance and async_producer.go:539 which calls broker.Produce as adding some printfs tells me b.conf is only nil for the first of many messages I am trying to send. Note the code in the stack trace above is not run for any other version.

Or it could be that I need to initialize things some other way before I start producing.

eapache:
Yes, I believe it is a race, I have a fix in the works.

PR #685, opened 2016-06-22
client: fix returning the wrong error

dgryski:
err goes out of scope at the end of the if-block

Bug found with ineffassign.


eapache:
Thanks!

PR #687, opened 2016-06-23
Check the received value in producers (synch/asych) mocks agains a regexp

### LINKED ISSUE #667: Check the received message value in producer mocks
Hi guys, not sure if this is a valid issue.
When developing a unit test for a program that uses the sync/asyc producer, I can only say that I expect to receive a message and that I want the mock response to be success/error.
For my test I need to check the value inside the message, so I propose to extend the `producerExpectation struct` in mocks.go and to add new methods to the producer mock, that receive in input a string (regexp) and check the received value against this regexp.
I am ok to propose an implementation if it makes sense.

jacksilv:
Fixes #667. Thanks to have a look guys

### DIFF
--- a/mocks/async_producer.go
+++ b/mocks/async_producer.go
@@ -1,7 +1,6 @@
 package mocks
 
 import (
-	"regexp"
 	"sync"
 
 	"github.com/Shopify/sarama"
@@ -11,8 +10,8 @@ import (
 // Before you can send messages to it's Input channel, you have to set expectations
 // so it knows how to handle the input; it returns and error if the numer of messages
 // received is bigger then the number of expectations set. You can also set a
-// regexp in each expectation so that the message value is matched against this
-// regexp and an error is returned if the match fails.
+// function in each expectation so that the message value is checked by this function
+// and an error is returned if the match fails.
 type AsyncProducer struct {
 	l            sync.Mutex
 	t            ErrorReporter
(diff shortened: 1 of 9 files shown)


eapache:
I would prefer not adding a method to `Encoder`, not all messages are encodable as strings. Instead I think the more general solution is to support passing a callback `func ([]byte) error` into the mock. Then if you know your messages are strings you can cast yourself and do whatever regexp checks you want; if not you can do whatever other sanity checks you want. Does this make sense?

jacksilv:
Thanks @eapache for your comments, I updated my code

mocks/mocks.go
```diff
@@ -25,6 +27,21 @@ type ErrorReporter interface {
 	Errorf(string, ...interface{})
 }
 
+type ValueChecker func(val []byte) error
```
eapache: could use a (brief) godoc comment

mocks/mocks.go
```diff
@@ -25,6 +27,21 @@ type ErrorReporter interface {
...
+
+func generateRegexpChecker(re string) func([]byte) error {
```
eapache: not exported, so if you want to actually use it in your app you won't be able to... I think it might just be easier to implement this directly in your app instead of here anyway

mocks/async_producer.go
```diff
@@ -8,8 +8,10 @@ import (
 
 // AsyncProducer implements sarama's Producer interface for testing purposes.
 // Before you can send messages to it's Input channel, you have to set expectations
-// so it knows how to handle the input. This way you can easily test success and
-// failure scenarios.
+// so it knows how to handle the input; it returns and error if the numer of messages
```
eapache: types ('and', 'numer')

eapache:
Thanks, this is pretty good now, just a few little things.

mocks/mocks.go
jacksilv: I put this there on purpuse just for the unit tests, thanks.

Issue #683, opened 2016-06-20
Expose metrics using go-metrics


slaunay:
##### Versions

Sarama Version: v1.9.0 
Kafka Version: *
Go Version: *
##### Problem Description

On the contrary to the Java client, no internal metrics are provided in order to monitor or optimize the performances (e.g. latency, throughput) like:
- the average number of bytes sent per partition per-request.
- the average time in ms record batches spent in the various channels before getting send on the wire.  
- the average compression rate of record batches.
- the average number of outgoing bytes sent per second to all servers.
- the average number of requests sent per second.
- see http://kafka.apache.org/documentation.html#monitoring for the (very) long list of metrics

Some of those metrics are keys for tuning a Kafka producer over long fat networks or just benchmarking a consumer without relying on the broker metrics.

I was thinking of exposing a few useful metrics to the producer through Richard Crowley's [Go port of Coda Hale's Metrics library](https://github.com/rcrowley/go-metrics).
That library do not have transitive dependencies but provides built-in stats reporter (Graphite, OpenTSDB) as well as third party ones (e.g. InfluxDB) and can be disable using `UseNilMetrics` variable.

The idea would be to add a new public `MetricRegistry Registry` field to the `Config` struct that would defaults to `metrics.DefaultRegistry` and enrich it with metrics using similar name that the official documentation.

What do you think, any ideas on which metrics needs to be exposed first?

eapache:
Sounds reasonable to me. I would start with a few easy ones as a proof of concept (for example bytes per producer request). Some of the others (such as ms spent in channels) would require tracking much more internal timing information, so I'd prefer to start with something less invasive.

slaunay:
Sounds good, I'll try to work on a PR so we can discuss on the details.

hasnickl:
maybe out of scope here but prometheus is also a popular metrics sink and being able to configure for either go-metrics or prometheus metrics would be useful

slaunay:
From what I have read, Prometheus provides a time series database and client integration.
It looks like the model is a bit different and requires sample information rather than summarized information that Dropwizard/go-metrics typically expose to reporters.

I believe that it is possible to publish metrics from go-metrics into Prometheus using their go integration but at the cost of losing data precision when using something else than a gauge or a counter.

It is an issue for Dropwizard's metrics library as well that might get tackle in future version as far as I know:
https://github.com/dropwizard/metrics/issues/712

[...]

PR #688, opened 2016-06-29
Expose metrics using go-metrics (#683)

slaunay:
- add `MetricRegistry` configuration parameter that defaults to `metrics.DefaultRegistry`
  - `incoming-byte-rate` meter (global and per registered broker)
  - `request-rate` meter (global and per registered broker)
  - `request-size` histogram (global and per registered broker)
  - `outgoing-byte-rate` meter (global and per registered broker)
  - `response-rate` meter (global and per registered broker)
  - `response-size` histogram (global and per registered broker)
  - `batch-size` histogram (global and per topic)
  - `record-send-rate` meter (global and per topic)
  - `records-per-request` histogram (global and per topic)
  - `compression-rate` histogram (global and per topic)
- add metrics flag to `kafka-console-producer` to output metrics
- validate metrics in `functional_producer_test`

This is a **work in progress**, next steps:
- rebase with master to fix merge issue
- unit tests
- documentation
- example of exporting metrics to Graphite using a prefix

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+
+	"github.com/rcrowley/go-metrics"
 )
 
 // Broker represents a single Kafka broker connection. All operations on this object are entirely concurrency-safe.
@@ -26,6 +28,19 @@ type Broker struct {
 
 	responses chan responsePromise
 	done      chan bool
+
+	incomingByteRate       metrics.Meter
+	requestRate            metrics.Meter
+	requestSize            metrics.Histogram
+	outgoingByteRate       metrics.Meter
+	responseRate           metrics.Meter
+	responseSize           metrics.Histogram
+	brokerIncomingByteRate metrics.Meter
+	brokerRequestRate      metrics.Meter
+	brokerRequestSize      metrics.Histogram
+	brokerOutgoingByteRate metrics.Meter
+	brokerResponseRate     metrics.Meter
+	brokerResponseSize     metrics.Histogram
 }
 
 type responsePromise struct {
@@ -83,6 +98,23 @@ func (b *Broker) Open(conf *Config) error {
 		b.conn = newBufConn(b.conn)
 
 		b.conf = conf
+
+		// Create or reuse the metrics shared between brokers
+		b.incomingByteRate = metrics.GetOrRegisterMeter("incoming-byte-rate", conf.MetricRegistry)
+		b.requestRate = metrics.GetOrRegisterMeter("request-rate", conf.MetricRegistry)
+		b.requestSize = getOrRegisterHistogram("request-size", conf.MetricRegistry)
+		b.outgoingByteRate = metrics.GetOrRegisterMeter("outgoing-byte-rate", conf.MetricRegistry)
+		b.responseRate = metrics.GetOrRegisterMeter("response-rate", conf.MetricRegistry)
+		b.responseSize = getOrRegisterHistogram("response-size", conf.MetricRegistry)
+		// Do not gather metrics for seeded broker (-1 id)
+		if b.id >= 0 {
+			b.brokerIncomingByteRate = getOrRegisterBrokerMeter("incoming-byte-rate", b.id, conf.MetricRegistry)
+			b.brokerRequestRate = getOrRegisterBrokerMeter("request-rate", b.id, conf.MetricRegistry)
+			b.brokerRequestSize = getOrRegisterBrokerHistogram("request-size", b.id, conf.MetricRegistry)
+			b.brokerOutgoingByteRate = getOrRegisterBrokerMeter("outgoing-byte-rate", b.id, conf.MetricRegistry)
+			b.brokerResponseRate = getOrRegisterBrokerMeter("response-rate", b.id, conf.MetricRegistry)
+			b.brokerResponseSize = getOrRegisterBrokerHistogram("response-size", b.id, conf.MetricRegistry)
+		}
 
 		if conf.Net.SASL.Enable {
 			b.connErr = b.sendAndReceiveSASLPlainAuth()
@@ -333,10 +365,12 @@ func (b *Broker) send(rb protocolBody, promiseResponse bool) (*responsePromise, 
 	}
 
 	req := &request{correlationID: b.correlationID, clientID: b.conf.ClientID, body: rb}
-	buf, err := encode(req)
+	buf, err := encodeWithMetrics(req, b.conf.MetricRegistry)
 	if err != nil {
 		return nil, err
 	}
+
+	b.updateOutgoingCommunicationMetrics(len(buf))
 
 	err = b.conn.SetWriteDeadline(time.Now().Add(b.conf.Net.WriteTimeout))
 	if err != nil {
@@ -441,7 +475,7 @@ func (b *Broker) responseReceiver() {
 			continue
 		}
 
-		_, err = io.ReadFull(b.conn, header)
+		headerSize, err := io.ReadFull(b.conn, header)
 		if err != nil {
 			dead = err
 			response.errors <- err
@@ -464,12 +498,14 @@ func (b *Broker) responseReceiver() {
 		}
 
 		buf := make([]byte, decodedHeader.length-4)
-		_, err = io.ReadFull(b.conn, buf)
+		remainingPayloadSize, err := io.ReadFull(b.conn, buf)
 		if err != nil {
 			dead = err
 			response.errors <- err
 			continue
 		}
+
+		b.updateIncomingCommunicationMetrics(headerSize + remainingPayloadSize)
 
 		response.packets <- buf
 	}
@@ -500,6 +536,8 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
 	binary.BigEndian.PutUint32(authBytes, uint32(length))
 	copy(authBytes[4:], []byte("\x00"+b.conf.Net.SASL.User+"\x00"+b.conf.Net.SASL.Password))
 
+	b.updateOutgoingCommunicationMetrics(len(authBytes))
+
 	err := b.conn.SetWriteDeadline(time.Now().Add(b.conf.Net.WriteTimeout))
 	if err != nil {
 		Logger.Printf("Failed to set write deadline when doing SASL auth with broker %s: %s\n", b.addr, err.Error())
@@ -521,6 +559,40 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
 		return err
 	}
 
+	b.updateIncomingCommunicationMetrics(n)
+
 	Logger.Printf("SASL authentication successful with broker %s:%v - %v\n", b.addr, n, header)
 	return nil
+}
+
+func (b *Broker) updateIncomingCommunicationMetrics(bytes int) {
+	b.responseRate.Mark(1)
+	if b.brokerResponseRate != nil {
+		b.brokerResponseRate.Mark(1)
+	}
+	responseSize := int64(bytes)
+	b.incomingByteRate.Mark(responseSize)
+	if b.brokerIncomingByteRate != nil {
+		b.brokerIncomingByteRate.Mark(responseSize)
+	}
+	b.responseSize.Update(responseSize)
+	if b.brokerResponseSize != nil {
+		b.brokerResponseSize.Update(responseSize)
+	}
+}
+
+func (b *Broker) updateOutgoingCommunicationMetrics(bytes int) {
+	b.requestRate.Mark(1)
+	if b.brokerRequestRate != nil {
+		b.brokerRequestRate.Mark(1)
+	}
+	requestSize := int64(bytes)
+	b.outgoingByteRate.Mark(requestSize)
+	if b.brokerOutgoingByteRate != nil {
+		b.brokerOutgoingByteRate.Mark(requestSize)
+	}
+	b.requestSize.Update(requestSize)
+	if b.brokerRequestSize != nil {
+		b.brokerRequestSize.Update(requestSize)
+	}
 }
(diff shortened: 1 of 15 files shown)


slaunay:
Example of metrics output when using `kafka-console-producer`:
```
./kafka-console-producer -brokers 192.168.100.67:9091 -topic test.1 -value message -metrics
topic=test.1	partition=0	offset=10003
histogram batch-size
  count:               1
  min:                41
  max:                41
  mean:               41.00
  stddev:              0.00
  median:             41.00
  75%:                41.00
  95%:                41.00
  99%:                41.00
  99.9%:              41.00
histogram batch-size-for-topic-test.1
  count:               1
  min:                41
  max:                41
  mean:               41.00
  stddev:              0.00
  median:             41.00
  75%:                41.00
  95%:                41.00
  99%:                41.00
  99.9%:              41.00
histogram compression-rate
  count:               1
  min:               100
  max:               100
  mean:              100.00
  stddev:              0.00
  median:            100.00
  75%:               100.00
  95%:               100.00
  99%:               100.00
  99.9%:             100.00
histogram compression-rate-for-topic-test.1
  count:               1
  min:               100
  max:               100
  mean:              100.00
  stddev:              0.00
  median:            100.00
  75%:               100.00
  95%:               100.00
  99%:               100.00
  99.9%:             100.00
meter incoming-byte-rate
  count:            5241
  1-min rate:          0.00
  5-min rate:          0.00
  15-min rate:         0.00
  mean rate:      356437.67
meter incoming-byte-rate-for-broker-9093
  count:              38
  1-min rate:          0.00
  5-min rate:          0.00
  15-min rate:         0.00
  mean rate:        6004.04
meter outgoing-byte-rate
  count:             107
  1-min rate:          0.00
  5-min rate:          0.00
  15-min rate:         0.00
  mean rate:       12559.68
meter outgoing-byte-rate-for-broker-9093
  count:              83
  1-min rate:          0.00
  5-min rate:          0.00
  15-min rate:         0.00
  mean rate:      510750.37
meter record-send-rate
  count:               1
  1-min rate:          0.00
  5-min rate:          0.00
  15-min rate:         0.00
  mean rate:      857632.93
meter record-send-rate-for-topic-test.1
  count:               1
  1-min rate:          0.00
  5-min rate:          0.00
  15-min rate:         0.00
  mean rate:      519480.52
histogram records-per-request
  count:               1
  min:                 1
  max:                 1
  mean:                1.00
  stddev:              0.00
  median:              1.00
  75%:                 1.00
  95%:                 1.00
  99%:                 1.00
  99.9%:               1.00
histogram records-per-request-for-topic-test.1
  count:               1
  min:                 1
  max:                 1
  mean:                1.00
  stddev:              0.00
  median:              1.00
  75%:                 1.00
  95%:                 1.00
  99%:                 1.00
  99.9%:               1.00
meter request-rate
  count:               2
  1-min rate:          0.00
  5-min rate:          0.00
  15-min rate:         0.00
  mean rate:         234.18
meter request-rate-for-broker-9093
  count:               1
  1-min rate:          0.00
  5-min rate:          0.00
  15-min rate:         0.00
  mean rate:        5795.62
histogram request-size
  count:               2
  min:                24
  max:                83
  mean:               53.50
  stddev:             29.50
  median:             53.50
  75%:                83.00
  95%:                83.00
  99%:                83.00
  99.9%:              83.00
histogram request-size-for-broker-9093
  count:               1
  min:                83
  max:                83
  mean:               83.00
  stddev:              0.00
  median:             83.00
  75%:                83.00
  95%:                83.00
  99%:                83.00
  99.9%:              83.00
meter response-rate
  count:               2
  1-min rate:          0.00
  5-min rate:          0.00
  15-min rate:         0.00
  mean rate:         136.45
meter response-rate-for-broker-9093
  count:               1
  1-min rate:          0.00
  5-min rate:          0.00
  15-min rate:         0.00
  mean rate:         158.71
histogram response-size
  count:               2
  min:                38
  max:              5203
  mean:             2620.50
  stddev:           2582.50
  median:           2620.50
  75%:              5203.00
  95%:              5203.00
  99%:              5203.00
  99.9%:            5203.00
histogram response-size-for-broker-9093
  count:               1
  min:                38
  max:                38
  mean:               38.00
  stddev:              0.00
  median:             38.00
  75%:                38.00
  95%:                38.00
  99%:                38.00
  99.9%:              38.00
```

broker.go
```diff
@@ -84,6 +99,23 @@ func (b *Broker) Open(conf *Config) error {
 
 		b.conf = conf
 
+		// Create or reuse the metrics shared between brokers
+		b.incomingByteRate = metrics.GetOrRegisterMeter("incoming-byte-rate", conf.MetricRegistry)
```
eapache: Are the `metrics.GetOrRegister*` methods concurrency-safe? Because many brokers can call these at once.

broker.go
```diff
@@ -84,6 +99,23 @@ func (b *Broker) Open(conf *Config) error {
...
+		b.requestRate = metrics.GetOrRegisterMeter("request-rate", conf.MetricRegistry)
+		b.requestSize = getOrRegisterHistogram("request-size", conf.MetricRegistry)
+		b.outgoingByteRate = metrics.GetOrRegisterMeter("outgoing-byte-rate", conf.MetricRegistry)
+		b.responseRate = metrics.GetOrRegisterMeter("response-rate", conf.MetricRegistry)
+		b.responseSize = getOrRegisterHistogram("response-size", conf.MetricRegistry)
+		// Do not gather metrics for seeded broker (-1 id)
+		if b.id >= 0 {
```
eapache: why don't you want metrics for the seeded broker?

message_set.go
```diff
@@ -48,9 +48,11 @@ func (msb *MessageBlock) decode(pd packetDecoder) (err error) {
 type MessageSet struct {
 	PartialTrailingMessage bool // whether the set on the wire contained an incomplete trailing MessageBlock
 	Messages               []*MessageBlock
+	UnderlyingMessageCount int // in case of compressed messages wraps in a single fake "message"
```
eapache: I'm not sure I understand why this is needed isn't it always just equal to `len(Messages)`?

eapache: Oh I need to learn to read, it's for compressed messages; I'm still not sure what purposes it servers when you could just ask for `len` of the compressed message-set's messages?

packet_encoder.go
```diff
@@ -22,6 +22,13 @@ type packetEncoder interface {
 	// Stacks, see PushEncoder
 	push(in pushEncoder)
 	pop() error
+
+	// Statistics in order to expose metrics
+	offset() int
+	recordTopic(topic string)
```
eapache: this isn't at the right level; the protocol-level encoding/decoding shouldn't have any knowledge of semantic values like topic names

broker.go
slaunay: Yes they are through a `sync.Mutex`.
I was thinking about caching the metrics resolved to avoid hitting that mutex but after running the existing benchmarks for the master branch and this branch I did not notice any slowness or contention.

broker.go
slaunay: They would appear with `-for-broker--1` without knowing which broker i used and metrics are already accumulated in the metrics for all brokers (without `-for-broker-<Broker.ID()>`.
Maybe using the hostname of the broker is more explicit but the Java client uses the broker id I believe.

packet_encoder.go
slaunay: I agree this add unwanted coupling but I am sure not sure how to be aware of a context (topic here) when recording statistics.
What cleaner approach would you suggest?

message_set.go
slaunay: I believe that once N compressed messages are stored inside a single fake "message" we lose the amount of messages that was serialized and `len(ms.Messages)` would then return 1 instead of N.

broker.go
eapache: Does java provide metrics for seed brokers? It just seems dangerous to omit them since it will mean your statistics will be incomplete.

message_set.go
eapache: You should still be able to ask for `len(set.Messages[0].Msg.Set.Messages)`? I'm not sure why you even need this value to be honest?

packet_encoder.go
eapache: I'm not sure why it's going through the packet-encoder at all? Can't the various callers (e.g. in `produce_request.go`) fetch and update the metric objects directly? They should have all the necessary context.

broker.go
slaunay: They do not provide the network related metrics for seed brokers either:
- https://github.com/apache/kafka/blob/d7de59a579af5ba4ecb1aec8fed84054f8b86443/clients/src/main/java/org/apache/kafka/common/network/Selector.java#L302
- https://github.com/apache/kafka/blob/d7de59a579af5ba4ecb1aec8fed84054f8b86443/clients/src/main/java/org/apache/kafka/common/network/Selector.java#L655
- https://github.com/apache/kafka/blob/d7de59a579af5ba4ecb1aec8fed84054f8b86443/clients/src/main/java/org/apache/kafka/common/Node.java#L20

I do not know why they chose not to but I found this comment on a JIRA ticket:
https://issues.apache.org/jira/browse/KAFKA-1251?focusedCommentId=13948165&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13948165

slaunay:
I moved most of the metric gathering code to `produce_request.go` but I had to expose a couple of things to make it work:
- `len(set.Messages[0].Msg.Set.Messages)` was not working as the Set field was not set for fake compressed "message"
- the current offset of the `packetEncoder` is necessary to compute the batch size
- a flag to know if this is the second pass in order to collect metrics once

If this looks better, I'll rebase against master, provide unit tests, documentation and some examples.

broker.go
eapache: Sounds like java uses the ID because IPv6 addresses contain characters that aren't permitted in jmx metric names. I'm OK with following them on this one.

I don't know what limitations this metrics package has, but if it supports all the necessary names then I'm also OK actually using the address:port combo instead of the ID.

eapache: If you stick with IDs, an explanatory comment here would be nice.

broker.go
```diff
@@ -464,13 +498,15 @@ func (b *Broker) responseReceiver() {
 		}
 
 		buf := make([]byte, decodedHeader.length-4)
-		_, err = io.ReadFull(b.conn, buf)
+		remainingPayloadSize, err := io.ReadFull(b.conn, buf)
```
eapache: You don't need to capture this value (or `headerSize` above) since `ReadFull` will return an error unless it completely fills the buffer, and we'll never even get to the metric call in that case. You should be able to use `len(header) + len(buf)` instead.

broker.go
```diff
@@ -521,6 +559,40 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
...
 
+	b.updateIncomingCommunicationMetrics(n)
+
 	Logger.Printf("SASL authentication successful with broker %s:%v - %v\n", b.addr, n, header)
 	return nil
 }
+
+func (b *Broker) updateIncomingCommunicationMetrics(bytes int) {
+	b.responseRate.Mark(1)
+	if b.brokerResponseRate != nil {
```
eapache: Can the metric creation methods return nil? Or is there some weird case where we could try and measure metrics before opening the connection?

eapache: Oh of course for seed-brokers we don't even generate some metrics. Oh well.

packet_encoder.go
```diff
@@ -19,9 +19,15 @@ type packetEncoder interface {
...
 
+	// Provide the current offset to record the batch size metric
+	offset() int
+
 	// Stacks, see PushEncoder
 	push(in pushEncoder)
 	pop() error
+
+	// To only record metrics during the real encoder pass
+	doRecordMetrics() bool
```
eapache: `do` feels too much like an action; maybe `shouldRecordMetrics` or even just `recordMetrics`?

eapache: Or even better, have a method `metricsRegistry() *metrics.Registry` that returns the registry or nil depending; this way we can pass the registry in once at the broker and won't have to add it to each request/response struct.

produce_set.go
```diff
@@ -85,10 +86,12 @@ func (ps *produceSet) buildRequest() *ProduceRequest {
 					Logger.Println(err) // if this happens, it's basically our fault.
 					panic(err)
 				}
+				// Provide the underlying message set for accurate metrics
```
eapache: good comment, put it right on the `Set: set.setToSend,` line please

eapache:
This is definitely better, thanks for all your work! It's getting large enough that when you rebase etc. I would ask you split it in two please: one PR for the config and docs and broker metrics, and then a follow-up PR adding producer metrics. It will be easier to review in chunks like that.

packet_encoder.go
slaunay: Sounds good for using `metricsRegistry() metrics.Registry` (it's an interface) to the `packetEncoder` interface.

If I understand correctly, the broker would pass the `metrics.Registry` reference during encoding that would eventually get stored into the `realEncoder struct`:
```
broker.send(...) > encode(req) > {req.encode(prep_encoder); req.encode(real_encoder)}
```
Do you prefer to add a `metrics.Registry` argument to `encode(e encoder) ([]byte, error)` or add another function `encodeWithMetrics(e encoder, metricsRegistry metrics.Registry) ([]byte, error)` like in the initial 1e5e537 commit?

The other places that uses `encode(e encoder) ([]byte, error)` would then either pass `nil` or call the new `encodeWithMetrics` function.

eapache: > (it's an interface)

Ah, missed that the first time round.

> Do you prefer to add a metrics.Registry argument to encode(e encoder) ([]byte, error) or add another function encodeWithMetrics(e encoder, metricsRegistry metrics.Registry) ([]byte, error) like in the initial 1e5e537 commit?

Just add an argument to `encode` and have the other callers pass `nil`, this doesn't feel important enough to warrant a whole method.

PR #695, opened 2016-07-07
Passing ProducerMessage.Timestamp to broker when producing to v0.10

dkjer:
This allows producing messages with correct CreateTime timestamps in 0.10


eapache:
Thanks!

PR #697, opened 2016-07-07
Do not reset ProducerMessage timestamp when it's CreateTime

yingnanliu:
ProduceResponse doesn't always guarantee a valid timestamp. When broker is set to "CreateTime", the field will be -1.

[http://kafka.apache.org/protocol.html](http://kafka.apache.org/protocol.html)
> If CreateTime is used for the topic, the timestamp will be -1.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -727,7 +727,7 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
 		switch block.Err {
 		// Success
 		case ErrNoError:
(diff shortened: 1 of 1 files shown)


eapache:
Thanks for pointing this out! The proper place to handle this change is in the produce_response.go parsing code, not in the high-level consumer, so I made that change in https://github.com/Shopify/sarama/commit/7f0fe2ddaf7ab1bb53058d3d1244a58bbc8bb110

PR #699, opened 2016-07-09
Don't propagate zero timestamp to ProducerMessage

yingnanliu:
Thank you for your response to #697. It makes much more sense to handle -1 timestamp in produce_response.go.
It would be nice to let ProducerMessage keep its timestamp, not overridden after being sent, so I made this pull request. Look forward to knowing your opinion.


eapache:
Yes of course, I'd forgot that the client is also allowed to specify the timestamp it wants on the wire. We haven't started using any of the timestamp features ourselves yet so I haven't got a full idea of how it works. Thanks!

PR #701, opened 2016-07-12
Expose broker metrics with go-metrics

slaunay:
- add `MetricRegistry` configuration parameter that defaults to `metrics.DefaultRegistry`
  - `incoming-byte-rate` meter (global and per registered broker)
  - `request-rate` meter (global and per registered broker)
  - `request-size` histogram (global and per registered broker)
  - `outgoing-byte-rate` meter (global and per registered broker)
  - `response-rate meter` (global and per registered broker)
  - `response-size histogram` (global and per registered broker)
- add `metrics` flag to `kafka-console-producer` to output metrics
- add bytes read to `decodeRequest`
- count socket bytes read and written in `MockBroker` for unit testing
- add unit tests and example
- functional tests in `functional_producer_test`
- documentation of metrics in main package

Related issue #683 and original PR #688.
I'll submit a PR for producer related metrics once this one is merged (should be shorter 😉).

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+
+	"github.com/rcrowley/go-metrics"
 )
 
 // Broker represents a single Kafka broker connection. All operations on this object are entirely concurrency-safe.
@@ -26,6 +28,19 @@ type Broker struct {
 
 	responses chan responsePromise
 	done      chan bool
+
+	incomingByteRate       metrics.Meter
+	requestRate            metrics.Meter
+	requestSize            metrics.Histogram
+	outgoingByteRate       metrics.Meter
+	responseRate           metrics.Meter
+	responseSize           metrics.Histogram
+	brokerIncomingByteRate metrics.Meter
+	brokerRequestRate      metrics.Meter
+	brokerRequestSize      metrics.Histogram
+	brokerOutgoingByteRate metrics.Meter
+	brokerResponseRate     metrics.Meter
+	brokerResponseSize     metrics.Histogram
 }
 
 type responsePromise struct {
@@ -83,6 +98,24 @@ func (b *Broker) Open(conf *Config) error {
 		b.conn = newBufConn(b.conn)
 
 		b.conf = conf
+
+		// Create or reuse the global metrics shared between brokers
+		b.incomingByteRate = metrics.GetOrRegisterMeter("incoming-byte-rate", conf.MetricRegistry)
+		b.requestRate = metrics.GetOrRegisterMeter("request-rate", conf.MetricRegistry)
+		b.requestSize = getOrRegisterHistogram("request-size", conf.MetricRegistry)
+		b.outgoingByteRate = metrics.GetOrRegisterMeter("outgoing-byte-rate", conf.MetricRegistry)
+		b.responseRate = metrics.GetOrRegisterMeter("response-rate", conf.MetricRegistry)
+		b.responseSize = getOrRegisterHistogram("response-size", conf.MetricRegistry)
+		// Do not gather metrics for seeded broker (only used during bootstrap) because they share
+		// the same id (-1) and are already exposed through the global metrics above
+		if b.id >= 0 {
+			b.brokerIncomingByteRate = getOrRegisterBrokerMeter("incoming-byte-rate", b, conf.MetricRegistry)
+			b.brokerRequestRate = getOrRegisterBrokerMeter("request-rate", b, conf.MetricRegistry)
+			b.brokerRequestSize = getOrRegisterBrokerHistogram("request-size", b, conf.MetricRegistry)
+			b.brokerOutgoingByteRate = getOrRegisterBrokerMeter("outgoing-byte-rate", b, conf.MetricRegistry)
+			b.brokerResponseRate = getOrRegisterBrokerMeter("response-rate", b, conf.MetricRegistry)
+			b.brokerResponseSize = getOrRegisterBrokerHistogram("response-size", b, conf.MetricRegistry)
+		}
 
 		if conf.Net.SASL.Enable {
 			b.connErr = b.sendAndReceiveSASLPlainAuth()
@@ -338,6 +371,8 @@ func (b *Broker) send(rb protocolBody, promiseResponse bool) (*responsePromise, 
 		return nil, err
 	}
 
+	b.updateOutgoingCommunicationMetrics(len(buf))
+
 	err = b.conn.SetWriteDeadline(time.Now().Add(b.conf.Net.WriteTimeout))
 	if err != nil {
 		return nil, err
@@ -471,6 +506,8 @@ func (b *Broker) responseReceiver() {
 			continue
 		}
 
+		b.updateIncomingCommunicationMetrics(len(header) + len(buf))
+
 		response.packets <- buf
 	}
 	close(b.done)
@@ -500,6 +537,8 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
 	binary.BigEndian.PutUint32(authBytes, uint32(length))
 	copy(authBytes[4:], []byte("\x00"+b.conf.Net.SASL.User+"\x00"+b.conf.Net.SASL.Password))
 
+	b.updateOutgoingCommunicationMetrics(len(authBytes))
+
 	err := b.conn.SetWriteDeadline(time.Now().Add(b.conf.Net.WriteTimeout))
 	if err != nil {
 		Logger.Printf("Failed to set write deadline when doing SASL auth with broker %s: %s\n", b.addr, err.Error())
@@ -521,6 +560,40 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
 		return err
 	}
 
+	b.updateIncomingCommunicationMetrics(n)
+
 	Logger.Printf("SASL authentication successful with broker %s:%v - %v\n", b.addr, n, header)
 	return nil
+}
+
+func (b *Broker) updateIncomingCommunicationMetrics(bytes int) {
+	b.responseRate.Mark(1)
+	if b.brokerResponseRate != nil {
+		b.brokerResponseRate.Mark(1)
+	}
+	responseSize := int64(bytes)
+	b.incomingByteRate.Mark(responseSize)
+	if b.brokerIncomingByteRate != nil {
+		b.brokerIncomingByteRate.Mark(responseSize)
+	}
+	b.responseSize.Update(responseSize)
+	if b.brokerResponseSize != nil {
+		b.brokerResponseSize.Update(responseSize)
+	}
+}
+
+func (b *Broker) updateOutgoingCommunicationMetrics(bytes int) {
+	b.requestRate.Mark(1)
+	if b.brokerRequestRate != nil {
+		b.brokerRequestRate.Mark(1)
+	}
+	requestSize := int64(bytes)
+	b.outgoingByteRate.Mark(requestSize)
+	if b.brokerOutgoingByteRate != nil {
+		b.brokerOutgoingByteRate.Mark(requestSize)
+	}
+	b.requestSize.Update(requestSize)
+	if b.brokerRequestSize != nil {
+		b.brokerRequestSize.Update(requestSize)
+	}
 }
--- a/broker.go
+++ b/broker.go
@@ -371,14 +371,13 @@ func (b *Broker) send(rb protocolBody, promiseResponse bool) (*responsePromise, 
 		return nil, err
 	}
 
-	b.updateOutgoingCommunicationMetrics(len(buf))
-
 	err = b.conn.SetWriteDeadline(time.Now().Add(b.conf.Net.WriteTimeout))
 	if err != nil {
 		return nil, err
 	}
 
-	_, err = b.conn.Write(buf)
+	bytes, err := b.conn.Write(buf)
+	b.updateOutgoingCommunicationMetrics(bytes)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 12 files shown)


eapache:
Sorry for the delays in responding to this - between vacations and some internal deadlines there hasn't been a lot of time for Sarama recently. Hopefully I'll be able to review this properly next week.

slaunay:
We have been running that branch to push logs for a couple of weeks without issues.

The metrics matches what we see coming out of the network interfaces:
grafana-dashboard

The only artifact we saw is that the min and max of the response size metrics tends to fluctuate most likely because of the way sampling is done and the periodic metadata refresh that is significantly bigger than regular produce responses:
grafana-response-size

Let me know if you need more information.

broker.go
```diff
@@ -338,6 +371,8 @@ func (b *Broker) send(rb protocolBody, promiseResponse bool) (*responsePromise,
 		return nil, err
 	}
 
+	b.updateOutgoingCommunicationMetrics(len(buf))
```
eapache: probably want to move this down below the `SetWriteDeadline` else you could miscount in the (admittedly exceptional) case where that fails

broker_test.go
```diff
@@ -52,36 +55,45 @@ func TestBrokerAccessors(t *testing.T) {
...
+			t.Fatal(err)
+		}
 		tt.runner(t, broker)
+		err = broker.Close()
+		if err != nil {
+			t.Error(err)
+		}
+		// Wait up to 500 ms for the remote broker to process requests
+		// in order to have consistent metrics
+		if err := mb.WaitForExpectations(500 * time.Millisecond); err != nil {
```
eapache: I'm not sure why this is necessary, isn't the request guaranteed to be fully processed by the time `tt.runner` finishes?

config.go
```diff
@@ -233,6 +235,10 @@ type Config struct {
 	// latest features. Setting it to a version greater than you are actually
 	// running may lead to random breakage.
 	Version KafkaVersion
+	// The registry to define metrics into.
+	// Defaults to metrics.DefaultRegistry.
+	// See Examples on how to use the metrics registry
+	MetricRegistry metrics.Registry
```
eapache: Can this be null? Should it be, for users who don't want metrics? If not, should we validate that?

metrics.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"fmt"
+
+	"github.com/rcrowley/go-metrics"
+)
+
+func getOrRegisterHistogram(name string, r metrics.Registry) metrics.Histogram {
+	return r.GetOrRegister(name, func() metrics.Histogram {
+		return metrics.NewHistogram(metrics.NewExpDecaySample(1028, 0.015))
```
eapache: Can these numbers be pulled out into consts so that they're slightly less magic?

mockbroker.go
```diff
@@ -50,18 +51,22 @@ type MockBroker struct {
...
 	stopper      chan none
 	expectations chan encoder
+	done         sync.WaitGroup
 	listener     net.Listener
 	t            TestReporter
 	latency      time.Duration
 	handler      requestHandlerFunc
-	history      []RequestResponse
+	origHandler  bool
+	history      []*RequestResponse
```
eapache: Why do we suddenly need to store pointers here?

mockbroker.go
```diff
@@ -192,8 +215,10 @@ func (b *MockBroker) handleRequests(conn net.Conn, idx int, wg *sync.WaitGroup)
 		}
 
 		b.lock.Lock()
+		originalHandlerUsed := b.origHandler
```
eapache: Not a huge fan of all this code:
- I'm not sure what it's for, the tests already failed if an expectation was not consumed?
- Is there a way to encapsulate the expectation-counting code inside the handler (perhaps via a closure) so that we don't have to track whether we're using the original or not, the handler just does the right thing?

eapache:
Thanks for all your work on this! Most of the code review points are pretty minor, it's just the mockbroker changes I'm not really convinced of. Apologies again for the delay.

broker_test.go
slaunay: I use a dedicated `MockBroker` for each runner otherwise the statistics to validate metrics against would contains data from previous runs.

But I found a race condition that often manifests itself for unacknowledged produce request (i.e. `RequiredAcks` set to `NoResponse`) but can happen for standard request too when using a `MockBroker` for each runner.
That is the runner would return early because the produce request has been written to the socket and invoking `MockBroker.Close()` gets invoked before `MockBroker` is able to process that request (the request might not even have been read by `MockBroker` but already acknowledged by the TCP stack) and therefore consume from the expectations channel leading to an error about expectation not being satisfied or that the statistics about the bytes written by the `MockBroker` being 0.

On a regular produce request, the runner would block on reading the response and by then it is very likely that the `MockBroker` has consumed his expectations as well as update his statistics but there is a very tight window where the runner would be reading the statistics to validate the metrics before the `MockBroker` updates them.

My first fix was to add a magic 100 ms sleep for the `ProduceRequest (NoResponse)` but like every magic sleep that was futile and then I believe I ran into the second case once probably because the integration server being loaded.
The working fix I came with consists of adding that `WaitForExpectations` method to be sure that we received and processed the expected requests including updating statistics.
I added a timeout of 500ms to be sure that it would not blocked on failing test or very loaded integration server, it generally finishes under a couple of ms.

I guess it would be good to document those scenarios better on both `broker_test.go` and `mockbroker.go`.

mockbroker.go
```diff
@@ -207,19 +232,24 @@ func (b *MockBroker) handleRequests(conn net.Conn, idx int, wg *sync.WaitGroup)
...
+			if _, err = conn.Write(resHeader); err != nil {
+				b.serverError(err)
+				break
+			}
+			if _, err = conn.Write(encodedRes); err != nil {
+				b.serverError(err)
+				break
+			}
+			b.lock.Lock()
+			requestResponse.ResponseSize = len(resHeader) + len(encodedRes)
```
slaunay: By using a reference to a `RequestResponse` I am able to update the `ResponseSize` field without keeping the index of the entry that was added to the slice and then updating the entry later.

config.go
slaunay: I think it must not be null as it would panic when invoking methods on it.
Although we could add `if ... != nil` conditions when using the registry I think the official way is to set `metrics.UseNilMetrics` to `true`:
https://godoc.org/github.com/rcrowley/go-metrics#pkg-variables

We can have that setting the default but I remember the overhead being minimal during my tests and when you need high performance you generally wants the metrics to capture those performance.
The Java client has metrics turn on by default too.

I can update some benchmarks to see the impact of metrics being enabled vs disabled.

mockbroker.go
slaunay: It is because of the way I am updating the `ResponseSize` field, I first add an entry to the `history` field with the default value 0 and keep a reference to that `RequestResponse` before potentially updating it at the end of the handling.

mockbroker.go
slaunay: It makes sense to have that logic into the handler but the current handler profile do not provide the request size.
Changing the handler function profile to provide such information would help but it will break backward compatibility.
But then we would need the response size as well, I guess it can be computed in the handler before returning it but in case the Write fails the data would be inconsistent.

The `origHandler` field is used for signaling when all the expected requests have been fully processed through a `WaitGroup`.

Maybe having just an optional notifier function called after a request has been handled is the way to go to separate concerns, signal when requests have been processed and avoid braking compatibility:
```
type requestHandledNotifierFunc func(err error, int bytesRead, int bytesWritten)
```

config.go
eapache: Benchmarks would be nice, but leaving it enabled by default is fine as long as the impact isn't huge.

broker_test.go
eapache: The race condition for unacknowledged produce requests is a very nice catch but I'm not sure I understand how it can possibly manifest for any other requests? The runner will not return in those cases until the mock broker has actually responded, which means it has to have consumed the expectation?

If I am understanding the other issue properly, it is because the mockbroker updates its metrics *after* sending the response, so the test can get 0 metrics if it gets run in that gap? Would not the simpler fix be to block `mb.Close()` until all metrics are updated?

mockbroker.go
eapache: You don't even have to keep the index though do you? It's always just the last element in the slice.

mockbroker.go
eapache: An optional callback function does feel cleaner to me, I think.

broker_test.go
slaunay: You are right, it should not happen and `mb.Close()` is the proper fix for having consistent metrics for regular requests and that is actually the way it is done in the last PR commit.

I remembered that corner case while retracing my commits because I got bit by it when reusing original `defer mb.Close()` that would be executing after checking the metrics (see intermediate commit f9642ad), sorry for the confusion.

mockbroker.go
slaunay: It would be if the responses are served in serial order but I do not know if this is always the case (third parties using the `MockBroker` concurrently).
But the cleaner implementation through a handler should discard that detail I believe.

slaunay:
Here are the results from the benchmark when running them twice on the original master commit, against the PR with the no-op implementation and then with the effective implementation:
```
$ git checkout daa549b755c85cf347258a4c22e0baba1e8d2cc2
$ go test -v -bench=. -run=^$
PASS
BenchmarkProducerSmall-8               	  200000	      9338 ns/op
BenchmarkProducerMedium-8              	   50000	     38441 ns/op
BenchmarkProducerLarge-8               	   10000	    259523 ns/op
BenchmarkProducerSmallSinglePartition-8	  200000	      7422 ns/op
BenchmarkProducerMediumSnappy-8        	  200000	     11344 ns/op
ok  	github.com/Shopify/sarama	14.704s
$ go test -v -bench=. -run=^$
PASS
BenchmarkProducerSmall-8               	  100000	     10699 ns/op
BenchmarkProducerMedium-8              	   30000	     42789 ns/op
BenchmarkProducerLarge-8               	    5000	    259906 ns/op
BenchmarkProducerSmallSinglePartition-8	  200000	      5422 ns/op
BenchmarkProducerMediumSnappy-8        	  100000	     11666 ns/op
ok  	github.com/Shopify/sarama	10.226s

$ git checkout enhancement/broker-metrics 
$ export METRICS_DISABLE=1
$ go test -v -bench=. -run=^$
PASS
BenchmarkProducerSmall-8               	  200000	      9156 ns/op
BenchmarkProducerMedium-8              	   30000	     38275 ns/op
BenchmarkProducerLarge-8               	    5000	    240086 ns/op
BenchmarkProducerSmallSinglePartition-8	  200000	      6304 ns/op
BenchmarkProducerMediumSnappy-8        	  100000	     10657 ns/op
ok  	github.com/Shopify/sarama	9.356s
$ go test -v -bench=. -run=^$
PASS
BenchmarkProducerSmall-8               	  100000	     10308 ns/op
BenchmarkProducerMedium-8              	   30000	     38828 ns/op
BenchmarkProducerLarge-8               	    5000	    260416 ns/op
BenchmarkProducerSmallSinglePartition-8	  300000	      4705 ns/op
BenchmarkProducerMediumSnappy-8        	  200000	      9466 ns/op
ok  	github.com/Shopify/sarama	10.890s

$ unset METRICS_DISABLE
$ go test -v -bench=. -run=^$
PASS
BenchmarkProducerSmall-8               	  200000	      9181 ns/op
BenchmarkProducerMedium-8              	   30000	     40766 ns/op
BenchmarkProducerLarge-8               	    5000	    284455 ns/op
BenchmarkProducerSmallSinglePartition-8	  200000	      6352 ns/op
BenchmarkProducerMediumSnappy-8        	  100000	     11113 ns/op
ok  	github.com/Shopify/sarama	11.409s
$ go test -v -bench=. -run=^$
PASS
BenchmarkProducerSmall-8               	  200000	      8389 ns/op
BenchmarkProducerMedium-8              	   50000	     37401 ns/op
BenchmarkProducerLarge-8               	    5000	    265853 ns/op
BenchmarkProducerSmallSinglePartition-8	  300000	      5418 ns/op
BenchmarkProducerMediumSnappy-8        	  200000	     10612 ns/op
ok  	github.com/Shopify/sarama	12.507s
```

Results are very similar so it probably comes down to I/O and caching on my VM,  that's why I believe the overall impact is minimal.

broker.go
```diff
@@ -441,8 +475,9 @@ func (b *Broker) responseReceiver() {
 			continue
 		}
 
-		_, err = io.ReadFull(b.conn, header)
+		bytesReadHeader, err := io.ReadFull(b.conn, header)
 		if err != nil {
+			b.updateIncomingCommunicationMetrics(bytesReadHeader)
```
eapache: you could just do this immediately after calling `ReadFull` rather than in all three error branches?

eapache:
This is ready to go, thanks again for all your patience. I'll probably wait until I'm sure the consumer time issues (#730, #733, #734) have settled out, release 1.10.1 with that fix, then merge this.

broker.go
slaunay: Yes but then I would have to call it twice for the regular case where we read the header and then read the body.
This would result in the same metric for byte rate but will double the rate for response rate metric and mess up with response size histogram metric too.

That being said I think using `defer` with a `totalBytesRead` variable might simplify the code.

slaunay:
Glad to hear that, I think this feature is gonna lead to more optimizations and easier tuning in the future as we will be able to quantify those.

broker.go
slaunay: I forgot that this code is running inside a `for` loop so using `defer` would require wrapping the `for` code block inside an function, not sure if it makes it easier to read.

eapache: Oh, OK I missed the fact it would double-count packets in the success case. This is fine as-is.

PR #703, opened 2016-07-13
Update logo to not use HTTP, but also allows HTTPS

patrickdappollonio:
If you access `https://shopify.github.io/sarama/` (note HTTPS), the page loads the icon via HTTP, breaking the seal. This PR makes it so it'll pick either HTTP or HTTPS, like the javascripts files at the bottom of the page.


eapache:
Thanks!

PR #709, opened 2016-07-18
Fix HashPartitioner's negative result

ttapjinda:
Issue: I have found that for some Key, the Partition function of HashPartitioner return negative value. After I investigate the issue, I have found that if the fnv's Sum32 function returns 2147483648, which is -2147483648 in int32 and the negative of this will still be -2147483648 causing the negative result returned.

Reproduce: this issue can be reproduced by using "1468509572224" as message's key.

Change: 
- partitioner.go: Apply modulo before checking the negative value to ensure the positive result.
- partitioner_test.go: Add TestHashPartitionerMinInt32 to test the error case.

Please feel free to suggest the better way to fix this issue.

### DIFF
--- a/partitioner.go
+++ b/partitioner.go
@@ -112,10 +112,11 @@ func (p *hashPartitioner) Partition(message *ProducerMessage, numPartitions int3
 		return -1, err
 	}
 	hash := int32(p.hasher.Sum32())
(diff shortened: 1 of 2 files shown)


eapache:
Nice find, thanks!

PR #713, opened 2016-07-26
OffsetManager: match upstream mark/next behaviour

### LINKED ISSUE #705: sarama committed offsets don't match java consumer offsets
When we consume everything from kafka topic, kafka's tool (bin/kafka-consumer-groups.sh) still shows lag of 1 message. If we run java consumer with the same group it consumes 1 remaining message (which was already consumed by sarama consumer).

Does sarama use own offset schema (one less than java)?

##### Versions

Sarama Version: master
Kafka Version: 0.10
Go Version: 1.6.2

eapache:
Upstream requires you mark last-consumed+1 and then returns that value directly.
We were requiring you mark last-consumed and then adding one to the returned
value.

Match upstream's behaviour so that our offset tracking is interoperable.

Fixes #705.

@Kane-Sendgrid

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ features that may not be compatible with old Kafka versions. If you don't
 specify this value it will default to 0.8.2 (the minimum supported), and trying
 to use more recent features (like the offset manager) will fail with an error.
 
+_Also:_ The offset-manager's behaviour has been changed to match the upstream
+java consumer (see [#705](https://github.com/Shopify/sarama/pull/705) and
+[#713](https://github.com/Shopify/sarama/pull/713)). If you use the
+offset-manager, please ensure that you are committing one *greater* than the
+last consumed message offset or else you may end up consuming duplicate
+messages.
+
 New Features:
  - Support for Kafka 0.10
    ([#672](https://github.com/Shopify/sarama/pull/672),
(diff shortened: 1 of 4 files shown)


tcrayford:
@eapache does this mean that all consumers that use the offset manager that upgrade past this patch will drop a single message during the upgrade?

eapache:
It depends on what you were using before. If you were sharing the offset topic with an upstream java producer you were already potentially inconsistent because of the different expectations.

If you were using the previous Sarama behaviour as documented (storing the offset of the last-consumed message) then an upgrade should cause you to consume a single message twice (the previous code will consume X and mark X, then the new code will read X, consume X, and mark X+1). Your consumer should already be able to handle that safely because it can occur in other scenarios.

Note that either way you should ensure that code using the behaviour in this PR commits `last-consumed + 1` instead of just `last-consumed`. Otherwise you will see duplicate messages consumed every time your consumer restarts.

eapache:
I've added a more explicit warning to the changelog, but since consumers should already be able to handle duplicates (and this change will at worst cause a single duplicate message when a consumer restarts) I'm not too worried.

eapache:
cc @dim FYI - I don't think sarama-cluster is using our offset-manager but I know you took some inspiration from it so you might have the same issue.

tcrayford:
@eapache ok, that seems fine

dim:
@eapache thanks for the heads up - fixed in https://github.com/bsm/sarama-cluster/commit/6fdf904abaa5808f155458033d50feec6c1af8a9

PR #716, opened 2016-07-29
Fix download link in gh-pages

carsonip:
The current download link in gh-pages is for downloading the zip of gh-pages branch, which should not be intended.


eapache:
Nice catch, thanks!

PR #721, opened 2016-08-08
add relative offset handling for 0.10+ compressed msgs

dynamix:
attempted fix for https://github.com/Shopify/sarama/issues/720

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -479,14 +479,18 @@ func (child *partitionConsumer) parseResponse(response *FetchResponse) ([]*Consu
 	prelude := true
 	var messages []*ConsumerMessage
 	for _, msgBlock := range block.MsgSet.Messages {
-
 		for _, msg := range msgBlock.Messages() {
-			if prelude && msg.Offset < child.offset {
+			offset := msg.Offset
+			if child.conf.Version.IsAtLeast(V0_10_0_0) {
+				// calculate correct offset from relative offset
+				offset = msgBlock.Offset - int64(len(msgBlock.Messages())) + msg.Offset + 1
+			}
+			if prelude && offset < child.offset {
 				continue
 			}
 			prelude = false
 
-			if msg.Offset >= child.offset {
+			if offset >= child.offset {
 				messages = append(messages, &ConsumerMessage{
 					Topic:     child.topic,
 					Partition: child.partition,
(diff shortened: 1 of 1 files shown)


consumer.go
```diff
@@ -479,23 +479,27 @@ func (child *partitionConsumer) parseResponse(response *FetchResponse) ([]*Consu
 	prelude := true
 	var messages []*ConsumerMessage
 	for _, msgBlock := range block.MsgSet.Messages {
-
 		for _, msg := range msgBlock.Messages() {
-			if prelude && msg.Offset < child.offset {
+			offset := msg.Offset
+			if child.conf.Version.IsAtLeast(V0_10_0_0) {
```
eapache: If I'm reading the docs properly, this condition should be based on `msg.Msg.Version` since 0.10 brokers can still use the old format in some configurations.

consumer.go
```diff
@@ -479,23 +479,27 @@ func (child *partitionConsumer) parseResponse(response *FetchResponse) ([]*Consu
...
+				// calculate correct offset from relative offset
+				offset = msgBlock.Offset - int64(len(msgBlock.Messages())) + msg.Offset + 1
```
eapache: Per the comment in the java version (https://github.com/apache/kafka/blob/e49b3aee6b25b9b26275849c693d5dec41d6cf68/clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java#L248-L249) and a careful reading of the KIP that @edenhill linked, I don't know if we can just assume that the offset of the final message is equal to the number of messages in the set? It sounds like the broker is supposed to re-assign discontiguous compressed relative offsets, but the java client handles them being discontiguous anyway?

Tangentially, having the base outer offset be the offset of the *last* message instead of the *first* message feels really bizarre. That's definitely the way it works, but I can't for the life of me figure out why.

eapache:
Thanks for figuring this out, and sorry for the delay, I've been unfortunately busy with other projects recently. I left a few notes to ensure we handle all the cases properly, but I think this is effectively the right fix.

consumer.go
tcrayford: I think compaction of compressed messages will play weirdly here as well.

consumer.go
dynamix: good point - will change this

consumer.go
dynamix: I think I need to read through that again as well :-)

eapache:
@dynamix I opened #735 which incorporates my most up-to-date understanding of all of these corner cases. Please let me know if that works for you.

dynamix:
#735 does is indeed the correct fix. 👍

PR #724, opened 2016-08-10
newmockresponses.go

chandradeepak:
with new kafka version V0_10_0_0 if you are using sarama-cluster and want to mock the consumer. there are no mock responses which are mandatory. those are added now.


chandradeepak:
will fix the tests and reopen again.

PR #727, opened 2016-08-14
Expose Encode in encode_decode.go

chenziliang:
Hi,

I would like to expose a public `Encode` interface to allow low level clients to do something like below. Basically low level clients would like to create their own Producer based on sarama low level components.

When producing message to Kafka:
1) Create a sarama.MessageSet, add sarama.MessageBlock
2) Call this new `sarama.Encode` against the MessageSet in 1) to generate encoded payload
3) Add the encoded payload to sarama.ProduceRequest
4) Call broker.Produce(req) against req in 3) to produce the message to Kafka

For now, sarama.encode in encoder_decoder.go doesn't expose to clients.

### DIFF
--- a/encoder_decoder.go
+++ b/encoder_decoder.go
@@ -8,6 +8,10 @@ type encoder interface {
 	encode(pe packetEncoder) error
 }
 
+func Encode(e encoder) ([]byte, error) {
+	return encode(e)
+}
+
 // Encode takes an Encoder and turns it into bytes.
 func encode(e encoder) ([]byte, error) {
 	if e == nil {


eapache:
Hmm, I understand the intent, but the encoding framework was never intended to be public and if we expose `encode` like this we really need to expose `encoder` which means we need to expose `packetEncoder` and it just becomes a mess.

As far as I know, compressed produce requests are the only case where a low-level client needs to be able to get raw bytes back. Since the consumer response structs already handle compressed sets transparently, I think it would be better to do the same for producing by adding an `AddCompressedSet()` method to the `ProduceRequest` which takes the set and the compression key and does the right thing.

Tangentially: I am curious what the sarama producer does (or doesn't do) which requires you to write your own?

chenziliang:
@eapache Thanks very much for the reply. I agree exposing encoder is kinda hacky. I will take a look at your suggestion of adding 'AddCompressedSet' method.

Why i am writing low level consumer/producer even though sarams provides good implementation ?
What I would like to do is multiplexing requests to different topic/partition by using only one consumer / producer, and the low level consumer and producer are totally stateless and consumer/producer are not bound to any particular topic/partition when init. 
For instance:
> consumer := NewKafkaLowLevelConsumer(...) 
> results  := consumer.Read(topic, partition, ....)
>
> producer := NewKafkaLowLevelProducer(...)
> producer.Write(topic, events, ...)

PR #729, opened 2016-08-15
SASL/Kerberos using SASL2 library

rickyspeak:
I wrote a basic way of using the SASL2 library to do SASL/Kerberos authentication.  I added values to the Config to enable the feature.   Please take a look and tell me how you'd like it modified. I think this would be very useful to people.  Currently this only works for principles in the ticket cache.


eapache:
I am not entirely fond of using a C dependency for this; is there a Go equivalent library we could use instead?

If the C library is the only choice, we should at the very least make it optional so that sarama continues to work on platforms where the library is not available, and in cases where it simply is not installed.

rickyspeak:
I understand your concern.  Going with sasl2 was literally the last thing I wanted to do, but all other options failed to authenticate.  There is no go only implementation that I can find.  I'm relatively new to go, but I think I could refactor this an extension of your library.  The only internal thing that I need to change is to replace your (Broker *) Open() function.

eapache:
Closing as stale.

PR #733, opened 2016-08-23
Fix Racy call to Reset in feederLoop

### LINKED ISSUE #732: Performance Regression on Master
##### Versions

Sarama Version: 4f47ee4924e147c781bf3f9aee2f7e9845dc2532
Kafka Version: 0.8.2.1
Go Version: 1.6

##### Problem Description

We build off Sarama's master for various reasons that were needed at one point but probably not now. Everything has been smooth up until a week ago, where performance seemed to fall off a cliff for us (from 60k msg/s to 3k msg/s).

I began to notice it started to take almost *exactly* 100ms for us to retrieve a new message off the Consumer channel (which is a problem for us because it takes ~3ms for us to process a message). Since it took 100ms, I guessed the problem was with `MaxProcessingTime`, and seeing how a recent change included this value, we reverted to an earlier commit and that fixed it.  

I traced it back to this commit: 906ed729ae756fc01e0c9c9d78b31a2840022022 so there seems to be a logic issue in how this was implemented.

nemothekid:
Checking Reset()'s active return value is racy according to [Golang docs](https://golang.org/pkg/time/#Timer.Reset). 

>Note that it is not possible to use Reset's return value correctly, as there is a race condition between draining the channel and the new timer expiring. Reset should always be used in concert with Stop, as described above. The return value exists to preserve compatibility with existing programs.

Docs advice to call Stop in order to reuse a timer. Should fix #732.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -420,10 +420,11 @@ feederLoop:
 		msgs, child.responseResult = child.parseResponse(response)
 
 		for i, msg := range msgs {
-			if !expiryTimer.Reset(child.conf.Consumer.MaxProcessingTime) {
+			if !expiryTimer.Stop() {
 				// expiryTimer was expired; clear out the waiting msg
 				<-expiryTimer.C
 			}
+			expiryTimer.Reset(child.conf.Consumer.MaxProcessingTime)
 
 			select {
 			case child.messages <- msg:


eapache:
Thanks! Stuff like this makes me wish Go had some sort of deprecation flag so the compiler would have warned when using the return value of `Reset`.

nemothekid:
I believe this patch is problematic as well - this is now happening to me in prod:

```
goroutine 273 [chan receive, 10 minutes]:
github.com/Shopify/sarama.(*partitionConsumer).responseFeeder(0xc820bb3880)
	/go/src/github.com/Shopify/sarama/consumer.go:425 +0x1da
github.com/Shopify/sarama.(*partitionConsumer).(github.com/Shopify/sarama.responseFeeder)-fm()
	/go/src/github.com/Shopify/sarama/consumer.go:158 +0x20
github.com/Shopify/sarama.withRecover(0xc820c4c060)
	/go/src/github.com/Shopify/sarama/utils.go:46 +0x3a
created by github.com/Shopify/sarama.(*consumer).ConsumePartition
	/go/src/github.com/Shopify/sarama/consumer.go:158 +0x4a8
```

PR #744, opened 2016-09-07
Use a local registry for metrics

slaunay:
- use a new dedicated registry per config instead of using the default global registry provided by go-metrics
- update documentation and unit tests

### DIFF
--- a/broker_test.go
+++ b/broker_test.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"testing"
 	"time"
-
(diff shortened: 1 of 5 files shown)


eapache:
@justinas does this work for you?

justinas:
@eapache Absolutely. Package-global registry would be another alternative, but that's up to you :)

eapache:
Trying to stay away from package globals going forward.

PR #756, opened 2016-10-04
add missing metadata and assignment methods in describegroups

pcallewaert:
The code did already exists in sync_group_response.go, but added them also in describe_groups_response.go.
Now you can read the MemberAssignment and MemberMetadata also (before it was raw bytes, and not that easy to get the actual structs)

### DIFF
--- a/describe_groups_response.go
+++ b/describe_groups_response.go
@@ -172,3 +172,15 @@ func (gmd *GroupMemberDescription) decode(pd packetDecoder) (err error) {
 
 	return nil
 }
+
+func (r *GroupMemberDescription) GetMemberAssignment() (*ConsumerGroupMemberAssignment, error) {
+	assignment := new(ConsumerGroupMemberAssignment)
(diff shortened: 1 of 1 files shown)


wvanbergen:
Can you add a unit test for this?

eapache:
Also please be consistent with the variable name you assign the `GroupMemberDescription`; the rest of the file uses `gmd`.

pcallewaert:
@eapache my bad, fixed

@wvanbergen I started adding them, but the encoder/decoder is already tested in consumer_group_members_test.go. This is also being used in the SyncGroupResponse. In SyncGroupResponse only the bytes are tested, so following this practice I did not add them.

pcallewaert:
Any update on this? :)

eapache:
Sorry about that. Looks fine to me. @wvanbergen ?

wvanbergen:
:+1: - I'd still like to see some tests for this but if we're currently not doing that elsewhere either, I am OK with leaving that for now.

Issue #758, opened 2016-10-04
Timestamp in ProducerMessage is incompatible with compression


AlbinOS:
##### Versions

Sarama Version: master
Kafka Version: 0.10.0.1
Go Version: 1.7

##### Configuration

Default for Sarama and Kafka

##### Logs

Sarama producer : 

```go
ERRO[0000] Failed to write entry:kafka: Failed to produce message to topic cyclop: kafka server: Unexpected (unknown?) server error.
```

Kafka :

```java
java.lang.IllegalStateException: Compressed message has magic value 0 but inner message has magic value 1
```

##### Problem Description

If I use the following Producer configuration, message are not accepted by Kafka :

```go
	config := sarama.NewConfig()
	config.Producer.RequiredAcks = sarama.WaitForLocal       // Only wait for the leader to ack
	config.Producer.Compression = sarama.CompressionSnappy   // Compress messages
	config.Producer.Flush.Frequency = 500 * time.Millisecond // Flush batches every 500ms
	config.Version = sarama.V0_10_0_0

	producer, err := sarama.NewAsyncProducer(brokerList, config)
	if err != nil {
		return nil, err
	}

       producer.Input() <- &sarama.ProducerMessage{
		Topic:     topic,
		Value:     sarama.StringEncoder("foo"),
		Timestamp: time.Now(),
	}
```

According to @eapache in #750, the problem lies in the usage of compression in the producer compression.

By removing `config.Producer.Compression = sarama.CompressionSnappy`, messages are indeed accepted by Kafka and the timestamp is well set in the message.

AlbinOS:
Duplicate of #757, sorry for this, I was opening this in reaction of #750 latest comments without looking at the issue list.

PR #759, opened 2016-10-05
Fix compression timestamps

### LINKED ISSUE #757: Can't produce compressed messages with timestamp
##### Versions

Sarama Version: [Sarama af0513c](https://github.com/Shopify/sarama/commit/af0513c4d97979eb876d201391fac6475ec08268)
Kafka Version: 0.10.0.1
Go Version: 1.7

##### Configuration

I've got a example snippet over [here](https://gist.github.com/niels-s/2381bca40d006fbf75c1e2cbfddfdbbc)

For Kafka I'm using the following configuration: 

```
INFO KafkaConfig values:
	advertised.host.name = null
	metric.reporters = []
	quota.producer.default = 9223372036854775807
	offsets.topic.num.partitions = 50
	log.flush.interval.messages = 9223372036854775807
	auto.create.topics.enable = true
	controller.socket.timeout.ms = 30000
	log.flush.interval.ms = null
	principal.builder.class = class org.apache.kafka.common.security.auth.DefaultPrincipalBuilder
	replica.socket.receive.buffer.bytes = 65536
	min.insync.replicas = 1
	replica.fetch.wait.max.ms = 500
	num.recovery.threads.per.data.dir = 1
	ssl.keystore.type = JKS
	sasl.mechanism.inter.broker.protocol = GSSAPI
	default.replication.factor = 1
	ssl.truststore.password = null
	log.preallocate = false
	sasl.kerberos.principal.to.local.rules = [DEFAULT]
	fetch.purgatory.purge.interval.requests = 1000
	ssl.endpoint.identification.algorithm = null
	replica.socket.timeout.ms = 30000
	message.max.bytes = 1000012
	num.io.threads = 8
	offsets.commit.required.acks = -1
	log.flush.offset.checkpoint.interval.ms = 60000
	delete.topic.enable = false
	quota.window.size.seconds = 1
	ssl.truststore.type = JKS
	offsets.commit.timeout.ms = 5000
	quota.window.num = 11
	zookeeper.connect = localhost:2181
	authorizer.class.name =
	num.replica.fetchers = 1
	log.retention.ms = null
	log.roll.jitter.hours = 0
	log.cleaner.enable = true
	offsets.load.buffer.size = 5242880
	log.cleaner.delete.retention.ms = 86400000
	ssl.client.auth = none
	controlled.shutdown.max.retries = 3
	queued.max.requests = 500
	offsets.topic.replication.factor = 3
	log.cleaner.threads = 1
	sasl.kerberos.service.name = null
	sasl.kerberos.ticket.renew.jitter = 0.05
	socket.request.max.bytes = 104857600
	ssl.trustmanager.algorithm = PKIX
	zookeeper.session.timeout.ms = 6000
	log.retention.bytes = -1
	log.message.timestamp.type = CreateTime
	sasl.kerberos.min.time.before.relogin = 60000
	zookeeper.set.acl = false
	connections.max.idle.ms = 600000
	offsets.retention.minutes = 1440
	replica.fetch.backoff.ms = 1000
	inter.broker.protocol.version = 0.10.0-IV1
	log.retention.hours = 168
	num.partitions = 1
	broker.id.generation.enable = true
	listeners = null
	ssl.provider = null
	ssl.enabled.protocols = [TLSv1.2, TLSv1.1, TLSv1]
	log.roll.ms = null
	log.flush.scheduler.interval.ms = 9223372036854775807
	ssl.cipher.suites = null
	log.index.size.max.bytes = 10485760
	ssl.keymanager.algorithm = SunX509
	security.inter.broker.protocol = PLAINTEXT
	replica.fetch.max.bytes = 1048576
	advertised.port = null
	log.cleaner.dedupe.buffer.size = 134217728
	replica.high.watermark.checkpoint.interval.ms = 5000
	log.cleaner.io.buffer.size = 524288
	sasl.kerberos.ticket.renew.window.factor = 0.8
	zookeeper.connection.timeout.ms = 6000
	controlled.shutdown.retry.backoff.ms = 5000
	log.roll.hours = 168
	log.cleanup.policy = delete
	host.name =
	log.roll.jitter.ms = null
	max.connections.per.ip = 2147483647
	offsets.topic.segment.bytes = 104857600
	background.threads = 10
	quota.consumer.default = 9223372036854775807
	request.timeout.ms = 30000
	log.message.format.version = 0.10.0-IV1
	log.index.interval.bytes = 4096
	log.dir = /tmp/kafka-logs
	log.segment.bytes = 1073741824
	log.cleaner.backoff.ms = 15000
	offset.metadata.max.bytes = 4096
	ssl.truststore.location = null
	group.max.session.timeout.ms = 300000
	ssl.keystore.password = null
	zookeeper.sync.time.ms = 2000
	port = 9092
	log.retention.minutes = null
	log.segment.delete.delay.ms = 60000
	log.dirs = /tmp/kafka-logs
	controlled.shutdown.enable = true
	compression.type = producer
	max.connections.per.ip.overrides =
	log.message.timestamp.difference.max.ms = 9223372036854775807
	sasl.kerberos.kinit.cmd = /usr/bin/kinit
	log.cleaner.io.max.bytes.per.second = 1.7976931348623157E308
	auto.leader.rebalance.enable = true
	leader.imbalance.check.interval.seconds = 300
	log.cleaner.min.cleanable.ratio = 0.5
	replica.lag.time.max.ms = 10000
	num.network.threads = 3
	ssl.key.password = null
	reserved.broker.max.id = 1000
	metrics.num.samples = 2
	socket.send.buffer.bytes = 102400
	ssl.protocol = TLS
	socket.receive.buffer.bytes = 102400
	ssl.keystore.location = null
	replica.fetch.min.bytes = 1
	broker.rack = null
	unclean.leader.election.enable = true
	sasl.enabled.mechanisms = [GSSAPI]
	group.min.session.timeout.ms = 6000
	log.cleaner.io.buffer.load.factor = 0.9
	offsets.retention.check.interval.ms = 600000
	producer.purgatory.purge.interval.requests = 1000
	metrics.sample.window.ms = 30000
	broker.id = 0
	offsets.topic.compression.codec = 0
	log.retention.check.interval.ms = 300000
	advertised.listeners = null
	leader.imbalance.per.broker.percentage = 10
```

##### Logs

```
[sarama] 2016/10/04 16:11:36 Initializing new client
[sarama] 2016/10/04 16:11:36 client/metadata fetching metadata for all topics from broker localhost:9092
[sarama] 2016/10/04 16:11:36 Connected to broker at localhost:9092 (unregistered)
[sarama] 2016/10/04 16:11:36 client/brokers registered new broker #0 at 192.168.8.101:9092
[sarama] 2016/10/04 16:11:36 Successfully initialized new client
[sarama] 2016/10/04 16:11:36 Producer shutting down.
[sarama] 2016/10/04 16:11:36 producer/broker/0 starting up
[sarama] 2016/10/04 16:11:36 producer/broker/0 state change to [open] on bla/0
[sarama] 2016/10/04 16:11:36 Connected to broker at 192.168.8.101:9092 (registered as #0)
[sarama] 2016/10/04 16:11:36 Closing Client
```

Kafka logging

```
[2016-10-04 15:50:52,568] ERROR [Replica Manager on Broker 0]: Error processing append operation on partition test-1 (kafka.server.ReplicaManager)
java.lang.IllegalStateException: Compressed message has magic value 0 but inner message has magic value 1
	at kafka.message.ByteBufferMessageSet$$anon$1.readMessageFromStream(ByteBufferMessageSet.scala:143)
	at kafka.message.ByteBufferMessageSet$$anon$1.liftedTree2$1(ByteBufferMessageSet.scala:111)
	at kafka.message.ByteBufferMessageSet$$anon$1.<init>(ByteBufferMessageSet.scala:109)
	at kafka.message.ByteBufferMessageSet$.deepIterator(ByteBufferMessageSet.scala:85)
	at kafka.message.ByteBufferMessageSet$$anon$2.makeNextOuter(ByteBufferMessageSet.scala:367)
	at kafka.message.ByteBufferMessageSet$$anon$2.makeNext(ByteBufferMessageSet.scala:380)
	at kafka.message.ByteBufferMessageSet$$anon$2.makeNext(ByteBufferMessageSet.scala:335)
	at kafka.utils.IteratorTemplate.maybeComputeNext(IteratorTemplate.scala:64)
	at kafka.utils.IteratorTemplate.hasNext(IteratorTemplate.scala:56)
	at scala.collection.Iterator$class.foreach(Iterator.scala:893)
	at kafka.utils.IteratorTemplate.foreach(IteratorTemplate.scala:30)
	at kafka.message.ByteBufferMessageSet.validateMessagesAndAssignOffsets(ByteBufferMessageSet.scala:438)
	at kafka.log.Log.liftedTree1$1(Log.scala:339)
	at kafka.log.Log.append(Log.scala:338)
	at kafka.cluster.Partition$$anonfun$11.apply(Partition.scala:441)
	at kafka.cluster.Partition$$anonfun$11.apply(Partition.scala:427)
	at kafka.utils.CoreUtils$.inLock(CoreUtils.scala:231)
	at kafka.utils.CoreUtils$.inReadLock(CoreUtils.scala:237)
	at kafka.cluster.Partition.appendMessagesToLeader(Partition.scala:427)
	at kafka.server.ReplicaManager$$anonfun$appendToLocalLog$2.apply(ReplicaManager.scala:409)
	at kafka.server.ReplicaManager$$anonfun$appendToLocalLog$2.apply(ReplicaManager.scala:395)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.mutable.HashMap$$anonfun$foreach$1.apply(HashMap.scala:99)
	at scala.collection.mutable.HashMap$$anonfun$foreach$1.apply(HashMap.scala:99)
	at scala.collection.mutable.HashTable$class.foreachEntry(HashTable.scala:230)
	at scala.collection.mutable.HashMap.foreachEntry(HashMap.scala:40)
	at scala.collection.mutable.HashMap.foreach(HashMap.scala:99)
	at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at kafka.server.ReplicaManager.appendToLocalLog(ReplicaManager.scala:395)
	at kafka.server.ReplicaManager.appendMessages(ReplicaManager.scala:331)
	at kafka.server.KafkaApis.handleProducerRequest(KafkaApis.scala:405)
	at kafka.server.KafkaApis.handle(KafkaApis.scala:76)
	at kafka.server.KafkaRequestHandler.run(KafkaRequestHandler.scala:60)
	at java.lang.Thread.run(Thread.java:745)
```
##### Problem Description

When I try to produce a message with Timestamp set and compression enabled for the producer it will give an error! When I try producing without compression enabled everything goes well and the right timestamp is applied to the kafka message

rtreffer:
The wrapper message has to use the "new" format if the compressed messages use it as well.
This will fix #757 #758

### DIFF
--- a/produce_set.go
+++ b/produce_set.go
@@ -101,7 +101,7 @@ func (ps *produceSet) buildRequest() *ProduceRequest {
 				}
 				if ps.parent.conf.Version.IsAtLeast(V0_10_0_0) {
 					compMsg.Version = 1
-					compMsg.Timestamp = time.Now()
+					compMsg.Timestamp = set.setToSend.Messages[0].Msg.Timestamp
 				}
 				req.AddMessage(topic, partition, compMsg)
 			}
--- a/produce_set.go
+++ b/produce_set.go
@@ -53,8 +53,12 @@ func (ps *produceSet) add(msg *ProducerMessage) error {
 
 	set.msgs = append(set.msgs, msg)
 	msgToSend := &Message{Codec: CompressionNone, Key: key, Value: val}
-	if ps.parent.conf.Version.IsAtLeast(V0_10_0_0) && !msg.Timestamp.IsZero() {
-		msgToSend.Timestamp = msg.Timestamp
+	if ps.parent.conf.Version.IsAtLeast(V0_10_0_0) {
+		if msg.Timestamp.IsZero() {
+			msgToSend.Timestamp = time.Now()
+		} else {
+			msgToSend.Timestamp = msg.Timestamp
+		}
 		msgToSend.Version = 1
 	}
 	set.setToSend.addMessage(msgToSend)
@@ -96,21 +100,8 @@ func (ps *produceSet) buildRequest() *ProduceRequest {
 					Value: payload,
 				}
 				if ps.parent.conf.Version.IsAtLeast(V0_10_0_0) {
-					// Compressed messages must use a protocol version
-					// that is newer than the inner messages version.
-					// Due to a lack of better timestamp notation copy the oldest
-					// (earliest) timestamp to message.
-					for _, msgBlock := range set.setToSend.Messages {
-						msg := msgBlock.Msg
-						if msg.Version > compMsg.Version {
-							compMsg.Version = msg.Version
-						}
-						if !msg.Timestamp.IsZero() &&
-							(compMsg.Timestamp.IsZero() ||
-								compMsg.Timestamp.After(msg.Timestamp)) {
-							compMsg.Timestamp = msg.Timestamp
-						}
-					}
+					compMsg.Version = 1
+					compMsg.Timestamp = time.Now()
 				}
 				req.AddMessage(topic, partition, compMsg)
 			}
--- a/produce_set.go
+++ b/produce_set.go
@@ -90,11 +90,29 @@ func (ps *produceSet) buildRequest() *ProduceRequest {
 					Logger.Println(err) // if this happens, it's basically our fault.
 					panic(err)
 				}
-				req.AddMessage(topic, partition, &Message{
+				compMsg := &Message{
 					Codec: ps.parent.conf.Producer.Compression,
 					Key:   nil,
 					Value: payload,
-				})
+				}
+				if ps.parent.conf.Version.IsAtLeast(V0_10_0_0) {
+					// Compressed messages must use a protocol version
+					// that is newer than the inner messages version.
+					// Due to a lack of better timestamp notation copy the oldest
+					// (earliest) timestamp to message.
+					for _, msgBlock := range set.setToSend.Messages {
+						msg := msgBlock.Msg
+						if msg.Version > compMsg.Version {
+							compMsg.Version = msg.Version
+						}
+						if !msg.Timestamp.IsZero() &&
+							(compMsg.Timestamp.IsZero() ||
+								compMsg.Timestamp.After(msg.Timestamp)) {
+							compMsg.Timestamp = msg.Timestamp
+						}
+					}
+				}
+				req.AddMessage(topic, partition, compMsg)
 			}
 		}
 	}
--- a/produce_set_test.go
+++ b/produce_set_test.go
@@ -141,3 +141,45 @@ func TestProduceSetRequestBuilding(t *testing.T) {
 		t.Error("Wrong number of topics in request")
 	}
 }
+
+func TestProduceSetCompressedRequestBuilding(t *testing.T) {
+	parent, ps := makeProduceSet()
+	parent.conf.Producer.RequiredAcks = WaitForAll
+	parent.conf.Producer.Timeout = 10 * time.Second
+	parent.conf.Producer.Compression = CompressionGZIP
+	parent.conf.Version = V0_10_0_0
+
+	msg := &ProducerMessage{
+		Topic:     "t1",
+		Partition: 0,
+		Key:       StringEncoder(TestMessage),
+		Value:     StringEncoder(TestMessage),
+		Timestamp: time.Now(),
+	}
+	for i := 0; i < 10; i++ {
+		safeAddMessage(t, ps, msg)
+	}
+
+	req := ps.buildRequest()
+
+	if req.Version != 2 {
+		t.Error("Wrong request version")
+	}
+
+	for _, msgBlock := range req.msgSets["t1"][0].Messages {
+		msg := msgBlock.Msg
+		err := msg.decodeSet()
+		if err != nil {
+			t.Error("Failed to decode set from payload")
+		}
+		for _, compMsgBlock := range msg.Set.Messages {
+			compMsg := compMsgBlock.Msg
+			if compMsg.Version != 1 {
+				t.Error("Wrong compressed message version")
+			}
+		}
+		if msg.Version != 1 {
+			t.Error("Wrong compressed parent message version")
+		}
+	}
+}


produce_set.go
```diff
@@ -90,11 +90,29 @@ func (ps *produceSet) buildRequest() *ProduceRequest {
...
-				req.AddMessage(topic, partition, &Message{
+				compMsg := &Message{
 					Codec: ps.parent.conf.Producer.Compression,
 					Key:   nil,
 					Value: payload,
-				})
+				}
+				if ps.parent.conf.Version.IsAtLeast(V0_10_0_0) {
+					// Compressed messages must use a protocol version
+					// that is newer than the inner messages version.
```
eapache: minor nit-pick: as I understand it (and as you've implemented here) the outer version has to be at least as new as the inner version, not strictly newer.

produce_set.go
```diff
@@ -90,11 +90,29 @@ func (ps *produceSet) buildRequest() *ProduceRequest {
...
+					// Due to a lack of better timestamp notation copy the oldest
+					// (earliest) timestamp to message.
+					for _, msgBlock := range set.setToSend.Messages {
+						msg := msgBlock.Msg
+						if msg.Version > compMsg.Version {
+							compMsg.Version = msg.Version
```
eapache: Hmm, I was thinking about better ways to accomplish this and I wonder now if the code in `produceSet#add` is even correct; is it safe to mix v0 and v1 messages in a single compressed batch which we'll do right now if some timestamps are 0? Or should we change to just always using v1 if the version supports it, and let the timestamps be 0 if necessary?

edit: actually this issue goes away if we provide a timestamp of `time.Now()` by default if you leave the timestamp as 0, which we should probably do for convenience anyway

rtreffer: I think always using v1 would be great, I don't see a reason to switch back at all. Providing a default of time.Now() sounds like a good idea, too.

rtreffer:
@eapache how about the updated version? Pushing message version 1 if possible and default timestamp of time.Now if unspecified?

eapache:
Definitely like this version better, thanks for adjusting it. Just one minor comment.

rtreffer:
@eapache you are right, the first timestamp would usually be a good proxy for the timestamp of the group.

I've updated the PR, I guess it needs a squash but github can do that on merge anyway.

eapache:
Thanks!

PR #764, opened 2016-10-11
re-partition messages on retry

nelhage:
If our partitioner doesn't require consistency, it's safe to do retries
on a different partition, and doing so can greatly improve availability
in the case of broker failures; Since we'll only try on partitions that
are known to have an active leader, we can route messages to a partition
homed on a healthy broker.

This is technically a behavior change, but it seems like a "safe" one in that `RequiresConsistency() == false` partitioners never had any strong guarantees on where messages would route in the case of failure.

I'm happy to write a test and/or hide this behavior behind a config flag if that'd be desired. But I think that the gain in availability when publishing is well-worth implementing this behavior in some form, so I'm opening this to start the discussion.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -300,7 +300,7 @@ func (p *asyncProducer) newTopicProducer(topic string) chan<- *ProducerMessage {
 
 func (tp *topicProducer) dispatch() {
 	for msg := range tp.input {
-		if msg.retries == 0 {
+		if msg.retries == 0 || (msg.flags == 0 && !tp.partitioner.RequiresConsistency()) {
 			if err := tp.partitionMessage(msg); err != nil {
 				tp.parent.returnError(msg, err)
 				continue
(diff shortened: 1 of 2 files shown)


eapache:
The (admittedly subtle) problem with this approach is that it breaks the producer's ordering guarantee in failure cases: if you put two messages into the producer and they end up on the same partition, they *must* end up on the broker in the same order as they were submitted to the producer.

eapache:
An actual example: consider brokers A,B,C and messages 1,2:
- user sends 1 and then 2 on the channel
- 1 is routed to A
- 2 is routed to B
- B fails fast, 2 is retried
- 2 is routed to C
- A fails with a timeout, 1 is retried
- 1 is routed to C

Broker C now contains message 2 before message 1, which violates the ordering guarantee.

eapache:
I would be happy to consider re-routing messages to other partitions if we can solve this problem, but the simple approach is unfortunately insufficient.

nelhage:
Yeah, I see the problem now. I was misunderstanding the ordering guarantee to be something weaker than it is. I'll give this some more thought.

PR #769, opened 2016-10-20
Add Consumer.HighWaterMarks()

mathpl:
This adds a way for a consumer user to access the PartitionConsumers high water marks.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -63,6 +63,9 @@ type Consumer interface {
 	// or OffsetOldest
 	ConsumePartition(topic string, partition int32, offset int64) (PartitionConsumer, error)
 
+	// HighWaterMarks returns the current high water marks for each topic and partitions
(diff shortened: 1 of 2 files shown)


eapache:
I think I'm OK with this, but it is worth noting that the data returned by this call is not guaranteed to be consistent between partitions. It's probably worth noting that caveat in the documentation.

consumer.go
```diff
@@ -63,6 +63,9 @@ type Consumer interface {
 	// or OffsetOldest
 	ConsumePartition(topic string, partition int32, offset int64) (PartitionConsumer, error)
 
+	// HighWaterMarks returns the current high water marks for each topic and partitions
```
eapache: grammatically, you don't need to pluralize "partitions"

mathpl:
Modified the comments and now the CI fails, not sure what to do.

PR #774, opened 2016-10-27
 Console Consumer Example

robertojrojas:
Please accept this example and added to your repo. I think it would help someone learn how to use the Consumer. Let me know if you need to make any changes.


eapache:
How is this different from `tools/kafka-console-consumer`?

robertojrojas:
@eapache thanks for reviewing this PR. You're right. They're similar. As matter of fact, I used the tools/kafka-console-consumer .sh as an inspiration for this. It's a scaled-down version of it, but, I think this could help someone "kick the tires" of this API with something they may be already familiar with. I'm open to any suggestions for changes.

eapache:
>  I used the tools/kafka-console-consumer.sh as an inspiration for this

I was referring to the go `kafka-console-consumer` program already in this repo. I am simply wondering if it might be better/simpler to improve the documentation and code style of that program rather than submitting a new one, since they are very similar.

robertojrojas:
You're right. It's better to improve the documentation. Closing this PR...

PR #786, opened 2016-11-22
Add lz4 compression

rtreffer:
This patch may cause issues with kafka < 0.10.0 as it doesn't handle "broken" framing.
It should thus only be used with 0.10+.

### DIFF
--- a/config.go
+++ b/config.go
@@ -375,9 +375,7 @@ func (c *Config) Validate() error {
 		return ConfigurationError("Producer.Retry.Backoff must be >= 0")
 	}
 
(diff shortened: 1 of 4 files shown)


rtreffer:
config.Validate checks for versions < V0_10 and errs if LZ4 is enabled. Will add a message test later today.

rtreffer:
@eapache encode / decode test, config check + tests.

rtreffer:
Hm, the failed test looks like a false positive to me.

eapache:
Ya, I've rerun it. Sometimes the broker just doesn't start properly on CI for some reason.

eapache:
Thanks!

PR #787, opened 2016-11-22
Producer: block in `Close` without `Return.Errors`

### LINKED ISSUE #784: AsyncProducer.Close() doesn't flush if conf.Producer.Return.Errors is false
##### Versions

Sarama Version: fd498173ae2bfc49a2b4eb19353241317785e7e6 (latest as of today)

##### Problem Description

The documentation says about Close():

> Close shuts down the producer and flushes any messages it may have buffered.

and about Errors():

> Alternatively, you can set Producer.Return.Errors in your config to false, which prevents errors to be returned.

To me, this means that if I don't care about errors, I can simply set `Return.Errors = false` with no ill-effects, then when I'm finished `Close()` the `AsyncProducer` and I expect it to block until it is flushed (as opposed to `AsyncClose()`.

This is not the case. `Close()` is simply a call to `AsyncClose()` followed by blocking on the `Errors()` channel, if and only if `Return.Errors` is true, if it is false then it simply won't block and won't flush if the program exits shortly after.

This is definitely not what I expect from the documentation. In my opinion, `Close()` should be implemented in such a way that it does actually block, regardless of the values of (what one should expect to be) orthogonal configuration variables.

Alternatively, a big caveat should be made clear in the doc: if `Return.Errors` is false then `Close()` won't block, you have to set `Return.Successes` to true and block on that yourself, and if `Return.Successes` is false also, then there is _no way_ to flush the producer reliably.

eapache:
If neither `Return.Errors` nor `Return.Successes` was set then `Close` wouldn't
actually block at all. This is arguably correct given the send-it-into-the-void-
and-pray configuration but it was still surprising.

It turns out to be a very minimal change to block anyway (since the `Errors`
channel is still present and closed at the appropriate moment) so just do that.
Also add a test for this configuration, since it's not one I remember to think
about on the regular.

@hchargois after poking around the code a bit I decided it was simpler to just fix this the way you suggested. Let me know if this works for you.

Closes #784.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -210,6 +210,8 @@ func (p *asyncProducer) Close() error {
 		for event := range p.errors {
 			errors = append(errors, event)
 		}
+	} else {
+		<-p.errors
 	}
 
 	if len(errors) > 0 {
--- a/async_producer_test.go
+++ b/async_producer_test.go
@@ -709,6 +709,46 @@ func TestAsyncProducerRetryShutdown(t *testing.T) {
 	}
 }
 
+func TestAsyncProducerNoReturns(t *testing.T) {
+	seedBroker := NewMockBroker(t, 1)
+	leader := NewMockBroker(t, 2)
+
+	metadataLeader := new(MetadataResponse)
+	metadataLeader.AddBroker(leader.Addr(), leader.BrokerID())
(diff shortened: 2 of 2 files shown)


hchargois:
This is awesome, I love how simple the solution was and it works beautifully. Thanks :)

PR #790, opened 2016-11-24
Sync Producer: Don't change config in constructor

### LINKED ISSUE #789: Race issue when using sarama.NewSyncProducerFromClient with a shared client
##### Versions

Sarama Version: gopkg.in/Shopify/sarama.v1      v1.10.1
Kafka Version: 0.10.0.1
Go Version: 1.7.3

##### Configuration

Default both for sarama and kafka

##### Problem Description

Our code uses shared sarama.Client instance in two goroutines. Each goroutine tries to create a sync publisher using the shared client. When we run our code with race detection enabled, we receive the following error:

```
WARNING: DATA RACE
Write at 0x00c4200788e0 by goroutine 61:
  vendor/gopkg.in/Shopify/sarama%2ev1.newSyncProducerFromAsyncProducer()
      /depot/src/vendor/gopkg.in/Shopify/sarama.v1/sync_producer.go:57 +0x61
  vendor/gopkg.in/Shopify/sarama%2ev1.NewSyncProducerFromClient()
      /depot/src/vendor/gopkg.in/Shopify/sarama.v1/sync_producer.go:53 +0xdf
...

Previous read at 0x00c4200788e0 by goroutine 89:
  vendor/gopkg.in/Shopify/sarama%2ev1.(*asyncProducer).returnSuccesses()
      /depot/src/vendor/gopkg.in/Shopify/sarama.v1/async_producer.go:843 +0x91
  vendor/gopkg.in/Shopify/sarama%2ev1.(*brokerProducer).handleSuccess.func1()
      /depot/src/vendor/gopkg.in/Shopify/sarama.v1/async_producer.go:738 +0x294
  vendor/gopkg.in/Shopify/sarama%2ev1.(*produceSet).eachPartition()
      /depot/src/vendor/gopkg.in/Shopify/sarama.v1/produce_set.go:108 +0x1e0
  vendor/gopkg.in/Shopify/sarama%2ev1.(*brokerProducer).handleSuccess()
      /depot/src/vendor/gopkg.in/Shopify/sarama.v1/async_producer.go:751 +0x90
  vendor/gopkg.in/Shopify/sarama%2ev1.(*brokerProducer).handleResponse()
      /depot/src/vendor/gopkg.in/Shopify/sarama.v1/async_producer.go:703 +0x2f0
  vendor/gopkg.in/Shopify/sarama%2ev1.(*brokerProducer).run()
      /depot/src/vendor/gopkg.in/Shopify/sarama.v1/async_producer.go:637 +0x1383
  vendor/gopkg.in/Shopify/sarama%2ev1.(*brokerProducer).(vendor/gopkg.in/Shopify/sarama%2ev1.run)-fm()
      /depot/src/vendor/gopkg.in/Shopify/sarama.v1/async_producer.go:536 +0x41
  vendor/gopkg.in/Shopify/sarama%2ev1.withRecover()
      /depot/src/vendor/gopkg.in/Shopify/sarama.v1/utils.go:46 +0x50
```

I believe the issue is caused by the fact that when NewSyncProducerFromClient creates an async producer, the producer sets its conf field to the shared client's config (see https://github.com/Shopify/sarama/blob/master/async_producer.go#L89). Then, newSyncProducerFromAsyncProducer function tries to record sync producer's specific config settings to the shared instance of the client's config (https://github.com/Shopify/sarama/blob/master/sync_producer.go#L57) and this causes the data race.

Could you please let us know if this is a bug in sarama or do we need to use the library in some special fashion (as far as I understand shared usage of a client reference is allowed by documentation).

Thank you

eapache:
It's surprising, for one thing, and as Alexander Zhuravlev discovered it can
actually lead to data races if you already have async producers active using the
same config object.

Instead, validate that it is pre-set correctly and return a config error if it
is not, so it is on the user to set things up properly (it's only one flag away
from default so it's not that big a deal).

Fixes #789.

cc @zaa

### DIFF
--- a/sync_producer.go
+++ b/sync_producer.go
@@ -9,6 +9,9 @@ import "sync"
 // The SyncProducer comes with two caveats: it will generally be less efficient than the AsyncProducer, and the actual
 // durability guarantee provided when a message is acknowledged depend on the configured value of `Producer.RequiredAcks`.
 // There are configurations where a message acknowledged by the SyncProducer can still sometimes be lost.
+//
+// For implementation reasons, the SyncProducer requires `Producer.Return.Errors` and `Producer.Return.Successes` to
+// be set to true in its configuration.
 type SyncProducer interface {
 
 	// SendMessage produces a given message, and returns only when it either has
@@ -36,6 +39,15 @@ type syncProducer struct {
 
 // NewSyncProducer creates a new SyncProducer using the given broker addresses and configuration.
 func NewSyncProducer(addrs []string, config *Config) (SyncProducer, error) {
+	if config == nil {
+		config = NewConfig()
+		config.Producer.Return.Successes = true
+	}
+
+	if err := verifyProducerConfig(config); err != nil {
+		return nil, err
+	}
+
 	p, err := NewAsyncProducer(addrs, config)
 	if err != nil {
 		return nil, err
@@ -46,6 +58,10 @@ func NewSyncProducer(addrs []string, config *Config) (SyncProducer, error) {
 // NewSyncProducerFromClient creates a new SyncProducer using the given client. It is still
 // necessary to call Close() on the underlying client when shutting down this producer.
 func NewSyncProducerFromClient(client Client) (SyncProducer, error) {
+	if err := verifyProducerConfig(client.Config()); err != nil {
+		return nil, err
+	}
+
 	p, err := NewAsyncProducerFromClient(client)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


wvanbergen:
There's no easy way to clone a config, and set the properties to `true` on the cloned value?

eapache:
We considered that in #789; there are a bunch of config values where it doesn't make any sense to have different instances of the config.

zaa:
@eapache Looks great, thanks.

PR #792, opened 2016-11-28
send errTimedOut to partition consumer

JohnnyLuoCM:
At the moment , when errTimeout happens, usually that means the client take too long to respond message , the behavior is silently stop consuming messages , which in my case it might happen , and that's not a good behavior,  instead of doing that, it will be better to dispatch the error message to the correct partition consumer, so it can take some appropriate actions.  

just one line changes, so it is quite straightforward.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -652,6 +652,7 @@ func (bc *brokerConsumer) handleResponses() {
 		case errTimedOut:
 			Logger.Printf("consumer/broker/%d abandoned subscription to %s/%d because consuming was taking too long\n",
 				bc.broker.ID(), child.topic, child.partition)
+			child.sendError(errTimedOut)
 			delete(bc.subscriptions, child)
 		case ErrOffsetOutOfRange:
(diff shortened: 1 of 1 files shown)


eapache:
> when errTimeout happens, usually that means the client take too long to respond message

`errTimedout` occurs when the user consumer code (i.e. the code you wrote to read from the channel of messages) is taking too long. This usually occurs if you are trying to make a request to some other service for each message, and that service is slow or failed.

> the behavior is silently stop consuming messages

No, the behaviour is to pause consuming messages until the user code is able to process the backlog of messages that have already been consumed.

> it will be better to dispatch the error message to the correct partition consumer

I think this is unlikely. If the user code is not reading from its channel quickly enough (or at all), then trying to send it *another* message is liable to make the problem worse.

If you are seeing a lot of the "consuming was taking too long" logs in your application, then you may have a bug, or you may simply want to adjust the `Consumer.MaxProcessingTime` configuration value.

PR #795, opened 2016-12-01
Fix segmentation fault on SASL failed

ueokande:
A segmentation violation occurs when initializing producer using SASL fails.


eapache:
Nice catch, thanks!

PR #796, opened 2016-12-08
add config setting

doubleceston:
in order for the Successes channel to be populated,have to set config.Producer.Return.Successes to true


eapache:
Nice catch, thanks

PR #811, opened 2017-01-24
Added preliminary support for create topics

cmshetty:


eapache:
Is this still something that just isn't ready yet?

cmshetty:
Yes. It isn't complete. I am hoping I can finish it and add some tests before creating a PR.

PR #813, opened 2017-01-25
make list of brokers available through sarama.Client

hollow:
Useful in general, but needed to iterate through all brokers and get consumer group offsets.

### DIFF
--- a/client.go
+++ b/client.go
@@ -17,6 +17,9 @@ type Client interface {
 	// altered after it has been created.
 	Config() *Config
 
+	// Brokers returns the current set of active brokers as retrieved from cluster metadata.
+	Brokers() []*Broker
+
 	// Topics returns the set of available topics as retrieved from cluster metadata.
 	Topics() ([]string, error)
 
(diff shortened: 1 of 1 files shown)


eapache:
> needed to iterate through all brokers and get consumer group offsets

I'm not sure I understand this use case. `Client.Coordinator` can already be used to retrieve the broker for a specific consumer group.

Tangential note: you should be able to use https://golang.org/pkg/builtin/#copy instead of manually looping over the array.

hollow:
> I'm not sure I understand this use case. Client.Coordinator can already be used to retrieve the broker for a specific consumer group.

The use case is to retrieve all consumer groups from all brokers with `Broker.ListGroups` without knowing the groups in advance.

> Tangential note: you should be able to use https://golang.org/pkg/builtin/#copy instead of manually looping over the array.

This only works with slices, but `client.brokers` is a map.

eapache:
> This only works with slices, but client.brokers is a map.

Oops, my mistake.

> The use case is to retrieve all consumer groups from all brokers with Broker.ListGroups without knowing the groups in advance.

Cool, makes sense. You need to hold the read lock (`client.lock.RLock()`) before reading `client.brokers`, otherwise this looks good.

hollow:
> You need to hold the read lock (`client.lock.RLock()`) before reading `client.brokers`

thanks, fixed :)

PR #814, opened 2017-01-25
Go back to hash/crc32

edsrzf:
As of Go 1.7, hash/crc32 has all the same optimizations. This means anyone on
an older Go version will see a performance hit, but anyone who cares about
performance should really be updating to Go 1.7 anyway.

### DIFF
--- a/crc32_field.go
+++ b/crc32_field.go
@@ -2,8 +2,7 @@ package sarama
 
 import (
 	"encoding/binary"
-
-	"github.com/klauspost/crc32"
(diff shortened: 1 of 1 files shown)


eapache:
We drop official support for go 1.6 two months after 1.8 releases, so I'll merge this then.

wvanbergen:
:+1:

eapache:
1.8 was released 2017-02-16, so we can drop 1.6 and merge this sometime after 2017-04-16.

PR #819, opened 2017-02-07
Add project LICENSE

DiSiqueira:
Adding a MIT LICENSE file to the project.


eapache:
We already have a license file (named `MIT-LICENSE`). There's an argument it should just be named `LICENSE` but I don't think we need two.

DiSiqueira:
@eapache I renamed MIT-LICENSE file to just LICENSE and removed the LICENSE file i created. If any other changes are needed, please let me know.

eapache:
Thanks!

PR #821, opened 2017-02-07
Formatting all files with gofmt -s

DiSiqueira:
Just runned gofmt -w -f in all files to get the simplified and formatted version of the code.


wvanbergen:
:+1: - we had to keep those around while we were supporting older go versions, but because we now no longer support them this is fine to update now.

PR #822, opened 2017-02-07
Changing some code to match golint coding style

DiSiqueira:
Executed golint on every file and followed the tips.

I will create another Pull Request with all comments related tips from golint.


DiSiqueira:
Thank you for your feedback @wvanbergen!! I will slip this in multiple PRs, making sure that don't break anything!

PR #826, opened 2017-02-09
Renaming local vars following tips from golint

DiSiqueira:


errors.go
```diff
@@ -96,8 +96,8 @@ const (
 	ErrInvalidRequiredAcks             KError = 21
 	ErrIllegalGeneration               KError = 22
 	ErrInconsistentGroupProtocol       KError = 23
-	ErrInvalidGroupId                  KError = 24
-	ErrUnknownMemberId                 KError = 25
```
wvanbergen: We cannot rename these - it breaks backwards compatibility.

wvanbergen:
I have mixed feelings. While many of the names of local variables can be corrected without issue, for the ones that are exported, we can't make those changes. This would make the code inconsistent - using both `Id` and `ID`. If we can't be consistent, I am not sure it's worth it.

eapache:
Local consistency is more important than golint in this case.

PR #827, opened 2017-02-09
Correcting two funcs comments

DiSiqueira:
This two funcs were with wrong names in their comments, just corrected this.


wvanbergen:
Thanks!

PR #828, opened 2017-02-09
Renaming structs following tips from GoLint

DiSiqueira:


wvanbergen:
Closing this for the same reason as https://github.com/Shopify/sarama/pull/829

PR #829, opened 2017-02-09
Renaming fields following tips from GoLint

DiSiqueira:


wvanbergen:
Because these types are exported (names are capitalized), this would break backwards compatibility. Following the style guide is not a good enough reason to warrant that.

PR #830, opened 2017-02-09
Removing unnecessary attribution

DiSiqueira:
Nil is already the default value for error vars, so removing it

### DIFF
--- a/mocks/mocks.go
+++ b/mocks/mocks.go
@@ -30,9 +30,9 @@ type ErrorReporter interface {
 type ValueChecker func(val []byte) error
 
 var (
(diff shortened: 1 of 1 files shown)


wvanbergen:
It is technically correct, but for clarity I prefer it being set the `nil` explicitly. 

For all intents and purposes, the variable is a constant - go just won't let us do that because of its type. By leaving off the `= nil`, it would imply it's a variable which will get assigned later in the code, which is not true.

So I am 👎 on this one. I'll defer to @eapache though.

eapache:
I prefer being explicit in this case as well.

Issue #834, opened 2017-02-23
AsyncProducer: i/o timeout error when message is big


agis-:
##### Versions

Sarama Version: 1416bd78f804d523005322194994f08c2a0ad797
Kafka Version: 0.10.1
Go Version: 1.8

##### Configuration

Sarama: *(see below)*

Kafka:
```
socket.send.buffer.bytes=1048576
socket.receive.buffer.bytes=1048576
replica.fetch.max.bytes=1048576
zookeeper.connection.timeout.ms=6000
zookeeper.session.timeout.ms=6000
```

##### Logs

```log
20:09:16 client.go:110: Initializing new client
20:09:16 client.go:601: client/metadata fetching metadata for all topics from broker kafka-b.example.com:9092
20:09:16 broker.go:146: Connected to broker at kafka-b.example.com:9092 (unregistered)
20:09:16 client.go:397: client/brokers registered new broker #2 at kafka-b.example.com:9092
20:09:16 client.go:397: client/brokers registered new broker #1 at kafka-a.example.com:9092
20:09:16 client.go:397: client/brokers registered new broker #3 at kafka-c.example.com:9092
20:09:16 client.go:154: Successfully initialized new client
0...
20:09:16 async_producer.go:585: producer/broker/2 starting up
20:09:16 async_producer.go:597: producer/broker/2 state change to [open] on mytopic/0
20:09:16 async_producer.go:585: producer/broker/3 starting up
20:09:16 async_producer.go:597: producer/broker/3 state change to [open] on mytopic/1
1000...
2000...
3000...
4000...
5000...
6000...
7000...
8000...
9000...
10000...
11000...
12000...
13000...
14000...
15000...
16000...
17000...
18000...
19000...
20:09:16 broker.go:144: Connected to broker at kafka-c.example.com:9092 (registered as #3)
20:09:16 broker.go:144: Connected to broker at kafka-b.example.com:9092 (registered as #2)
20:09:19 async_producer.go:763: producer/broker/3 state change to [closing] because write tcp [xxx.xxx.xxx.xxx]:34488->[xxx.xxx.xxx.xxx::16]:9092: i/o timeout
20:09:19 async_producer.go:763: producer/broker/2 state change to [closing] because write tcp [xxx.xxx.xxx.xxx]:34487->[xxx.xxx.xxx.xxx::54]:9092: i/o timeout
20:09:19 broker.go:182: Closed connection to broker kafka-c.example.com:9092
[...]

PR #835, opened 2017-02-24
Tweak documentation of Close/AsyncClose methods

eapache:
AsyncClose has nothing to do with flushes at all.

Close *waits* for messages to be flushed, but does not *trigger* a flush
itself.

Clarifications as suggested in #834.

@agis- @wvanbergen


agis-:
Looks good to me 👍

PR #837, opened 2017-02-27
Add functionality for custom hash partitioner.

manolovl:
The function NewCustomHashPartitioner is a wrapper around the PartitionerConstructor, allowing for specifying the hash algorithm, which will be used.

Background for the proposed change: messages published on compacted topics by services, written in different languages (Go, Java, etc) end up on different partitions, although they have the same key. We were able to narrow the problem down to different hashing algorithms used by the different libraries, when determining the partition: murmur2 for a Java lib, FNV in sarama, crc32 for .NET and xxHash in librdkafka.  
A quick workaround is of course to copy/paste the hashPartitioner implementation to change the implementation of the hash.Hash32 interface ...which is basically one line out of ~50. So I decided I'd propose this PR to facilitate the use of different hashing algorithms, based on specific needs.

### DIFF
--- a/partitioner.go
+++ b/partitioner.go
@@ -87,6 +87,17 @@ type hashPartitioner struct {
 	hasher hash.Hash32
 }
 
+// NewCustomHashPartitioner is a wrapper around NewHashPartitioner,
+// allowing the use of custom hasher
+func NewCustomHashPartitioner(hasher hash.Hash32) PartitionerConstructor {
+	return func(topic string) Partitioner {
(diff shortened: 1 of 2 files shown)


eapache:
Thanks, this makes sense to me.

It was a couple of years ago now, but I swear we chose FNV at the time because that's what the upstream kafka libs were using. I'm surprised to see so much variation now. Anyway.

manolovl:
Thanks! I can now update to the new version and get rid of the crufty current solution I mentioned above.

PR #839, opened 2017-02-28
Fix incorrect "larger than MaxRequestSize" message, closes #836

### LINKED ISSUE #836: Incorrect "larger than MaxRequestSize" log output
##### Versions

Sarama Version: https://github.com/Shopify/sarama/commit/1416bd78f804d523005322194994f08c2a0ad797

##### Logs

> Producer.MaxMessageBytes is larger than MaxRequestSize; it will be ignored.

##### Problem Description

The logger prints out this message even when the two values are equal. Same goes for `Producer.Flush.Bytes` and `MaxRequestSize`.

[These are the corresponding lines](https://github.com/Shopify/sarama/blob/e6df07fb9cf315a30d9a8d665185090d5bf84b22/config.go#L307-L312) in `config.go`.

##### Problem Solution

Two scenarios here:

1. Equality is not a problem: Turn `>=` to `>` in both cases.
2. Equality is a problem: Update the logger message accordingly.

Let me know which one is it (I am guessing the former), and I can push a PR for it.

Thank you.

kchristidis:


eapache:
Thanks!

PR #841, opened 2017-03-02
Fix CustomHasher

manolovl:
I had to realise that even though now the same hashing algorithm was used across services to determine the partition, the problem that messages published under the same key on log compact topics still ended up on different partitions. Unfortunately I found out that I made a huge mistake by passing an already instantiated hasher. Thus the proposed fix: pass a callback so the hasher is instantiated just before needed (thanks to the [wiki page](https://github.com/Shopify/sarama/wiki/Producer-implementation#message-flow) for helping my understanding of the publisher's flow).

### DIFF
--- a/partitioner.go
+++ b/partitioner.go
@@ -86,7 +86,7 @@ type hasherFunc func() hash.Hash32
 
 type hashPartitioner struct {
 	random Partitioner
-	hasher hasherFunc
+	hasher hash.Hash32
 }
 
 // NewCustomHashPartitioner is a wrapper around NewHashPartitioner,
@@ -95,7 +95,7 @@ func NewCustomHashPartitioner(hasher hasherFunc) PartitionerConstructor {
 	return func(topic string) Partitioner {
 		p := new(hashPartitioner)
 		p.random = NewRandomPartitioner(topic)
-		p.hasher = hasher
+		p.hasher = hasher()
 		return p
 	}
 }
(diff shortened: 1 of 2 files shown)


partitioner.go
```diff
@@ -82,18 +82,20 @@ func (p *roundRobinPartitioner) RequiresConsistency() bool {
 	return false
 }
 
+type hasherFunc func() hash.Hash32
```
eapache: I wouldn't even bother naming this type, especially since it's weird for a non-exported interface to be an argument to an exported method.

partitioner.go
```diff
@@ -82,18 +82,20 @@ func (p *roundRobinPartitioner) RequiresConsistency() bool {
...
+
 type hashPartitioner struct {
 	random Partitioner
 	hasher hash.Hash32
 }
 
 // NewCustomHashPartitioner is a wrapper around NewHashPartitioner,
 // allowing the use of custom hasher
```
eapache: might be worth documenting here that it expects the method reference not an instance of the hasher

partitioner.go
manolovl: Yeah, true.

partitioner.go
manolovl: Absolutely. I might have gotten a bit in a hurry to open the PR, since I realised my mistake.

eapache:
Thanks! While technically this is a breaking API change the method was just introduced, broken in its current format, and never included in an actual release, so I think we can skirt the go rules for this one.

manolovl:
Thank you too! Those were also my thoughts, plus the fact that without this change, it was simply broken. Also learned a lesson or two.

PR #842, opened 2017-03-02
Bump supported/tested go and kafka versions

eapache:


eapache:
From the golang 1.8 release notes:

> There have been some minor fixes to the encoder to improve the compression ratio in certain situations. As a result, the exact encoded output of DEFLATE may be different from Go 1.7. Since DEFLATE is the underlying compression of gzip, png, zlib, and zip, those formats may have changed outputs.

PR #843, opened 2017-03-02
Permit protocol ordering in JoinGroup requests

### LINKED ISSUE #812: Cannot specify JoinGroupRequest protocol priority
##### Versions

Present on every version that supports JoinGroupRequest

##### Problem Description

The kafka protocol for JoinGroupRequest allows you to specify priority on group protocols for seamless rollout of new protocols while a consumer group is running. This is done by changing the ordering on the wire for the GroupProtocols. See [here](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-JoinGroupRequest) for more details

Sarama exposes this field as a [map](https://github.com/Shopify/sarama/blob/0fb560e5f7fbcaee2f75e3c34174320709f69944/join_group_request.go#L8), making it impossible to specify an ordering.

eapache:
The kafka protocol for JoinGroupRequest allows you to specify priority
on group protocols for seamless rollout of new protocols while a
consumer group is running. The old map was not supporting that use case.

Add an array field (OrderedGroupProtocols) and deprecate the map.

Fixes #812.

@DanielMorsing @wvanbergen

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -50,7 +50,11 @@ func testVersionDecodable(t *testing.T, name string, out versionedDecoder, in []
 }
 
 func testRequest(t *testing.T, name string, rb protocolBody, expected []byte) {
-	// Encoder request
+	packet := testRequestEncode(t, name, rb, expected)
+	testRequestDecode(t, name, rb, packet)
+}
+
+func testRequestEncode(t *testing.T, name string, rb protocolBody, expected []byte) []byte {
 	req := &request{correlationID: 123, clientID: "foo", body: rb}
 	packet, err := encode(req, nil)
 	headerSize := 14 + len("foo")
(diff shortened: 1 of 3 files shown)


request_test.go
```diff
@@ -50,7 +50,11 @@ func testVersionDecodable(t *testing.T, name string, out versionedDecoder, in []
 }
 
 func testRequest(t *testing.T, name string, rb protocolBody, expected []byte) {
```
wvanbergen: `testRequestEncode?`

eapache: I don't follow? All I did was split `testRequest` into the two pieces `testRequestEncode` and `testRequestDecode`; this method just calls them both.

PR #861, opened 2017-04-05
add an interface to create client from broker list. 

hanbing0715:
add this interface to make create client for an group available


hanbing0715:
maybe ci have something wrong,i can pass the "go fmt"but ci can't

wvanbergen:
CI, including `make fmt` was working fine yesterday. I also rebuilt master just now (https://travis-ci.org/Shopify/sarama/builds/218407653), and it is working fine. So it must be something with your branch.

wvanbergen:
What is the use case you have for this function? You can already create a client with the broker address as strings, which is pretty easy to obtain from a list of Broker objects (which you normally only have access to once you already have a client anyway)

hanbing0715:
i have make a commit to pass the ci
https://travis-ci.org/hanbing0715/sarama
and i will add my use case later

hanbing0715:
close the pr for a bug

PR #872, opened 2017-05-01
Add client.Isr to determine in-sync replicas

pd:
This restores the commit from @funkygao in PR #566, with an additional commit to expose the function on the `Client` interface and add some basic tests around its functionality.

The PR was previously closed with the suggestion that Zookeeper should be used directly, because, at the time, it was possible to receive stale information from Kafka. If I'm reading the [linked issue](https://issues.apache.org/jira/browse/KAFKA-1367?jql=text%20~%20%22ISR%20out%20of%20sync%22) correctly, that issue has been resolved since Kafka 0.9, so I'd like this patch to be reconsidered now.

I personally want this so that I can then add ISR support to the excellent [kt](https://github.com/fgeller/kt), which is the workhorse of a fair amount of the Kafka automation tooling I use.

### DIFF
--- a/client.go
+++ b/client.go
@@ -38,8 +38,10 @@ type Client interface {
 	// Replicas returns the set of all replica IDs for the given partition.
 	Replicas(topic string, partitionID int32) ([]int32, error)
 
-	// Isr returns the set of in-sync replica IDs for the given partition.
-	Isr(topic string, partitionID int32) ([]int32, error)
+	// InSyncReplicas returns the set of all in-sync replica IDs for the given
+	// partition. In-sync replicas are replicas which are fully caught up with
+	// the partition leader.
+	InSyncReplicas(topic string, partitionID int32) ([]int32, error)
 
 	// RefreshMetadata takes a list of topics and queries the cluster to refresh the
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -38,6 +38,9 @@ type Client interface {
 	// Replicas returns the set of all replica IDs for the given partition.
 	Replicas(topic string, partitionID int32) ([]int32, error)
 
+	// Isr returns the set of in-sync replica IDs for the given partition.
+	Isr(topic string, partitionID int32) ([]int32, error)
```
eapache: For a public API I would prefer this be a bit more explicit: `ReplicasInSync` or `InSyncReplicas` or something like that. It would also be nice if the docs explained (or linked to an explanation) of what it means to be in sync.

pd:
@eapache I've renamed it to `InSyncReplicas`, and added some semblance of an explanation of what that means to the godoc. I wasn't able to find any particularly useful, official documentation of what ISR actually means, aside from some things on cwiki that haven't been edited since 2013.

Improved documentation suggestions welcome. =)

eapache:
This is great, thanks!

PR #874, opened 2017-05-06
Fix memory alignment of int64 variable used as atomic.

dlzr:
This fixes a test crash on 32-bit architectures.

https://bugs.debian.org/cgi-bin/bugreport.cgi?archive=no&bug=860705


eapache:
For my own future reference: https://golang.org/pkg/sync/atomic/#pkg-note-BUG

> On both ARM and x86-32, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically. The first word in a global variable or in an allocated struct or slice can be relied upon to be 64-bit aligned.

eapache:
Thanks!

PR #905, opened 2017-06-29
Added support for FetchRequest protocol version 3.

sp1ff:
This commit will add the 'MaxBytes' field to FetchRequests when the Kafka version is
0.10.1 or better (i.e. it adds support for V3 of the FetchRequest API). On send, it will be set to MaxResponseSize.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -728,7 +728,7 @@ func (bc *brokerConsumer) fetchNewMessages() (*FetchResponse, error) {
 	}
 	if bc.consumer.conf.Version.IsAtLeast(V0_10_1_0) {
 		request.Version = 3
-		request.MaxBytes = MaxResponseSize - 47 // TODO(mherstin): WTF?!
+		request.MaxBytes = MaxResponseSize
 	}
 
 	for child := range bc.subscriptions {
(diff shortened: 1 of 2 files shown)


fetch_request.go
```diff
@@ -32,6 +36,9 @@ func (r *FetchRequest) encode(pe packetEncoder) (err error) {
 	pe.putInt32(-1) // replica ID is always -1 for clients
 	pe.putInt32(r.MaxWaitTime)
 	pe.putInt32(r.MinBytes)
+	if 3 == r.Version {
```
eapache: `r.Version == 3` is more consistent with the rest of the code, and with the check you added below

consumer.go
```diff
@@ -726,6 +726,10 @@ func (bc *brokerConsumer) fetchNewMessages() (*FetchResponse, error) {
 	if bc.consumer.conf.Version.IsAtLeast(V0_10_0_0) {
 		request.Version = 2
 	}
+	if bc.consumer.conf.Version.IsAtLeast(V0_10_1_0) {
+		request.Version = 3
+		request.MaxBytes = MaxResponseSize - 47 // TODO(mherstin): WTF?!
```
eapache: I agree with the comment: where did the value of 47 come from?

fetch_request.go
sp1ff: Old habits die hard; fixed in my latest commit.

consumer.go
sp1ff: That's embarrassing-- this was removed in a subsequent commit which I forgot to push. Fixed.

eapache:
Thanks!

PR #906, opened 2017-06-30
closes #854 clear up doc confusion on producer config

### LINKED ISSUE #854: Sync Producer config doc confusion
Context: https://github.com/Shopify/sarama/commit/30a4affd855f22040dadf4267e392c6cae4bb45b#commitcomment-21504927

The sync producer verifies and uses the `Producer.Return.*` config values internally, but the documentation on those specific fields mentions reading from their channels, which is not necessary in that case (the sync producer does this internally). I believe the behaviour of the code is correct as-is, but documentation definitely needs clarifying around this.

makkes:


eapache:
Thanks!

PR #907, opened 2017-06-30
negative timestamps on messages fix

SamiHiltunen:
* Adds checks for negative timestamps on messages. Negative
  timestamps are invalid. On decoding, return a zero time for easy
  existence checking. On encoding, set timestamp to be -1 if the
  given time describes a time before unix epoch.

### DIFF
--- a/message.go
+++ b/message.go
@@ -45,7 +45,15 @@ func (m *Message) encode(pe packetEncoder) error {
 	pe.putInt8(attributes)
 
 	if m.Version >= 1 {
-		pe.putInt64(m.Timestamp.UnixNano() / int64(time.Millisecond))
+		timestamp := int64(-1)
+
+		if !m.Timestamp.Before(time.Unix(0, 0)) {
+			timestamp = m.Timestamp.UnixNano() / int64(time.Millisecond)
+		} else if !m.Timestamp.IsZero() {
+			return PacketEncodingError{fmt.Sprintf("invalid timestamp (%v)", m.Timestamp)}
+		}
+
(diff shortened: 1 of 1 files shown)


SamiHiltunen:
I assume you mean to return an error when trying to encode a bad packet? Kafka is returning messages with ```-1``` timestamps so the decoding part should probably just return a zero time if it encounters anything below ```0```. For encoding, returning errors would make sense. 

I'll make the following changes for the encoding part:

1. Except for zero time, anything before unix epoch will return a ```PacketEncodingError{"invalid timestamp"}```
2. If timestamp is a zero time, timestamp will be set to ```-1```
3. Time value on or after unix epoch will use the current formula

Does this sound good?

message.go
```diff
@@ -45,7 +45,15 @@ func (m *Message) encode(pe packetEncoder) error {
 	pe.putInt8(attributes)
 
 	if m.Version >= 1 {
-		pe.putInt64(m.Timestamp.UnixNano() / int64(time.Millisecond))
+		timestamp := int64(-1)
+
+		if m.Timestamp.After(time.Unix(0, -1)) {
```
eapache: looking at the actual implementation (https://golang.org/src/time/time.go?s=29125:29162#L991), did you mean `time.Unix(0, 0)`?

eapache: or perhaps it would be simpler to just do `!Before` a zero time, which is cleaner to construct

SamiHiltunen: I inverted the condition to use ```!m.Timestamp.Before``` as you suggested. Let me know if there is anything else.

PR #927, opened 2017-08-10
Docstring change only to PartitionConsumer.

sp1ff:
As per the discussion in https://github.com/Shopify/sarama/issues/920,
I've added some verbiage to the docstrings around Close & AsyncClose
on interface PartitionConsumer.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -247,9 +247,9 @@ func (c *consumer) abandonBrokerConsumer(brokerWorker *brokerConsumer) {
 
 // PartitionConsumer
 
-// PartitionConsumer processes Kafka messages from a given topic and partition. You MUST call Close()
-// or AsyncClose() on a PartitionConsumer to avoid leaks, it will not be garbage-collected automatically
-// when it passes out of scope.
+// PartitionConsumer processes Kafka messages from a given topic and partition. You MUST call one of Close() or
+// AsyncClose() on a PartitionConsumer to avoid leaks; it will not be garbage-collected automatically when it passes out
+// of scope.
 //
 // The simplest way of using a PartitionConsumer is to loop over its Messages channel using a for/range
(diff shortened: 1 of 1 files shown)


consumer.go
```diff
@@ -258,19 +258,25 @@ func (c *consumer) abandonBrokerConsumer(brokerWorker *brokerConsumer) {
 // By default, it logs these errors to sarama.Logger; if you want to be notified directly of all errors, set
 // your config's Consumer.Return.Errors to true and read from the Errors channel, using a select statement
 // or a separate goroutine. Check out the Consumer examples to see implementations of these different approaches.
+//
+// To terminate such a for/range loop while the loop is executing, call AsyncClose. This will kick off the process of
+// consumer tear-down & return imediately. Continue to loop until the Messages channel has been drained. If you've
```
eapache: I think this part might be a touch unclear; the call to `AsyncClose` will return immediately, but the loop could take some time to exit depending.

Might also be worth clarifying that `AsyncClose` will actually close the `messages` channel when cleanup is done, so the `range` will stop automatically.

consumer.go
```diff
@@ -258,19 +258,25 @@ func (c *consumer) abandonBrokerConsumer(brokerWorker *brokerConsumer) {
...
+// already ceased reading Messages, call Close; this will signal the PartitionConsumer's goroutines to begin shutting
+// down (just like AysncClose), but will also drain the Messages channel, harvest all errors & return them once cleanup
```
eapache: typo `AsyncClose`

eapache:
This looks really nice, thanks. Just one typo and one other comment.

sp1ff:
Thanks, Evan. Updated as per your suggestions.

eapache:
Thanks!

PR #930, opened 2017-08-11
FIX: Remove dupeAndSort in Replicas and InSyncReplicas

git-hulk:
It shouldn't sort the replicas and isr while the order is meaningful to Kafka(preferred leader)


eapache:
Hmm, I'm not sure why I originally added this logic. The sorting does seem incorrect.

Creating a copy of the array is a good practice for API safety though to avoid complications when callers mutate the result. Please leave the duplication part in place.

git-hulk:
@eapache Fixed and rebased, thanks for the review.

eapache:
Thanks!

Issue #554, opened 2015-10-20
OffsetManager, ability to mark previous offset


yulrizka:
SHA: a23cf43069c4a88e12887ac68250bb69e0fe8c84

in https://github.com/Shopify/sarama/blob/master/offset_manager.go#L321-L325
Offset manager only commit the offset if it's larger than current offset.

I'm having a problem in this one particular use-case where my program accept an option to read kafka from beginning. If for example current recorded offset is 10000 and I start from the beginning but stop at 500, next time I run the process again it will start at 10000 instead.

What do you think if we could add an extra function / parameter to skip this check ?

wvanbergen:
Interesting. It's a legitimate use case to "reset" the offset, but that currently cannot be done with the offset manager. You well have to hand-craft a CommitOffset request to do this (for an example, see [here](https://github.com/wvanbergen/kafka/pull/72/files#diff-46af69c208f0f522187167dc3f923eb6R307)).

I am not necessarily against removing the if statement. Maybe adding a `ResetOffset` method would be a better idea, to make it explicit that you're trying to commit a lower offset? Unsure.

hasnickl:
another option is a ```SetOffset``` method to make it explicit that you're committing an arbitrary offset, instead of doing ```ResetOffset``` followed by a ```MarkOffset```

amitech74:
Can someone confirm this - In order to "hand-craft a CommitOffset request" (to reset offset), it needs access to the latest generationID, without which resetting an offset doesn't seem to work, if consumers are running already.

Xaelias:
Hi guys. I'm trying to do the same thing.

I tried to remove the if just to test, and the program just blocked...
So now I'm trying to manually commit the offset but it doesn't look like it's working reliably either.

Here is an example of what I'm trying to achieve:
```go
package main

import (
	"fmt"
	"github.com/Shopify/sarama"
	"os"
)

func main() {
	config := sarama.NewConfig()
	config.ClientID, _ = os.Hostname()
	config.Version = sarama.V0_10_1_0
	kafkaClient, _ := sarama.NewClient([]string{"127.0.0.1:9092"}, config)

	request := &sarama.OffsetCommitRequest{}
	request.ConsumerGroup = "blah"
	request.ConsumerGroupGeneration = sarama.GroupGenerationUndefined
	request.RetentionTime = 1000 // not sure what to put here
	request.Version = 2
	request.AddBlock("blah", 0, 5, 0, "Reseting offset to 5")
[...]

PR #932, opened 2017-08-18
adds ResetOffset to reset to earlier offset values.

fgeller:
This is mostly meant as a start of a discussion, happy to change it. cf #554 for initial discussions and use cases.

Added ResetOffset on PartitionOffsetManager that allows setting the offset to a value less than the current value.

From a downstream perspective I would prefer to have a single func to call that allows setting an offset to a smaller or greater value, regardless of the current value - but thought API-wise it'd make more sense to have MarkOffset and ResetOffset act as duals/complements. What do you think?

### DIFF
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -151,6 +151,13 @@ type PartitionOffsetManager interface {
 	// message twice, and your processing should ideally be idempotent.
 	MarkOffset(offset int64, metadata string)
 
+	// ResetOffset resets to the provided offset, alongside a metadata string that
+	// represents the state of the partition consumer at that point in time. Reset
+	// acts as a counterpart to MarkOffset, the difference being that it allows to
+	// reset an offset to an earlier or smaller value, where MarkOffset only
+	// allows incrementing the offset. cf MarkOffset for more details.
+	ResetOffset(offset int64, metadata string)
+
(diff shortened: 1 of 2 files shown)


eapache:
I'm quite happy with how this looks, thanks!

@wvanbergen you had some thoughts on this originally.

offset_manager.go
```diff
@@ -329,6 +336,17 @@ func (pom *partitionOffsetManager) MarkOffset(offset int64, metadata string) {
 	}
 }
 
+func (pom *partitionOffsetManager) ResetOffset(offset int64, metadata string) {
+	pom.lock.Lock()
+	defer pom.lock.Unlock()
+
+	if offset < pom.offset {
```
hasnickl: @eapache @fgeller sorry I'm late to the party on this, just now getting around to implement our usage of this feature.

Any reason for this condition? this is preventing me from updating the metadata without changing the offset.

eapache: > updating the metadata without changing the offset.

I don't think either of us ever considered a use case for that. I suppose it would be easy enough to make this a `<=`

eapache: https://github.com/Shopify/sarama/commit/96fa1c837819761fc5c706010a9677aab15ca46d

Issue #931, opened 2017-08-17
sarama leaking memory in (*TopicMetadata).decode


chandradeepak:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*
Sarama Version: v1.12.0-2-gac03dfa , ac03dfa0c99a940a48357980a4ebfafed5f23f9a
Kafka Version: 0.10.0.1
Go Version:1.8.3

##### Configuration

What configuration values are you using for Sarama and Kafka?

default configuration . no changes here and this have been stable for a while. 

##### Logs

When filing an issue please provide logs from Sarama and Kafka if at all
possible. You can set `sarama.Logger` to a `log.Logger` to capture Sarama debug
output.

##### Problem Description
we are seeing memory leak in sarama. this is the pprof output for it .  This happened recently. the leak is happening in decoding of the metadata from kafka . 

I see there is a similar issue addressed before https://github.com/Shopify/sarama/issues/560

But i think this is a new problem . do you think restarting kafka will get rid of this issue. But still it is strange sarama is leaking memory. i don't see any weird partition problemw in kafka . Is there any thing specific i have to check to give you more details. Below i pasted the pprof logs which showed the growth and their code lines. as you can observe , decode has 1gb in use space. 
 1.17GB    144:		err = r.Topics[i].decode(pd)

Please let me know if you need any more details . 

> 680.54MB 29.10% 29.10%  1199.06MB 51.27%  github.com/Shopify/sarama.(*TopicMetadata).decode
  329.01MB 14.07% 43.17%   329.01MB 14.07%  github.com/Shopify/sarama.(*realDecoder).getInt32Array
  306.21MB 13.09% 56.26%   306.21MB 13.09%  runtime.hashGrow
  264.04MB 11.29% 67.55%   572.75MB 24.49%  runtime.mapassign
  196.51MB  8.40% 75.95%   196.51MB  8.40%  runtime.rawstringtmp
  124.50MB  5.32% 81.27%   125.50MB  5.37%  Shopify/sarama.(*client).setPartitionCache
  102.51MB  4.38% 85.66%   102.51MB  4.38%  runtime.makemap

> /github.com/Shopify/sarama/metadata_response.go
  680.54MB     1.17GB (flat, cum) 51.27% of Total
         .          .     67:	if err != nil {
         .          .     68:		return err
         .          .     69:	}
         .          .     70:	tm.Err = KError(tmp)
         .          .     71:
[...]

PR #937, opened 2017-08-23
Add a config option to store only minimal metadata

eapache:
Typically the default is just to fetch and store all topics, as it is
simpler and more convenient not to have to go to the network for every
new topic that is needed. However this can use a lot of memory for
clusters with many topics and partitions, so add a flag which can be
unset in order to just store the minimal topic information needed so
far.

As suggested in #931.

@chandradeepak does this solve your problem?


chandradeepak:
i will check with this pull request and let you know . thanks for considering it .

chandradeepak:
the pull request  kind of stopped the growth . i think it is good to merge to master.  Thanks for the quick fix .

Issue #938, opened 2017-08-24
Can MockBroker return v2 messages?


sanchezl:
##### Versions

Sarama Version: `v1.12.0`
Kafka Version: `0.10.1.0`
Go Version: `1.8.3`

##### Configuration

Kafka: using `sarama.MockBroker`
Sarama:  `Config.Version = sarama.V0_10_1_0`

##### Logs

```
=== RUN   TestVersionedConsumer
=== RUN   TestVersionedConsumer/0.9.0.1
[sarama] 2017/08/24 12:36:29 *** mockbroker/0 listening on 127.0.0.1:54921
[sarama] 2017/08/24 12:36:29 Initializing new client
[sarama] 2017/08/24 12:36:29 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2017/08/24 12:36:29 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2017/08/24 12:36:29 client/metadata fetching metadata for all topics from broker 127.0.0.1:54921
[sarama] 2017/08/24 12:36:29 *** mockbroker/0/0: connection opened
[sarama] 2017/08/24 12:36:29 Connected to broker at 127.0.0.1:54921 (unregistered)
[sarama] 2017/08/24 12:36:29 *** mockbroker/0/0: served &{0 sarama 0xc42016a020} -> &{[0xc420172000] [0xc42016c060]}
[sarama] 2017/08/24 12:36:29 client/brokers registered new broker #0 at 127.0.0.1:54921
[sarama] 2017/08/24 12:36:29 Successfully initialized new client
[sarama] 2017/08/24 12:36:29 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2017/08/24 12:36:29 Connected to broker at 127.0.0.1:54921 (registered as #0)
[sarama] 2017/08/24 12:36:29 *** mockbroker/0/1: connection opened
[sarama] 2017/08/24 12:36:29 *** mockbroker/0/1: served &{0 sarama 0xc4201ae040} -> &{0 map[my_topic:map[0:0xc4201a6180]]}
[sarama] 2017/08/24 12:36:29 *** mockbroker/0/1: served &{1 sarama 0xc4201702a0} -> &{0 map[my_topic:map[0:0xc42016c7e0]]}
[sarama] 2017/08/24 12:36:29 consumer/broker/0 added subscription to my_topic/0
[...]

PR #939, opened 2017-08-24
Permit setting version on mock fetch response

eapache:
@sanchezl this should address #938.

### DIFF
--- a/mockresponses.go
+++ b/mockresponses.go
@@ -180,6 +180,7 @@ type MockFetchResponse struct {
 	highWaterMarks map[string]map[int32]int64
 	t              TestReporter
 	batchSize      int
+	version        int16
 }
 
 func NewMockFetchResponse(t TestReporter, batchSize int) *MockFetchResponse {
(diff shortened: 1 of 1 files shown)


sanchezl:
Thanks @eapache. Would you consider an additional API such as:
```go
func (mfr *MockFetchResponse) SetMaxVersion(kafkaVersion KafkaVersion) *MockFetchResponse {
switch {
  case kafkaVersion.IsAtLeast(sarama.V0_11_0_0): // e.g. when supported
    mfr.version = 3
  case kafkaVersion.IsAtLeast(sarama.V0_10_0_0):
    mfr.version = 2
  default:
    mfr.version = 1
}
return mfr
}
```

eapache:
Hmm, actually would it just be simpler to return the same version of FetchResponse as was received in the FetchRequest? That would be simpler, but I'm not sure if they line up in all cases?

sanchezl:
What about just adding property to the MockBroker itself (e.g. `MockBroker.setMessageFormat(kafkaVersion KafkaVersion)` that would correspond to setting the `log.message.format.version` property on a real kafka server.

I advocate for the use of `KafkaVersion`, not just the message version since that is what the `log.message.format.version` property expects.

eapache:
> I advocate for the use of KafkaVersion, not just the message version since that is what the log.message.format.version property expects.

AFAICT log.message.format.version refers to the version of the actual `Message` data struct, not the Request/Response wire protocol. They are versioned differently.

eapache:
I've reviewed this some more and I think the PR I have here is correct. The wire protocol version should be flexible even knowing the broker version.

PR #940, opened 2017-08-25
make decoding unknown message versions error

tcrayford:
If sarama encounters an unknown message version somehow (e.g. due to a
bug in Kafka 0.11's downconversion), right now sarama will carry on
decoding the message, resulting in much harder to understand errors.

Instead, bail out early with a friendly error message.

I added a test, but struggled running tests on my laptop due to (I think) out of date dependencies, so will rely on CI running here. I also added a test for decoding a message with the `1` magic byte, which sarama supports but I could not see a test in `message_test.go` for.

### DIFF
--- a/crc32_field.go
+++ b/crc32_field.go
@@ -2,6 +2,7 @@ package sarama
 
 import (
 	"encoding/binary"
+	"fmt"
 	"hash/crc32"
 )
 
@@ -27,8 +28,9 @@ func (c *crc32Field) run(curOffset int, buf []byte) error {
 func (c *crc32Field) check(curOffset int, buf []byte) error {
 	crc := crc32.ChecksumIEEE(buf[c.startOffset+4 : curOffset])
 
-	if crc != binary.BigEndian.Uint32(buf[c.startOffset:]) {
-		return PacketDecodingError{"CRC didn't match"}
+	expected := binary.BigEndian.Uint32(buf[c.startOffset:])
+	if crc != expected {
+		return PacketDecodingError{fmt.Sprintf("CRC didn't match expected %#x got %#x", expected, crc)}
 	}
 
 	return nil
--- a/message.go
+++ b/message.go
@@ -122,13 +122,17 @@ func (m *Message) decode(pd packetDecoder) (err error) {
 		return err
 	}
 
+	if m.Version > 1 {
+		return PacketDecodingError{fmt.Sprintf("unknown magic byte (%v)", m.Version)}
+	}
+
 	attribute, err := pd.getInt8()
 	if err != nil {
 		return err
 	}
 	m.Codec = CompressionCodec(attribute & compressionCodecMask)
 
-	if m.Version >= 1 {
+	if m.Version == 1 {
 		millis, err := pd.getInt64()
 		if err != nil {
 			return err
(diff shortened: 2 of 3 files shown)


message.go
```diff
@@ -144,6 +144,10 @@ func (m *Message) decode(pd packetDecoder) (err error) {
 		m.Timestamp = timestamp
 	}
 
+	if m.Version > 1 {
```
eapache: might as well put this check immediately after extracting the version

message_test.go
```diff
@@ -14,6 +14,21 @@ var (
 		0xFF, 0xFF, 0xFF, 0xFF, // key
 		0xFF, 0xFF, 0xFF, 0xFF} // value
 
+	emptyV1Message = []byte{
+		167, 236, 104, 3, // CRC
+		0x01,                   // magic version byte
+		0x00,                   // attribute flags
+		0x00, 0x00, 0x00, 0x00, // timestamp
```
eapache: timestamp is an int64, aka 8 bytes not 4

tcrayford:
@eapache thanks! I rebased the changes into this branch.

tcrayford:
@eapache the test failure now is due to a v1 message failing with a crc mismatch. Do you have a good way of recomputing the CRC?

eapache:
> Do you have a good way of recomputing the CRC?

I think the last time I had to do one of these I just stuck a print into the test and ran it locally. Are you still unable to run the local tests?

tcrayford:
@eapache yeah, still confused about my local state. Instead, I pushed an `Sprintf` to the error to print out the CRC codes, which seems better anyway?

crc32_field.go
```diff
@@ -27,8 +28,9 @@ func (c *crc32Field) run(curOffset int, buf []byte) error {
 func (c *crc32Field) check(curOffset int, buf []byte) error {
 	crc := crc32.ChecksumIEEE(buf[c.startOffset+4 : curOffset])
 
-	if crc != binary.BigEndian.Uint32(buf[c.startOffset:]) {
-		return PacketDecodingError{"CRC didn't match"}
+	expected := binary.BigEndian.Uint32(buf[c.startOffset:])
+	if crc != expected {
+		return PacketDecodingError{fmt.Sprintf("CRC didn't match expected %v got %v", expected, crc)}
```
eapache: :+1: you probably want to print these as hex though for convenience

tcrayford:
@eapache ok, all ready! I got tests working locally eventually.

eapache:
Thanks!

catlittlechen:
Hello, @tcrayford @eapache When my process running with kafka0.11, I catch this error. If I ignore this error, some messages are missing. But I can't found any reason about this. Can you give me a favor? Maybe some archives about a
bug in Kafka 0.11's downconversion?

eapache:
I am not aware of any bugs in the downconversion, but it's possible. I'd also note that Sarama master branch just got support for talking in the 0.11 record format, so you might be able to avoid the whole problem with that.

tcrayford:
@eapache @catlittlechen you're likely hitting a thing that *is not* a bug, but a documented fact about 0.11 - if your "max message size" on the consumer's `FetchMessages` isn't large enough for a 0.11 record batch, then downconversion does *not* take place and kafka sends back the new formatted version: http://kafka.apache.org/documentation.html#upgrade_11_message_format

catlittlechen:
Thank you.  @eapache @tcrayford

PR #942, opened 2017-08-30
Restrict client metadata update more

eapache:
If Metadata.Full is set to false, and no topics have ever been
requested, the first background refresh would load them all because go's
varags don't distinguish between nothing and an explicitly splatted
empty array.

Just break instead.

@chandradeepak this *might* fix https://github.com/Shopify/sarama/issues/931#issuecomment-325829238? I'm just guessing.

### DIFF
--- a/client.go
+++ b/client.go
@@ -612,6 +612,9 @@ func (client *client) backgroundMetadataUpdater() {
 				if specificTopics, err := client.Topics(); err != nil {
 					Logger.Println("Client background metadata topic load:", err)
 					break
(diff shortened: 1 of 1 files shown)


chandradeepak:
@eapache , i updated the issue with my comments there. i think this alone won't fix the problem.

eapache:
I think you're right, but this is worth fixing anyway.

chandradeepak:
@eapache , yeah this makes sense . it would be good to have it.

PR #952, opened 2017-09-20
travis: update go versions

dvrkps:
As discussed at https://github.com/Shopify/sarama/pull/950.


eapache:
And of course the gzip encoding in go 1.9 is also different, sorry about that. Could you add a check for 1.9.* in the failing test just like for 1.8?

eapache:
Thanks!

PR #965, opened 2017-10-06
Added more tests for config validation

agnivade:


agnivade:
@eapache  - 
```
Error while executing topic command : replication factor: 3 larger than available brokers: 2
[2017-10-06 10:11:17,820] ERROR org.apache.kafka.common.errors.InvalidReplicationFactorException: replication factor: 3 larger than available brokers: 2
```
Does not look like related to my change.

config_test.go
```diff
@@ -33,6 +33,205 @@ func TestEmptyClientIDConfigValidates(t *testing.T) {
 	}
 }
 
+func TestNetConfigValidates(t *testing.T) {
+	tests := []struct {
+		name string
+		cfg  func() *Config // resorting to using a function as a param because of internal composite structs
```
eapache: change the method sig to `func(*Config)` and have the loop create the new config each time; will reduce the size of all the function bodies

eapache:
Sometimes that CI error happens if the kafka cluster does not boot properly, I can trigger a re-run if it happens again.

config_test.go
agnivade: Thanks. Updated.

eapache:
Thanks!

PR #967, opened 2017-10-10
Removed some redundant error checks in real_encoder

agnivade:
Several functions were returning nil error to satisfy the packet encoder
interface. Therefore, removed the redundant error checking.


eapache:
The errcheck tool requires it and it costs very little.

agnivade:
Ah I see. Had forgot about the errcheck. I probably went overboard anyway trying to over optimize :stuck_out_tongue_closed_eyes:

Issue #901, opened 2017-06-23
Plans to support upcoming Kafka v0.11 ?


dvsekhvalnov:
Any plans so far? 0.11 is about to be released fairly soon, adding support for message headers.

Thanks.

eapache:
I would like to, but I have no time to work on it in the next little while. I'm happy to accept pull requests though.

Kane-Sendgrid:
We plan to add support for that, should we agree about transaction protocol before any work starts on this?

tomsapps:
It'd be awesome if we could start using the projects tab to make a 0.11 roadmap. I'm sure that'd bring more people together to work on v0.11 support

17twenty:
What's the extent of work on this? Is there anything outlined as would be happy to pick it up if there's some low hanging fruit?

eapache:
There's a bunch of work that can be done here, and a lot of the initial stuff is pretty straightforward low-hanging fruit.

The first batch of work is all foundational:
- Add all the new errors (codes 45-55) to `errors.go` (easy).
- Implement all the new request/response objects (keys 19-33) with the appropriate encode/decode methods and tests (easy). Each request/response pair can be done separately/in-parallel.

With that in place there's then a bunch of other improvements we can make to expose new functionality using the new request/response objects. Some of this should be pretty easy (e.g. a `Client.AddTopic` method) and some of it will be quite complicated (e.g. transaction support in the producer).

The other piece of work for 0.11 is to adapt to the new `Message` and `MessageSet` structures. The pure wire format changes shouldn't be too hard (though doing them without breaking old versions won't be entirely trivial). However, the switch to a non-recursive compressions structure and varint encoding looks really messy if we want to maintain backwards compatibility. I haven't had time to really dig into what would be required here.

wladh:
I have already started (and almost finished) the work on using the new wire format for messages (now called `Record` and `RecordBatch`), but of course without all the transactional support. We mostly need it for the `Headers` extension. The way I implemented it for now was to add a separate field to the `FetchResponseBlock` for `RecordBatch` and I split `consumer.parseResponse` into separate functions dealing with old and new versions (based on the API version). And I'm implementing similar thing for producer path.
I thought I replied to this issue before, but it seems I didn't.

dvsekhvalnov:
Just curious, if anybody made `Headers` part work? Or still planning to submit patch request?

Thanks.

wladh:
[...]

PR #969, opened 2017-10-18
Add support for Kafka 0.11 Record and RecordBatch

wladh:
The new records format (including headers) is supported, however there's
no support for transactions or idempotent messages.
This is the first step towards issue #901


eapache:
Thanks so much for this contribution, it looks pretty good. Unfortunately it's an enormous PR (1600 lines!) and I'm having trouble untangling the pieces to review it properly.

Could you split it up and submit it in a couple of smaller PRs please? Off the top of my head there are at least the following chunks that are separable:
- varint support in the encoder layer
- Castagnoli support for CRCs
- implementation of the new record and recordbatch structs at the protocol layer
- support for record/recordbatch in the producer
- support for record/recordbatch in the consumer

Don't be afraid to split it up even further though if you find good divisions. The time it takes me to review something is usually about quadratic on the number of lines of code, so the smaller the better (within reason).

wladh:
Sure, I prefer small commits myself.

wladh:
Do you want multiple PRs or is multiple commits in a single PR fine?

eapache:
Multiple PRs would be preferable, but I can deal with either

PR #971, opened 2017-10-24
Add support for Castagnoli polynomial for CRC32 computation

wladh:
This is needed for Kafka 0.11 support as the new records' checksum is
computed with Castagnoli polynomial (previous versions of Kafka use
IEEE polynomial).
Issue #901

### DIFF
--- a/crc32_field.go
+++ b/crc32_field.go
@@ -6,8 +6,10 @@ import (
 	"hash/crc32"
 )
 
+type crcPolynomial int8
+
 const (
-	crcIEEE = iota
+	crcIEEE crcPolynomial = iota
 	crcCastagnoli
 )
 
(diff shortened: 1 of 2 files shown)


crc32_field.go
```diff
@@ -6,9 +6,17 @@ import (
 	"hash/crc32"
 )
 
+const (
+	crcIEEE = iota
+	crcCastagnoli
```
eapache: Why define these? you can just have `newCRC32Field` take the `crc32`-defined constant directly?

wladh: Do you mean I should use the polynomial constants directly? I could, as they are just ints, but it didn't seem very clean to me as the intent here is to represent an enumeration of supported polynomials in the context of Kafka protocol, not the polynomial itself.
But if you feel strongly about it, I can remove them.

eapache: That's what I meant, but I see your point. In that case it probably makes sense to explicitly type it (e.g. the way that compression codecs are) rather than using ints, since the polynomials themselves are just ints.

eapache:
Looks like TravisCI is having infrastructure issues, but this is fine. Thanks!

PR #972, opened 2017-10-25
Add support for varint encoded fields

wladh:
Kafka 0.11 Record uses varint encoding for most fields.
Issue #901


eapache:
Travis is still having problems, and codecov is complaining, but this will get coverage when the actual Record struct is added with tests.

PR #973, opened 2017-10-26
Add implementation of Kafka 0.11 Records

wladh:
This PR introduces implementation of the new Record and RecordBatch formats from Kafka 0.11. Also it introduces an union type to deal with request/responses that can contain either records or messages.
Issue #901

### DIFF
--- a/length_field.go
+++ b/length_field.go
@@ -31,22 +31,43 @@ func (l *lengthField) check(curOffset int, buf []byte) error {
 type varintLengthField struct {
 	startOffset int
 	length      int64
+	adjusted    bool
+	size        int
 }
 
-func newVarintLengthField(pd packetDecoder) (*varintLengthField, error) {
-	n, err := pd.getVarint()
-	if err != nil {
-		return nil, err
-	}
-	return &varintLengthField{length: n}, nil
+func (l *varintLengthField) decode(pd packetDecoder) error {
+	var err error
+	l.length, err = pd.getVarint()
+	return err
 }
 
 func (l *varintLengthField) saveOffset(in int) {
 	l.startOffset = in
 }
 
+func (l *varintLengthField) adjustLength(currOffset int) int {
+	l.adjusted = true
+
+	var tmp [binary.MaxVarintLen64]byte
+	l.length = int64(currOffset - l.startOffset - l.size)
+
+	newSize := binary.PutVarint(tmp[:], l.length)
+	diff := newSize - l.size
+	l.size = newSize
+
+	return diff
+}
+
 func (l *varintLengthField) reserveLength() int {
-	return 0
+	return l.size
+}
+
+func (l *varintLengthField) run(curOffset int, buf []byte) error {
+	if !l.adjusted {
+		return PacketEncodingError{"varintLengthField.run called before adjustLength"}
+	}
+	binary.PutVarint(buf[l.startOffset:], l.length)
+	return nil
 }
 
 func (l *varintLengthField) check(curOffset int, buf []byte) error {
--- a/length_field.go
+++ b/length_field.go
@@ -31,10 +31,12 @@ func (l *lengthField) check(curOffset int, buf []byte) error {
 type varintLengthField struct {
 	startOffset int
 	length      int64
+	consumed    bool
 }
 
 func (l *varintLengthField) decode(pd packetDecoder) error {
 	var err error
+	l.consumed = true
 	l.length, err = pd.getVarint()
 	return err
 }
@@ -51,6 +53,10 @@ func (l *varintLengthField) adjustLength(currOffset int) int {
 }
 
 func (l *varintLengthField) reserveLength() int {
+	if l.consumed { // the field was consumed during the decode
+		l.consumed = false
+		return 0
+	}
 	var tmp [binary.MaxVarintLen64]byte
 	return binary.PutVarint(tmp[:], l.length)
 }
--- a/length_field.go
+++ b/length_field.go
@@ -31,12 +31,10 @@ func (l *lengthField) check(curOffset int, buf []byte) error {
 type varintLengthField struct {
 	startOffset int
 	length      int64
-	consumed    bool
 }
 
 func (l *varintLengthField) decode(pd packetDecoder) error {
 	var err error
-	l.consumed = true
 	l.length, err = pd.getVarint()
 	return err
 }
@@ -53,10 +51,6 @@ func (l *varintLengthField) adjustLength(currOffset int) int {
 }
 
 func (l *varintLengthField) reserveLength() int {
-	if l.consumed { // the field was consumed during the decode
-		l.consumed = false
-		return 0
-	}
 	var tmp [binary.MaxVarintLen64]byte
 	return binary.PutVarint(tmp[:], l.length)
 }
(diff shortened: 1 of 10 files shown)


real_decoder.go
```diff
@@ -79,7 +79,7 @@ func (rd *realDecoder) getArrayLength() (int, error) {
 		rd.off = len(rd.raw)
 		return -1, ErrInsufficientData
 	}
-	tmp := int(binary.BigEndian.Uint32(rd.raw[rd.off:]))
+	tmp := int(int32(binary.BigEndian.Uint32(rd.raw[rd.off:])))
```
eapache: Why cast to int32 and then int?

record.go
```diff
@@ -0,0 +1,149 @@
+package sarama
+
+import "encoding/binary"
+
+const (
+	controlMask = 0x20
+)
+
+type Header struct {
```
eapache: should we call this `RecordHeader` to be a bit clearer?

records.go
```diff
@@ -0,0 +1,96 @@
...
+	case defaultRecords:
+		if r.recordBatch == nil {
+			return 0, nil
+		}
+		return len(r.recordBatch.Records), nil
+	}
+	return 0, fmt.Errorf("unknown records type: %v", r.recordsType)
+}
+
+func (r *Records) isPartial() (bool, error) {
```
eapache: this method (and isControl below) don't have any coverage AFAICT

real_decoder.go
wladh: Because we want to convert it to a signed value, so we're converting it to its signed counterpart (int usually 64 bits on 64bit platforms, so the sign conversion won't happen).

record.go
wladh: Ok

records.go
wladh: Added some coverage for them.

wladh:
Indeed the uncompressed records will be prep-encoded twice. But I didn't find a clean solution (that doesn't involve checking in encode() whether the encoder is real or prep).
I could make varintLengthField also a pushEncoder and keep it around in Record between the encoders runs.

eapache:
> I could make varintLengthField also a pushEncoder and keep it around in Record between the encoders runs.

I think that sounds like it might be simpler? If `pushEncoder.run` could also return additional length then I think this becomes very nice?

wladh:
Actually looking again at the code that wouldn't work without a few more changes. `pushEncoder.run` is not called by `prepEncoder.pop` because the `run` methods are actually writing the field. Maybe we can define a `pushEncoder.prepRun` method that will be called by `prepEncoder.pop` but it's not very appealing to me because we're re-introducing the knowledge of encoding phase into the encoders that shouldn't care about it. That was one thing I tried to avoid in my solution (and it should work even if we do away with `prepEncoder` for instance and use vectored I/O or builders).

eapache:
Hmm, ya. What if `reserveLength()` was allowed to return a special `const dynamicLength = -1`? Then the `prepEncoder` could just call it again in `pop` if it was flagged as dynamic during the `push`? Or even a separate `isDynamic() bool`  method on the interface or something...

wladh:
So I added this additional interface `dynamicPushEncoder` that defines an additional method that will be called during `prepEncoder.pop()` to get the actual field length. I'm not thrilled with this solution as it implies a prep phase, but at least the knowledge of it is limited to the `varintLengthField`.

record_batch.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	var re packetEncoder
+	var raw []byte
+
+	switch b.Codec {
+	case CompressionNone:
+		re = pe
+	case CompressionGZIP, CompressionLZ4, CompressionSnappy:
+		if err := b.computeRecordsLength(); err != nil {
```
eapache: This block here is doing basically the same thing as the global `encode(encoder) ([]byte, err)` method, i.e. using a prep-encoder to calculate the length and then making a byte array and using a real-encoder to encode it?

length_field.go
```diff
@@ -31,22 +31,43 @@ func (l *lengthField) check(curOffset int, buf []byte) error {
 type varintLengthField struct {
 	startOffset int
 	length      int64
+	adjusted    bool
+	size        int
```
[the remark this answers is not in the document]
wladh: size could always be calculated, but it involves encoding the varint into a buffer, so not very cheap.

length_field.go
```diff
@@ -31,22 +31,43 @@ func (l *lengthField) check(curOffset int, buf []byte) error {
...
+	return diff
+}
+
 func (l *varintLengthField) reserveLength() int {
-	return 0
+	return l.size
+}
+
+func (l *varintLengthField) run(curOffset int, buf []byte) error {
+	if !l.adjusted {
```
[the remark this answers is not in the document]
wladh: I can remove it, but I think the failure won't be too obvious (it will have length 0, which could happen in other scenarios as well).

packet_encoder.go
```diff
@@ -50,3 +50,13 @@ type pushEncoder interface {
...
 }
+
+// dynamicPushEncoder extends the interface of pushEncoder for uses cases where the length of the
+// fields itself is unknown until its value was computed (for instance varint encoded lenght
+// fields).
+type dynamicPushEncoder interface {
+	pushEncoder
+
+	// Called during pop() to adjust the length of the field.
+	adjustLength(currOffset int) int
```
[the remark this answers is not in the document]
wladh: That will complicate the caller, since in practice most of the time they will call `reserveLength` at `push` time and then again at `pop` time just to readjust.
The initial call to `reserveLength` might not return 0 (especially if we go with your suggestion of computing the size on demand). So, most of the time at the adjust time you want the diff.
I will update the comment.

record.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (r *Record) getTotalLength() (int, error) {
+	var prep prepEncoder
+	if !r.length.adjusted {
+		if err := r.encode(&prep); err != nil {
+			return 0, err
+		}
+	}
+	return int(r.length.length) + r.length.size, nil
```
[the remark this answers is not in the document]
wladh: We could do that, but then we also want to cache it, because the total uncompressed length will be used for producer metrics and we don't want to re-encode it there.

record_batch.go
wladh: You're right, I can reuse that.

wladh:
It seems that `travis-ci` failed because one broker didn't come up. Not sure how to retrigger the check.

length_field.go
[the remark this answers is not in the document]
eapache: It looks pretty cheap to me? The buffer doesn't escape so it will be allocated on the stack, and the actual encoding itself is just a couple of bit-ops in a very short for-loop.

eapache:
I can retrigger CI as needed, I've never been able to figure out how to let contributors do that without giving them full access to everything.

packet_encoder.go
```diff
@@ -50,3 +50,14 @@ type pushEncoder interface {
 	// of data to the saved offset, based on the data between the saved offset and curOffset.
 	run(curOffset int, buf []byte) error
 }
+
+// dynamicPushEncoder extends the interface of pushEncoder for uses cases where the length of the
+// fields itself is unknown until its value was computed (for instance varint encoded lenght
```
eapache: typo s/lenght/length

record_batch.go
```diff
@@ -0,0 +1,250 @@
...
+		if err := pe.putRawBytes(b.compressedRecords); err != nil {
+			return err
+		}
+		if err := pe.pop(); err != nil {
+			return err
+		}
+		return pe.pop()
+	}
+
+	var raw []byte
```
eapache: if you move this whole block (down to line ~126) into a method, and invert the check on `b.compressedRecords` you can get rid of lines 77-85 which are duplicates of 127-135.

wladh:
Maybe I should add a `dynamicPushDecoder` that will call a `decodeField` method during push and skip `reserveLen`.

eapache:
This is really nice, thanks!

PR #974, opened 2017-10-31
Add consumer support for Kafka 0.11 messages

wladh:
This change adds support for new record format of Kafka 0.11 in consumer. Transactions are not supported (so isolation level is always ReadUncommitted) and neither are idempotent messages.
Issue #901


fetch_request.go
```diff
@@ -29,16 +29,25 @@ type FetchRequest struct {
 	MinBytes    int32
 	MaxBytes    int32
 	Version     int16
+	Isolation   int8
 	blocks      map[string]map[int32]*fetchRequestBlock
 }
 
+const (
+	ReadUncommitted = 0
```
eapache: it's worth strongly-typing these

consumer.go
```diff
@@ -542,7 +511,52 @@ func (child *partitionConsumer) parseResponse(response *FetchResponse) ([]*Consu
...
+	batch := block.Records.recordBatch
+
+	for _, rec := range batch.Records {
+		offset := batch.FirstOffset + rec.OffsetDelta
+		if prelude && offset < child.offset {
+			continue
+		}
+		prelude = false
+
+		millis := batch.FirstTimestamp + rec.TimestampDelta
```
eapache: ah sorry, I didn't notice this in your previous PRs, but these fields should be represented in the actual wire structs as `time.Time` and the encode/decode methods should take care of doing the millis-to-`Time` conversion

eapache:
Thanks!

consumer.go
```diff
@@ -14,8 +14,9 @@ type ConsumerMessage struct {
 	Topic          string
 	Partition      int32
 	Offset         int64
-	Timestamp      time.Time // only set if kafka is version 0.10+, inner message timestamp
-	BlockTimestamp time.Time // only set if kafka is version 0.10+, outer (compressed) block timestamp
+	Timestamp      time.Time       // only set if kafka is version 0.10+, inner message timestamp
```
tutengdihuang, line 17: Can u tell me the this timeStamp is the time when the message is consumed?

PR #976, opened 2017-11-02
Add producer support for Kafka 0.11 records

wladh:
This change adds support for new record format of Kafka 0.11 in consumer. Transactions are not supported and neither are idempotent messages.
Issue #901


eapache:
Thanks!

PR #978, opened 2017-11-02
add format of an GetOffset argument to documentation

mkevac:
We have spent two damn days debugging this... :-(


eapache:
Sorry :(

It should really be a `time.Time` except it made the special values for newest/oldest more awkward, and it's too late now.

Thanks for the documentation!

mkevac:
No problem!

PR #981, opened 2017-11-16
add block.LastStableOffset check ,only block.LastStableOffset >0 mean…

xhbpiao:
When spending 0.11 kafka, block.LastStableOffset has been -1, so each time you get the data can only be parsed one. Cause serious consumer congestion

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -547,7 +547,7 @@ func (child *partitionConsumer) parseRecords(block *FetchResponseBlock) ([]*Cons
 			incomplete = true
 		}
 
-		if child.offset > block.LastStableOffset {
+		if block.LastStableOffset > 0 && child.offset > block.LastStableOffset {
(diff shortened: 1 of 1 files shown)


eapache:
Is this the cause of https://github.com/Shopify/sarama/issues/980?

cc @wladh

xhbpiao:
@eapache  yes

xhbpiao:
I use filebeat to collect the log, But it's official website supports the latest version is 0.10.  I have to use 0.11 protocol to parse the message, otherwise it will cause congestion。
May be the result of the reason of filebeat LastStableOffset resolution fail

wladh:
Actually we can do away with the whole check for now. Since we're always at `ReadUncommitted` isolation level, the stable offset doesn't matter as we're reading everything.
Also, it's wrong as it should have been performed before returning one messages.

PR #985, opened 2017-11-17
add CreatePartitionsRequest/Response

buyology:
this adds the request/response pair for `CreatePartitions`

### DIFF
--- a/create_partitions_request.go
+++ b/create_partitions_request.go
@@ -1,0 +1,119 @@
+package sarama
+
+import "time"
+
+type CreatePartitionsRequest struct {
+	TopicPartitions map[string]*TopicPartition
+	Timeout         time.Duration
+	ValidateOnly    bool
+}
+
+func (c *CreatePartitionsRequest) encode(pe packetEncoder) error {
+	if err := pe.putArrayLength(len(c.TopicPartitions)); err != nil {
+		return err
+	}
+
+	for topic, partition := range c.TopicPartitions {
+		if err := pe.putString(topic); err != nil {
+			return err
+		}
+		if err := partition.encode(pe); err != nil {
+			return err
+		}
+	}
+
+	pe.putInt32(int32(c.Timeout / time.Millisecond))
+
+	pe.putBool(c.ValidateOnly)
+
+	return nil
+}
+
+func (c *CreatePartitionsRequest) decode(pd packetDecoder, version int16) (err error) {
+	n, err := pd.getArrayLength()
+	if err != nil {
+		return err
+	}
+	c.TopicPartitions = make(map[string]*TopicPartition, n)
+	for i := 0; i < n; i++ {
+		topic, err := pd.getString()
+		if err != nil {
+			return err
+		}
+		c.TopicPartitions[topic] = new(TopicPartition)
+		if err := c.TopicPartitions[topic].decode(pd, version); err != nil {
+			return err
+		}
+	}
+
+	timeout, err := pd.getInt32()
+	if err != nil {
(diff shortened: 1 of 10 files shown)


buyology:
sorry for the multiple pushes. realised some style-related things and fixed a small error in the `getNullableStrings`-that had previously prevented me from using it, while writing up some of the other new request/response pairs (I have all of the `acl`-request/responses, 29-31, almost done as well).

re: travis — seems to fail already at the cluster install?

prep_encoder.go
```diff
@@ -44,6 +44,10 @@ func (pe *prepEncoder) putArrayLength(in int) error {
 	return nil
 }
 
+func (pe *prepEncoder) putBool(in bool) {
+	pe.length++
```
eapache: might be tempted to just delegate this to `putInt8` since that's the effective implementation elsewhere

real_decoder.go
```diff
@@ -143,11 +155,10 @@ func (rd *realDecoder) getString() (string, error) {
 }
 
 func (rd *realDecoder) getNullableString() (*string, error) {
-	tmp, err := rd.getInt16()
-	if err != nil || tmp == -1 {
+	str, err := rd.getString()
+	if err != nil || str == "" {
```
eapache: I think the whole point of nullable strings was to distinguish between a length of `-1` (null) and a length of `0` (`""`) which this check doesn't do, because `getString` already turns them both into `""`.

create_partitions_request.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (t *TopicPartition) encode(pe packetEncoder) error {
+	pe.putInt32(t.Count)
+
+	if len(t.Assignment) == 0 {
+		pe.putInt32(-1)
+		return nil
+	}
+
+	pe.putInt32(int32(len(t.Assignment)))
```
eapache: `putArrayLength`

create_partitions_request.go
```diff
@@ -0,0 +1,119 @@
...
+func (r *CreatePartitionsRequest) version() int16 {
+	return 0
+}
+
+func (r *CreatePartitionsRequest) requiredVersion() KafkaVersion {
+	return V1_0_0_0
+}
+
+type TopicPartition struct {
+	Count      int32
```
eapache: I'm looking at https://kafka.apache.org/protocol.html#The_Messages_CreateTopics and this doesn't seem quite right. Aren't you missing the number of replication factor?

create_partitions_request.go
buyology: Done

real_decoder.go
buyology: I factored out `getStringLength()` now to allow for exactly this logic. the problem with the former `getNullableString()` impl was that it consumed the length so when calling out to `getString()` the output became corrupted.

prep_encoder.go
buyology: the above is mostly for symmetry :) bool is quite common in the new request/responses + just thinking it's nice to have `getBool()` sanity check for valid values. what do you think. is it ok to leave in?

create_partitions_request.go
buyology: Hm. This is actually https://kafka.apache.org/protocol#The_Messages_CreatePartitions, so it's only about adding about new partitions to an already existing topic. But maybe there is a better name than the official `"Count"`.

eapache: 🤦‍♂️ alright then

eapache:
Thanks!

PR #989, opened 2017-11-24
String functions for KafkaVersion

backlin:
Added protobuf style functionality for 

* parsing a string to `KafkaVersion`, allowing it for example to be set in a config file, and
* rendering a `KafkaVersion` as a string, allowing it for example to be printed to a log.

Sorry for the messy commit log. Please squash if merging.


eapache:
Thanks. I think it probably makes more sense to use printf/scanf for these rather than map constants.

backlin:
Yes, you're right. I thought of the version as an enum with only certain allowed values, but that also places a requirement on explicitly stating all Kafka versions as public variables. The scanf way is more flexible, and it really doesn't matter if you set a non-existent version since it is still comparable to other versions.

Will fix the build errors in a little while and update.

eapache:
Thanks!

PR #997, opened 2017-12-04
Escape periods in Kafka version regex

cassidykelly:
This change makes error logging more informative for invalid version strings.


eapache:
Nice catch, thanks.

Issue #962, opened 2017-10-05
High Rate Producer Corruption With Large Messages


wingedpig:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*
Sarama Version: 1.13.0
Kafka Version: kafka_2.11-0.11.0.1
Go Version: 1.9

##### Configuration

What configuration values are you using for Sarama and Kafka?

Default installation for Kafka, except that `delete.topic.enable=true`. One instance.

##### Problem Description

Run the test producer at https://gist.github.com/wingedpig/c635e77fcd42d95cfa7923a541724f6b. It creates a test partition and fills it with 15 messages. Then run the test consumer, at https://gist.github.com/wingedpig/6d4999591fc305b683b9fd7cbeaed240, which tries to read the partition. It fails with the following errors:

```$ ./test
[sarama] 2017/10/05 09:45:53 Initializing new client
[sarama] 2017/10/05 09:45:53 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2017/10/05 09:45:53 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2017/10/05 09:45:53 client/metadata fetching metadata for all topics from broker 127.0.0.1:9092
[sarama] 2017/10/05 09:45:53 Connected to broker at 127.0.0.1:9092 (unregistered)
[sarama] 2017/10/05 09:45:53 client/brokers registered new broker #0 at localhost:9092
[sarama] 2017/10/05 09:45:53 Successfully initialized new client
[sarama] 2017/10/05 09:45:53 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2017/10/05 09:45:53 Connected to broker at localhost:9092 (registered as #0)
[sarama] 2017/10/05 09:45:53 consumer/broker/0 added subscription to test/0
[sarama] 2017/10/05 09:45:53 consumer/broker/0 disconnecting due to error processing FetchRequest: kafka: error decoding packet: unknown magic byte (2)
[sarama] 2017/10/05 09:45:53 Closed connection to broker localhost:9092
[sarama] 2017/10/05 09:45:53 kafka: error while consuming test/0: kafka: error decoding packet: unknown magic byte (2)
```
If you delete the test partition, uncomment the Sleep line in the test producer, and re-run the test, the consumer is able to read the messages without issue. 

[...]

PR #998, opened 2017-12-05
Check for insufficient data when we try to get the records.

wladh:
In some corner cases (like v2 records produced on the topic but the
sarama consumer uses v0 Fetch API with a limit less than one record size)
we might receive a truncated record batch. We want to be able to make
progress by increasing the fetch limit, so don't return an error, but
rater mark the batch as partial trailing record.
Fixes issue #962

### DIFF
--- a/record_batch.go
+++ b/record_batch.go
@@ -161,6 +161,11 @@ func (b *RecordBatch) decode(pd packetDecoder) (err error) {
 	bufSize := int(batchLen) - recordBatchOverhead
 	recBuffer, err := pd.getRawBytes(bufSize)
 	if err != nil {
+		if err == ErrInsufficientData {
+			b.PartialTrailingRecord = true
+			b.Records = nil
+			return nil
+		}
 		return err
 	}
 


tcrayford:
@wladh what does the java client do here?

eapache:
This fix makes sense to me, though I'm also curious what the java client does here.

eapache:
Thanks for digging into this!

wladh:
From what I gather (not familiar with Kafka codebase) the java client doesn't have this incremental fetch size logic. So it basically fails the way sarama was failing before this commit: https://github.com/apache/kafka/blob/4b8a29f12a142d02be0b64eac71975b6a129d04a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java#L840-L855

tcrayford:
@wladh that doesn't line up with my testing, which seemed to indicate everything "just worked" with the jvm clients.

wladh:
@tcrayford like I said I'm not familiar with the code base and have never used the Java client (this was just looking at the code at HEAD in their repo, maybe things were changed) so I might be missing something. One difference that I see is that the default fetch size for Java client is 1MB, whereas sarama starts at 32KB. But I don't know how big your messages were or if you changed any of these params.

PR #1006, opened 2017-12-15
partition consumer offset to last offset delta

### LINKED ISSUE #1005: Consuming a compacted topic fails with ErrIncompleteResponse
##### Versions

Sarama Version: v1.15.0 - Bump changelog - 3b1b38866a79f06deddf0487d5c27ba0697ccd65
Kafka Version: 1.0.0
Go Version: 1.9.2

##### Configuration
Sarama: 
```
cfg.Version = sarama.V1_0_0_0
```

Kafka itself has nothing noteworthy. The topic I am consuming is log compacted, which seems to be the critical thing here.
##### Logs

I couldn't find anything related in Kafka logs. 

Offsets in ```compacted-topic```:
```
38
40
42
43
45
```
Logs from my example: 

```
go run main.go --kafka broker0:9092 --topic compacted-topic --partition 16 --offset 41
SARAMA 2017/12/15 11:12:44 Initializing new client
SARAMA 2017/12/15 11:12:44 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
SARAMA 2017/12/15 11:12:44 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
SARAMA 2017/12/15 11:12:44 client/metadata fetching metadata for all topics from broker broker0:9092
SARAMA 2017/12/15 11:12:44 Connected to broker at broker0:9092 (unregistered)
SARAMA 2017/12/15 11:12:44 client/brokers registered new broker #2 at broker2:9092
SARAMA 2017/12/15 11:12:44 client/brokers registered new broker #4 at broker4:9092
SARAMA 2017/12/15 11:12:44 client/brokers registered new broker #1 at broker1:9092
SARAMA 2017/12/15 11:12:44 client/brokers registered new broker #3 at broker3:9092
SARAMA 2017/12/15 11:12:44 client/brokers registered new broker #0 at broker0:9092
SARAMA 2017/12/15 11:12:44 Successfully initialized new client
SARAMA 2017/12/15 11:12:44 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
SARAMA 2017/12/15 11:12:44 Connected to broker at broker2:9092 (registered as #2)
SARAMA 2017/12/15 11:12:45 consumer/broker/2 added subscription to compacted-topic/16
2017/12/15 11:12:45 
FETCH:
	compacted-topic/16/41
2017/12/15 11:12:45 partitionConsumer[16].chooseStartingOffset: offset: 41
sarama.partitionConsumer[compacted-topic/16].parseRecords: prelude: true, batch.FirstOffset: 39, rec.OffsetDelta: 1, offset: 40, child.offset: 41
sarama.partitionConsumer[compacted-topic/16].parseRecords: incomplete: false, len(messages): 0, batch: &{FirstOffset:39 PartitionLeaderEpoch:0 Version:2 Codec:2 Control:false LastOffsetDelta:2 FirstTimestamp:2017-12-13 10:19:18.805 +0100 CET MaxTimestamp:2017-12-13 10:19:18.805 +0100 CET ProducerID:-1 ProducerEpoch:-1 FirstSequence:-1 Records:[0xc4200aed20] PartialTrailingRecord:false compressedRecords:[] recordsLen:5343}

SARAMA 2017/12/15 11:12:45 kafka: error while consuming compacted-topic/16: kafka: response did not contain all the expected topic/partition blocks
SARAMA 2017/12/15 11:12:45 consumer/broker/2 abandoned subscription to compacted-topic/16 because kafka: response did not contain all the expected topic/partition blocks
SARAMA 2017/12/15 11:12:47 consumer/compacted-topic/16 finding new broker
SARAMA 2017/12/15 11:12:47 client/metadata fetching metadata for [compacted-topic] from broker broker0:9092
SARAMA 2017/12/15 11:12:47 consumer/broker/2 added subscription to compacted-topic/16
2017/12/15 11:12:47 
FETCH:
	compacted-topic/16/41
SARAMA 2017/12/15 11:12:47 kafka: error while consuming compacted-topic/16: kafka: response did not contain all the expected topic/partition blocks
SARAMA 2017/12/15 11:12:47 consumer/broker/2 abandoned subscription to compacted-topic/16 because kafka: response did not contain all the expected topic/partition blocks
^Csignal: interrupt

```
##### Problem Description

Sarama ```partitionConsumer``` fails with a ```ErrIncompleteResponse``` in ```parseRecords```. This happens when the consumer requests an offset that has been compacted away. As far as I can see, Sarama doesn't advance the requested offset but rather keeps requesting the compacted offset and keeps failing. I guess the expected behaviour would be to ask for the next available offset which in the example is offset `42`.

#### Example

```go
package main

import (
	"log"
	"os"

	"github.com/Shopify/sarama"
	"github.com/spf13/pflag"
)

var (
	kafka     = pflag.StringSlice("kafka", []string{"broker0:9092"}, "kafka address")
	topic     = pflag.String("topic", "compacted-topic", "topic to consume")
	partition = pflag.Int32("partition", 16, "partition to consume")
	offset    = pflag.Int64("offset", 41, "starting offset")
)

func init() {
	pflag.Parse()

}

func main() {
	cfg := sarama.NewConfig()
	cfg.Version = sarama.V1_0_0_0

	sarama.Logger = log.New(os.Stdout, "SARAMA ", log.LstdFlags)

	client, err := sarama.NewClient(*kafka, cfg)
	if err != nil {
		log.Fatalf("error creating client: %v", err)
	}

	consumer, err := sarama.NewConsumerFromClient(client)
	if err != nil {
		log.Fatalf("error creating consumer: %v", err)
	}

	pc, err := consumer.ConsumePartition(*topic, *partition, *offset)
	if err != nil {
		log.Fatalf("error creating partition consumer: %v", err)
	}

	for msg := range pc.Messages() {
		log.Printf("received msg offset: %v", msg.Offset)
	}
}

```

SamiHiltunen:
Hey, 

This seems to fix #1005. As far as I understood, we should advance partition consumer's offset with the LastOffsetDelta as this might be higher than the last record's offset in the batch if some of the offsets were compacted away.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -523,6 +523,7 @@ func (child *partitionConsumer) parseRecords(batch *RecordBatch) ([]*ConsumerMes
 	var messages []*ConsumerMessage
 	var incomplete bool
 	prelude := true
+	originalOffset := child.offset
 
 	for _, rec := range batch.Records {
 		offset := batch.FirstOffset + rec.OffsetDelta
@@ -547,9 +548,15 @@ func (child *partitionConsumer) parseRecords(batch *RecordBatch) ([]*ConsumerMes
 		}
 	}
 
-	if incomplete || len(messages) == 0 {
+	if incomplete {
 		return nil, ErrIncompleteResponse
 	}
+
+	child.offset = batch.FirstOffset + int64(batch.LastOffsetDelta) + 1
+	if child.offset <= originalOffset {
+		return nil, ErrConsumerOffsetNotAdvanced
+	}
+
 	return messages, nil
 }
 
--- a/consumer_test.go
+++ b/consumer_test.go
@@ -389,10 +389,12 @@ func TestConsumerExtraOffsets(t *testing.T) {
 	newFetchResponse.AddRecord("my_topic", 0, nil, testMsg, 2)
 	newFetchResponse.AddRecord("my_topic", 0, nil, testMsg, 3)
 	newFetchResponse.AddRecord("my_topic", 0, nil, testMsg, 4)
+	newFetchResponse.SetLastOffsetDelta("my_topic", 0, 4)
 	newFetchResponse.SetLastStableOffset("my_topic", 0, 4)
 	for _, fetchResponse1 := range []*FetchResponse{legacyFetchResponse, newFetchResponse} {
 		var offsetResponseVersion int16
 		cfg := NewConfig()
+		cfg.Consumer.Return.Errors = true
 		if fetchResponse1.Version >= 4 {
 			cfg.Version = V0_11_0_0
 			offsetResponseVersion = 1
@@ -426,8 +428,19 @@ func TestConsumerExtraOffsets(t *testing.T) {
 
 		// Then: messages with offsets 1 and 2 are not returned even though they
 		// are present in the response.
-		assertMessageOffset(t, <-consumer.Messages(), 3)
-		assertMessageOffset(t, <-consumer.Messages(), 4)
+		select {
+		case msg := <-consumer.Messages():
+			assertMessageOffset(t, msg, 3)
+		case err := <-consumer.Errors():
+			t.Fatal(err)
+		}
+
+		select {
+		case msg := <-consumer.Messages():
+			assertMessageOffset(t, msg, 4)
+		case err := <-consumer.Errors():
+			t.Fatal(err)
+		}
 
 		safeClose(t, consumer)
 		safeClose(t, master)
@@ -490,6 +503,7 @@ func TestConsumerNonSequentialOffsets(t *testing.T) {
 	newFetchResponse.AddRecord("my_topic", 0, nil, testMsg, 5)
 	newFetchResponse.AddRecord("my_topic", 0, nil, testMsg, 7)
 	newFetchResponse.AddRecord("my_topic", 0, nil, testMsg, 11)
+	newFetchResponse.SetLastOffsetDelta("my_topic", 0, 11)
 	newFetchResponse.SetLastStableOffset("my_topic", 0, 11)
 	for _, fetchResponse1 := range []*FetchResponse{legacyFetchResponse, newFetchResponse} {
 		var offsetResponseVersion int16
--- a/errors.go
+++ b/errors.go
@@ -37,6 +37,10 @@ var ErrShuttingDown = errors.New("kafka: message received by producer in process
 // ErrMessageTooLarge is returned when the next message to consume is larger than the configured Consumer.Fetch.Max
 var ErrMessageTooLarge = errors.New("kafka: message is larger than Consumer.Fetch.Max")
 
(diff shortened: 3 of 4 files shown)


SamiHiltunen:
Tests are passing locally, not quite sure what is up with Travis.

eapache:
Travis seems to be experiencing some issues (https://github.com/travis-ci/travis-ci/issues/8898) - I've played with a few of the workarounds listed in that issue but none of them have worked for me so far.

consumer.go
```diff
@@ -547,9 +547,11 @@ func (child *partitionConsumer) parseRecords(batch *RecordBatch) ([]*ConsumerMes
 		}
 	}
 
-	if incomplete || len(messages) == 0 {
+	if incomplete {
 		return nil, ErrIncompleteResponse
 	}
+
+	child.offset = batch.FirstOffset + int64(batch.LastOffsetDelta) + 1
```
eapache: Should this be guarded by a check that this actually advances the offset? If it doesn't and/or if messages were present which don't line up, we should probably still return an error of some sort.

eapache:
@wladh I think this makes sense, but curious for your opinion here; looks like another subtle difference between the old and new record-set formats.

consumer.go
SamiHiltunen: Good call, I added a check for that now. The check is after the assignment rather than guarding it as L554 should not advance the offset when processing a normal non-compacted topic.

wladh:
I think this fix makes sense. It seems to be what the Java consumer is doing as well  (`nextOffset()` for new records format returns `lastOffset()+1`).

consumer.go
```diff
@@ -547,9 +548,15 @@ func (child *partitionConsumer) parseRecords(batch *RecordBatch) ([]*ConsumerMes
...
+	if child.offset == originalOffset {
```
eapache: `<=`?

consumer.go
```diff
@@ -523,6 +523,7 @@ func (child *partitionConsumer) parseRecords(batch *RecordBatch) ([]*ConsumerMes
 	var messages []*ConsumerMessage
 	var incomplete bool
 	prelude := true
+	originalOffset := child.offset
```
eapache: should we save this value around L550 after we've already processed the messages?

consumer.go
SamiHiltunen: Not sure if this would happen but it is definitely better to check this case. I've the operator to ```<=``` now.

consumer.go
SamiHiltunen: I don't think we should do that. If we are processing a non-compacted stream, the final message would set the offset correctly in the loop. In this case, the offset advancing check would fail incorrectly.

eapache:
Looks like `TestConsumerExtraOffsets` is failing now. Otherwise this looks pretty good to me.

SamiHiltunen:
I've fixed the failing tests now and rebased on master.

eapache:
Thanks!

PR #852, opened 2017-03-27
Support Create Topics

cmshetty:
We use this change to create Kafka topics in our deployment process.

PR #1007, opened 2017-12-18
add CreateTopicsRequest/Response

buyology:
follows on from #852 with the necessary request/response pair for creating topics.


eapache:
Thanks!

PR #1009, opened 2017-12-18
add Describe/Create/DeleteAcls

buyology:
continuing the `AdminClient` request-response party with `Describe`/`Create`/`DeleteAcls`.

These share some common stuff so easier to PR them together.

### DIFF
--- a/acl_bindings.go
+++ b/acl_bindings.go
@@ -1,0 +1,119 @@
+package sarama
+
+type Resource struct {
+	ResourceType AclResourceType
+	ResourceName string
+}
+
+func (r *Resource) encode(pe packetEncoder) error {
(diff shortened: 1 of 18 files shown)


acl_bindings.go
```diff
@@ -0,0 +1,119 @@
...
+	permissionType, err := pd.getInt8()
+	if err != nil {
+		return err
+	}
+	a.PermissionType = AclPermissionType(permissionType)
+
+	return nil
+}
+
+type ResourceAcls struct {
```
eapache: I'm not sure why this is in the general file, it seems specific to the one request type?

eapache:
One comment but it makes basically no difference. Thanks for your work on this and sorry it's taken me so long to get to it.

PR #1010, opened 2017-12-18
refactor ConsumerMetadataReq/Resp -> FindCoordinatorReq/Resp

buyology:
As part of `KIP-98` and for the impl of the transactional producer the `GroupCoordinatorRequest`/`ConsumerMetadataRequest` has been renamed `FindCoordinatorRequest` and now serves the dual purpose of both looking up consumer group _and_ transaction coordinator.

This PR moves the logic to `FindCoordinatorRequest`, transitions `getConsumerMetadata` to use `FindCoordinatorRequest`, and provides transparent backwards compatibility for the `ConsumerMetadataRequest`.

### DIFF
--- a/find_coordinator_response.go
+++ b/find_coordinator_response.go
@@ -1,20 +1,15 @@
 package sarama
 
 import (
-	"net"
-	"strconv"
 	"time"
 )
 
 type FindCoordinatorResponse struct {
-	Version         int16
-	ThrottleTime    time.Duration
-	Err             KError
-	ErrMsg          *string
-	Coordinator     *Broker
-	CoordinatorID   int32  // deprecated: use Coordinator.ID()
-	CoordinatorHost string // deprecated: use Coordinator.Addr()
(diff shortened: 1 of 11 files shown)


find_coordinator_response.go
```diff
@@ -0,0 +1,121 @@
...
+
+type FindCoordinatorResponse struct {
+	Version         int16
+	ThrottleTime    time.Duration
+	Err             KError
+	ErrMsg          *string
+	Coordinator     *Broker
+	CoordinatorID   int32  // deprecated: use Coordinator.ID()
+	CoordinatorHost string // deprecated: use Coordinator.Addr()
+	CoordinatorPort int32  // deprecated: use Coordinator.Addr()
```
eapache: Can we take this opportunity to remove these, since this is technically a new struct?

buyology: Good call. Done.

eapache:
Thanks! Sorry I kind of forgot about this PR for a while.

PR #1014, opened 2017-12-28
Add Describe + AlterConfigs

Mongey:


alter_configs_request.go
```diff
@@ -0,0 +1,146 @@
+package sarama
+
+type AlterConfigsRequest struct {
+	Resources    []*AlterConfigsResource
+	ValidateOnly bool
+}
+
+type AlterConfigsResource struct {
+	T             ResourceType
```
eapache: I'd prefer a more useful name here... I guess `Type` is reserved and `ResourceType` is already the name of the struct, but maybe `ConfigResourceType` or something? Or make that the name of the struct and call the element just `ResourceType`?

alter_configs_request.go
```diff
@@ -0,0 +1,146 @@
...
+	Name          string
+	ConfigEntries []*ConfigEntryKV
+}
+
+type ConfigEntryKV struct {
+	Name  string
+	Value string
```
eapache: values are supposed to be nullable, this should probably be a `*string` or something?

resource_type.go
```diff
@@ -0,0 +1,12 @@
+package sarama
+
+type ResourceType int8
+
+const (
+	UnknownResource ResourceType = 0
```
eapache: curious where you got this set of constants? I don't see it in the docs anywhere?

eapache:
Thanks, this looks pretty good, just a few minor things.

resource_type.go
Mongey: Took me a while to track it down again. It's from [KIP-133](https://cwiki.apache.org/confluence/display/KAFKA/KIP-133%3A+Describe+and+Alter+Configs+Admin+APIs#KIP-133:DescribeandAlterConfigsAdminAPIs-WireFormattypes) 

I can add a comment in the code

alter_configs_request.go
Mongey: Whoops! you're right, I'll update

alter_configs_request.go
Mongey: 😊 I meant to do a second pass to rename this. I _think_ we can use `Type` here, since it's `type` that's reserved

alter_configs_response.go
```diff
@@ -0,0 +1,95 @@
...
+import "time"
+
+type AlterConfigsResponse struct {
+	ThrottleTime time.Duration
+	Resources    []*AlterConfigsResourceResponse
+}
+
+type AlterConfigsResourceResponse struct {
+	ErrorCode int16
+	ErrorMsg  string
```
Mongey: 🤔 Maybe I should squash `ErrorCode` and `ErrorMsg` into a `KError`

eapache: It’s not clear to me from the docs whether this *is* a KError or if it’s  some other error status set? Normally KErrors don’t include a string in the actual message, just the code.

describe_configs_request.go
```diff
@@ -0,0 +1,91 @@
+package sarama
+
+type Resource struct {
```
eapache: CI is failing because there is already a type named Resource being used for the ACL stuff

eapache:
Thanks!

andrejvanderzee:
@Mongey It looks like you have everything for the terraform plugin for Kafka topics now. Great stuff really appreciate your work, if I can help you out in one way or another (code review or whatever), I would love too!

PR #1019, opened 2018-01-09
Add NewMockBrokerListener() so that it's possible to test TLS connections

aeijdenberg:
and add some basic TLS tests now that we have it.

Our motivation is that we want to put similar tests in a library that wraps this one (https://github.com/rakutentech/kafka-firehose-nozzle/pull/21/), so are keen to get the `NewMockBrokerListener()` exposed.

### DIFF
--- a/client_tls_test.go
+++ b/client_tls_test.go
@@ -1,0 +1,206 @@
+package sarama
+
+import (
+	"math/big"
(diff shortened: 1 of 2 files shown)


eapache:
Cool thanks!

Issue #1013, opened 2017-12-28
Server throws exception while producing with low-level api


imjustfly:
##### Versions

Sarama Version: 1.14.0
Kafka Version: 1.0
Go Version: 1.8

##### Configuration

default configuration

##### Logs

```
[2017-12-27 19:08:00,897] ERROR [KafkaApi-2] Error when handling request {replica_id=-1,max_wait_time=250,min_bytes=1,topics=[{topic=drc.codis.codis-demo,partitions=[{partition=0,fetch_offset=0,max_bytes=32768}]}]} (kafka.server.KafkaApis)
java.lang.IllegalArgumentException: Illegal offset 1 following previous offset 1 (Offsets must increase monotonically).
	at org.apache.kafka.common.record.MemoryRecordsBuilder.appendWithOffset(MemoryRecordsBuilder.java:396)
	at org.apache.kafka.common.record.MemoryRecordsBuilder.append(MemoryRecordsBuilder.java:586)
	at org.apache.kafka.common.record.AbstractRecords.convertRecordBatch(AbstractRecords.java:134)
	at org.apache.kafka.common.record.AbstractRecords.downConvert(AbstractRecords.java:109)
	at org.apache.kafka.common.record.FileRecords.downConvert(FileRecords.java:253)
	at kafka.server.KafkaApis$$anonfun$kafka$server$KafkaApis$$convertedPartitionData$1$1$$anonfun$apply$4.apply(KafkaApis.scala:520)
	at kafka.server.KafkaApis$$anonfun$kafka$server$KafkaApis$$convertedPartitionData$1$1$$anonfun$apply$4.apply(KafkaApis.scala:518)
	at scala.Option.map(Option.scala:146)
	at kafka.server.KafkaApis$$anonfun$kafka$server$KafkaApis$$convertedPartitionData$1$1.apply(KafkaApis.scala:518)
	at kafka.server.KafkaApis$$anonfun$kafka$server$KafkaApis$$convertedPartitionData$1$1.apply(KafkaApis.scala:508)
	at scala.Option.flatMap(Option.scala:171)
	at kafka.server.KafkaApis.kafka$server$KafkaApis$$convertedPartitionData$1(KafkaApis.scala:508)
	at kafka.server.KafkaApis$$anonfun$kafka$server$KafkaApis$$createResponse$2$1.apply(KafkaApis.scala:556)
[...]

PR #1026, opened 2018-01-20
Correctly encode RecordBatch.LastOffsetDelta

bobrik:
Existing code always encodes zero. I'm not sure what the implications are, since consumers are working just fine, but `kafka.tools.DumpLogSegments` reports invalid data where all record batches have the same one record size.

cc @wladh


eapache:
Could this be the cause of #1013?

bobrik:
Seems possible, although I can't say for sure.

PR #1030, opened 2018-02-01
Do not choke on MessageSet without version

bobrik:
If we end un unlucky and MessageSet buffer is truncated right before
we can read magic value, we just return `ErrInsufficientData`, even
if some messages were previously successfully parsed.

This propagates all the way up to `partitionConsumer`, where we zero
parsed messages, which results in `ErrMessageTooLarge` returned
to the consumer.

This commit fixes the issue by setting `PartialTrailingMessage` to `true`
on `MessageSet`, making it possible to read successfully parsed messages.


eapache:
Nice find, thanks.

PR #1033, opened 2018-02-04
Clear all metadata when we have the latest topic info

Mongey:
Currently, If you list topics, then delete a topic, and poll to find out when the topic is actually deleted, it will never appear as deleted. 
This is because `updateMetadata` doesn't remove topics that no longer exist from the metadata.

An example can be seen over at [terraform-provider-kafka](https://github.com/Mongey/terraform-provider-kafka/blob/bde19c591305bc385f6349d58f7b7183a2df8def/kafka/resource_kafka_topic.go#L115-L157)

Here's a very naive solution that deletes the previous metadata, if you've got all the topic information from the broker.

### DIFF
--- a/client.go
+++ b/client.go
@@ -649,8 +649,9 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int)
 
 		switch err.(type) {
 		case nil:
+			allKnownMetaData := len(topics) == 0
 			// valid response, use it
-			shouldRetry, err := client.updateMetadata(response)
+			shouldRetry, err := client.updateMetadata(response, allKnownMetaData)
 			if shouldRetry {
 				Logger.Println("client/metadata found some partitions to be leaderless")
 				return retry(err) // note: err can be nil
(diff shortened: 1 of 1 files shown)


eapache:
Hmm, I would have thought this would already work? If you refresh metadata for a topic that has been deleted, you should presumably get back `ErrUnknownTopicOrPartition` which will result in the `delete(client.metadata, topic.Name)` being the only line run for that topic.

Oh, I guess this fix is *only* needed for the refresh-all case? It seems like polling the specific topic would be more efficient for your client anyway? But I suppose this PR is still more correct.

Mongey:
Yeah, this is only in the refresh all case, and I should actually just poll for the specific topic 😅, but this seems like undesired behaviour.

client.go
```diff
@@ -685,7 +686,9 @@ func (client *client) updateMetadata(data *MetadataResponse) (retry bool, err er
 	for _, broker := range data.Brokers {
 		client.registerBroker(broker)
 	}
-
+	if allKnownMetaData {
+		client.metadata = make(map[string]map[int32]*PartitionMetadata)
```
eapache: we should probably clear `client.cachedPartitionResults` here too then?

PR #1035, opened 2018-02-06
Proper mock expectations in SyncProducer.SendMessages

adamliesko:
Fixing incorrect expectations checking for the SyncProducer mock
in SendMessages call.

1) SendMessages now checks the CheckFunction on each and
every single message/expectation. It's consistent with singular
SendMessage mocked implementation.

2) SendMessages now correctly copies the right size of expectations
to use when matching with produced messages. Fixes off by one error.


adamliesko:
seems like it fixes https://github.com/Shopify/sarama/issues/776

eapache:
Looks great, thanks!

PR #1037, opened 2018-02-07
Account for possible offset skew

horkhe:
In our testing with Kafka v0.11.0.0 - v1.0.0 we discovered that compacted records lie about their offsets and `LastOffsetDelta` can be used to correct that error. According to the Kafka protocol [document](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-Responses), that is indeed the intended use for this field:
> LastOffsetDelta: The offset of the last message in the RecordBatch. This is used by the broker to ensure correct behavior even when Records within a batch are compacted out.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -524,10 +524,20 @@ func (child *partitionConsumer) parseRecords(batch *RecordBatch) ([]*ConsumerMes
 	var messages []*ConsumerMessage
 	var incomplete bool
 	prelude := true
-	originalOffset := child.offset
+
+	// Compacted records can lie about their offset delta, but that can be
+	// detected comparing the last offset delta from the record batch with
+	// actual offset delta of the last record. The difference (skew) should be
+	// applied to all records in the batch.
+	var offsetSkew int64
+	recordCount := len(batch.Records)
+	if recordCount > 0 {
+		lastRecordOffsetDelta := batch.Records[recordCount-1].OffsetDelta
+		offsetSkew = int64(batch.LastOffsetDelta) - lastRecordOffsetDelta
+	}
 
 	for _, rec := range batch.Records {
-		offset := batch.FirstOffset + rec.OffsetDelta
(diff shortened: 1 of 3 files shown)


consumer.go
```diff
@@ -524,10 +524,20 @@ func (child *partitionConsumer) parseRecords(batch *RecordBatch) ([]*ConsumerMes
 	var messages []*ConsumerMessage
 	var incomplete bool
 	prelude := true
-	originalOffset := child.offset
+
+	// Compacted records can lie about their offset delta, but that can be
+	// detected comparing the last offset delta from the record batch with
+	// actual offset delta of the last record. The difference (skew) should be
+	// applied to all records in the batch.
```
eapache: I'm not sure this is true? Isn't the calculated skew only guaranteed to be true for the final record in the batch? Depending on where the records got compacted out the skew could be arbitrarily distributed in the batch.

consumer.go
```diff
@@ -552,12 +562,6 @@ func (child *partitionConsumer) parseRecords(batch *RecordBatch) ([]*ConsumerMes
 	if incomplete {
 		return nil, ErrIncompleteResponse
 	}
-
-	child.offset = batch.FirstOffset + int64(batch.LastOffsetDelta) + 1
-	if child.offset <= originalOffset {
```
eapache: Perhaps the condition needs to change, but don't we still need to account for some case where the offset doesn't advance? Or is that handled by the `incomplete` logic now?

edit: the `incomplete` logic looks like it might be wrong now anyway, I'm not sure

consumer.go
horkhe: You are right. This is not correct. The problem is most likely with v0.11 producer. Please take a look at #1040 for more details. I am closing this PR.

Issue #1032, opened 2018-02-02
Not seeing all messages when syncProducer called from goroutines


kenschneider18:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*
Sarama Version: 1.15.0
Kafka Version: 0.11.0
Go Version: 1.9.2 darwin/amd64

##### Configuration

What configuration values are you using for Sarama and Kafka?

Sarama configuration
```go
config.Producer.RequiredAcks = sarama.WaitForAll
config.Producer.Retry.Max = 10
config.Producer.Return.Successes = true
config.Producer.Return.Errors = true
config.Version = sarama.V0_11_0_0
```

Kafka configuration
```properties
broker.id=0
delete.topic.enable=true
listeners=PLAINTEXT://localhost:9092
num.network.threads=3
num.io.threads=8
socket.send.buffer.bytes=102400
socket.receive.buffer.bytes=102400
socket.request.max.bytes=104857600
log.dirs=/usr/local/Cellar/kafka/kafka-log-1
num.partitions=1
num.recovery.threads.per.data.dir=1
offsets.topic.replication.factor=1
transaction.state.log.replication.factor=1
transaction.state.log.min.isr=1
auto.create.topics.enable=false 
log.retention.hours=168
log.segment.bytes=1073741824
log.retention.check.interval.ms=300000
zookeeper.connection.timeout.ms=6000
group.initial.rebalance.delay.ms=0
```

##### Logs

When filing an issue please provide logs from Sarama and Kafka if at all
possible. You can set `sarama.Logger` to a `log.Logger` to capture Sarama debug
output.

```
sarama: client.go:115: Initializing new client
sarama: config.go:351: ClientID is the default of 'sarama', you should consider setting it to something application-specific.
sarama: config.go:351: ClientID is the default of 'sarama', you should consider setting it to something application-specific.
sarama: client.go:646: client/metadata fetching metadata for all topics from broker localhost:9092
sarama: broker.go:146: Connected to broker at localhost:9092 (unregistered)
sarama: client.go:429: client/brokers registered new broker #0 at localhost:9092
sarama: client.go:655: client/metadata found some partitions to be leaderless
sarama: client.go:635: client/metadata retrying after 250ms... (3 attempts remaining)
sarama: client.go:646: client/metadata fetching metadata for all topics from broker localhost:9092
sarama: client.go:655: client/metadata found some partitions to be leaderless
[...]

PR #960, opened 2017-10-05
Added coverage badge

agnivade:
Linking up with codecov.io to show coverage. 

Added the badge in the README.

Issue #959, opened 2017-10-03
high rate producer error 


lghinet:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*
Sarama Version: latest 
Kafka Version: 0.11
Go Version: 1.9 / windows 

##### Configuration

What configuration values are you using for Sarama and Kafka?

##### Logs
```
2017/10/03 18:07:34 client/brokers registered new broker #11 at 10.1.3.90:9094
2017/10/03 18:07:34 client/brokers registered new broker #10 at 10.1.3.86:9094
2017/10/03 18:07:34 client/brokers registered new broker #9 at 10.1.3.88:9094
2017/10/03 18:07:34 Successfully initialized new client
2017/10/03 18:07:34 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
2017/10/03 18:07:34 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
2017/10/03 18:07:34 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
2017/10/03 18:07:34 producer/broker/11 starting up
2017/10/03 18:07:34 producer/broker/10 starting up
2017/10/03 18:07:34 producer/broker/9 starting up
2017/10/03 18:07:34 producer/broker/11 state change to [open] on go2_teskbb/0
2017/10/03 18:07:34 producer/broker/11 state change to [open] on go2_teskbb/9
2017/10/03 18:07:34 producer/broker/11 state change to [open] on go2_teskbb/6
2017/10/03 18:07:34 producer/broker/11 state change to [open] on go2_teskbb/3
2017/10/03 18:07:34 producer/broker/11 state change to [open] on go2_teskbb/12
2017/10/03 18:07:34 producer/broker/9 state change to [open] on go2_teskbb/4
2017/10/03 18:07:34 producer/broker/10 state change to [open] on go2_teskbb/14
2017/10/03 18:07:34 producer/broker/9 state change to [open] on go2_teskbb/1
2017/10/03 18:07:34 producer/broker/10 state change to [open] on go2_teskbb/8
2017/10/03 18:07:34 producer/broker/10 state change to [open] on go2_teskbb/11
2017/10/03 18:07:34 producer/broker/10 state change to [open] on go2_teskbb/2
2017/10/03 18:07:34 producer/broker/9 state change to [open] on go2_teskbb/10
[...]

PR #1040, opened 2018-02-09
 Add full version matrix produce/consume test

horkhe:
This PR adds a test that is checking that messages produced by clients of all supported versions can be consumed by clients of all supported version, also throwing all possible compressions into the mix. When the test runs it uses the value of the `KAFKA_VERSION` environment variable as the upper version bound. First it produces with clients of all version up to the upper limit and for each version it produces messages with different compression algorithms. Then it reads all produced messages with clients of all supported versions.

The test revealed several issues:

1. ~`kafka: response did not contain all the expected topic/partition blocks`~: 
  EDIT: I fixed that in this PR by only reporting `ErrIncompleteResponse` when there is no complete messages retrieved;
2. ~`lz4.Read: invalid header checksum: got 26 expected 130`: basically lz4 compression does not work at all, I commented it out in the test for now~; 
EDIT: My mistake, since lz4 was introduced in v0.10.0.0 older clients obviously cannot read it. I moved LZ4 testing to another test case that features version matrix starting with v0.10.0.0;
3. ~`Consumed unexpected offset: version=<any version>, index=3219, want={offset: 3220, value: msg:0.11.0.0:none:35}, got={offset: 3221, value: msg:0.11.0.0:none:35}` messages produced with client version 0.11.0.0 and higher are fetched with incorrect offset when consumed by any client version. So my earlier PR that suggested offsetSkew for 0.11.0.0-1.0.0 is incomplete. The fix is most likely should be in producer, because errors are returned by clients of all supported versions.~
EDIT: Hurray, #1041, fixed this issue, thank you @pkedy.

~WARNING: For some reason Travis-CI thinks that build succeeds for all targets even though tests fail.~ 
EDIT: I fixed that. Now test failure makes entire build fail. ~So you can see that builds for version 0.11.0.0 and higher fail.~
EIDT2: Builds for version 0.11.0.0 and higher do not fail anymore, thanks to #1041.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -482,9 +482,6 @@ feederLoop:
 
 func (child *partitionConsumer) parseMessages(msgSet *MessageSet) ([]*ConsumerMessage, error) {
 	var messages []*ConsumerMessage
-	var incomplete bool
-	prelude := true
-
 	for _, msgBlock := range msgSet.Messages {
 		for _, msg := range msgBlock.Messages() {
 			offset := msg.Offset
@@ -492,29 +489,22 @@ func (child *partitionConsumer) parseMessages(msgSet *MessageSet) ([]*ConsumerMe
 				baseOffset := msgBlock.Offset - msgBlock.Messages()[len(msgBlock.Messages())-1].Offset
 				offset += baseOffset
 			}
-			if prelude && offset < child.offset {
+			if offset < child.offset {
 				continue
 			}
-			prelude = false
-
-			if offset >= child.offset {
-				messages = append(messages, &ConsumerMessage{
-					Topic:          child.topic,
-					Partition:      child.partition,
-					Key:            msg.Msg.Key,
-					Value:          msg.Msg.Value,
-					Offset:         offset,
-					Timestamp:      msg.Msg.Timestamp,
-					BlockTimestamp: msgBlock.Msg.Timestamp,
-				})
-				child.offset = offset + 1
-			} else {
-				incomplete = true
-			}
+			messages = append(messages, &ConsumerMessage{
+				Topic:          child.topic,
+				Partition:      child.partition,
+				Key:            msg.Msg.Key,
+				Value:          msg.Msg.Value,
+				Offset:         offset,
+				Timestamp:      msg.Msg.Timestamp,
+				BlockTimestamp: msgBlock.Msg.Timestamp,
+			})
+			child.offset = offset + 1
 		}
 	}
-
-	if incomplete && len(messages) == 0 {
+	if len(messages) == 0 {
 		return nil, ErrIncompleteResponse
 	}
 	return messages, nil
@@ -522,42 +512,25 @@ func (child *partitionConsumer) parseMessages(msgSet *MessageSet) ([]*ConsumerMe
 
 func (child *partitionConsumer) parseRecords(batch *RecordBatch) ([]*ConsumerMessage, error) {
 	var messages []*ConsumerMessage
-	var incomplete bool
-	prelude := true
-	originalOffset := child.offset
-
 	for _, rec := range batch.Records {
 		offset := batch.FirstOffset + rec.OffsetDelta
-		if prelude && offset < child.offset {
+		if offset < child.offset {
 			continue
 		}
-		prelude = false
-
-		if offset >= child.offset {
-			messages = append(messages, &ConsumerMessage{
-				Topic:     child.topic,
-				Partition: child.partition,
-				Key:       rec.Key,
-				Value:     rec.Value,
-				Offset:    offset,
-				Timestamp: batch.FirstTimestamp.Add(rec.TimestampDelta),
-				Headers:   rec.Headers,
-			})
-			child.offset = offset + 1
-		} else {
-			incomplete = true
-		}
+		messages = append(messages, &ConsumerMessage{
+			Topic:     child.topic,
+			Partition: child.partition,
+			Key:       rec.Key,
+			Value:     rec.Value,
+			Offset:    offset,
+			Timestamp: batch.FirstTimestamp.Add(rec.TimestampDelta),
+			Headers:   rec.Headers,
+		})
+		child.offset = offset + 1
 	}
-
-	if incomplete && len(messages) == 0 {
+	if len(messages) == 0 {
 		return nil, ErrIncompleteResponse
 	}
-
-	child.offset = batch.FirstOffset + int64(batch.LastOffsetDelta) + 1
-	if child.offset <= originalOffset {
-		return nil, ErrConsumerOffsetNotAdvanced
-	}
-
 	return messages, nil
 }
 
--- a/consumer.go
+++ b/consumer.go
@@ -514,7 +514,7 @@ func (child *partitionConsumer) parseMessages(msgSet *MessageSet) ([]*ConsumerMe
 		}
 	}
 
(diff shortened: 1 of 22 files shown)


horkhe:
@eapache please take a look.

consumer.go
```diff
@@ -514,7 +514,7 @@ func (child *partitionConsumer) parseMessages(msgSet *MessageSet) ([]*ConsumerMe
 		}
 	}
 
-	if incomplete || len(messages) == 0 {
+	if incomplete && len(messages) == 0 {
```
eapache: I don't understand this change; `incomplete` can only be set if `messages` contains at least one complete message already, so this condition becomes impossible to satisfy? I assume this is the `kafka: response did not contain all the expected topic/partition blocks` fix you mention; I'd appreciate more details on that.

I'm looking at this method again, and I'm wondering if the `incomplete` boolean is even necessary anymore. I'm not sure what it was trying to accomplish originally (guarding against decreasing offsets maybe?), but a simple check on `len(messages) == 0` seems to be sufficient on its own? I'm not sure. That might even mean we don't need `prelude` anymore either, and can just unconditionally skip messages with `offset < child.offset`.

functional_consumer_test.go
```diff
@@ -59,3 +64,163 @@ func TestConsumerHighWaterMarkOffset(t *testing.T) {
...
+	for _, prodVer := range clientVersions {
+		for _, codec := range codecs {
+			prodCfg := NewConfig()
+			prodCfg.Version = prodVer
+			prodCfg.Producer.Return.Successes = true
+			prodCfg.Producer.Return.Errors = true
+			prodCfg.Producer.Flush.MaxMessages = flush
+			prodCfg.Producer.Compression = codec
+
+			p, err := NewSyncProducer(kafkaBrokers, prodCfg)
```
eapache: It might be simpler to run this test with an AsyncProducer so you can avoid spinning up a bunch of goroutines and tracking them with a waitgroup.

utils.go
```diff
@@ -139,21 +139,47 @@ func (v KafkaVersion) IsAtLeast(other KafkaVersion) bool {
...
+	V0_10_1_0 = newKafkaVersion(0, 10, 1, 0)
+	V0_10_1_1 = newKafkaVersion(0, 10, 1, 1)
+	V0_10_2_0 = newKafkaVersion(0, 10, 2, 0)
+	V0_10_2_1 = newKafkaVersion(0, 10, 2, 1)
+	V0_11_0_0 = newKafkaVersion(0, 11, 0, 0)
+	V0_11_0_1 = newKafkaVersion(0, 11, 0, 1)
+	V0_11_0_2 = newKafkaVersion(0, 11, 0, 2)
+	V1_0_0_0  = newKafkaVersion(1, 0, 0, 0)
+
+	SupportedVersions = []KafkaVersion{
```
eapache: Curious why you decided to expose these values publicly? I don't see a use for them outside the tests you wrote.

functional_consumer_test.go
horkhe: The point was to have several producers produce different types of messageSets/batchRecords concurrently. I certainly can use AsyncProducers for that although then I would need to store them somewhere and collect produced messages from all of their Success() channels, not sure if this implementation is going to be shorter though. I can implement it this way if you want.

eapache: Oh ok, I missed the `wg.Wait()` being fully outside the outer loop and thought the purpose was only for batching. This is fine then.

utils.go
horkhe: Sarama exposes the Protocol implementation that is used in my project [mailgun/kafka-pixy](https://github.com/mailgun/kafka-pixy), but some things are missing. And this is one of them. I realize that this is a selfish reason, so if you are against it, I will roll it back.

consumer.go
horkhe: You right, the error got away because this condition is now impossible to achieve. My reasoning was that incomplete record is supposed to be claimed as error only if there are no complete messages in the batch/set. But now looking at incomplete I understand that it actually does not mean partial trailing message as I thought.

I do not think that event `len(message) == 0` should be treated as an error. If Kafka broker decides to return an empty block, so what? it just means that there are no new message and sarama should just make another long polling fetch request.

utils.go
eapache: No, that's also fine, I just couldn't think of a use case for it but you provided one.

consumer.go
eapache: Ah, so the case where there are legitimately no messages in the block is already handled in `parseResponse`. If we get here and there are no messages we could extract, then I think it's legitimate to conclude an error of some sort since the server presumably sent us only messages we didn't ask for? I don't know if it's OK for the server to respond with a prelude but no actual messages...

horkhe: I think the reason why I saw `kafka: response did not contain all the expected topic/partition blocks` error while running this test, was exactly because the server was replying with a sets/batches containing prelude only. I will double check that tomorrow when I am back in the office.

pkedy:
@horkhe - Do your producer tests set `Headers` in the `ProducerMessage`s?  Even after the fix for #1032, if we set `Headers`, the following error occurs for at least half of the messages.

```
kafka server: Message was too large, server rejected it to avoid allocation error.
```

@eapache - There are other closed issues from the past that have the same error message but this might be a unique case just for when `Headers` are set.  Worth opening a new issue?

eapache:
Ya, please open a new issue. The implementation of `Headers` is fairly new, and I don't recall any past issues of "Message was too large" that didn't just end up being messages actually too large.

pkedy:
@eapache Now I'm not sure this is unique to `Headers`.  I was able to get the error to occur without errors also.  Referring back to the older issues on this (e.g. #959), I tried setting `config.Producer.MaxMessageBytes = 104857500` (100 less than the default value of `sarama.MaxRequestSize`) and it seems to be stable.  Under load, why would the defaults produce this error?

horkhe:
@pkedy nope, I do not set headers. Later we can add a test that sets headers, although that would be for a much smaller subset of versions.

functional_consumer_test.go
horkhe: You know what, I cannot reproduce `kafka: response did not contain all the expected topic/partition blocks` now at all. Looks like #1041 fixed the root cause of this issue too. So I suggest simplifying messageSet/recordBatch parsing a bit to make it easier to understand.

Note that I left:
```
	if len(messages) == 0 {
	}
```
intact. But honestly I think it would be better just ignore this case. Because what do you expect the library user to do about that error? Nothing but retry, and sarama can do that on its own.

horkhe:
@eapache After rebase to the latest master, the gzip compression encode/decode test started to fail. The test was broken by #1044, but a deeper problem is that CI tells that build is successful even if tests fail. In this PR that is fixed. But be careful merging PRs before the fix is in master. I fixed the gzip test by the way.

eapache:
Thank you for all of your work on this!

> CI tells that build is successful even if tests fail. In this PR that is fixed

That's really weird, I swear it used to fail properly and I haven't changed the config recently. Anyway, thank you for fixing this, I'm glad the only thing I missed due to this was a minor test failure.

horkhe:
Test errors stopped failing builds since #960. Because in bash if commands in a pipeline are separated with `;` the status of the pipeline is the status of the last command, so test execution status was shadowed by the last command.

PR #1044, opened 2018-02-12
Add support for compression levels, closes #1042

### LINKED ISSUE #1042: Configurable compression level
Currently there's no way to configure compression level for Gzip. We also want to add Zstd support to Kafka itself, where there are many compression levels:

* https://github.com/apache/kafka/pull/2267

There should be a setting to set compression level in producer config. I'm not sure whether we want separate config per codec (`CompressionLevelGZIP`, `CompressionLevelZSTD`) or one setting to rule them all (`CompressionLevel`). The former seems more flexible.

bobrik:
This makes it possible to set compression level for gzip:

```
config.Producer.CompressionLevel = flate.BestCompression
```

There are no other compression codecs that support levels,
but the code is generic to easily add other codecs with
numeric levels.

Default compression level tracks what Go libraries
set as the default level.

### DIFF
--- a/config.go
+++ b/config.go
@@ -99,6 +99,10 @@ type Config struct {
 		// The type of compression to use on messages (defaults to no compression).
 		// Similar to `compression.codec` setting of the JVM producer.
 		Compression CompressionCodec
+		// The level of compression to use on messages. The meaning depends
+		// on the actual compression type used and defaults to default compression
+		// level for the codec.
+		CompressionLevel int
 		// Generates partitioners for choosing the partition to send messages to
 		// (defaults to hashing the message key). Similar to the `partitioner.class`
 		// setting for the JVM producer.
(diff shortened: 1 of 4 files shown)


bobrik:
I started with `*int` for compression level with `nil` being the default, but that turned out to be ugly, because you can't take address of a constant like `flat.BestCompression`.

eapache:
I'm not sure if it's worth throwing any checks into `Validate()` or if we should just rely on the libraries we call during `encode()`?

bobrik:
@eapache I added another commit with validation check:

```
2018/02/12 19:51:34 Error creating producer: kafka: invalid configuration (gzip compression does not work with level 12: gzip: invalid compression level: 12)
```

eapache:
CI is going to fail because you're not accounting for the default value the same way in `Validate` as in `encode`

bobrik:
I just added the check.

eapache:
Thanks!

PR #1047, opened 2018-02-13
Added support for Metadata Request/Response v1

mimaison:
V1 enables to discover the cluster controller which is required to send CreateTopics, DeleteTopics and CreatePartitions requests

So far we've just added the v1 request/response "definition" just enough so v1 requests can be sent using `Broker.GetMetadata()` to find the controller.

`Client` is still using v0 metadata and also it would be nice to add a `Controller` field to `Broker.` We can do that in another PR.

Co-authored-by: Adrian Preston <prestona@uk.ibm.com>
Co-authored-by: Mickael Maison <mickael.maison@gmail.com>

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -18,6 +18,7 @@ import (
 type Broker struct {
 	id   int32
 	addr string
+	rack *string
 
 	conf          *Config
 	correlationID int32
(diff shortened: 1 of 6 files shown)


eapache:
Looks pretty good to me; for symmetry it would be nice to have the broker's `encode` also support the new version.

mimaison:
Thanks for the review, I'll add that and rebase

mimaison:
@eapache I've rebased on top of master and added the version field to broker's `encode`. Can you have another look ?

PR #1050, opened 2018-02-14
 Fix FindCoordinatorResponse.encode to allow nil Coordinator

horkhe:
`FindCoordinatorResponse.encode` function panics if `Coordinator` is `nil`, so I fixed that. Besides I added MockFindCoordinatorResponse to use in tests.

### DIFF
--- a/find_coordinator_response.go
+++ b/find_coordinator_response.go
@@ -4,6 +4,8 @@ import (
 	"time"
 )
 
+var NoNode = &Broker{id: -1, addr: ":-1"}
+
 type FindCoordinatorResponse struct {
 	Version      int16
 	ThrottleTime time.Duration
@@ -36,7 +38,7 @@ func (f *FindCoordinatorResponse) decode(pd packetDecoder, version int16) (err e
 	}
 
 	coordinator := new(Broker)
-	if err := coordinator.decode(pd, 0); err != nil {
+	if err := coordinator.decode(pd, version); err != nil {
 		return err
 	}
 	if coordinator.addr == ":0" {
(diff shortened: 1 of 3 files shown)


find_coordinator_response.go
```diff
@@ -60,10 +60,22 @@ func (f *FindCoordinatorResponse) encode(pe packetEncoder) error {
 		}
 	}
 
+	if f.Coordinator == nil {
+		pe.putInt32(0)
+		if err := pe.putString(""); err != nil {
+			return nil
+		}
+		pe.putInt32(0)
+		if f.Version >= 1 {
```
eapache: This is not matched by the fact that we hard-code passing version 0 into `Coordinator.encode`?

eapache:
It would be nice if you could add a test for this case.

buyology:
Good catch. Maybe we should line up the test of a nil value from the broker with the `Java` code which defines `NO_NODE` [like so](https://github.com/apache/kafka/blob/e31c0c9bdbad432bc21b583bd3c084f05323f642/clients/src/main/java/org/apache/kafka/common/Node.java#L24):

```java
    private static final Node NO_NODE = new Node(-1, "", -1);
```

find_coordinator_response.go
buyology: Hm. We are currently sharing the same decode/encode method for `Metadata*` and `FindCoordinator*` but since version 1 of the `MetadataResponse` they are not the equal (`Metadata` has rack whereas `FindCoordinator` does not.)

horkhe:
@eapache @buyology done. Version is not properly used in encoding/decoding. NoNode was introduced and used in encoding if Coordinator is nil. The test covers both versions.

buyology:
Sorry if I was unclear about my comment re: the shared `decode/encode`.

The tl;dr was that the `FindCoordinatorRequest/Response` does not have a `rack`-field:
https://kafka.apache.org/protocol#The_Messages_FindCoordinator

Sending a followup patch so let's continue there

horkhe:
Good catch, thank you @buyology.

PR #1052, opened 2018-02-14
Added support for v1 DeleteTopicsRequest

mimaison:
Interestingly, support for DeleteTopicsResponse v1 was there but not for the request !


eapache:
Thanks!

Issue #1048, opened 2018-02-13
proposal: create AdminClient interface


buyology:
Now that we have almost all of the administrative messages in place (and one in the pipeline #1027) and the updated `Metadata`-message for finding the coordinator as a PR as necessitated by some requests (#1047)  — a natural next step would be to provide some higher level interfaces (as requested in e.g. #1039). 

I therefore propose a new `AdminClient` interface, with a draft looking something like below. It could potentially be split up into smaller units if that makes sense for increased composability.

```go
type AdminClient interface {
	CreateTopic(topic string, detail *TopicDetail) error
	DeleteTopic(topic string) error
	CreatePartitions(topic string, count int32, assignment [][]int32, validateOnly bool) error
	DeleteRecords(topic string, partitionOffsets map[int32]int64) error

	DescribeConfig(resource ConfigResource) ([]ConfigEntry, error)
	AlterConfig(resourceType ConfigResourceType, name string, entries map[string]*string, validateOnly bool) error

	CreateAcl(resource Resource, acl Acl) error
	ListAcls(filter AclFilter) ([]ResourceAcls, error)
	DeleteAcl(filter AclFilter, validateOnly bool) ([]MatchingAcl, error)
}
```

cc @Mongey @mimaison @ongardie-ebay @chandradeepak

eapache:
Something like this makes a lot of sense to me, 👍.

My question is around how this interacts with `Client`; does it build on top (like the producer/consumer) or does it stand alone? How does it interact with the client's cache (e.g. if you use the admin client to delete a topic, does that get busted from the underlying client's cache somehow)?

eapache:
Having just read what I wrote here, I'm also wondering if we can stay away from `Client` in the naming just to avoid ambiguity. `ClusterAdmin` or `BrokerAdmin` or something?

mimaison:
Thanks for coming up with a proposal. It looks good overall. Just a few comments:
- we also need an API to list/describe topics
- many of the Kafka APIs can accept a list of topics/resources so it would be nice to be able to support this. For example create/delete several topics in 1 call
- we probably want to return tuples like `(*TopicError, error)` to differentiate failing to send the request and Kafka errors.

buyology:
> My question is around how this interacts with Client; does it build on top (like the producer/consumer) or does it stand alone? How does it interact with the client's cache (e.g. if you use the admin client to delete a topic, does that get busted from the underlying client's cache somehow)?
[...]

PR #1055, opened 2018-02-16
ClusterAdmin Support 

chandradeepak:
@eapache @buyology @Mongey @mimaison @ongardie-ebay 

I started working on ClusterAdmin support and wrote an initial draft version. Want to get your comments that I am moving in the right direction. After your comments i will close to the pull request.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -1,0 +1,375 @@
+package sarama
+
+import "errors"
+
+// ClusterAdmin is the administrative client for Kafka, which supports managing and inspecting topics,
+// brokers, configurations and ACLs. The minimum broker version required is 0.10.0.0.
+// Methods with stricter requirements will specify the minimum broker version required.
+// You MUST call Close() on a client to avoid leaks
+type ClusterAdmin interface {
+	// Creates a new topic. This operation is supported by brokers with version 0.10.1.0 or higher.
+	// It may take several seconds after CreateTopic returns success for all the brokers
+	// to become aware that the topic has been created. During this time, listTopics
+	// may not return information about the new topic.The validateOnly option is supported from version 0.10.2.0.
+	CreateTopic(topic string, detail *TopicDetail, validateOnly bool) error
+
+	// Delete a topic. It may take several seconds after the DeleteTopic to returns success
+	// and for all the brokers to become aware that the topics are gone.
+	// During this time, listTopics  may continue to return information about the deleted topic.
+	// If delete.topic.enable is false on the brokers, deleteTopic will mark
+	// the topic for deletion, but not actually delete them.
+	// This operation is supported by brokers with version 0.10.1.0 or higher.
+	DeleteTopic(topic string) error
+
+	// Increase the number of partitions of the topics  according to the corresponding values.
+	// If partitions are increased for a topic that has a key, the partition logic or ordering of
+	// the messages will be affected. It may take several seconds after this method returns
+	// success for all the brokers to become aware that the partitions have been created.
+	// During this time, ClusterAdmin#describeTopics may not return information about the
+	// new partitions. This operation is supported by brokers with version 1.0.0 or higher.
+	CreatePartitions(topic string, count int32, assignment [][]int32, validateOnly bool) error
+
+	// Delete records whose offset is smaller than the given offset of the corresponding partition.
+	// This operation is supported by brokers with version 0.11.0.0 or higher.
+	DeleteRecords(topic string, partitionOffsets map[int32]int64) error
+
+	// Get the configuration for the specified resources.
+	// The returned configuration includes default values and the Default is true
+	// can be used to distinguish them from user supplied values.
+	// Config entries where ReadOnly is true cannot be updated.
+	// The value of config entries where Sensitive is true is always nil so
+	// sensitive information is not disclosed.
+	// This operation is supported by brokers with version 0.11.0.0 or higher.
+	DescribeConfig(resource ConfigResource) ([]ConfigEntry, error)
+
+	// Update the configuration for the specified resources with the default options.
+	// This operation is supported by brokers with version 0.11.0.0 or higher.
+	// The resources with their configs (topic is the only resource type with configs
+	// that can be updated currently Updates are not transactional so they may succeed
+	// for some resources while fail for others. The configs for a particular resource are updated automatically.
+	AlterConfig(resourceType ConfigResourceType, name string, entries map[string]*string, validateOnly bool) error
+
+	// Creates access control lists (ACLs) which are bound to specific resources.
+	// This operation is not transactional so it may succeed for some ACLs while fail for others.
+	// If you attempt to add an ACL that duplicates an existing ACL, no error will be raised, but
+	// no changes will be made. This operation is supported by brokers with version 0.11.0.0 or higher.
+	CreateACL(resource Resource, acl Acl) error
+
+	// Lists access control lists (ACLs) according to the supplied filter.
+	// it may take some time for changes made by createAcls or deleteAcls to be reflected in the output of ListAcls
+	// This operation is supported by brokers with version 0.11.0.0 or higher.
+	ListAcls(filter AclFilter) ([]ResourceAcls, error)
+
+	// Deletes access control lists (ACLs) according to the supplied filters.
+	// This operation is not transactional so it may succeed for some ACLs while fail for others.
+	// This operation is supported by brokers with version 0.11.0.0 or higher.
+	DeleteACL(filter AclFilter, validateOnly bool) ([]MatchingAcl, error)
+
+	// Close shuts down the admin and closes underlying client.
+	Close() error
+}
+
+type clusterAdmin struct {
+	client Client
+	conf   *Config
+}
+
+// NewClusterAdmin creates a new ClusterAdmin using the given broker addresses and configuration.
+func NewClusterAdmin(addrs []string, conf *Config) (ClusterAdmin, error) {
+	client, err := NewClient(addrs, conf)
+	if err != nil {
+		return nil, err
+	}
+
+	//make sure we can retrieve the controller
+	_, err = client.Controller()
+	if err != nil {
+		return nil, err
+	}
+
+	ca := &clusterAdmin{
+		client: client,
+		conf:   client.Config(),
+	}
+	return ca, nil
+}
+
+func (ca *clusterAdmin) Close() error {
+	return ca.client.Close()
+}
+
+func (ca *clusterAdmin) Controller() (*Broker, error) {
+	return ca.client.Controller()
+}
+
+func (ca *clusterAdmin) CreateTopic(topic string, detail *TopicDetail, validateOnly bool) error {
+
+	if topic == "" {
+		return ErrInvalidTopic
+	}
+
+	if detail == nil {
+		return errors.New("You must specify topic details")
+	}
+
+	topicDetails := make(map[string]*TopicDetail)
+	topicDetails[topic] = detail
+
+	request := &CreateTopicsRequest{
+		TopicDetails: topicDetails,
+		ValidateOnly: validateOnly,
+	}
+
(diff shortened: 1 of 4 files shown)


admin.go
```diff
@@ -0,0 +1,162 @@
...
+
+// NewClusterAdmin creates a new ClusterAdmin using the given broker addresses and configuration.
+func NewClusterAdmin(addrs []string, conf *Config) (ClusterAdmin, error) {
+	client, err := NewClient(addrs, conf)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check that we are not dealing with a closed Client before processing any other arguments
+	if client.Closed() {
```
eapache: unnecessary since you literally just created it

admin.go
```diff
@@ -0,0 +1,162 @@
...
+
+	ca := &clusterAdmin{
+		client: client,
+		conf:   client.Config(),
+	}
+
+	return ca, nil
+}
+
+func (ca *clusterAdmin) handleResponses(rsp interface{}) {
```
eapache: I don't understand the motivation for grouping all the response-handling in a giant switch here

client.go
```diff
@@ -35,6 +35,8 @@ type Client interface {
 	// topic/partition, as determined by querying the cluster metadata.
 	Leader(topic string, partitionID int32) (*Broker, error)
 
+	Any() *Broker
```
eapache: I don't think there's a good reason to expose this publicly, it's never been a well-defined method in the first place.

admin.go
chandradeepak: Will delete that

admin.go
chandradeepak: I will handle all the responses . For now I clubbed it but I will handle them individually.

client.go
chandradeepak: I need to call broker.Createtopic, broker.Deletetopic etc from admin.go . But I need a broker that is open so I thought of exposing this and accessing a broker which is open. Is there a way to access a broker which is open easily or we need to do that in this file. Which do you recommend ?

chandradeepak: Will use client.Brokers instead of any.

chandradeepak:
@eapache  what is the convenient way to access broker (CreateTopic , DeleteTopic from client. I exposed Any() from client interface but i agree it is not needed.I tried using brokers but it is not guaranteed they are open always. My tests failed with broker not connected.  Can you please suggest any other way to access the underlying brokers which are opened .  I will make  those changes .

buyology:
I would suggest that we first settle on the design in #1048 before finalising the implementation. 

Also, the answer to the question about how to access the broker is that some of the methods (`CreateTopics`, `DeleteTopics`, `CreatePartitions`) must go through the `controller` — and the method for retrieving the controller is not yet implemented.

chandradeepak:
Thanks @buyology  , will wait for the design to be finalized and method for retrieving the controller.

chandradeepak:
@eapache  @buyology , I integrated the controller for this pull request. let me know if you have any comments or changes that needs to be done .

admin.go
```diff
@@ -0,0 +1,360 @@
+package sarama
+
+import "errors"
+
+// ClusterAdmin is the administrative client for Kafka, which supports managing and inspecting topics,
+// brokers, configurations and ACLs. The minimum broker version required is 0.10.0.0.
+// Methods with stricter requirements will specify the minimum broker version required.
+// This client was introduced in 0.11.0.0 and the API is still evolving.
```
eapache: I don't think this line is necessary in the documentation?

admin.go
```diff
@@ -0,0 +1,360 @@
...
+// You MUST call Close() on a client to avoid leaks
```
eapache: s/client/cluster admin

admin.go
```diff
@@ -0,0 +1,360 @@
...
+type ClusterAdmin interface {
+	// Creates a new topic .This operation is supported by brokers with version 0.10.1.0 or higher.
```
eapache: typo on the period

eapache: You don't need to specify the version requirement here, I prefer it at the end of the description where it already is for other methods.

admin.go
```diff
@@ -0,0 +1,360 @@
...
+	// It may take several seconds after CreateTopic returns success for all the brokers
+	// to become aware that the topics have been created. During this time, ClusterAdmin#listTopics
+	// and ClusterAdmin#describeTopics may not return information about the new topics.
+	// This operation is supported by brokers with version 0.10.1.0 or higher.
+	// The validateOnly option is supported from version 0.10.2.0.
+	CreateTopic(topic string, detail *TopicDetail) error
+
+	// Delete a batch of topics.
```
eapache: the method only takes one topic

admin.go
```diff
@@ -0,0 +1,360 @@
...
+// Methods with stricter requirements will specify the minimum broker version required.
+// This client was introduced in 0.11.0.0 and the API is still evolving.
+// You MUST call Close() on a client to avoid leaks
+type ClusterAdmin interface {
+	// Creates a new topic .This operation is supported by brokers with version 0.10.1.0 or higher.
+	// It may take several seconds after CreateTopic returns success for all the brokers
+	// to become aware that the topics have been created. During this time, ClusterAdmin#listTopics
+	// and ClusterAdmin#describeTopics may not return information about the new topics.
+	// This operation is supported by brokers with version 0.10.1.0 or higher.
+	// The validateOnly option is supported from version 0.10.2.0.
```
eapache: the method doesn't take a `validateOnly` flag

admin.go
```diff
@@ -0,0 +1,360 @@
...
+	// The resources with their configs (topic is the only resource type with configs
+	// that can be updated currently Updates are not transactional so they may succeed
+	// for some resources while fail for others. The configs for
+	// a particular resource are updated atomically.
+	AlterConfig(resourceType ConfigResourceType, name string, entries map[string]*string, validateOnly bool) error
+
+	// Creates access control lists (ACLs) which are bound to specific resources.
+	// This operation is not transactional so it may succeed for some ACLs while fail for others.
+	// If you attempt to add an ACL that duplicates an existing ACL, no error will be raised, but
+	// no changes will be made.This operation is supported by brokers with version 0.11.0.0 or higher.
```
eapache: missing space

admin.go
```diff
@@ -0,0 +1,360 @@
...
+	CreateACL(resource Resource, acl Acl) error
+
+	ListAcls(filter AclFilter) ([]ResourceAcls, error)
```
eapache: missing description

admin.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func (ca *clusterAdmin) CreateTopic(topic string, detail *TopicDetail) error {
+
+	if topic == "" {
+		return ErrInvalidTopic
+	}
+
+	if detail == nil {
+		return ErrInvalidInput
```
eapache: I don't think it's worth naming an error for this, it's better to give something less computer-friendly and more human-friendly (e.g. `errors.New("You must specify topic details")`)

admin.go
chandradeepak: ok. will remove that . I was thinking mentioning the version dependency would be needed especially admin related stuff is supported only in latest versions of kafka.

admin.go
chandradeepak: will address that.

chandradeepak: ok , will do that.

admin.go
chandradeepak: that is what we though in the initial design. So i went with that . Should i change it to take multiple topics ?

admin.go
chandradeepak: ok sure , will do that

admin.go
eapache: Probably? At the very least, the code and the documentation should be consistent.

admin.go
chandradeepak: removed it

admin.go
chandradeepak: done

admin.go
chandradeepak: gave option for  validateOnly flag

admin.go
chandradeepak: done

admin.go
chandradeepak: gave support for ListAcls and added the description for it.

admin.go
chandradeepak: removed the ErrInvalidInput from errors.go

eapache:
Did you intend to push https://github.com/Shopify/sarama/pull/1116 to this PR instead?

admin.go
```diff
@@ -0,0 +1,376 @@
+package sarama
+
+import "errors"
+
+type ClusterAdmin interface {
```
eapache: Sorry my previous comment wasn't clear. I still think we should have a description for `ClusterAdmin`, I just think it doesn't need to mention the bit about "This client was introduced in 0.11.0.0 and the API is still evolving."

admin.go
```diff
@@ -0,0 +1,376 @@
...
+	// Creates a new topic. This operation is supported by brokers with version 0.10.1.0 or higher.
```
eapache: You've still got a duplicate version sentence here.

admin.go
```diff
@@ -0,0 +1,376 @@
...
+	// It may take several seconds after CreateTopic returns success for all the brokers
+	// to become aware that the topic has been created. During this time, ClusterAdmin#listTopics
+	// and ClusterAdmin#describeTopics may not return information about the new topic.
+	// This operation is supported by brokers with version 0.10.1.0 or higher.
+	// The validateOnly option is supported from version 0.10.2.0.
+	CreateTopic(topic string, detail *TopicDetail, validateOnly bool) error
+
+	// Delete a batch of topics.
```
eapache: This sentence needs to be singular still since the method only deletes one at a time.

admin.go
```diff
@@ -0,0 +1,376 @@
...
+
+	// Increase the number of partitions of the topics  according to the corresponding values.
+	// If partitions are increased for a topic that has a key, the partition logic or ordering of
+	// the messages will be affected. It may take several seconds after this method returns
+	// success for all the brokers to become aware that the partitions have been created.
+	// During this time, ClusterAdmin#describeTopics may not return information about the
+	// new partitions. This operation is supported by brokers with version 1.0.0 or higher.
+	CreatePartitions(topic string, count int32, assignment [][]int32, validateOnly bool) error
+
+	//Delete records whose offset is smaller than the given offset of the corresponding partition.
```
eapache: spacing after the `//`

errors.go
```diff
@@ -49,6 +49,8 @@ var ErrControllerNotAvailable = errors.New("kafka: controller is not avaiable")
 // the metadata.
 var ErrNoTopicsToUpdateMetadata = errors.New("kafka: no specific topics to update metadata")
 
+var ErrReferenceNodeNotFound = errors.New("kafka: response did not contain a reference to node")
```
eapache: Is this fundamentally any different from `ErrIncompleteResponse`?

chandradeepak: i was looking at the java documentation in kafka when ever topics creation is not successful or the response we dont get a response to the topic which we requested changes for it was returning some thing like that . I can change it to use ErrIncompleteResponse but it doesn't say the intent that the specified resource is not found.  I will wait for your response and take that action.

eapache: I'm not sure I fully understand, if something doesn't work the broker should still reply with that topic/partition, just with an error code set? If it's omitting the topic/partition entirely that's new behaviour.

At least as written, NodeNotFound doesn't contain any more information beyond IncompleteResponse about the "why", so I'd prefer to stick with IncompleteResponse until we have something more meaningful to put here.

chandradeepak: ok sure will switch to do that .

eapache:
This looks pretty good to me now. Could you please squash/rebase your commits?

chandradeepak:
Done squashing all the commits .

eapache:
Thanks!

PR #1059, opened 2018-02-23
'Config.MetricRegistry=nil' to mute sarama metrics

demaagd:
Histograms and Meters are very useful metrics to collect, but the current implemetation has two drawbacks:
1) They are also heavy in comparison to their Gauge and Counter peers, and the volume of data and the number of topics and brokers being addressed may warrant disabling them for throughput reasons.  
2) Entirely disabling rcrowlee/go-metrics is an awfully big hammer to use when just setting it to nil feels a lot more idiomatic.  

High level view of change is that all instances of metrics.GetOrRegister$FOO were guarded by a test for a nil registry, and if the registry was nil, return the corresponding metrics.Nil$FOO to preserve proper interface mechanics.  

Seem OK: 
- 'go test' passed
- 'go vet' doesn't complain about anything besides printf format strings
- Have consumed a couple hundred thousand non-synthetic messages through it so far

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -109,6 +109,11 @@ func (b *Broker) Open(conf *Config) error {
 			b.requestRate = metrics.GetOrRegisterMeter("request-rate", conf.MetricRegistry)
 			b.outgoingByteRate = metrics.GetOrRegisterMeter("outgoing-byte-rate", conf.MetricRegistry)
 			b.responseRate = metrics.GetOrRegisterMeter("response-rate", conf.MetricRegistry)
+		} else {
+			b.incomingByteRate = metrics.NilMeter{}
+			b.requestRate = metrics.NilMeter{}
+			b.responseRate = metrics.NilMeter{}
+			b.outgoingByteRate = metrics.NilMeter{}
 		}
 		b.requestSize = getOrRegisterHistogram("request-size", conf.MetricRegistry)
 		b.requestLatency = getOrRegisterHistogram("request-latency-in-ms", conf.MetricRegistry)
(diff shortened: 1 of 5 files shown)


eapache:
I thought the metrics lib provided some sort of `NilRegistry` mock you could use for this, but I can't find it now. That seems almost more idiomatic to me, especially since it's basically what `UseNilMetrics` already does.

demaagd:
That was what I thought a NilRegistry as well, but the only traces I found were the individual nil types. 

The problem with UseNilMetrics is that it is global - everything registered in all registries will be the nil types as long as it is set.  If it were the case that I could 'UseNilMetrics=true, do all the sarama setup, UseNilMetrics=false' and know that things would get handled properly then that would be an option. But if either the sarama or not-sarama code registers metrics at runtime, that won't cover things without a lock around UseNilMetrics. 

This change is library-scoped version of UseNilMetrics.

demaagd:
oops

demaagd:
Found a better way to do it, will deal with the metrics load in a more general way later...

wblakecaldwell:
@demaagd - hey - any update? I'm not using the metrics, but it's taking up a considerable amount of CPU, so I'd like to disable it. Unfortunately, I'm also using go-metrics :)

Thanks!

wblakecaldwell:
@demaagd - is this still planned? Thanks :)

demaagd:
Hack up a nil registry and use that. I have not pursued this line of
questioning further than that as my project is moving away from Kafka as a
transport mechanism.

On Nov 28, 2018 08:49, "Blake Caldwell" <notifications@github.com> wrote:

@demaagd <https://github.com/demaagd> - is this still planned? Thanks :)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/Shopify/sarama/pull/1059#issuecomment-442519105>, or mute
the thread

PR #1063, opened 2018-03-07
add Controller() method to Client interface

imjustfly:
Sometimes we need to know the controller broker of a kafka cluster to do something, such as creating or deleting topics.

### DIFF
--- a/client.go
+++ b/client.go
@@ -17,6 +17,9 @@ type Client interface {
 	// altered after it has been created.
 	Config() *Config
 
+	// Controller returns the cluster controller broker.
+	Controller() (*Broker, error)
+
 	// Brokers returns the current set of active brokers as retrieved from cluster metadata.
 	Brokers() []*Broker
 
@@ -97,6 +100,7 @@ type client struct {
 	seedBrokers []*Broker
 	deadSeeds   []*Broker
 
+	controllerID int32                                   // cluster controller broker id
 	brokers      map[int32]*Broker                       // maps broker ids to brokers
 	metadata     map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
 	coordinators map[string]int32                        // Maps consumer group names to coordinating broker IDs
@@ -379,6 +383,27 @@ func (client *client) GetOffset(topic string, partitionID int32, time int64) (in
 	return offset, err
 }
 
+func (client *client) Controller() (*Broker, error) {
+	if client.Closed() {
+		return nil, ErrClosedClient
+	}
+
+	controller := client.cachedController()
+	if controller == nil {
+		if err := client.refreshMetadata(); err != nil {
+			return nil, err
+		}
+		controller = client.cachedController()
+	}
+
+	if controller == nil {
+		return nil, ErrControllerNotAvailable
+	}
+
+	_ = controller.Open(client.conf)
+	return controller, nil
+}
+
 func (client *client) Coordinator(consumerGroup string) (*Broker, error) {
 	if client.Closed() {
 		return nil, ErrClosedClient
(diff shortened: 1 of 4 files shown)


imjustfly:
um. The unit test issue is not introduced by this  PR.

client.go
```diff
@@ -645,7 +662,12 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int)
 		} else {
 			Logger.Printf("client/metadata fetching metadata for all topics from broker %s\n", broker.addr)
 		}
-		response, err := broker.GetMetadata(&MetadataRequest{Topics: topics})
+
+		req := &MetadataRequest{Topics: topics}
+		if client.conf.Version.IsAtLeast(V0_10_0_0) {
+			req.Version = 1
```
eapache: I believe it is this line that is causing the tests to fail. The commit you are based on runs successfully, and this branch fails consistently even when I rerun.

client.go
```diff
@@ -17,6 +17,9 @@ type Client interface {
 	// altered after it has been created.
 	Config() *Config
 
+	// Controller returns the cluster controller broker
```
eapache: Full sentences ending in a period please.

client.go
```diff
@@ -379,6 +383,19 @@ func (client *client) GetOffset(topic string, partitionID int32, time int64) (in
...
 }
 
+func (client *client) Controller() (*Broker, error) {
+	if client.Closed() {
+		return nil, ErrClosedClient
+	}
+
+	controller := client.cachedController()
+	if controller == nil {
+		return nil, ErrControllerNotAvailable
```
eapache: Rather than returning an error immediately we should try refreshing the metadata once to see if we can fetch the controller - being nil may just mean `RefreshMetadata` has never been run.

client.go
```diff
@@ -97,6 +100,7 @@ type client struct {
 	seedBrokers []*Broker
 	deadSeeds   []*Broker
 
+	controller   *Broker                                 // cluster controller broker
```
eapache: As with the coordinator map, I'd prefer to store the int32 id here instead of the broker itself. It's easier to reason about object ownership and garbage-collection that way.

client.go
imjustfly: Ok, I will work on it. But it's a little complicated，I need separating a new method `refreshMetadata` from `backgroundMetadataUpdater` to refresh the meta data because of the configuration `Metadata.Full`.

client.go
imjustfly: Fixed the test fails about consumer by adding a `SetVersion` method to `MockMetadataResponse`

imjustfly:
OK. I fixed all issues and added an unit test.

chandradeepak:
@eapache  any update on when this pull request will be merged ,  that would unblock the ClusterAdmin pull request.

errors.go
```diff
@@ -41,6 +41,14 @@ var ErrMessageTooLarge = errors.New("kafka: message is larger than Consumer.Fetc
 // a RecordBatch.
 var ErrConsumerOffsetNotAdvanced = errors.New("kafka: consumer offset was not advanced after a RecordBatch")
 
+// ErrControllerNotAvailable is returned when server didn't give correect controller id. May be kafka server's version
```
eapache: typo: "correct"

errors.go
```diff
@@ -41,6 +41,14 @@ var ErrMessageTooLarge = errors.New("kafka: message is larger than Consumer.Fetc
...
+// is lower than 0.10.0.0.
+var ErrControllerNotAvailable = errors.New("kafka: controller is not avaiable")
+
+// ErrNoTopicsToUpdateMetaData is returned when Meta.Full is set to false but no specific topics were found to update
```
eapache: Please spell out `Metadata` (one word) here and in the error message itself

client.go
```diff
@@ -607,28 +632,35 @@ func (client *client) backgroundMetadataUpdater() {
...
-					break
-				} else {
-					topics = specificTopics
-				}
-			}
-
-			if err := client.RefreshMetadata(topics...); err != nil {
-				Logger.Println("Client background metadata update:", err)
+			if err := client.refreshMetadata(); err != nil {
+				Logger.Printf("Client background metadata update:%s", err)
```
eapache: This will not format properly (there's no space between the `:` and the error, and there's no newline at the end). You can fix the format string, but it's probably simpler to just go back to `Println` instead of `Printf` which does the right thing.

imjustfly:
Thanks for your review!  

 I have fixed the spelling and logging issue. Sorry for that :(

eapache:
Thanks!

PR #1064, opened 2018-03-07
Drop kafka 0.10, add go 1.10

eapache:


eapache:
Looks like gzip encoding has changed again in Go 1.10. Sigh.

ctrochalakis:
> Looks like gzip encoding has changed again in Go 1.10. Sigh.

Since support for go 1.7 is dropped (and tests in Go 1.7 are already failing, due to the missing `sort.Slice`), there is probably no need to maintain the pre1.8 gzip message.

PR #1065, opened 2018-03-08
Add CreatePartitions to Broker

Mongey:
So that you can add more partitions 😄


eapache:
Thanks, not sure how I missed this method.

PR #1066, opened 2018-03-09
Drop testing empty gzip encoding with the pre-1.8 encoder

ctrochalakis:
Support for Go 1.7 is dropped and tests are already failing on 1.7 (due
to the missing `sort.Slice`), so it might not make sense to keep the
fragile `runtime.Version` test.

Currently tests are failing on Go 1.10 (see #1064)

### DIFF
--- a/message_test.go
+++ b/message_test.go
@@ -1,8 +1,6 @@
 package sarama
 
 import (
-	"runtime"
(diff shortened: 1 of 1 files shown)


eapache:
Thanks!

PR #1073, opened 2018-04-01
Add support for Kafka 1.1.0

bai:
Add support for Kafka 1.1.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
   - KAFKA_HOSTNAME=localhost
   - DEBUG=true
   matrix:
+  - KAFKA_VERSION=0.11.0.2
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -12,8 +12,8 @@ env:
   - KAFKA_HOSTNAME=localhost
   - DEBUG=true
   matrix:
-  - KAFKA_VERSION=0.11.0.2
```
eapache: Our stated support commitment means we shouldn't remove 0.11 from CI until 2 months after 1.1 comes out. If you're looking to trim CI load we'll be able to drop Go 1.8 in a few weeks.

bai: Thanks, updated.

eapache:
Thanks!

PR #1077, opened 2018-04-10
Add context to sync producer

MatthewDolan:


MatthewDolan:
I need to think about this some more. I can't quite get the code to work and pass the race detector.

PR #1086, opened 2018-04-24
Unblock consumer when receiving invalid FetchResponse (response did not contain all the expected topic/partition blocks)

adwinsky:
Increment offsets to avoid consumer blocking. The issue is described here https://github.com/Shopify/sarama/issues/1087

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -528,7 +528,7 @@ func (child *partitionConsumer) parseRecords(batch *RecordBatch) ([]*ConsumerMes
 		child.offset = offset + 1
 	}
 	if len(messages) == 0 {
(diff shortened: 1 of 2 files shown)


eapache:
Sorry it took me a while to get back to this. Thanks for the fix and the test!

gburanov:
@adwinsky @eapache I also experience this issue now with sarama. 

I cherry picked your commit, and it did not help. But than I made the fix in the place very similar to your and it started to work https://github.com/Shopify/sarama/pull/1227

gburanov:
WDYT? Does it seem reasonable to fix in this place as well?

PR #1092, opened 2018-04-29
Make PartitionConsumer.Close idempotent

dim:
There is a use-case in sarama-cluster where I want to `defer pcm.Close()` as well as call it explicitly to ensure it has actually stopped before proceeding with the next steps. I am currently wrapping the call in a `sync.Once` myself, but moving this to sarama seems generally a good idea.


eapache:
Fair enough, thanks.

PR #1096, opened 2018-05-02
Add support for DeleteGroups

### LINKED ISSUE #1095: Support the DeleteGroups API
Currently `sarama` does not support the [DeleteGroups API](https://kafka.apache.org/protocol#The_Messages_DeleteGroups), which is very useful for implementing ephemeral consumer groups. (See [the discussion](https://github.com/confluentinc/kafka-rest/issues/148) of the use case.)

RussellLuo:
Fixed #1095.

### DIFF
--- a/delete_groups_response.go
+++ b/delete_groups_response.go
@@ -1,0 +1,67 @@
+package sarama
+
+type DeleteGroupsResponse struct {
+	ThrottleTimeMs  int32
+	GroupErrorCodes map[string]KError
+}
+
+func (r *DeleteGroupsResponse) encode(pe packetEncoder) error {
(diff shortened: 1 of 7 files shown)


delete_groups_response.go
```diff
@@ -0,0 +1,67 @@
+package sarama
+
+type DeleteGroupsResponse struct {
+	ThrottleTimeMs  int32
```
eapache: For time fields we prefer to expose them as native go types, see e.g. https://github.com/Shopify/sarama/blob/0f333e6f2449e00440f7a53d3d2f006d702abe23/end_txn_response.go#L8

eapache:
Thanks

PR #1100, opened 2018-05-05
Optional replica id in offset request

urso:
This change adds `SetReplicaID` to the OffsetRequest structure. If `SetReplicaID` has not been used on an instance of `OffsetRequest`, `-1` will be used when serializing the request.


eapache:
Great idea, thanks. Sorry for taking so long to get back to this, but it would be great to have a test for this.

jsoriano:
@eapache we have finally added tests and signed the CLA :slightly_smiling_face: Could you take another look? Thanks!

bai:
Thanks for contributing!

PR #1110, opened 2018-06-02
Do not modify ProducerMessage.MetaData

the729:
In SyncProducer, do not modify ProducerMessage.MetaData during the whole process of publishing the message to Kafka. 
MetaData could be used in a user-defined Partitioner. The original design saves the MetaData and reuse the field as a signal channel to wait for completion of publish. It's better to add a non-exeported field for this purpose in the ProducerMessage struct.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -145,8 +145,9 @@ type ProducerMessage struct {
 	// least version 0.10.0.
 	Timestamp time.Time
 
-	retries int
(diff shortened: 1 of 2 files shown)


eapache:
Nice, thanks! I never considered that the metadata field might actually be useful during the message's lifecycle, instead of just after being returned on errors/successes.

PR #1114, opened 2018-06-12
Adjust paritioning / hash behavior to comply with java lib

mier85:
So we had a problem here with java and go services working together. Even though they are using the same hash function, the behavior afterwards is slightly different.

While java is doing this:
`Abs(Hash) % numPartitions`

Sarama was doing this: 
`Abs(Hash % numPartitions`

I updated that according to how the java lib is doing it : 

https://github.com/apache/kafka/blob/fa1d0383902260576132e09bdf9efcc2784b55b4/clients/src/main/java/org/apache/kafka/clients/producer/internals/DefaultPartitioner.java#L69

https://github.com/apache/kafka/blob/fa1d0383902260576132e09bdf9efcc2784b55b4/clients/src/main/java/org/apache/kafka/common/utils/Utils.java#L821

### DIFF
--- a/partitioner.go
+++ b/partitioner.go
@@ -123,10 +123,7 @@ func (p *hashPartitioner) Partition(message *ProducerMessage, numPartitions int3
 	if err != nil {
 		return -1, err
 	}
-	partition := int32(p.hasher.Sum32()) % numPartitions
-	if partition < 0 {
-		partition = -partition
-	}
+	partition :=  (int32(p.hasher.Sum32()) & 0x7fffffff) % numPartitions
 	return partition, nil
 }
 
--- a/partitioner.go
+++ b/partitioner.go
@@ -123,7 +123,7 @@ func (p *hashPartitioner) Partition(message *ProducerMessage, numPartitions int3
 	if err != nil {
 		return -1, err
 	}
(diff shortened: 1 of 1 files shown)


mier85:
Yes, because of the bitwise operation, the problem described in  https://github.com/Shopify/sarama/pull/709 will not happen again and the tests for it are green. The problem with that was with the max integer values. 

Even though this modification halves the basic set, it shouldn't affect the even total distribution amongst the partitions. 

On the other hand, since the implementation differs in the official client and presumably other languages, the partitioning doesn't match which leads to other problems which has a more grave effect on how the whole partitioning works and leads to inconsistency.

eapache:
Makes sense, I agree this is a worthwhile change. I'm wondering if it should be shipped as a new partitioner though, in case there are users depending on the old implementation now without realizing it (e.g. because they never used the official client at all).

mier85:
Making it an own partitioner would make sense for making a non-breaking migration possible even though, naturally, for us it means that we have to use this new hash function in every service.

Only other possible solution would be adjusting the version major numbers.

eapache:
Thanks for reporting this, https://github.com/Shopify/sarama/commit/6967cdb516216f65c91b9ae11431901c8b12ecf5 adds a `NewReferenceHashPartitioner` based on this PR.

PR #1116, opened 2018-06-15
Increasecoverage2

chandradeepak:


chandradeepak:
sorry my bad. it was supposed to fix the issue with coverage not being shown more than 80%.  It was supposed to be opened against my master not Shopify master

PR #1118, opened 2018-06-15
add customizable partitioner

mier85:
So the partitioner has a lot of different partitioner constructors now, and they contain functionality which cannot all be created without creating new custom constructor functions for each.

This PR introduces a customizable partitioner which works with functions as options. New added options will not break the current behavior nor will it break the function signature. That will let us introduce new options while we are still backwards compatible.

For reference: 
https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

### DIFF
--- a/partitioner.go
+++ b/partitioner.go
@@ -60,8 +60,8 @@ func WithCustomHashFunction(hasher func() hash.Hash32) HashPartitionerOption {
 	}
 }
 
-// WithCustomRandomPartitioner lets you specify what HashPartitioner should be used in case a Distribution Key is empty
-func WithCustomRandomPartitioner(randomHP *hashPartitioner) HashPartitionerOption {
+// WithCustomFallbackPartitioner lets you specify what HashPartitioner should be used in case a Distribution Key is empty
+func WithCustomFallbackPartitioner(randomHP *hashPartitioner) HashPartitionerOption {
 	return func(hp *hashPartitioner) {
 		hp.random = hp
 	}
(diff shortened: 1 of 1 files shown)


eapache:
Ya, I was noticing this issue as well. In a from-scratch world (or if I ever do a major version bump) I think I might simply provide a single `NewHashPartitioner` which is properly compatible with the upstream/reference implementation, and if you want anything else you can implement your own partitioner. They're not exactly large or complicated.

In the meantime, I suppose this is probably the best we can do.

partitioner.go
```diff
@@ -42,6 +42,31 @@ type PartitionerConstructor func(topic string) Partitioner
...
+}
+
+// WithCustomHashFunction lets you specify what hash function to use for the partitioning
+func WithCustomHashFunction(hasher func() hash.Hash32) HashPartitionerOption {
+	return func(hp *hashPartitioner) {
+		hp.hasher = hasher()
+	}
+}
+
+// WithCustomRandomPartitioner lets you specify what HashPartitioner should be used in case a Distribution Key is empty
```
eapache: s/HashPartitioner/Partitioner

partitioner.go
```diff
@@ -42,6 +42,31 @@ type PartitionerConstructor func(topic string) Partitioner
...
+func WithCustomRandomPartitioner(randomHP *hashPartitioner) HashPartitionerOption {
```
eapache: I might call this `WithCustomFallbackPartitioner`

mier85: makes much more sense, I'll update it.

PR #1123, opened 2018-07-03
Add LocalAddr to Config struct and honor in broker.Open().

PSUdaemon:
We have hosts that are multi-homed and being able to pick a source address is important to contend with things like network ACLs.


eapache:
Nice, thanks!

PR #1125, opened 2018-07-06
fix producer topic metadata on-demand fetch when topic error happens

andyxning:
This PR can help fix producer topic metadata on-demand fetch when topic error happens in metadata response.

There is an option [`Full`](https://github.com/Shopify/sarama/blame/f7df95cff1bc4d14a4194a5be4c88a86253afb98/config.go#L86-L90) in [`Config`](https://github.com/Shopify/sarama/blob/master/config.go#L285-L324) to config whether producer should fetch all topic metadata in a cluster or do it on-demand.
This feature is added in #937. 

But there is an situation where we may make on-demand topic metadata fetching fall back to empty once an topic error happens in metadata update.

For example, if `ErrUnknownTopicOrPartition` or `ErrInvalidTopic` or `ErrTopicAuthorizationFailed` or any other errors that will make the `switch` case `continue` happens, the topic name will be deleted from `client.metadata`. This will result in an empty topic slice when refresh topic metadata next time in https://github.com/Shopify/sarama/blob/master/client.go#L648-L666 from `client.Topics()` which actually read topics from `client.metadata`. Thus we lost the topic name which we need to periodically refresh metadata.

We should not update `client.metadata` or delete any topic key from it.

### DIFF
--- a/client.go
+++ b/client.go
@@ -100,10 +100,11 @@ type client struct {
 	seedBrokers []*Broker
 	deadSeeds   []*Broker
 
-	controllerID int32                                   // cluster controller broker id
-	brokers      map[int32]*Broker                       // maps broker ids to brokers
-	metadata     map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
-	coordinators map[string]int32                        // Maps consumer group names to coordinating broker IDs
+	controllerID   int32                                   // cluster controller broker id
+	brokers        map[int32]*Broker                       // maps broker ids to brokers
+	metadata       map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
+	metadataTopics map[string]none                         // topics that need to collect metadata
+	coordinators   map[string]int32                        // Maps consumer group names to coordinating broker IDs
 
 	// If the number of partitions is large, we can get some churn calling cachedPartitions,
 	// so the result is cached.  It is important to update this value whenever metadata is changed
@@ -136,6 +137,7 @@ func NewClient(addrs []string, conf *Config) (Client, error) {
 		closed:                  make(chan none),
 		brokers:                 make(map[int32]*Broker),
 		metadata:                make(map[string]map[int32]*PartitionMetadata),
+		metadataTopics:          make(map[string]none),
 		cachedPartitionsResults: make(map[string][maxPartitionIndex][]int32),
 		coordinators:            make(map[string]int32),
 	}
@@ -207,6 +209,7 @@ func (client *client) Close() error {
 
 	client.brokers = nil
 	client.metadata = nil
+	client.metadataTopics = nil
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


eapache:
If `ErrLeaderNotAvailable` happens, the topic will be deleted from the map but will immediately be re-added at line 756. I don't understand how this is a bug.

andyxning:
>If ErrLeaderNotAvailable happens, the topic will be deleted from the map but will immediately be re-added at line 756. I don't understand how this is a bug.

@eapache Sorry for not making the description clear and correct. The error maybe `ErrUnknownTopicOrPartition` or `ErrInvalidTopic` or `ErrTopicAuthorizationFailed` or any other errors that will make the `switch` case `continue`.

eapache:
I'm not sure I agree. If the broker returns one of those errors the topic either doesn't exist or is fundamentally inaccessible. Leaving it in the metadata hash at that point would be extremely misleading to callers and would likely lead to other problems when trying to iterate over the set of topics.

> But there is an situation where we may make on-demand topic metadata fetching fall back to empty once an topic error happens in metadata update.

I'm also not sure how this can happen. The `refreshMetadata` method already checks if the list of `Topics` is empty when trying to do a partial refresh: https://github.com/Shopify/sarama/blob/f7df95cff1bc4d14a4194a5be4c88a86253afb98/client.go#L654-L655

andyxning:
@eapache 
> I'm also not sure how this can happen. The refreshMetadata method already checks if the list of Topics is empty when trying to do a partial refresh

Yes. That's right. But `refreshMetadata` will return `ErrNoTopicsToUpdateMetadata` forever since the topic error happens. This is a problem because the topic error may disappear, but the topic name has been removed from metadata and it will not recover automatically alongside the topic error recover. A restart of producer must be done. This hits me.

>  If the broker returns one of those errors the topic either doesn't exist or is fundamentally inaccessible.

The error maybe not all related with producer side configuration. Some errors are related with server side operations or status.

eapache:
OK, I understand. As mentioned, leaving it in the metadata would cause other issues. A better solution would be to keep a separate copy of the topics when not doing full metadata, and use that (in combination with `client.Topics()` still) in `refreshMetadata()`.

andyxning:
> As mentioned, leaving it in the metadata would cause other issues. A better solution would be to keep a separate copy of the topics when not doing full metadata, and use that (in combination with client.Topics() still) in refreshMetadata().

I have thought that before i propose this PR. I think directly update the `client.Topics` is fine. According to the discuss above, it seems that a new field should be added to store topic names whose metadata are need to be updated periodically. @eapache 
Will update this later today.

andyxning:
@eapache PTAL.

client.go
```diff
@@ -100,10 +100,11 @@ type client struct {
...
 	deadSeeds   []*Broker
 
-	controllerID int32                                   // cluster controller broker id
-	brokers      map[int32]*Broker                       // maps broker ids to brokers
-	metadata     map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
-	coordinators map[string]int32                        // Maps consumer group names to coordinating broker IDs
+	controllerID   int32                                   // cluster controller broker id
+	brokers        map[int32]*Broker                       // maps broker ids to brokers
+	metadata       map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
+	metadataTopics map[string]struct{}                     // topics that need to collect metadata
```
eapache: sarama defines a helper type called `none` to make uses of `struct{}` like this more explicit

andyxning: @eapache Done. PTAL.

eapache:
thanks!

PR #1127, opened 2018-07-10
Simpler offset management, fixed minor race

dim:
This is a necessary prerequisite for #1099, please also see the inline comments in the diff.

### DIFF
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -25,10 +25,17 @@ type offsetManager struct {
 	client Client
 	conf   *Config
 	group  string
+	ticker *time.Ticker
 
-	lock sync.Mutex
-	poms map[string]map[int32]*partitionOffsetManager
-	boms map[*Broker]*brokerOffsetManager
+	broker   *Broker
+	brokerMu sync.RWMutex
+
+	poms   map[string]map[int32]*partitionOffsetManager
+	pomsMu sync.Mutex
+
+	closeOnce sync.Once
+	closing   chan none
+	closed    chan none
 }
 
 // NewOffsetManagerFromClient creates a new OffsetManager from the given client.
@@ -39,13 +46,18 @@ func NewOffsetManagerFromClient(group string, client Client) (OffsetManager, err
 		return nil, ErrClosedClient
 	}
 
+	conf := client.Config()
 	om := &offsetManager{
 		client: client,
-		conf:   client.Config(),
+		conf:   conf,
 		group:  group,
+		ticker: time.NewTicker(conf.Consumer.Offsets.CommitInterval),
 		poms:   make(map[string]map[int32]*partitionOffsetManager),
-		boms:   make(map[*Broker]*brokerOffsetManager),
+
+		closing: make(chan none),
+		closed:  make(chan none),
 	}
+	go withRecover(om.mainLoop)
 
 	return om, nil
 }
@@ -56,8 +68,8 @@ func (om *offsetManager) ManagePartition(topic string, partition int32) (Partiti
 		return nil, err
 	}
 
-	om.lock.Lock()
-	defer om.lock.Unlock()
+	om.pomsMu.Lock()
+	defer om.pomsMu.Unlock()
 
 	topicManagers := om.poms[topic]
 	if topicManagers == nil {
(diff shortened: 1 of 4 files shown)


offset_manager.go
```diff
@@ -74,52 +86,298 @@ func (om *offsetManager) ManagePartition(topic string, partition int32) (Partiti
...
+	}
+
+	return nil
+}
+
+func (om *offsetManager) handleError(err error) {
+	om.pomsMu.Lock()
+	defer om.pomsMu.Unlock()
+
+	for _, topicManagers := range om.poms {
```
dim: Should global commit errors be propagated to each individual POM or just to the first one?

offset_manager_test.go
```diff
@@ -353,24 +352,14 @@ func TestPartitionOffsetManagerCommitErr(t *testing.T) {
 	ocResponse2 := new(OffsetCommitResponse)
 	newCoordinator.Returns(ocResponse2)
 
-	// For RefreshCoordinator()
-	broker.Returns(&ConsumerMetadataResponse{
-		CoordinatorID:   newCoordinator.BrokerID(),
-		CoordinatorHost: "127.0.0.1",
-		CoordinatorPort: newCoordinator.Port(),
-	})
+	// No error, no need to refresh coordinator
```
dim: I changed the logic to only `RefreshCoordinator` after errors and not on every commit attempt.

offset_manager.go
```diff
@@ -369,16 +514,9 @@ func (pom *partitionOffsetManager) NextOffset() (int64, string) {
 }
 
 func (pom *partitionOffsetManager) AsyncClose() {
-	go func() {
-		pom.lock.Lock()
-		defer pom.lock.Unlock()
-
-		for pom.dirty {
```
dim: So I think this has actually never worked properly, as you are creating a deadlock here. The state is blocked as long as `pom.dirty`, but the flag will never actually change to `false` nor will `pom.clean` ever fire, because `pom.updateCommitted` will never be able to acquire `pom.lock.Lock()`

offset_manager.go
```diff
@@ -25,10 +25,17 @@ type offsetManager struct {
...
+	ticker *time.Ticker
 
-	lock sync.Mutex
-	poms map[string]map[int32]*partitionOffsetManager
-	boms map[*Broker]*brokerOffsetManager
+	broker   *Broker
+	brokerMu sync.RWMutex
+
+	poms   map[string]map[int32]*partitionOffsetManager
+	pomsMu sync.Mutex
```
eapache: f

offset_manager.go
```diff
@@ -74,52 +86,292 @@ func (om *offsetManager) ManagePartition(topic string, partition int32) (Partiti
...
+		return false
+	}
+
+	success = true
+
+	om.pomsMu.Lock()
+	defer om.pomsMu.Unlock()
 
-	delete(om.boms, bom.broker)
+	for _, topicManagers := range om.poms {
```
eapache: Just for organization I'd prefer `handleResponse` or something as a separate method to mirror `constructRequest`

offset_manager.go
```diff
@@ -74,52 +86,292 @@ func (om *offsetManager) ManagePartition(topic string, partition int32) (Partiti
...
+	om.closeOnce.Do(func() {
+		// exit the mainLoop
+		close(om.closing)
+		<-om.closed
+
+		// mark all POMs as closed
+		om.asyncClosePOMs()
+
+		// flush one last time
+		for retries := om.conf.Metadata.Retry.Max; true; {
```
eapache: I think it makes more sense to define a separate set of Retry configurations in `Consumer.Offsets`?

offset_manager.go
```diff
@@ -74,52 +86,292 @@ func (om *offsetManager) ManagePartition(topic string, partition int32) (Partiti
...
+
+func (om *offsetManager) mainLoop() {
+	defer om.ticker.Stop()
+	defer close(om.closed)
+
+	for {
+		select {
+		case <-om.ticker.C:
+			om.flushToBroker()
+			om.releasePOMs(false)
```
eapache: It took me a while to figure out what this was doing, and I'm not much a fan of this pattern. An `updateSubscriptions` channel (as in the old BOM) seems easier to follow and less likely to result in races/deadlocks than polling a piece of state in all the POMs?

offset_manager.go
```diff
@@ -74,52 +86,292 @@ func (om *offsetManager) ManagePartition(topic string, partition int32) (Partiti
...
+			}
+
+			switch err {
+			case ErrNoError:
+				block := request.blocks[pom.topic][pom.partition]
+				pom.updateCommitted(block.offset, block.metadata)
+			case ErrNotLeaderForPartition, ErrLeaderNotAvailable,
+				ErrConsumerCoordinatorNotAvailable, ErrNotCoordinatorForConsumer:
+				// not a critical error, we just need to redispatch
+				om.releaseCoordinator(broker)
```
eapache: In cases like this I wonder if we should be doing something to trigger another dispatch before the next timer tick? There's no point in waiting really.

offset_manager.go
eapache: 🤔 all of them still, I think

offset_manager.go
dim: I thought about `updateSubscriptions` but it's not making things any more trivial in my opinion. The problem is that the eviction of POMs is closely linked to the flush cycles and we still need to coordinate those on `OffsetManager.Close`, i.e. keep retrying flush until either `Retry.Max` is reached or until all POMs are clean. I don't see how to implement this with an `updateSubscriptions` channel, we would have to do something like:

```
func (om *offsetManager) Close() error {
  om.closeOnce.Do(func() {
    // exit the mainLoop
    close(om.closing)
    <-om.closed

    // mark all POMs as closed
    om.asyncClosePOMs()

    deadline := time.NewTimer(10*time.Second) // just an example
    defer deadline.Stop()

    retry := time.NewTicker(100*time.Millisecond) // just an example
    defer retry.Stop()

  FlushLoop:
    for {
      om.flushToBroker()

      select {
      case <-deadline.C:
        break FlushLoop
      case pom := <-om.updateSubscription:
        om.pomsLock.Lock()
        delete(om.poms[pom.topic], pom.partition)
        if len(om.poms[pom.topic]) == 0 {
          delete(om.poms, pom.topic)
        }
        remaining := len(om.poms)
        om.pomsLock.Unlock()
        if remaining == 0 {
          break FlushLoop
        }
      case <-retry.C:
      }
    }

    // TODO: abandon any remaining POMs

    om.brokerLock.Lock()
    om.broker = nil
    om.brokerLock.Unlock()
  })
  return nil
}
```

Even with helper method, I don't think this is easier to follow than the polling.

offset_manager.go
dim: I think it's not the end of the world if we miss a cycle and wait for `mainLoop` to trigger the next flush. At the same time, we currently already retry immediately if triggered by `Close`

offset_manager.go
eapache: For `OffsetManager.Close` I think you're overthinking things. Instead of exiting the mainloop and them manually looping flush, just call `pom.AsyncClose()` on all of them (as you already do), set a flag and then wait for a waitgroup. The mainloop can respect the flag to count down the number of remaining flushes.

offset_manager.go
eapache: Hmm, good point, that wasn't obvious. However, the metadata refresh itself will retry this many times internal to the client, so by using this value here you're technically enabling the square of it as retries? You're also not sleeping the way the `Metadata.Retry` struct might imply.

I still think this might make the most sense as `Consumer.Offsets.MaxShutdownAttempts` or something. It's not the same thing as refreshing metadata.

offset_manager.go
dim: Letting `mainLoop` doing the work after calling `Close()` could delay things significantly, depending on `CommitInterval`. I'm not a fan of this, I prefer exiting the `mainLoop` to attempt an immediate final flush (with instant retries). The current approach will also simplify the integration with `ClusterClient`.

eapache: OK. It's not clear to me how this will simplify the `ClusterClient` integration, but I don't feel too strongly about this.

eapache:
I'm pretty happy with this now, if it suits your needs. CI has detected a race condition though.

eapache:
Nice, thanks!

PR #1128, opened 2018-07-13
Add NewConsistentHashPartitioner

bataras:
Uses Google's Jump Consistent Hash function to provide consistent key -> partition hashing when a topic's partition count changes. This is for the case where topic consumers need to rely as much as possible on messages with specific keys arriving in order by staying on the same partition even when the topic's partition count changes.

### DIFF
--- a/partitioner.go
+++ b/partitioner.go
@@ -56,7 +56,7 @@ func WithAbsFirst() HashPartitionerOption {
 // WithCustomHashFunction lets you specify what hash function to use for the partitioning
 func WithCustomHashFunction(hasher func() hash.Hash32) HashPartitionerOption {
 	return func(hp *hashPartitioner) {
-		hp.hasher = hasher()
+		hp.hasher = hasher
 	}
 }
 
@@ -124,7 +124,7 @@ func (p *roundRobinPartitioner) RequiresConsistency() bool {
 
 type hashPartitioner struct {
 	random       Partitioner
-	hasher       hash.Hash32
+	hasher       func() hash.Hash32
 	referenceAbs bool
 }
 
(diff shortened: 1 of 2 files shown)


eapache:
Thanks for the contribution, but I'm a bit confused why you can't just pass the custom hash function into the existing hash partitioner?

bataras:
Hmm. Should've read the code better. You're right. Also it appears my concern about thread safety with hashPartitioner calling hasher.Reset, Write is also invalid as msgs are passed through a channel/goroutine. I'll close this.

bataras:
Sorry for the noise. It appears that simply using NewCustomHashPartitioner and passing in a custom hash.Hash32 won't work. The consistent hashing function needs to know the number of partitions to hash against. See func jumpHash(key uint64, numBuckets int) int32 in this PR.

I can redo this PR to remove the thread-safety change for calling hasher since it's not needed by the way SyncProducer calls the partitioner. Let me know what you think.

eapache:
- Yes, please undo the thread-safety changes. We don't need them and it is more efficient to reuse a single hasher.
- Good point about the need for the number of partitions. Based on your actual implementation here, the jump hash isn't even the underlying hash algorithm (that's FNV64), it's yet another way of doing the modulus/abs-value, so I think it could still be implemented as an additional option call on the existing hash partitioner?

bataras:
It could be done that way. We'd need to change/add this..

`
func NewCustomHashPartitioner(hasher func() hash.Hash32) PartitionerConstructor {
`
to something like this...

`
func NewCustomHashPartitioner2(hasher func() hash.Hash32, makeConst func(hash uint32, partitions int) uint32) PartitionerConstructor {
`

eapache:
No, I think you can just add a method conforming to `HashPartitionerOption` and then add it to a call to `NewCustomPartitioner`.

bataras:
Maybe I'm missing something. But I don't see how it's possible to implement consistent hashing without changing the code somewhere. To me it fundamentally looks like `hashPartitioner.Partition(message, partitions)` needs to be made to call -some- method that takes both the message key (or its bytes) and the number of partitions in order to determine a partition number.

eapache:
Oh yes, you'll have to change the internals of `hashPartitioner.Partition` a bit to make this work; I was focusing on what the external API changes would look like.

bataras:
ok. will try to put something together when I can get time. Closing this for now

PR #1129, opened 2018-07-13
Threadsafe hasher

bataras:


bataras:
will move this to the other PR I created

PR #1132, opened 2018-07-16
Fix metadata requests on later versions

### LINKED ISSUE #1130: No topics returned on successive refreshing of full metadata on versions >= 0_10_0_0
##### Versions

Sarama Version: commit on master: 5cd4d8675b3b11c08e71529e7280c45f9fc3dafe
Kafka Version: 0.10.1.0 & 1.1.0
Go Version: 1.10.3

##### Configuration

	config.Consumer.Return.Errors = true
	config.ClientID = "consumer-test"
	config.Metadata.RefreshFrequency = time.Duration(2) * time.Second
	config.Metadata.Full = true
	config.Version = sarama.V0_10_0_0

##### Logs

```
[Sarama] 2018/07/13 11:05:19.935931 client.go:120: Initializing new client
[Sarama] 2018/07/13 11:05:19.936136 client.go:701: client/metadata fetching metadata for all topics from broker localhost:9092
[Sarama] 2018/07/13 11:05:19.939979 broker.go:148: Connected to broker at localhost:9092 (unregistered)
[Sarama] 2018/07/13 11:05:19.948910 client.go:477: client/brokers registered new broker #0 at localhost:9092
[Sarama] 2018/07/13 11:05:19.948945 client.go:167: Successfully initialized new client
client.Topics() [test2 test1 test3]
[Sarama] 2018/07/13 11:05:21.949261 client.go:701: client/metadata fetching metadata for all topics from broker localhost:9092
client.Topics() [test2 test1 test3]
client.Topics() []
[Sarama] 2018/07/13 11:05:23.952187 client.go:701: client/metadata fetching metadata for all topics from broker localhost:9092
client.Topics() []
client.Topics() []
client.Topics() []
[Sarama] 2018/07/13 11:05:25.951132 client.go:701: client/metadata fetching metadata for all topics from broker localhost:9092
```

##### Problem Description
Consuming using Kafka version 0_10_0_0 and above I've noticed that refreshing metadata does not seem to be returning a correct list of topics.

I'm calling client.Topics() in a loop about twice every refresh period and the first few calls return a list of all my topics. After the next refresh the topic list becomes empty from then on, as shown in my provided logs

I do not see these problems when using V0_9_0_1 and below and client.Topics() always returns the full list.

Trying to debug this issue I saw this code: https://github.com/Shopify/sarama/blob/master/client.go#L705
I commented out this IF statement for testing and the data returned correctly.
I don't know enough about kafka's versions so I can't seem to pinpoint why this might be causing this problem.

I can provide more info if needed.

eapache:
If the set of Topics is an empty array, we don't need to put it in the
request or we'll just get an empty set of data back which is useless.
As far as I can tell the nil check was for safety, but len(nil) is
already safe and the boolean logic was wrong anyway.

Thanks to @Preylien for figuring this one out.

Fixes #1130.

### DIFF
--- a/metadata_request.go
+++ b/metadata_request.go
@@ -10,7 +10,7 @@ func (r *MetadataRequest) encode(pe packetEncoder) error {
 	if r.Version < 0 || r.Version > 5 {
 		return PacketEncodingError{"invalid or unsupported MetadataRequest version field"}
 	}
-	if r.Version == 0 || r.Topics != nil || len(r.Topics) > 0 {
+	if r.Version == 0 || len(r.Topics) > 0 {
 		err := pe.putArrayLength(len(r.Topics))
 		if err != nil {
 			return err


varun06:
👍

PR #1141, opened 2018-07-26
Respect MaxMessageBytes limit for uncompressed messages

adwinsky:
This is the fix for the issue reported here https://github.com/Shopify/sarama/issues/1140


varun06:
@adwinsky this change looks okay to me. What do we need to do to merge this PR, it has been sitting from some time?

adwinsky:
@varun06 looks ok to me as well and lately more people are  interested in this going live (
https://github.com/Shopify/sarama/issues/114). It only needs some attention of the reviewers with merge privileges. Maybe @eapache can help?

varun06:
also @bai, please have a look and then we can get it merged.

Issue #1120, opened 2018-06-25
All the requests in admin.go don't specify a timeout and none is allowed to be set causing them to ErrRequestTimedOut


gadkrumholz:
##### Versions
Sarama Version: master@46cf3e2cf1acef7876068f66cf69ec31aad2d0b2
Kafka Version: 1.10
Go Version: 1.10

##### Configuration
N/A

What configuration values are you using for Sarama and Kafka?
N/A

##### Logs
N/A

##### Problem Description
When a long-running operation like creating a topic with a lot of partitions and replicasets or when changing a topic's # of partitions (the 2 examples we've run into this with) you may sometimes get ErrRequestTimedOut even though the operation completes successfully.

Could you please expose the Timeout parameter for the requests to be passed in with each method call or have it use the same Timeout parameter from the config passed into NewClusterAdmin(...)?

eapache:
I don't think there's an existing config value which makes sense for those timeouts, but I'm happy to take a PR adding one.

andy2046:
@eapache do you think adding Cluster Timeout into existing Config is the right way to solve this? or you have better idea, I'd like to raise a PR for this 

```go
type Config struct {
	Cluster struct {
		Timeout Time.Duration
	}
}
```

eapache:
Something like that. I could see an argument for putting this value under `Metadata`, or for creating a new `Admin` struct in config. The name `Cluster` on its own isn't specific enough for me.

andy2046:
@eapache Thanks for the advise, I raised a PR [1142](https://github.com/Shopify/sarama/pull/1142) with new `Admin` struct. please help to review and comment, thanks.

andy2046:
@eapache should we close this issue?

d1egoaz:
fixed by https://github.com/Shopify/sarama/pull/1142
thanks

PR #1142, opened 2018-07-29
add Timeout config for ClusterAdmin related Requests

andy2046:
This is the fix for the issue reported here #1120


eapache:
Thanks!

PR #1149, opened 2018-08-10
Fix partial messages handling

### LINKED ISSUE #1144: Error when consuming: error decoding packet: CRC didn't match expected 0x0
##### Versions

Sarama Version: e7238b119b7daab993720f0153eafb88e2b0ac1f
Kafka Version: 2.0.0
Go Version: 1.10.1

##### Configuration
Kafka is Wurstmeisters docker image file, nothing besides env variables changed.

Sarama SyncProducer for producing and consumer no configuration. Just NewConsumer, looping over Partitions and trying to get msg from ConsumePartition

##### Logs
kafka: error while consuming user.created/0: kafka: error decoding packet: CRC didn't match expected 0x0 got 0xe38a6876
##### Problem Description
Cant consume. Getting that error

wladh:
Kafka 2.0 introduced chunked message down conversions which might
produce partial messages at the end of the set. These partial messages
are not well formed beyond offset and length, so they might cause
strange decoding errors down the line.
This fix makes `lengthField` a `dynamicPushDecoder` so it checks that
the `packetDecoder` has at least that many bytes. It fixes the problem
above and I think it's a more robust check in general.

Fixes #1144

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,9 @@ env:
   - KAFKA_HOSTNAME=localhost
   - DEBUG=true
   matrix:
-  - KAFKA_VERSION=0.11.0.2
   - KAFKA_VERSION=1.0.0
   - KAFKA_VERSION=1.1.0
+  - KAFKA_VERSION=2.0.0
 
 before_install:
 - export REPOSITORY_ROOT=${TRAVIS_BUILD_DIR}
--- a/dev.yml
+++ b/dev.yml
@@ -2,7 +2,7 @@ name: sarama
 
 up:
   - go:
-      version: '1.9'
+      version: '1.10'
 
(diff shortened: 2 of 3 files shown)


bai:
Thanks for digging into this! I've pushed a [test branch on CI](https://github.com/Shopify/sarama/compare/wladh-master-test) but still seeing failures https://travis-ci.org/Shopify/sarama/jobs/415607362?

bai:
[Test branch](https://github.com/Shopify/sarama/compare/wladh-master-test) is 🍏 , could please chery-pick https://github.com/Shopify/sarama/commit/3de182b43ce4e6ac1599844866957655ba7d32a2 to enable Kafka 2.0 on CI?

PR #1152, opened 2018-08-20
Added support for Idempotent Producer

mimaison:
This PR adds support for the Idempotent Producer.

If the Idempotent configuration is enabled, upon startup the producer will retrieve a ProducerID and add it as well as sequence numbers to all produce requests.

This is still not complete and has not been properly tested yet. Quick tests showed that when retrying with idempotency, messages that have already been written to disk by brokers are correctly marked as duplicates.

This should be enough to start a discussion. Have a look and see if it's something you'd like merged. If so, I'll complete testing.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -797,6 +797,7 @@ func (bp *brokerProducer) handleResponse(response *brokerProducerResponse) {
 func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceResponse) {
 	// we iterate through the blocks in the request set, not the response, so that we notice
 	// if the response is missing a block completely
+	var retryTopics []string
 	sent.eachPartition(func(topic string, partition int32, pSet *partitionSet) {
 		if response == nil {
 			// this only happens when RequiredAcks is NoResponse, so we have to assume success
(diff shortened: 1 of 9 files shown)


mimaison:
@eapache Is this a feature you would be interested in? If I'm not mistaken, this would be the first 3rd party client to support it.

It still needs more work and I plan to complete it this week. So I'd like to know if for any reasons it's not something you'd want

birdayz:
i've also just started working on this, i was not aware that you already got the TX producer running.
i wonder if we can some up with a plan to get the EOS producer and consumer in, with everything that's required. i'm interested in helping.
@buyology mind to share your tx producer? i could't find a branch for this in your repo, just some other stuff like the Coordinator request refactor :)

mimaison:
@buyology We've not solved that issue yet.

Our plan was to try requeuing failed PartitionSets directly into the output channel of BrokerProducer. Then assuming MaxOpenRequests is 1 (like when Idempotent was first released in Java in 0.11), the producer _should_ recover. We've not tested that yet though.

Did you try something similar or some other methods?

edoardocomar:
@buyology we've reimplemented the PR with retry logic applied to whole batches
could you please help review ? 
cc @eapache

async_producer.go
```diff
@@ -740,62 +797,105 @@ func (bp *brokerProducer) handleResponse(response *brokerProducerResponse) {
...
+	produceSet.bufferCount += len(pSet.msgs)
+	for _, msg := range pSet.msgs {
+		if msg.retries >= p.conf.Producer.Retry.Max {
+			p.returnError(msg, kerr)
+			return
+		}
+		msg.retries++
+	}
+	// extremely pessimistic strategy - refreshing metadata for every batch retried. Should be improved
+	err := p.client.RefreshMetadata(topic)
```
buyology: I guess would basically be the `InvalidMetadataException`s?

async_producer.go
```diff
@@ -740,62 +797,105 @@ func (bp *brokerProducer) handleResponse(response *brokerProducerResponse) {
...
 			bp.currentRetries[topic][partition] = block.Err
-			bp.parent.retryMessages(msgs, block.Err)
+			// dropping the following messages has the side effect of incrementing their retry count
 			bp.parent.retryMessages(bp.buffer.dropPartition(topic, partition), block.Err)
+			bp.parent.retryBatch(topic, partition, pSet, block.Err)
+
 		// Other non-retriable errors
 		default:
-			bp.parent.returnErrors(msgs, block.Err)
+			bp.parent.returnErrors(pSet.msgs, block.Err)
```
buyology: The Java client compensates subsequent sequence numbers due to any non-retriable errors: 

https://github.com/apache/kafka/blob/05ba5aa00847b18b74369a821e972bbba9f155eb/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java#L508

I guess we would have to do something similar.

client.go
```diff
@@ -67,6 +67,9 @@ type Client interface {
 	// in local cache. This function only works on Kafka 0.8.2 and higher.
 	RefreshCoordinator(consumerGroup string) error
 
+	// InitProducerID retrieves information required for Idempotent Producer
+	InitProducerID() (*InitProducerIDResponse, error)
```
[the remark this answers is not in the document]
edoardocomar: We think a future implementation of the transactional producer shall change the implementation of `InitProducerID()` to get the TXID from the config.
Reusing a client with a TXID set in the config for creating a new producer would result in the old producer being fenced off as a zombie. Similar to the Java client if you were to reuse the config.

config.go
```diff
@@ -454,6 +457,21 @@ func (c *Config) Validate() error {
...
+		}
+		if c.Producer.Retry.Max == 0 {
+			return ConfigurationError("Idempotent producer requires Producer.Retry.Max >= 1")
+		}
+		if c.Producer.RequiredAcks != WaitForAll {
+			return ConfigurationError("Idempotent producer requires Producer.RequiredAcks to be WaitForAll")
+		}
+		if c.Net.MaxOpenRequests > 1 {
+			return ConfigurationError("Idempotent producer requires Net.MaxOpenRequests to be 1")
+		}
```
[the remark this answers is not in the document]
edoardocomar: yes we stuck to the sarama way of not rewriting the config under the covers

async_producer.go
edoardocomar: Correct, currently upon hitting a non retriable error, we do not attempt to recover automatically from it. Users will have to create a new Producer to get a new ProducerId.

As you pointed out, the Java client tries to recover automatically (by fetching a new producerId and even resequencing message in some cases) but we feel like this can be done as a further improvement/PR.

mimaison:
Sure, I'll do that this week (hopefully today or tomorrow!)

mimaison:
@bai We rebased the PR. One of the Travis jobs failed while installing dependencies but all the other passed. I don't seem to be able to retrigger a build so I can rerun it.

mimaison:
Thank you!

bai:
Many thanks for contributing this 👍

PR #1155, opened 2018-08-22
Add ListTopics func to ClusterAdmin interface

dnwe:
Currently the Sarama ClusterAdmin client (added under https://github.com/Shopify/sarama/pull/1055) didn't have anything analogous to the Java client's [AdminClient#listTopics()](https://kafka.apache.org/10/javadoc/org/apache/kafka/clients/admin/AdminClient.html#listTopics--) command.

This PR is to add one which will return an up-to-date map of all topics in the cluster and any non-default topic configuration that they have.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -1,6 +1,9 @@
 package sarama
 
-import "errors"
+import (
+	"errors"
+	"math/rand"
+)
 
 // ClusterAdmin is the administrative client for Kafka, which supports managing and inspecting topics,
 // brokers, configurations and ACLs. The minimum broker version required is 0.10.0.0.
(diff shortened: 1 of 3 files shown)


birdayz:
I feel the returned structure is insufficient.. TopicDetail is actually used to create topics, so i'm not sure if it's appropriate here.

For example, i'm missing the 'is internal' flag. 
I'm all for being rather close to the 'official' java kafka-client; for reference they return the following structure: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/admin/TopicListing.java.

So it just returns the names of the topics. Pretty much what your code does as well - fetch topics and then the configurations of the topics. I wonder, since this involves multiple network roundtrips - wouldn't it be better to leave the choice to fetch further details to the user, e.g. he calls ListTopics(), gets just the names, and then he calls DescribeTopic (i'm currently working on a PR to add this one) to fetch the details.

birdayz:
what's missing for getting this merged?

dnwe:
Thanks to @birdayz the commits from this PR were included in the follow on #1178 and have now been merged to master so closing this PR

PR #1161, opened 2018-08-31
Reuse objects in consume flow

imjustfly:
As mentioned in issue https://github.com/Shopify/sarama/issues/1131 , consumer allocs many objects which can be reused.

This PR reuses objects in the whole consume flow using `sync.Pool` and can significantly reduce gc frequency. 

Users should call `ReleaseConsumerMessage` after they finished all jobs with the `ConsumerMessage`.

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -275,7 +275,7 @@ func (b *Broker) Produce(request *ProduceRequest) (*ProduceResponse, error) {
 }
 
 func (b *Broker) Fetch(request *FetchRequest) (*FetchResponse, error) {
-	response := new(FetchResponse)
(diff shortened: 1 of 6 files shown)


imjustfly:
I have no idea what causes the test check in go 1.9.x and kafka 2.0 timeout while producing message .

 I didn't do anything related to producer. I can't reproduce it.

varun06:
@imjustfly what is the status for this PR, has been open from long time?

varun06:
There hasn't been any activity on this PR in some time. I would like to close it if don't see an objection in next day to two.

PR #1164, opened 2018-09-06
Allow the Consumer to disable auto-commit offsets

kjelle:
Adding configuration `Consumer.Offsets.Enable` (default `true`)
It blocks the call to `flushToBroker` in `OffsetManager`


kjelle:
One of the tests fail;

> coverage:  71.3% of statements
> panic: test timed out after 1m30s

How does one force a rerun @eapache  ?

snigle:
Hi, I also need to disable the autocommit but I need to commit when I want and at closing.
Do you have the same need ?
If yes this patch should be enough ?  https://github.com/Shopify/sarama/pull/1193

kjelle:
Given the new consumer groups one also needs to escape in the Close() of the offset manager as it requires MainLoop to run... 
I updated my patch @snigle

kjelle:
@taiyang-li I'll merge upstream and see if I is compatible with the latest tag, 1.23.1 - if so, I'll reissue this PR.

kjelle:
cla is signed. im unsure why only one of the travis builds fail, and not all..

varun06:
hey @kjelle 
even though the change is simple, I feel that there should be a test for it, don't want an edge case to pop up.

is it feasible to write a test for it?

config.go
```diff
@@ -423,6 +427,7 @@ func NewConfig() *Config {
 	c.Consumer.MaxWaitTime = 250 * time.Millisecond
 	c.Consumer.MaxProcessingTime = 100 * time.Millisecond
 	c.Consumer.Return.Errors = false
+	c.Consumer.Offsets.Enable = true
```
sam-obeid: Would suggest have a more descriptive name:

```suggestion
	c.Consumer.Offsets.AutoCommitEnable = true
```

Also `c.Consumer.Offsets.AutoCommit.Enable` would be fine.

varun06: @kjelle can you please make this change?

kjelle: > Also `c.Consumer.Offsets.AutoCommit.Enable` would be fine.

I totally agree @sam-obeid. I am making appropriate changes and reissuing the PR.

kjelle:
> is it feasible to write a test for it?

I am adding tests to check if setting the `Consumer.Offsets.AutoCommit.Enable` gives the expected results.

kjelle:
@varun06 I discovered some problems while writing the unit-tests. I am rewriting how the offset_manager.go will handle a disabled `Consumer.Offsets.AutoCommit.Enable`. I will test it in our production line before re-submitting the PR.

varun06:
@kjelle That's why I asked for test :)

Thanks for doing that and finding the issue.

varun06:
@sam-obeid can you please have a look again?

kjelle:
@varun06 given the amount of commits I would suggest a merge squash when it comes to that :)

varun06:
Done, Please test and let me know how it goes.

config.go
```diff
@@ -338,8 +338,15 @@ type Config struct {
 		// offsets. This currently requires the manual use of an OffsetManager
 		// but will eventually be automated.
 		Offsets struct {
-			// How frequently to commit updated offsets. Defaults to 1s.
-			CommitInterval time.Duration
```
d1egoaz: ohh shit, this was a breaking change, i.e: it breaks people using sarama-cluster with the latest version of sarama.
```
../../../../go/pkg/mod/github.com/bsm/sarama-cluster@v2.1.15+incompatible/consumer.go:452:59: c.client.config.Config.Consumer.Offsets.CommitInterval undefined (type struct { AutoCommit struct { Enable bool; Interval time.Duration }; Initial int64; Retention time.Duration; Retry struct { Max int } } has no field or method CommitInterval)
```

offset_manager.go
```diff
@@ -233,7 +233,12 @@ func (om *offsetManager) mainLoop() {
 	}
 }
 
+// flushToBroker is ignored if auto-commit offsets is disabled
 func (om *offsetManager) flushToBroker() {
+	if !om.conf.Consumer.Offsets.AutoCommit.Enable {
```
imaffe, line 238: Glad this line is fixed, it actually blocks all calls to Commit(), however sometimes we do wish to commit manually.

kjelle:
I'm sorry if this PR caused a lot of hazzle. I commit offsets manually by making my own offset manager sending a sarama.OffsetCommitRequest with my own block containing topic, partition and offset. I then send it to the correct broker coordinating that partition. If Sarama now supports manual offset commit I could move to that instead.

PR #1167, opened 2018-09-12
consumer: set BlockTimestamp when consuming V2 messages (Records)

champtar:
record.MaxTimestamp is set to "log append time" when using
message.timestamp.type=LogAppendTime

see https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/record/DefaultRecordBatch.java#L85

Signed-off-by: Etienne Champetier <champetier.etienne@gmail.com>

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -520,13 +520,14 @@ func (child *partitionConsumer) parseRecords(batch *RecordBatch) ([]*ConsumerMes
 			continue
 		}
 		messages = append(messages, &ConsumerMessage{
-			Topic:     child.topic,
-			Partition: child.partition,
-			Key:       rec.Key,
-			Value:     rec.Value,
-			Offset:    offset,
-			Timestamp: batch.FirstTimestamp.Add(rec.TimestampDelta),
-			Headers:   rec.Headers,
+			Topic:          child.topic,
+			Partition:      child.partition,
+			Key:            rec.Key,
+			Value:          rec.Value,
+			Offset:         offset,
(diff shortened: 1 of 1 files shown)


varun06:
Hi @champtar,  can you please fix the conflicts and have a test case for it?

champtar:
Hi @varun06,
Sadly I don't use sarama anymore, add to go with librdkafka for performance
I can rebase but I don't have time to do proper test case
Looking at master, https://github.com/Shopify/sarama/commit/f0e5fa7ff707c878111fad8b1ccce635c8be3286 seems to fix my issue

varun06:
Nice, Can we close it then?

varun06:
Closing as fixed in https://github.com/Shopify/sarama/commit/f0e5fa7ff707c878111fad8b1ccce635c8be3286

champtar:
Just to be clear, my initial issue was to have access to "log append time" when using " message.timestamp.type=LogAppendTime", this is fixed via https://github.com/Shopify/sarama/commit/f0e5fa7ff707c878111fad8b1ccce635c8be3286

Now you are still not setting the value of BlockTimestamp when consuming V2 messages, but you are when consuming V1 messages. Other kafka clients only expose 1 timestamp, so I would recommend either:
1) set BlockTimestamp for consistency between V1 and V2
2) remove BlockTimestamp

varun06:
Thanks for clarifying that. I shall create a new issue with this info.

Issue #1168, opened 2018-09-15
AsyncProducer rejects messages at high rates...


joshua-rutherford:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*
Sarama Version: 1.15.0
Kafka Version: 0.10.2.1
Go Version: 1.10.3

##### Configuration

```
brokerList := ...
producer, err := sarama.NewAsyncProducer(brokerList, sarama.NewConfig())
```

##### Logs

No logs from Sarama or Kafka that provide details on the issue.

##### Problem Description

When we publish two messages to the same producer in rapid succession the second message sometimes fails to be enqueued.  

After review of the Sarama source code for the AsyncProducer it is clear that there is a shared channel `AsyncProducer.Input()` to which all messages are originally enqueued and then channels for each topic which consume from that channel.  It seems that often the second message fails to be enqueued because the first message has not yet been popped off the input channel and into the topic channel.

Modifying the initialization of the input channel to be the same size as each of the topic channels solves this issue in local testing but I'm not sure that is the most desirable fix given that it will effectively double the per topic channel in the worst case. See below for the change at https://github.com/Shopify/sarama/blob/master/async_producer.go#L91:

```
		input:      make(chan *ProducerMessage, client.Config().ChannelBufferSize),
```

Clearly this is an easy fix, but I'm not sure if the it would be accepted or if I should buffer in my code instead.

Thanks.

joshua-rutherford:
Closing as this issue is with our use of Sarama, not Sarama itself.

PR #1169, opened 2018-09-15
Fix input channel overflow

joshua-rutherford:
This is a pretty straight forward fix for #1168 but it may be sub-optimal given it effectively means that in the worst case scenario the producer will accept `ChannelBufferSize * 2` for a given topic.  It may be more desirable to add another configuration option to set the shared channel buffer size independently of the topic channel buffer size.


joshua-rutherford:
Closing as this issue is with our usage of Sarama, not Sarama itself.

PR #1170, opened 2018-09-18
Zstd support for compression

bobrik:
[KIP-110](https://cwiki.apache.org/confluence/display/KAFKA/KIP-110%3A+Add+Codec+for+ZStandard+Compression) is finally [accepted](https://lists.apache.org/thread.html/7ab8407e6cff115eafab1b8e0750cbb906cc9414104b352e26b32e55@%3Cdev.kafka.apache.org%3E), so I can make a PR with something [we've been using](https://blog.cloudflare.com/squeezing-the-firehose/) for quite a while now.

cc @dqminh, @andrew-plunk

### DIFF
--- a/message.go
+++ b/message.go
@@ -116,15 +116,11 @@ func (m *Message) encode(pe packetEncoder) error {
 			m.compressedCache = buf.Bytes()
 			payload = m.compressedCache
 		case CompressionZSTD:
-			var buf bytes.Buffer
-			writer := zstd.NewWriterLevel(&buf, m.CompressionLevel)
-			if _, err = writer.Write(m.Value); err != nil {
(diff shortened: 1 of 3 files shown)


bai:
Thanks for contributing this. Could we add some tests, similarly to how lz4 is tested?

bobrik:
@bai good point. I started looking into that and hit https://github.com/DataDog/zstd/issues/41.

bobrik:
I've updated PR with tests and a workaround for zero sized messages.

bai:
❤️

lizthegrey:
This patch does not appear to actually work in practice without also incrementing the version of the producer and consumer protocol messages beyond `3` and `4` to `7` and `9`, respectively. However, there are potentially breaking spec changes (even if not field/proto changes) in the intermediate versions we'd be skipping over.

PR #1173, opened 2018-09-25
exported Broker's rack setting

stevevls:
This change enables library clients to use the sarama Client to implement
rack-aware features.


stevevls:
My pleasure!  The build looks a lot happier after picking up the CI test fixes commit.

PR #1174, opened 2018-09-26
fix encoding of MetadataResponse with version >= 2

birdayz:
ThrottleTimeMs and ClusterID have been added to decode(), but apparently
encode() was forgotten. This causes errors in tests, if versions >= 2 are used.


birdayz:
I fixed a race in the mock async producer, unrelated to this PR. should i move it to a separate PR?

bai:
Thanks!

PR #1178, opened 2018-09-27
admin: Add some missing admin methods

birdayz:
This PR adds the following admin client methods:
DescribeTopic
DescribeConsumerGroup
ListConsumerGroups
ListConsumerGroupOffsets

For ListConsumerGroups, all brokers are queried (the java client does it the same way). Since this will be super slow with multiple brokers, i made an implementation with parallel requests.
The others methods are rather straight forward.
I'll be happy about comments on the API, is it fine?
I'm not 100% sure - for the groups i chose it to be  'singular' - each method call is for one consumer group. For topics it's a slice of topics as parameter - in my case (building a kafka cli tool) it's very often the case that multiple topics are requested, so i kept it this way.

In #1155 (another addition to the admin client) i wondered how the API should look like. I kept it simple here. and reused types from the response messages.

I also added mock tests for these methods.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -1,6 +1,9 @@
 package sarama
 
-import "errors"
+import (
+	"errors"
+	"math/rand"
+)
 
 // ClusterAdmin is the administrative client for Kafka, which supports managing and inspecting topics,
 // brokers, configurations and ACLs. The minimum broker version required is 0.10.0.0.
@@ -12,6 +15,9 @@ type ClusterAdmin interface {
 	// to become aware that the topic has been created. During this time, listTopics
 	// may not return information about the new topic.The validateOnly option is supported from version 0.10.2.0.
 	CreateTopic(topic string, detail *TopicDetail, validateOnly bool) error
+
+	// List the topics available in the cluster with the default options.
+	ListTopics() (map[string]TopicDetail, error)
 
 	// Delete a topic. It may take several seconds after the DeleteTopic to returns success
 	// and for all the brokers to become aware that the topics are gone.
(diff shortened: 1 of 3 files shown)


birdayz:
i renamed it to DescribeTopics, so it's consistent with its parameters and return value.

birdayz:
@bai can you have a look?

jeffwidman:
I'm just checking out Sarama for the first time, so not sure how closely you try to follow the Java reference clients. However, I added these same methods to `kafka-python`, so had one design comment based on having already played with the kafka protocol here:
* A common scenario in monitoring situations is to fetch the offsets/calculate consumer lag for many/all consumer groups in the cluster. 
* `ListConsumerGroupOffsets` is a very chatty call because it will first lookup the group's coordinator. So it requires two network round trips.
* However, when listing consumer groups, if a broker returns a consumer group, you know it's the coordinator.
* If you allow `ListConsumerGroups` to have an optional param of specifying a single broker and `ListConsumerGroupOffsets` to have an optional param of specifying a known group coordinator than you can leverage this to nearly halve the number of network calls required to fetch offsets for all consumer groups:
    1. Loop through all brokers, calling `ListConsumerGroups` against each one.
    2. For the consumer groups returned, call `ListConsumerGroupOffsets`, using the group coordinator identified above. 

Java doesn't currently allow these overrides, but they were useful enough that we chose to include them in `kafka-python`, even though in general we try to follow the Java clients very closely: https://kafka-python.readthedocs.io/en/master/apidoc/KafkaAdminClient.html#kafka.admin.KafkaAdminClient.list_consumer_group_offsets

You can see how I then leveraged this when patching the Datadog check here:
https://github.com/DataDog/integrations-core/pull/2730/files#diff-deed983f73a04522ac621d7f6d0c8403R248

admin.go
```diff
@@ -65,6 +72,15 @@ type ClusterAdmin interface {
 	// This operation is supported by brokers with version 0.11.0.0 or higher.
 	DeleteACL(filter AclFilter, validateOnly bool) ([]MatchingAcl, error)
 
+	// List the consumer groups available in the cluster.
+	ListConsumerGroups() (map[string]string, error)
+
+	// Describe some group IDs in the cluster.
+	DescribeConsumerGroup(group string) (*GroupDescription, error)
```
varun06: Can you please update the comment, it is not very clear here.

admin.go
```diff
@@ -380,3 +417,87 @@ func (ca *clusterAdmin) DeleteACL(filter AclFilter, validateOnly bool) ([]Matchi
 	}
 	return mAcls, nil
 }
+
+func (ca *clusterAdmin) DescribeConsumerGroup(group string) (*GroupDescription, error) {
```
varun06: I wonder if it works better as `DescribeConsumerGroups(groups []string)`, coz controller takes a slice of string anyway, and that way if you only want it for one group, you can only pass that too?

admin.go
```diff
@@ -380,3 +417,87 @@ func (ca *clusterAdmin) DeleteACL(filter AclFilter, validateOnly bool) ([]Matchi
...
+	brokers := ca.client.Brokers()
+	groupMaps := make(chan map[string]string, len(brokers))
+	errors := make(chan error, len(brokers))
+	wg := sync.WaitGroup{}
+
+	for _, b := range brokers {
+		wg.Add(1)
+		go func(b *Broker, conf *Config) {
+			defer wg.Done()
+			_ = b.Open(conf) // Ensure that broker is opened
```
varun06: you can do 
```suggestion
			b.Open(conf) // Ensure that broker is opened
```

birdayz: i can, but then we don't document that there's an error (deliberately) ignored here.

birdayz:
@dnwe i added your commits of your PR #1155 ..since my work is based on that / related. since your PR is also pending and i require it as well, do you mind pushing this forward together? would be nice if we could get this finally merged :)
if it's not OK, i can remove it from my PR.

admin.go
birdayz: in fact errcheck will fail travis if i do what you suggest.. https://api.travis-ci.org/v3/job/490310704/log.txt i reverted that.

admin.go
birdayz: i agree. i now improved it to accept a slice of group names. in addition, the logic is now quite smart to first build a map of brokers -> []string groups, so each group coordinator is contacted only once. in a trivial test / cluster with just 8 groups this saves me a second (50% savings)..which will have much a bigger impact on larger clusters.

admin.go
birdayz: done

birdayz:
Ci failed with a strange error:
```
# cd .; git clone https://go.googlesource.com/tools /home/travis/gopath/src/golang.org/x/tools
Cloning into '/home/travis/gopath/src/golang.org/x/tools'...
fatal: remote error: Access denied to IP 35.188.1.99
package golang.org/x/tools/go/packages: exit status 128
```
how can i trigger a re-run?

varun06:
Try closing and reopen again.

Sent from my iPhone

On Feb 7, 2019, at 6:07 PM, Johannes Brüderl <notifications@github.com<mailto:notifications@github.com>> wrote:


Ci failed with a strange error:

# cd .; git clone https://go.googlesource.com/tools /home/travis/gopath/src/golang.org/x/tools<http://golang.org/x/tools>
Cloning into '/home/travis/gopath/src/golang.org/x/tools'<http://golang.org/x/tools'>...
fatal: remote error: Access denied to IP 35.188.1.99
package golang.org/x/tools/go/packages:<http://golang.org/x/tools/go/packages:> exit status 128


how can i trigger a re-run?

birdayz:
great. it's green now :)

bai:
Thanks!

bai:
This is lovely work 💯

PR #1180, opened 2018-10-01
Revert "Add workaround for an empty zstd message"

bobrik:
This reverts commit 0f30d9760beb21bf0fc07612077a21770917807a.

The issue was fixed in https://github.com/DataDog/zstd/pull/42

cc @Viq111 @jmoiron


message_test.go
```diff
@@ -60,7 +60,7 @@ var (
 		0xFF, 0xFF, 0xFF, 0xFF, // key
 		0x00, 0x00, 0x00, 0x0d, // len
 		// ZSTD data
-		0x28, 0xb5, 0x2f, 0xfd, 0x24, 0x00, 0x01, 0x00, 0x00, 0x99, 0xe9, 0xd8, 0x51,
```
bobrik: 4 missing bytes are an optional checksum. `zstd` cli adds it, but DataDog's package does not and I don't see an option to enable it.

There's a message level checksum and checksum, so I think it's fine to leave it out.

bobrik: Opened an issue, but it should not be a blocker here:

* https://github.com/DataDog/zstd/issues/43

message_test.go
```diff
@@ -53,14 +53,14 @@ var (
 	}
 
 	emptyZSTDMessage = []byte{
-		252, 62, 137, 23, // CRC
+		180, 172, 84, 179, // CRC
 		0x01,                          // version byte
 		0x04,                          // attribute flags: lz4
```
bai: nit: `attribute flags: zstd`

bobrik: Oops, fixed.

PR #1182, opened 2018-10-02
Fix build without cgo

### LINKED ISSUE #1179: Builds with CGO_ENABLED=0 are failing
##### Versions

Sarama Version: 74799833f433718a70bfd6fea1d1cc2d55ef6f37
Kafka Version: 
Go Version: go1.11 darwin/amd64

##### Configuration

##### Logs

```bash
$ CGO_ENABLED=0 go build
go build github.com/DataDog/zstd: build constraints exclude all Go files in .../go/src/github.com/DataDog/zstd
```

##### Problem Description

We build all our code in alpine with cgo disabled. Since recent commit `6bfdd61781f74c235e05b72e41308561394b8232` introduced `github.com/DataDog/zstd` which is a c lib wrapper our code with the current config won't compile.

I understand this is not a _real_ bug, but a change that probably breaks many team's builds. Hence, the report.

muirrn:
The github.com/DataDog/zstd library wraps a c library, so it requires
cgo to compile. Add zstd wrapper funcs to sarama and put the zstd
import behind a build flag requiring cgo. Without cgo, the wrapper
funcs return an error saying zstd doesn't work without cgo.

Resolves #1179

### DIFF
--- a/zstd_fallback.go
+++ b/zstd_fallback.go
@@ -1,0 +1,17 @@
+// +build !cgo
+
+package sarama
+
+import (
+	"errors"
+)
+
+var errZstdCgo = errors.New("zstd compression requires building with cgo enabled")
+
+func zstdDecompress(dst, src []byte) ([]byte, error) {
+	return nil, errZstdCgo
+}
+
+func zstdCompressLevel(dst, src []byte, level int) ([]byte, error) {
+	return nil, errZstdCgo
+}
--- a/message.go
+++ b/message.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"time"
 
-	"github.com/DataDog/zstd"
 	"github.com/eapache/go-xerial-snappy"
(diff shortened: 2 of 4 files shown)


bai:
Thank you!

cc: @bobrik -- could you please 👀 this one too?

zstd_fallback.go
```diff
@@ -0,0 +1,17 @@
+// +build !cgo
+
+package sarama
+
+import (
+	"errors"
+)
+
+var errZstdCgo = errors.New("zstd compression requires building with cgo enabled")
```
bobrik: Should this error be public?

muirrn: I considered that, but I didn't think it would be useful (what is the user going to do differently in this case?). If it were public, we would have to maintain it forever to keep backwards compatibility (e.g. even after switching to a pure-go zstd library).

PR #1183, opened 2018-10-03
Add option to propagate OffsetOutOfRange error

### LINKED ISSUE #1181: Consumer group offset can reset during rebalance if underreplicated
##### Versions

Sarama Version: 74799833f433718a70bfd6fea1d1cc2d55ef6f37
Kafka Version: 1.0, 2.0
Go Version: 1.11

##### Problem Description

If you have underreplication for whatever reason (e.g. publishing messages at sarama.WaitForLocal instead of WaitForAll), a rebalance can end up resetting a consumer group partition offset back to the initial position. This happens when you consume up to, say, offset 100 on replica A, but replica B only has data up to 99 due to temporary underreplication. When rebalancing to replica B, the client will return `ErrOffsetOutOfRange` trying to subscribe to offset 100, and that causes the consumer to reset to the initial offset.

Is there a reason ErrOffsetOutOfRange is not propagated up to the user? Are there any cases the user would want to silently reset the consumer offset?

Note that we experienced this behavior using sarama-cluster. I am not able to reproduce the error consistently and have not reproduced it with the sarama consumer yet. However, the code seems to behave the same as sarama-cluster in this regard.

muirdm:
When consuming a partition using a consumer group, the code
handles ErrOffsetOutOfRange errors by resetting to the "initial"
position, as specified by user (i.e. either oldest or newest available
offset). This, however, can be very dangerous. Say a consumer has
consumed up to offset 100 on replica A but replica B has only
replicated up to offset 99 due to temporary under-replication. During
a rebalance, sarama can end up with an offset out-of-range error if it
fetches partition metadata from replica B since the desired offset of
100 is greater than the newest offset of 99. The sarama consumer would
reset the offset in this case, which can cause reprocessing of old
data, especially if the initial offset is configured as "oldest".

This commit adds a config flag to disable this automatic reset. In the
above case, the consumer will be able to proceed normally after the
data replicates.

Resolves #1181

### DIFF
--- a/config.go
+++ b/config.go
@@ -223,6 +223,14 @@ type Config struct {
 				// coordinator for the group.
 				UserData []byte
 			}
+
+			// If true, consumer offsets will be automatically reset to configured Initial value
+			// if the fetched consumer offset is out of range of available offsets. Out of range
+			// can happen if the data has been deleted from the server, or during situations of
+			// under-replication where a replica does not have all the data yet. It can be
+			// dangerous to reset the offset automatically, particularly in the latter case. Defaults
+			// to true to maintain existing behavior.
+			ResetInvalidOffsets bool
 		}
 
 		Retry struct {
@@ -381,6 +389,7 @@ func NewConfig() *Config {
 	c.Consumer.Group.Rebalance.Timeout = 60 * time.Second
 	c.Consumer.Group.Rebalance.Retry.Max = 4
 	c.Consumer.Group.Rebalance.Retry.Backoff = 2 * time.Second
+	c.Consumer.Group.ResetInvalidOffsets = true
 
 	c.ClientID = defaultClientID
 	c.ChannelBufferSize = 256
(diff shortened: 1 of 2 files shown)


varun06:
@dim can you please have a look also @muirrn can you please add a test to it?

dim:
@varun06 I don't think this is easily testable. It's not something that can be easily re-created in a test scenario.

@muirrn thanks for the detailed explanation and the patch. One question: how would you handle these cases. Assuming the `ErrOffsetOutOfRange` error is returned to the user, how would you decide what to do next? I wonder if there is a more generic solution to the problem.

muirrn:
The error gets propagated and we retry consuming until it works.

Maybe a general solution is to only fetch offsets from the group coordinator, but I'm not really sure.

dim:
@muirrn the reason why we added this in the first place was an endless loop of `ErrOffsetOutOfRange` errors when a high-volume partition (with a very low TTL) was pruned and the stored offset was < the first available offset. I think your patch is safe to apply, but I would even go a bit further and make `ResetInvalidOffsets` to false by default

muirdm:
@dim are you still in favor of changing the default to `false` and proceeding with this change? I don't have a good understanding of how this might affect other users, so I was trying to be conservative.

dim:
@muirdm yes, I still think that `false` may be a better default. I doubt it will affect anyone TBH. Apologies for the delayed response.

PR #1185, opened 2018-10-05
Reuse compression objects

muirrn:
Reusing lz4/gzip reader and writer objects via sync.Pool greatly reduces CPU usage (saves allocations and creates much less garbage).

### DIFF
--- a/compress.go
+++ b/compress.go
@@ -1,0 +1,75 @@
+package sarama
+
+import (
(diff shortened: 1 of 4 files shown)


qiangyin:
Yes,I used 1.17.0 and gzip, will  lead to gc use cpu very high.

PR #1189, opened 2018-10-08
Maintain metadata in the producer even when retries are disabled.

stevevls:
Some applications may wish to disable retries so that they can implement custom
retry logic.  As currently written, though, setting Config.Producer.Retry.Max to
zero also disables the mechanism by which the producer updates its metadata in
response to an error.  If the producer gets an error such as
ErrNotLeaderForPartition or an io.EOF, it will never recover...even in response
to background metadata updates.  This commit resolves that issue by ensuring
that the partition producers still update the leader and corresponding input
channel even when retries are turned off.


stevevls:
I imagine that there aren't very many folks who set `Config.Producer.Retry.Max` to zero, but I'm actually working on a couple apps with very specific retry requirements at the moment. 😄  This patch addresses keeping the `partitionProducer`'s `brokerProducer` reference up to date in a minimally-invasive way.  I had initially attempted to splice the functionality into the retry high/low watermark code, but it got pretty complex to reason about the different scenarios, and I abandoned that in favor of this code.  Any feedback or suggestions on how to improve this patch are much appreciated.

PR #1191, opened 2018-10-11
Add error codes 61-72

greysond:
Saw these were missing and thought I'd add them.


greysond:
There seems to be something funky with the CI. Tests run successfully locally.

Issue #1158, opened 2018-08-24
Allow the Consumer to disable auto-commit offsets


kjelle:
##### Versions

Sarama Version: v1.17.0
Kafka Version: 2.2.0
Go Version: go1.10

##### Configuration

Not applicable

##### Logs

Not applicable

##### Problem Description

I have a scenario where I want to use the PartitionManager in Sarama, but disable the auto commit of offsets. I use the MarkOffset to calculate `lag` in my consumer.  I also see from a google search there are others with similar issues.

Since the `Consumer.Offsets.CommitInterval` is used in a time.NewTicker it cannot be disabled:
`The duration d must be greater than zero; if not, NewTicker will panic.`

##### Workarounds
A workaround would be to set this to a high number so the `bom.timer.C` never triggers. We run Go in docker containers which runs for months or years, I would prefer a disable.

Another workaround would be not to call `MarkOffset`. I then would need to keep my own store of current partition offsets in the consumer.

##### My own branch
I have run a branched Sarama version for some time, it has two additions:
- It adds a configuration `Consumer.Offsets.Enable` bool
- In https://github.com/Shopify/sarama/blob/v1.17.0/offset_manager.go#L417 we check if `Consumer.Offsets.Enable` is false, and then calls `bom.timer.Stop()`.

This might not be the ideal solution, or right place to put the check, but it works for me.

Since I need to reapply my patch for every release of Sarama, perhaps this could be addressed and fixed in the library?

kjelle:
A possibly better approach would be to check for `Consumer.Offsets.Enable` in https://github.com/Shopify/sarama/blob/v1.17.0/offset_manager.go#L445 so we do not tamper with the `bom.timer` but only block the `OffsetCommitRequest` as long as `Consumer.Offsets.Enable` is set.

I could prepare a PR

varun06:
@kjelle PR is definitely welcome.

kjelle:
@varun06 Any way of getting #1164  merged? CLA is signed

varun06:
Will do today. Thanks.

p3lim:
@varun06 What's the status on this?

varun06:
Hey Friend, there is one comment on PR, please have a look, once that is resolved, will merge it.

kjelle:
I am closing this issue since the PR #1164 was merged.

PR #1193, opened 2018-10-16
#1158 permits to disable offset autocommitment

snigle:
If you want to be sure that your consumers doesn't take a same message in parallel at rebalance, you need to block the until the process of the message is done.

For that, configure a rebalance timeout  : `conf.Config.Consumer.Group.Rebalance.Timeout = 5 * time.Minute`
When you receive the message, mark offset : `consumer.MarkOffset(msg, "")`
When you finish your process for this message, commit offsets : `consumer.CommitOffsets()`

To avoid loosing message if your application reboot when your message is processing or this there is error in process, we don't want to commit before the process is done.

This workflow can work only if the autocommit is disabled. So with this new conf :
`conf.Consumer.Offsets.CommitInterval = 0`

### DIFF
--- a/config.go
+++ b/config.go
@@ -292,7 +292,7 @@ type Config struct {
 		// offsets. This currently requires the manual use of an OffsetManager
 		// but will eventually be automated.
 		Offsets struct {
-			// How frequently to commit updated offsets. Defaults to 1s.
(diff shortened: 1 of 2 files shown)


snigle:
It seems this workflow doesn't work. The MarkOffset doesn't block rebalancing.

PR #1198, opened 2018-10-22
Support new OffsetFetch request/response

mimaison:
This adds support for OffsetFetch up to version 5 (v5 is added in Kafka 2.1).

We were especially interested in v2, which allows to query offsets for all partitions used by a group instead of explicitly passing the partitions but adding the extra versions was straight forward (very few new arguments).


mimaison:
@bai At the Kafka Summit last week, we had a chat with @sam-obeid. We are interested in contributing regularly to this project and we are wondering if you'd consider adding new maintainers/committers.

mimaison:
@bai I've just rebased this PR on master. Can you have a look? Thanks!

Issue #1216, opened 2018-11-15
Error codes are missing in the list of error codes


nrvnrvn:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*
Sarama Version: ec843464b50d4c8b56403ec9d589cf41ea30e722
Kafka Version: 1.1.1
Go Version: 1.11

##### Configuration

What configuration values are you using for Sarama and Kafka?

##### Logs

When filing an issue please provide logs from Sarama and Kafka if at all
possible. You can set `sarama.Logger` to a `log.Logger` to capture Sarama debug
output.

##### Problem Description

While implementing a simple CLI app for removing orphaned consumer groups got this message:
```
2018/11/15 13:46:37 Successfully initialized new client
2018/11/15 13:46:37 client/coordinator requesting coordinator for consumergroup woot from kafka:9092
2018/11/15 13:46:37 client/coordinator coordinator for consumergroup woot is #3 (10.20.5.14:9092)
2018/11/15 13:46:37 Connected to broker at 10.20.5.14:9092 (registered as #3)
&{0s map[woot:Unknown error, how did this happen? Error code = 69]}
2018/11/15 13:46:37 Closing Client
```

Corresponding code snippet:
```go
	config := sarama.NewConfig()
	config.Version = sarama.V1_1_0_0
	client, err := sarama.NewClient(strings.Split(*brokerList, ","), config)
	if err != nil {
		// deal with it
	}
	defer client.Close()

	broker, err := client.Coordinator(*consumerGroupID)
	if err != nil {
		// deal with it
	}

	resp, err := broker.DeleteGroups(&sarama.DeleteGroupsRequest{Groups: []string{*consumerGroupID}})
	if err != nil {
		// deal with it
	}
	fmt.Printf("%v\n", resp)
```

`fmt.Printf("%v\n", resp)` produces the `&{0s map[woot:Unknown error, how did this happen? Error code = 69]}` line. 
Error code `69` (`GROUP_ID_NOT_FOUND `) is absent apparently.

Moreover https://kafka.apache.org/protocol#protocol_error_codes and https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ErrorCodes do not match. And the latter is referred to in errors.go

The suggestion is:
1. add missing error codes
[...]

PR #1218, opened 2018-11-15
Synced error names and descriptions with the kafka's protocol

nrvnrvn:
Part of #1216 
This one is very opinionated since it makes a lot of renamings just for the sake of clear mapping between sarama's error codes and messages and kafka protocol.

Errors from https://kafka.apache.org/protocol#protocol_error_codes were converted to CamelCase style. Descriptions were taken as is. 
See also: 
https://github.com/apache/kafka/blob/741cb761c5239297029a446518c332f6c4ed08f6/clients/src/main/java/org/apache/kafka/common/protocol/Errors.java#L116

While it would be useful to keep this on par with the upstream kafka's error names and descriptions I understand that there is possibility of breaking backwards compatibility for users who already rely on error names that sarama has provided.

If you think this change is not worth it feel free to decline it.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -105,7 +105,7 @@ func (ca *clusterAdmin) Controller() (*Broker, error) {
 func (ca *clusterAdmin) CreateTopic(topic string, detail *TopicDetail, validateOnly bool) error {
 
 	if topic == "" {
-		return ErrInvalidTopic
+		return ErrInvalidTopicException
 	}
 
 	if detail == nil {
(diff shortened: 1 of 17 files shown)


varun06:
This one looks really helpful. I also had some hard time with error messages recently. 👍

varun06:
@nrvnrvn If you can get the tests passed, we will get it approved and merged.

nrvnrvn:
@sam-obeid fixed the failing tests. Not sure if I need to do anything about codecov though...

varun06:
@nrvnrvn I think it should be fine as the changes are mostly in naming. @bai what do you think, can we merge this one?

bai:
Yep, thanks so much for @nrvnrvn for contributing and thanks @sam-obeid and @varun06 for reviewing.

bai:
Any major objections to reverting this one to address public API regression?

varun06:
I am in favor of reverting and bring it back with proper process.

nrvnrvn:
> I am in favor of reverting and bring it back with proper process.

@bai @varun06 should I do anything about that?

varun06:
I think it has to be cherry picked in a new release. it's a braking change and need to communicated properly. @bai what's your thought?

PR #1222, opened 2018-11-22
Producer performance tool

TheSalmonTapes:
This tool mirrors `kafka-producer-perf-test.sh` in the Kafka distribution. I was using this myself for some performance testing on this library and thought I might as well drop this here in in case others find it useful.

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -6,5 +6,6 @@ Some of these tools mirror tools that ship with Kafka, but these tools won't req
 - [kafka-console-producer](./kafka-console-producer): a command line tool to produce a single message to your Kafka custer.
 - [kafka-console-partitionconsumer](./kafka-console-partitionconsumer): (deprecated) a command line tool to consume a single partition of a topic on your Kafka cluster.
 - [kafka-console-consumer](./kafka-console-consumer): a command line tool to consume arbitrary partitions of a topic on your Kafka cluster.
+- [kafka-producer-performance](./kafka-producer-performance): a command line tool to performance test production on your Kafka cluster.
 
 To install all tools, run `go get github.com/Shopify/sarama/tools/...`
--- a/tools/README.md
+++ b/tools/README.md
@@ -6,6 +6,6 @@ Some of these tools mirror tools that ship with Kafka, but these tools won't req
 - [kafka-console-producer](./kafka-console-producer): a command line tool to produce a single message to your Kafka custer.
 - [kafka-console-partitionconsumer](./kafka-console-partitionconsumer): (deprecated) a command line tool to consume a single partition of a topic on your Kafka cluster.
 - [kafka-console-consumer](./kafka-console-consumer): a command line tool to consume arbitrary partitions of a topic on your Kafka cluster.
-- [kafka-producer-performance](./kafka-producer-performance): a command line tool to performance test production on your Kafka cluster.
+- [kafka-producer-performance](./kafka-producer-performance): a command line tool to performance test producers (sync and async) on your Kafka cluster.
 
 To install all tools, run `go get github.com/Shopify/sarama/tools/...`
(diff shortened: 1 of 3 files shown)


varun06:
Looks great. 👍

tools/README.md
```diff
@@ -6,5 +6,6 @@ Some of these tools mirror tools that ship with Kafka, but these tools won't req
 - [kafka-console-producer](./kafka-console-producer): a command line tool to produce a single message to your Kafka custer.
 - [kafka-console-partitionconsumer](./kafka-console-partitionconsumer): (deprecated) a command line tool to consume a single partition of a topic on your Kafka cluster.
 - [kafka-console-consumer](./kafka-console-consumer): a command line tool to consume arbitrary partitions of a topic on your Kafka cluster.
+- [kafka-producer-performance](./kafka-producer-performance): a command line tool to performance test production on your Kafka cluster.
```
varun06:
```suggestion
- [kafka-producer-performance](./kafka-producer-performance): a command line tool to performance test your Kafka cluster.
```

tools/kafka-producer-performance/main.go
```diff
@@ -0,0 +1,358 @@
...
+	case "gzip":
+		return sarama.CompressionGZIP
+	case "snappy":
+		return sarama.CompressionSnappy
+	case "lz4":
+		return sarama.CompressionLZ4
+	default:
+		printUsageErrorAndExit(fmt.Sprintf("Unknown -compression: %s", scheme))
+	}
+	panic("should not happen")
```
varun06: any particular reason to use `panic`, what about `log.Fatal("should not happen")`.

tools/kafka-producer-performance/main.go
```diff
@@ -0,0 +1,358 @@
...
+	if err != nil {
+		printErrorAndExit(69, "Failed to create producer: %s", err)
+	}
+	defer func() {
+		if err := producer.Close(); err != nil {
+			printErrorAndExit(69, "Failed to close producer: %s", err)
+		}
+	}()
+
+	messagesDone := make(chan struct{})
```
varun06: you can pass context to these function and then exit on `context.Done()` from go routine.

tools/kafka-producer-performance/main.go
TheSalmonTapes: Because this is programming error and `panic` prints a stack trace; if we hit this line, there's clearly a bug since we would have expected to exit on line 136.

tools/kafka-producer-performance/main.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+		messages[i] = &sarama.ProducerMessage{
+			Topic:     *topic,
+			Partition: int32(*partition),
+			Value:     sarama.ByteEncoder(payload),
+		}
+	}
+
+	// Print out metrics periodically.
+	metricsDone := make(chan struct{})
```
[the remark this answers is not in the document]
TheSalmonTapes: Yes, you're completely right. When `cancel` is called we can wait on `ctx.Done` in both the goroutine here and in the main routine where I was waiting with `metricsDone` before. Thanks!

tools/kafka-producer-performance/main.go
TheSalmonTapes: Do you mean passing the context from the main routine to `runAsyncProducer` or instantiating a new context in this function and replacing `messagesDone` with it?  The first case is not desirable because we don't want to exit from this function until this goroutine has finished (i.e. all the messages have been successfully sent, or an error occurred). The second case is really just a matter of style; we can wait on `context.Done` at the end of the function or this `messagesDone` channel.  I don't have a huge preference here.

tools/README.md
TheSalmonTapes: Updated. Thanks!

varun06:
This can be helpful, can we get it going?

varun06:
@TheSalmonTapes should I merge it?

TheSalmonTapes:
> @TheSalmonTapes should I merge it?

By all means!

PR #1229, opened 2018-11-28
Add support for Kafka 2.1.0 and fix CI

bai:
Old download URLs for kafka return 404 so I updated them with a new mirror. 

Also fix `make fmt` condition to match all patch versions of Golang—instead of only matching 1.11.0 (`make fmt` would never run since we're already on 1.11.2).

Last but not least, add new version definitions and support for recently released Kafka 2.1.0.

@varun06 @muirrn could you please 👀this PR?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ script:
 - make test
 - make vet
 - make errcheck
-- if [ "$TRAVIS_GO_VERSION" = "1.11" ]; then make fmt; fi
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -28,7 +27,7 @@ script:
 - make test
 - make vet
 - make errcheck
-- if [ "$TRAVIS_GO_VERSION" = "1.11" ]; then make fmt; fi
+- if [[ "$TRAVIS_GO_VERSION" == 1.11* ]]; then make fmt; fi
```
varun06: 👍

PR #1230, opened 2018-11-29
Implement DescribeConfigs Request + Response v1 & v2

Mongey:
This is needed to determine if a Topic Config is actually a Default in Kafka 1.0 and above. The Default value in the DescribeConfigResponse should really be deprecated, if < 1.0 is no longer supported.

Not quite sure what the new struct fields in DescribeConfigResponse should include `Config` prefixes or not. (`ConfigSource` `ConfigSynonms` etc)

### DIFF
--- a/describe_configs_request.go
+++ b/describe_configs_request.go
@@ -1,13 +1,15 @@
 package sarama
 
+type DescribeConfigsRequest struct {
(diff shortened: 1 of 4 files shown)


bai:
Many thanks.

PR #1232, opened 2018-12-07
Unregister all broker metrics on broker stop

### LINKED ISSUE #1219: memory leak when using go-metrics
https://github.com/Shopify/sarama/blob/cebb5847efade319a00e1888acfd8bcdb393e478/broker.go#L108

b.incomingByteRate is inited by NewMeter, b.incomingByteRate.Stop() should be called when Broker was closed.

jsoriano:
Add methods for registering broker-specific metrics so it is possible to
keep track of them and they can be unregistered on broker stop.

Fixes #1219

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -30,6 +30,8 @@ type Broker struct {
 	responses chan responsePromise
 	done      chan bool
 
+	registeredMetrics []string
+
 	incomingByteRate       metrics.Meter
 	requestRate            metrics.Meter
 	requestSize            metrics.Histogram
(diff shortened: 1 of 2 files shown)


jsoriano:
I think there can still be another leak with topic metrics, as they are never unregistered, I'll take a look to this as a follow up.

mflpopescu:
@bai Checks appear to be passing, is it possible to approve this PR ?

varun06:
Can you please look into writing a test around it?

broker.go
```diff
@@ -180,12 +176,7 @@ func (b *Broker) Close() error {
 	b.done = nil
 	b.responses = nil
 
-	if b.id >= 0 {
-		b.conf.MetricRegistry.Unregister(getMetricNameForBroker("incoming-byte-rate", b))
-		b.conf.MetricRegistry.Unregister(getMetricNameForBroker("request-rate", b))
-		b.conf.MetricRegistry.Unregister(getMetricNameForBroker("outgoing-byte-rate", b))
-		b.conf.MetricRegistry.Unregister(getMetricNameForBroker("response-rate", b))
-	}
+	b.unregisterMetrics()
```
sam-obeid: I am not sure you need to maintain the list of registered metrics here, could you replace this method with:

```suggestion
	b.conf.MetricRegistry.UnregisterAll()
```

The MetricRegistry instance of scoped to a single broker instance, so once `broker.Close()` is called, we can Unregister all metrics attached to that broker.

jsoriano: Wouldn't `UnregisterAll()` also unregister the global metrics shared between brokers?

bai:
Thank you!

PR #1236, opened 2018-12-12
Implements v1 of {Create,Describe,Delete}AclRequest

Mongey:


acl_create_request.go
```diff
@@ -1,6 +1,7 @@
 package sarama
 
 type CreateAclsRequest struct {
+	Version      int
```
varun06: Does `Version` has to be int, if we are casting it to/fro int16?

acl_types.go
```diff
@@ -40,3 +40,15 @@ const (
 	AclResourceCluster         AclResourceType = 4
 	AclResourceTransactionalID AclResourceType = 5
 )
+
+type AclResourcePatternType int
+
+// ref: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/resource/PatternType.java
+
+const (
```
varun06: another way(idiomatic) to do it.

```suggestion
const (
	AclPatternUnknown  AclResourcePatternType = iota
	AclPatternAny      
	AclPatternMatch    
	AclPatternLiteral  
	AclPatternPrefixed 
	)
```

Mongey: TIL, thanks

bai:
Thanks @Mongey and @varun06 — as always 🙌

symaras:
It seems that the ClusterAdmin interface is not updated to cover the prefixed resources. The ListAcls method does not take a version argument. Thus, it creates a DescribeAclsRequest struct with no version defined and this defaults to V0_11_0_0. 

Or do I miss something here? Let me know, I am happy to submit a PR for this.

thanks!

symaras:
Opened an issue: https://github.com/Shopify/sarama/issues/1451

PR #1239, opened 2018-12-20
Add error codes 73-76

greysond:
Even more error codes!


bai:
Thanks! Could you please sign CLA?

Issue #1223, opened 2018-11-26
SASL/OAUTHBEARER authentication for Kafka 2.0


ggrcha:
**Versions**
Sarama Version: cebb5847efade319a00e1888acfd8bcdb393e478
Kafka Version: 2.0
Go Version: v1.11.1

**Problem Description**

As said by @mkmkaminski1988 on confluent-kafka-go#247, it would be important to have the OAUTHBEARER mechanism for authentication on Kafka brokers. Today, only Java clients can take advantage of this functionality, as edenhill/librdkafka still doesn't have SASL/OAUTHBEARER.

I'd like to discuss the design of the implementation before forking the project.

Thanks!

(cc @edenhill / @jairsjunior)

ggrcha:
Having some issues on this implementation:

- I successfully authenticate via OAUTH and the kafka broker finishes de SASL process (DEBUG Set SASL server state to COMPLETE)

- After that, I have an EOF exception and I can't figure it out what is missing. 

While testing de SASL Plain authentication, did anyone got some similar issue?

Thanks.

ggrcha:
@mkaminski1988 yes I did!!! Great job! It works like a charm :)

Closing the isue.

PR #1240, opened 2018-12-23
Implement SASL/OAUTHBEARER support

mkaminski1988:
Hello,

This commit implements SASL/OAUTHBEARER client authentication as described in [KIP-255](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876) and issue #1223.

To get started, the user must set `c.Net.SASL.Mechanism:  OAUTHBEARER` and assign a user-defined token generator to `config.Net.SASL.TokenProvider`. For backwards compatibility, a blank value for `c.Net.SASL.Mechanism` defaults to `PLAIN`.

Example:

```go
type TokenProvider struct {
}

func (t *TokenProvider) Token() (*sarama.AccessToken, error) {
	return &sarama.AccessToken{Token: "access-token-string"}, nil
}
// ...
config.Net.SASL.Enable = true
config.Net.SASL.Mechanism = sarama.SASLTypeOAuth
config.Net.SASL.TokenProvider = &TokenProvider{}
```

cc @ggrcha

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -159,14 +159,7 @@ func (b *Broker) Open(conf *Config) error {
 
 		if conf.Net.SASL.Enable {
 
-			switch conf.Net.SASL.Mechanism {
-			case SASLTypeOAuth:
-				b.connErr = b.sendAndReceiveSASLOAuth(conf.Net.SASL.TokenProvider, conf.Net.SASL.Extensions)
-			case SASLTypePlaintext:
-				b.connErr = b.sendAndReceiveSASLPlainAuth()
-			default:
-				b.connErr = b.sendAndReceiveSASLPlainAuth()
-			}
+			b.connErr = b.authenticateViaSASL()
 
 			if b.connErr != nil {
 				err = b.conn.Close()
@@ -784,6 +777,17 @@ func (b *Broker) responseReceiver() {
 		response.packets <- buf
 	}
 	close(b.done)
+}
+
+func (b *Broker) authenticateViaSASL() error {
+	switch b.conf.Net.SASL.Mechanism {
+	case SASLTypeOAuth:
+		return b.sendAndReceiveSASLOAuth(b.conf.Net.SASL.TokenProvider, b.conf.Net.SASL.Extensions)
+	case SASLTypePlaintext:
+		return b.sendAndReceiveSASLPlainAuth()
+	default:
+		return b.sendAndReceiveSASLPlainAuth()
+	}
 }
 
 func (b *Broker) sendAndReceiveSASLHandshake(saslType string, version int16) error {
--- a/broker.go
+++ b/broker.go
@@ -6,7 +6,9 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"sort"
 	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -61,6 +63,9 @@ const (
 	// SASLHandshakeV1 is v1 of the Kafka SASL handshake protocol. Client and
 	// server negotiate SASL by wrapping tokens with Kafka protocol headers.
 	SASLHandshakeV1 = int16(1)
+	// SASLExtKeyAuth is the reserved extension key name sent as part of the
+	// SASL/OAUTHBEARER intial client response
+	SASLExtKeyAuth = "auth"
 )
 
 // AccessTokenProvider is the interface that encapsulates how implementors
@@ -156,7 +161,7 @@ func (b *Broker) Open(conf *Config) error {
 
 			switch conf.Net.SASL.Mechanism {
 			case SASLTypeOAuth:
-				b.connErr = b.sendAndReceiveSASLOAuth(conf.Net.SASL.TokenProvider)
+				b.connErr = b.sendAndReceiveSASLOAuth(conf.Net.SASL.TokenProvider, conf.Net.SASL.Extensions)
 			case SASLTypePlaintext:
 				b.connErr = b.sendAndReceiveSASLPlainAuth()
 			default:
@@ -893,7 +898,7 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
 
 // sendAndReceiveSASLOAuth performs the authentication flow as described by KIP-255
 // https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876
-func (b *Broker) sendAndReceiveSASLOAuth(tokenProvider AccessTokenProvider) error {
+func (b *Broker) sendAndReceiveSASLOAuth(tokenProvider AccessTokenProvider, extensions map[string]string) error {
 
 	if err := b.sendAndReceiveSASLHandshake(SASLTypeOAuth, SASLHandshakeV1); err != nil {
 		return err
@@ -909,7 +914,7 @@ func (b *Broker) sendAndReceiveSASLOAuth(tokenProvider AccessTokenProvider) erro
 
 	correlationID := b.correlationID
 
-	bytesWritten, err := b.sendSASLOAuthBearerClientResponse(token, correlationID)
+	bytesWritten, err := b.sendSASLOAuthBearerClientResponse(token, extensions, correlationID)
 
 	if err != nil {
 		return err
@@ -931,13 +936,45 @@ func (b *Broker) sendAndReceiveSASLOAuth(tokenProvider AccessTokenProvider) erro
 	return nil
 }
 
-func (b *Broker) sendSASLOAuthBearerClientResponse(bearerToken string, correlationID int32) (int, error) {
+// Build SASL/OAUTHBEARER initial client response as described by RFC-7628
+// https://tools.ietf.org/html/rfc7628
+func buildClientInitialResponse(bearerToken string, extensions map[string]string) ([]byte, error) {
 
-	// Initial client response as described by RFC-7628.
-	// https://tools.ietf.org/html/rfc7628
-	oauthRequest := []byte(fmt.Sprintf("n,,\x01auth=Bearer %s\x01\x01", bearerToken))
+	if _, ok := extensions[SASLExtKeyAuth]; ok {
+		return []byte{}, fmt.Errorf("The extension `%s` is invalid", SASLExtKeyAuth)
+	}
 
-	rb := &SaslAuthenticateRequest{oauthRequest}
+	extensions[SASLExtKeyAuth] = "Bearer " + bearerToken
+
+	resp := []byte(fmt.Sprintf("n,,\x01%s\x01\x01", mapToString(extensions, "=", "\x01")))
+
+	return resp, nil
+}
+
+// mapToString returns a list of key-value pairs ordered by key.
+// keyValSep separates the key from the value. elemSep separates each pair.
+func mapToString(extensions map[string]string, keyValSep string, elemSep string) string {
+
+	buf := make([]string, 0, len(extensions))
+
+	for k, v := range extensions {
+		buf = append(buf, k+keyValSep+v)
+	}
+
+	sort.Strings(buf)
+
+	return strings.Join(buf, elemSep)
+}
+
+func (b *Broker) sendSASLOAuthBearerClientResponse(bearerToken string, extensions map[string]string, correlationID int32) (int, error) {
+
+	initialResp, err := buildClientInitialResponse(bearerToken, extensions)
+
+	if err != nil {
+		return 0, err
+	}
+
+	rb := &SaslAuthenticateRequest{initialResp}
 
 	req := &request{correlationID: correlationID, clientID: b.conf.ClientID, body: rb}
 
--- a/broker.go
+++ b/broker.go
@@ -11,6 +11,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/rcrowley/go-metrics"
 )
 
@@ -906,15 +907,21 @@ func (b *Broker) sendAndReceiveSASLOAuth(tokenProvider AccessTokenProvider) erro
 
 	requestTime := time.Now()
 
-	bytesWritten, err := b.sendSASLOAuthBearerClientResponse(token)
+	correlationID := b.correlationID
+
+	bytesWritten, err := b.sendSASLOAuthBearerClientResponse(token, correlationID)
 
 	if err != nil {
 		return err
 	}
 
+	Logger.Printf("Correlation ID %d ", b.correlationID)
+
 	b.updateOutgoingCommunicationMetrics(bytesWritten)
 
-	bytesRead, err := b.receiveSASLOAuthBearerServerResponse()
+	b.correlationID++
+
+	bytesRead, err := b.receiveSASLOAuthBearerServerResponse(correlationID)
 
 	if err != nil {
 		return err
@@ -926,7 +933,7 @@ func (b *Broker) sendAndReceiveSASLOAuth(tokenProvider AccessTokenProvider) erro
 	return nil
 }
 
-func (b *Broker) sendSASLOAuthBearerClientResponse(bearerToken string) (int, error) {
+func (b *Broker) sendSASLOAuthBearerClientResponse(bearerToken string, correlationID int32) (int, error) {
 
 	// Initial client response as described by RFC-7628
 	// https://tools.ietf.org/html/rfc7628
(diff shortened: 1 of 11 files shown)


broker.go
```diff
@@ -853,6 +881,124 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
...
+// sendAndReceiveSASLOAuth performs the authentication flow as described by KIP-255
+// https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=75968876
+func (b *Broker) sendAndReceiveSASLOAuth(tokenProvider OAuthBearerTokenProvider) error {
+
+	// This version allows us to wrap tokens in the Kafka protocol, as opposed
+	// to sending opaque packets
+	handshakeVersion := int16(1)
+
+	if err := b.sendAndReceiveSASLHandshake(SaslTypeOAuth, handshakeVersion); err != nil {
+		Logger.Printf("Error while performing SASL handshake %s\n", b.addr)
```
jprobinson: I'm not 100% sure of the standards of this project, but a nice rule for error handling in Go is either return the error or log it but don't do both.

Here's a nice blog post for some info on how you can handle these scenarios: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

broker.go
```diff
@@ -853,6 +881,124 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
...
+	b.correlationID++
+
+	return bytesWritten, nil
+}
+
+func (b *Broker) receiveSASLOAuthBearerServerResponse() (int, error) {
+
+	var bytesRead int
+	var err error
+	totalBytesRead := 0
```
jprobinson: You should be fine using just the 0 value of the int type, no need to declare a `0`:

```go
var totalByesRead int
```

Also, you could group all the variables here together in a single `var`  block:

```go
var (
 err error
 bytesRead int
 totalBytesRead int
)
```

varun06: I was also going to suggest grouping variables/types in a block, but then it is very personal style thing :) I however like it a lot.

broker.go
```diff
@@ -853,6 +881,124 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
...
+
+	header := make([]byte, 8)
+
+	if bytesRead, err = io.ReadFull(b.conn, header); err != nil {
+		Logger.Printf("Failed to read SASL/OAUTHBEARER intitial response header : %s\n", err.Error())
```
varun06: as suggested above, don't log if you are returning error, these errors have potential of filling the disks if not handled properly downstream.

sasl_authenticate_request.go
```diff
@@ -0,0 +1,33 @@
...
+func (r *SaslAuthenticateRequest) decode(pd packetDecoder, version int16) (err error) {
+	if r.SaslAuthBytes, err = pd.getBytes(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *SaslAuthenticateRequest) key() int16 {
+	return 36
```
varun06: `36` is a magic number, should be a named constant.

config.go
```diff
@@ -54,13 +54,19 @@ type Config struct {
 			// Whether or not to use SASL authentication when connecting to the broker
 			// (defaults to false).
 			Enable bool
+			// The type of SASL mechanism to enable. Possible values: OAUTHBEARER, PLAIN (defaults to PLAIN)
```
jprobinson: Comments should begin with the name of the field: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

config.go
```diff
@@ -454,10 +460,28 @@ func (c *Config) Validate() error {
 		return ConfigurationError("Net.WriteTimeout must be > 0")
 	case c.Net.KeepAlive < 0:
 		return ConfigurationError("Net.KeepAlive must be >= 0")
-	case c.Net.SASL.Enable == true && c.Net.SASL.User == "":
-		return ConfigurationError("Net.SASL.User must not be empty when SASL is enabled")
-	case c.Net.SASL.Enable == true && c.Net.SASL.Password == "":
-		return ConfigurationError("Net.SASL.Password must not be empty when SASL is enabled")
+	case c.Net.SASL.Enable == true:
```
jprobinson: No need for the `== true` here, `c.Net.SASL.Enable` is a bool so the case statement will honor it.

sasl_authenticate_request.go
mkaminski1988: I modeled this after the way the other API structs are defined. Shall I put the constant here or is there a central place where they should go? I'll also need to do this for the response struct.

varun06: package level constant should be fine for this.

broker.go
mkaminski1988: Makes sense. How about further down where the authentication failure is logged (`Invalid SASL/OAUTHBEARER request`)? This would help users figure out why their client doesn't work when they pass invalid credentials.

broker.go
```diff
@@ -853,6 +881,124 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
...
+	b.updateIncomingCommunicationMetrics(bytesRead, requestLatency)
+
+	return nil
+}
+
+func (b *Broker) sendSASLOAuthBearerClientResponse(bearerToken []byte) (int, error) {
+
+	// Initial client response as described by RFC-7628
+	// https://tools.ietf.org/html/rfc7628
+	oauthRequest := []byte(`n,,`)
```
varun06: `fmt.Sprintf("n,,\x01auth=Bearer %s\x01\x01", bearerToken)` is a simple and readable version can be bit slow though.

broker.go
mkaminski1988: I could make the caller log the auth message. Not sure if that's a disk space issue if the error is only emitted when debug mode is turned on?

broker.go
```diff
@@ -46,6 +46,24 @@ type Broker struct {
...
+	// SaslTypePlaintext represents the SASL/PLAIN mechanism
+	SaslTypePlaintext = "PLAIN"
+)
+
+// OAuthBearerTokenProvider is an interface that encapsualtes bearer token creation in an
+// unopinionated way. Users are free to decide how to construct bearer tokens and how often
+// they are be refreshed.
+type OAuthBearerTokenProvider interface {
+	// Return a valid bearer token.
+	Token() []byte
```
jprobinson: I'd recommend adding some more control over this definition by making it:

```go
Token(ctx context.Context) ([]byte, error)
```

The context will allow more control over timeouts and cancellation while fetching a token. The error also will allow implementors to alert the system if there were problems fetching/creating a new token. This is similar to [oauth2.TokenSource](https://godoc.org/golang.org/x/oauth2#TokenSource)

broker.go
```diff
@@ -853,6 +881,124 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
...
+		Logger.Printf("Error while performing SASL handshake %s\n", b.addr)
+		return err
+	}
+
+	requestTime := time.Now()
+
+	var bytesWritten int
+	var err error
+
+	if bytesWritten, err = b.sendSASLOAuthBearerClientResponse(tokenProvider.Token()); err != nil {
```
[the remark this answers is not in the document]
mkaminski1988: Then bytesWritten is not available to `updateOutgoingCommunicationMetrics()` unless the call is in an `else` block.

```
	if bytesWritten, err := b.sendSASLOAuthBearerClientResponse(tokenProvider.Token()); err != nil {
		return err
	} else {
		b.updateOutgoingCommunicationMetrics(bytesWritten)
	}
```

Do you prefer that style?

broker.go
mkaminski1988: Changed the interface to return an error. Unless someone feels strongly about this, I am going to hold off on the `context.Context` param for now because I think it will keep the configuration simpler (passing a user-configured context object to a user-defined callback seems redundant to me).

request.go
```diff
@@ -140,6 +140,8 @@ func allocateBody(key, version int16) protocolBody {
 		return &DescribeConfigsRequest{}
 	case 33:
 		return &AlterConfigsRequest{}
+	case 36:
```
mkaminski1988: @varun06 should we use the `APIKeySASLAuth` const defined in `SaslAuthenticateRequest` here? Seems like the Right Thing To Do, but also breaks with the style of the rest of the case statements.

broker.go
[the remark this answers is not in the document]
mkaminski1988: Nevermind, I misread this like twice!

request.go
varun06: It does, but for now we can keep it this way for the sake of consistency, and come back later.

varun06:
@mkaminski1988 PR is looking great now. But you might have to add some more tests for codecoverage to pass.

mkaminski1988:
Cool, tests in progress. I also mentioned this PR to @rondagostino, who implemented SASL/OAUTHBEARER in Kafka. He said he'll take a look once he gets back from vacation after the new year.

varun06:
Makes sense.

broker.go
```diff
@@ -853,6 +891,122 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
...
+	b.updateIncomingCommunicationMetrics(bytesRead, requestLatency)
+
+	return nil
+}
+
+func (b *Broker) sendSASLOAuthBearerClientResponse(bearerToken string, correlationID int32) (int, error) {
+
+	// Initial client response as described by RFC-7628
+	// https://tools.ietf.org/html/rfc7628
+	oauthRequest := []byte(fmt.Sprintf("n,,\x01auth=Bearer %s\x01\x01", bearerToken))
```
rondagostino: This is where SASL Extensions would be included as per https://tools.ietf.org/html/rfc7628#section-3.1 (it supports arbitrary `key=value\x01` pairs beyond the standard `auth=Bearer ...\x01` pair).

rondagostino:
@mkaminski1988 Hey Mike, looks good.  There are 2 things that come to mind.  One is support for SASL extensions (arbitrary name/value pairs that can be sent with the client response as per https://tools.ietf.org/html/rfc7628#section-3.1).  It looks like the code currently does not support such extensions.  I had originally left out that functionality from the original Java implementation, and it was added later.  I don't know if the same approach would be acceptable here.

The second thing that comes to mind is the lack of token reuse/refresh.  `AccessTokenProvider` is documented as "each invocation returns a new, unexpired token" which will work, but that means multiple separate connections to different brokers, one after the other in a short period of time, will result in multiple separate tokens being retrieved.  The Java client works differently -- it keeps a single "current" token and reuses it, and there is a background refresh thread that periodically replaces it before it expires.  Again, as with SASL Extensions above, I'm not sure what is acceptable here in an initial implementation.

broker.go
```diff
@@ -46,6 +47,33 @@ type Broker struct {
...
+	// SASLHandshakeV1 is v1 of the Kafka SASL handshake protocol. Client and
+	// server negotiate SASL by wrapping tokens with Kafka protocol headers.
+	SASLHandshakeV1 = int16(1)
+)
+
+// AccessTokenProvider is the interface that encapsulates how implementors
+// can generate access tokens for Kafka broker authentication.
+type AccessTokenProvider interface {
+	// Token returns an access token. Because this method may be called multiple
+	// times, each invocation returns a new, unexpired token. This method should
```
rondagostino: Token reuse/refresh?

mkaminski1988: My plan for this implementation is to make the user responsible for implementing the token reuse/refresh, which can be accomplished by using an OAuth client with built-in token refresh, such as [Google's OAuth2 client](https://github.com/golang/oauth2/blob/master/oauth2.go#L345).

I'm going to modify the contract to more accurately capture reuse/refresh requirements as you described.

I think it should be possible to support protocol-level token refresh by augmenting the configuration with a token expiration interval parameter that will instruct when the client should reauthenticate with the broker.

mkaminski1988:
@rondagostino I've implemented extensions. Couple of questions:

1) Is it necessary for the client to send an `authzid` header? I see that it's application-dependent according to RFC-7628. If I'm understanding the Kafka Java SASL auth code correctly, the broker _can_ deserialize the authorization ID _if_ the client sends one. The Java Kafka client, however, [doesn't appear to send one](https://github.com/apache/kafka/blob/2.1.0/clients/src/main/java/org/apache/kafka/common/security/oauthbearer/internals/OAuthBearerSaslClient.java#L97). I'm assuming that was done to keep the Kafak SASL server compliant with RFC-7628?

2) In this implementation, the SASL extensions are assigned to the client configuration and remain static for the duration of the client. This seems to mirror the behavior of the Java client, where the extensions are set in the JAAS configuration. Am I correct in my assumption that these extension parameters don't change after initial configuration?

rondagostino:
@mkaminski1988 Yes, you are correct about the `authzid` -- it is optional as per the spec, and the Java client currently does not send one.  [The broker does check for it in case it is there, and it ensures that any `authzid` matches the principal name as defined by the token](https://github.com/apache/kafka/blob/2.1.0/clients/src/main/java/org/apache/kafka/common/security/oauthbearer/internals/OAuthBearerSaslServer.java#L170).

With regard to any extensions that the Java client sends to the broker, it is implementation-dependent whether they are constant across token retrievals.  The [login callback handler is tasked with retrieving the extensions](https://github.com/apache/kafka/blob/2.1.0/clients/src/main/java/org/apache/kafka/common/security/oauthbearer/OAuthBearerLoginModule.java#L336), and the default is for there to be [no extensions if the login callback handler doesn't know how to handle `SaslExtensionsCallback`](https://github.com/apache/kafka/blob/2.1.0/clients/src/main/java/org/apache/kafka/common/security/oauthbearer/OAuthBearerLoginModule.java#L342).  The [unsecured login callback handler does look for extensions in the JAAS config](https://github.com/apache/kafka/blob/2.1.0/clients/src/main/java/org/apache/kafka/common/security/oauthbearer/internals/unsecured/OAuthBearerUnsecuredLoginCallbackHandler.java#L231) as you point out, but that of course can't be used in production, so it falls to the production login callback handler implementation that is plugged in to decide what (if anything) to do with `SaslExtensionsCallback` when it sees it (and it can throw `UnsupportedCallbackException` since that will be caught and empty extensions will be injected by default).

mkaminski1988:
Upon further investigation I've decided to make the `TokenProvider .Token()` method return a struct that contains both the access token and SASL extensions instead of a string, which gives flexibility for future changes.

mkaminski1988:
@varun06 This is ready to merge.

varun06:
🕺 

This is great @mkaminski1988 . I am going to have a look and may be @bai or @jprobinson can also have one more look. This is a big change and thanks a lot of all the work.

broker.go
```diff
@@ -853,6 +909,159 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
...
+	b.updateIncomingCommunicationMetrics(bytesRead, requestLatency)
+
+	return nil
+}
+
+// Build SASL/OAUTHBEARER initial client response as described by RFC-7628
+// https://tools.ietf.org/html/rfc7628
+func buildClientInitialResponse(token *AccessToken) ([]byte, error) {
+
+	ext := ""
```
jprobinson: Tiny style bit, totes optional. You can also just rely on the default value of the string here and just declare it as `var ext string`

broker.go
```diff
@@ -853,6 +909,159 @@ func (b *Broker) sendAndReceiveSASLPlainAuth() error {
...
+
+	if err != nil {
+		return 0, err
+	}
+
+	if err := b.conn.SetWriteDeadline(time.Now().Add(b.conf.Net.WriteTimeout)); err != nil {
+		return 0, err
+	}
+
+	bytesWritten, err := b.conn.Write(buf)
```
jprobinson: More small style things; You can just return here, no need to declare the variables and return them.

mkaminski1988:
Thanks for your expertise @jprobinson!

broker_test.go
```diff
@@ -105,6 +110,197 @@ func TestSimpleBrokerCommunication(t *testing.T) {
...
+					"auth": "auth-value",
+				},
+			},
+			[]byte(""),
+			true,
+		},
+	}
+
+	for i, test := range testTable {
+
```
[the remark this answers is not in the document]
mkaminski1988: Good to know about `t.Run()`. I think I'll match the style of the project, unless the maintainers would like it otherwise.

bsvingen:
This is great, would very much like to see this merged.

varun06:
@bai What is the strategy to get this merged. It's a big change and might need a release?

bai:
Thanks for getting this done, amazing work! I'm going to merge this one now and will compile changelog and do a release this week.

mkaminski1988:
My pleasure, thanks for your help everyone!

bai:
Sorry for the noise here, just to clarify: 1.20.1 with deadlock fix is coming up in a moment, this one will go into 1.21.

mkaminski1988:
@bai Do you have an idea of when the next release will be cut?

bai:
I intend to cut it tomorrow or Friday latest.

Issue #1252, opened 2019-01-11
Sending messages with ZStd compression enabled fails in multiple ways


akrennmair:
##### Versions

Sarama Version: v1.20.1
Kafka Version: 2.1.0 (Scala version 2.12)
Go Version: 1.11.4

##### Configuration

What configuration values are you using for Sarama and Kafka?

Kafka configuration:
[server.properties.txt](https://github.com/Shopify/sarama/files/2750161/server.properties.txt)

##### Logs

```
=== RUN   TestSaramaZSTD
SARAMA: Initializing new client
SARAMA: client/metadata fetching metadata for all topics from broker localhost:9092
SARAMA: Connected to broker at localhost:9092 (unregistered)
SARAMA: client/brokers registered new broker #1 at localhost:9092
SARAMA: Successfully initialized new client
SARAMA: producer/broker/1 starting up
SARAMA: producer/broker/1 state change to [open] on sarama-test/0
SARAMA: Connected to broker at localhost:9092 (registered as #1)
SARAMA: client/metadata fetching metadata for [sarama-test] from broker localhost:9092
SARAMA: producer/broker/1 state change to [retrying] on sarama-test/0 because kafka server: Message contents does not match its CRC.
SARAMA: Retrying batch for sarama-test-0 because of kafka server: Message contents does not match its CRC.
TEST:   sending message failed: kafka server: Message contents does not match its CRC.
SARAMA: Producer shutting down.
SARAMA: Closing Client
SARAMA: producer/broker/1 shut down
SARAMA: Closed connection to broker localhost:9092
SARAMA: Initializing new client
SARAMA: client/metadata fetching metadata for all topics from broker localhost:9092
SARAMA: Closed connection to broker localhost:9092
SARAMA: Connected to broker at localhost:9092 (unregistered)
SARAMA: client/brokers registered new broker #1 at localhost:9092
SARAMA: Successfully initialized new client
SARAMA: producer/broker/1 starting up
SARAMA: producer/broker/1 state change to [open] on sarama-test/0
SARAMA: Connected to broker at localhost:9092 (registered as #1)
SARAMA: producer/broker/1 state change to [closing] because EOF
SARAMA: Closed connection to broker localhost:9092
TEST:   sending message failed: EOF
SARAMA: Producer shutting down.
SARAMA: Closing Client
SARAMA: Closed connection to broker localhost:9092
SARAMA: producer/broker/1 shut down
--- PASS: TestSaramaZSTD (0.01s)
[...]

PR #1256, opened 2019-01-16
adding version for zstd compression

varun06:
This is a fix for issues mentioned in #1252 

- naming the import is done by one of the static analysis tool on my local, if that's a problem, I will revert it back.

### DIFF
--- a/produce_request.go
+++ b/produce_request.go
@@ -1,6 +1,6 @@
 package sarama
 
-import "github.com/rcrowley/go-metrics"
+import metrics "github.com/rcrowley/go-metrics"
 
 // RequiredAcks is used in Produce Requests to tell the broker how many replica acknowledgements
 // it must see before responding. Any of the constants defined here are valid. On broker versions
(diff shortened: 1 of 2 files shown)


varun06:
@bobrik does it look okay?

produce_request.go
```diff
@@ -214,6 +212,8 @@ func (r *ProduceRequest) requiredVersion() KafkaVersion {
 		return V0_10_0_0
 	case 3:
 		return V0_11_0_0
+	case 4:
```
bobrik: Shouldn't this be `7`? See [Kafka source](https://github.com/apache/kafka/blob/2c44e77e2f2056ee88b4b9cc8bac1f769748830a/clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java#L118-L121).

I think you also need to update [this code](https://github.com/Shopify/sarama/blob/4c93f1f002740d2d8604ba4f82d2c87a4d49a1ba/produce_set.go#L124-L129) to start encoding with that version.

Ultimately you want to test this against a live v2.1.0 Kafka and see if that works.

varun06:
Will bring it back once I have full understanding.

PR #1258, opened 2019-01-16
Support LogAppend timestamps

zimin2000:
Currently sarama doesn't seem to properly support LogAppend type of timestamps. The patch is addressing this issue. The following was used is the reference for the change:
   * https://github.com/dpkp/kafka-python/blob/34fcb11c1cf96d69573274104d3b746ce67a97f4/kafka/record/default_records.py#L204 and 
   * https://github.com/dpkp/kafka-python/blob/34fcb11c1cf96d69573274104d3b746ce67a97f4/kafka/record/legacy_records.py#L262

### DIFF
--- a/consumer_test.go
+++ b/consumer_test.go
@@ -985,6 +985,123 @@ func TestConsumerExpiryTicker(t *testing.T) {
 	broker0.Close()
 }
 
+func TestConsumerTimestamps(t *testing.T) {
+	now := time.Now().Truncate(time.Millisecond)
+	type testMessage struct {
+		key       Encoder
+		value     Encoder
+		offset    int64
+		timestamp time.Time
+	}
+	for _, d := range []struct {
+		kversion          KafkaVersion
+		logAppendTime     bool
+		messages          []testMessage
+		expectedTimestamp []time.Time
+	}{
+		{MinVersion, false, []testMessage{
+			{nil, testMsg, 1, now},
+			{nil, testMsg, 2, now},
+		}, []time.Time{{}, {}}},
+		{V0_9_0_0, false, []testMessage{
+			{nil, testMsg, 1, now},
+			{nil, testMsg, 2, now},
+		}, []time.Time{{}, {}}},
+		{V0_10_0_0, false, []testMessage{
+			{nil, testMsg, 1, now},
+			{nil, testMsg, 2, now},
+		}, []time.Time{{}, {}}},
+		{V0_10_2_1, false, []testMessage{
+			{nil, testMsg, 1, now.Add(time.Second)},
+			{nil, testMsg, 2, now.Add(2 * time.Second)},
+		}, []time.Time{now.Add(time.Second), now.Add(2 * time.Second)}},
+		{V0_10_2_1, true, []testMessage{
+			{nil, testMsg, 1, now.Add(time.Second)},
+			{nil, testMsg, 2, now.Add(2 * time.Second)},
+		}, []time.Time{now, now}},
+		{V0_11_0_0, false, []testMessage{
+			{nil, testMsg, 1, now.Add(time.Second)},
+			{nil, testMsg, 2, now.Add(2 * time.Second)},
+		}, []time.Time{now.Add(time.Second), now.Add(2 * time.Second)}},
+		{V0_11_0_0, true, []testMessage{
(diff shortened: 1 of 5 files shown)


message.go
```diff
@@ -19,6 +19,8 @@ const (
 	CompressionZSTD   CompressionCodec = 4
 )
 
+const timestampTypeMask = 0x08
+
```
mikejquinn: Move this below the `CompressionCodec.String` function and `CompressionLevelDefault` const.

message.go
```diff
@@ -52,6 +55,9 @@ func (m *Message) encode(pe packetEncoder) error {
 	pe.putInt8(m.Version)
 
 	attributes := int8(m.Codec) & compressionCodecMask
+	if m.LogAppendTime {
+		attributes |= timestampTypeMask
+	}
```
mikejquinn: Does this make sense to be set by a producer? I thought this was a server config setting.

varun06:
Looks okay to me, I wonder who else can review it?

consumer_test.go
```diff
@@ -985,6 +985,108 @@ func TestConsumerExpiryTicker(t *testing.T) {
...
+		}, []time.Time{now.Add(time.Second), now.Add(2 * time.Second)}},
+		{V0_10_2_1, true, []testMessage{
+			{nil, testMsg, 1, now.Add(time.Second)},
+			{nil, testMsg, 2, now.Add(2 * time.Second)},
+		}, []time.Time{now, now}},
+		{V0_11_0_0, false, []testMessage{
+			{nil, testMsg, 1, now.Add(time.Second)},
+			{nil, testMsg, 2, now.Add(2 * time.Second)},
+		}, []time.Time{now.Add(time.Second), now.Add(2 * time.Second)}},
+		{V0_11_0_0, true, []testMessage{
```
sam-obeid: Should you include other supported versions here?

https://github.com/bai/sarama/blob/master/utils.go#L163

zimin2000: I have adjusted the test to cover all versions for which format of fetch is different (from https://github.com/bai/sarama/blob/0a21d90df4f6266fdf28d603e5ef91f2426c362a/fetch_request.go#L141)

zimin2000:
@sam-obeid , @varun06 , how else I can help with the PR to make it through ?

fetch_response.go
```diff
@@ -355,10 +357,13 @@ func encodeKV(key, value Encoder) ([]byte, []byte) {
 	return kb, vb
 }
 
-func (r *FetchResponse) AddMessage(topic string, partition int32, key, value Encoder, offset int64) {
+func (r *FetchResponse) AddMessageWithTimestamp(topic string, partition int32, key, value Encoder, offset int64, timestamp time.Time, version int8) {
```
varun06: This and below are exported function and this change is going to be a breaking change. Let's see how we can handle them. @bai do you have an idea?

varun06:
@zimin2000 there are couple breaking changes(exported function names are changed), we got bitten by last merge, so let's walk carefully now and find a holistic approach to handle these changes.

fetch_response.go
sam-obeid: You'll need to keep both exported functions, and make one delegates to the other!

zimin2000: @varun06 , I think I did preserve the original API call reimplementing the original function using this one (https://github.com/Shopify/sarama/pull/1258/files/e1eda41894f763094e70b74f6dbf958c4cbbd315#diff-be942b2617e64413fb251f24e4bd4ea2R388). Same for `AddRecord`.

Am I missing something ?

zimin2000: The “AddMessage” method is still there (check the link I provided above). Also note that there are multiple tests that would have got broken otherwise.

varun06: Apologies, Do we have AddRecord method also?

mikejquinn: It looks like @zimin2000 preserved the original `AddRecord` and `AddMessage` names & signatures below starting on line 388. Perhaps they should each be moved right above their respective `WithTimestamp` versions.

zimin2000: Yes, “AddRecord” also kept.

varun06: Cool Bean. Let's get it going then. @bai can we merge this please?

bai:
Thanks for contributing!

PR #1268, opened 2019-02-01
fix memory accumulation

tanmq:
Because input and output channel in one select scope, so if output is not quickly as input, the messages will accumulate in memory and result in message too large. So we send batch message for sending goroutine before continue receive next message from input channel.


bai:
Thanks for your contribution. Could you please rebase your PR off of latest master?

bai:
Also, could you please sign CLA?

varun06:
Build is failing, can you please look what's up?

PR #1271, opened 2019-02-07
Add NextOffset for consumer group

Danm87:
Add the ability to know what is the next offset for the specific partition, without needed to use lower level or create a redundant PartitionOffsetManager

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -11,6 +11,7 @@ import (
 
 // ErrClosedConsumerGroup is the error returned when a method is called on a consumer group that has been closed.
 var ErrClosedConsumerGroup = errors.New("kafka: tried to use a consumer group that was closed")
+var ErrPartitionNotClaimed = errors.New("that topic/partition is not claimed by this consumer group session")
 
 // ConsumerGroup is responsible for dividing up processing of topics and partitions
 // over a collection of processes (the members of the consumer group).
@@ -433,6 +434,13 @@ type ConsumerGroupSession interface {
 	// allows incrementing the offset. cf MarkOffset for more details.
 	ResetOffset(topic string, partition int32, offset int64, metadata string)
 
+	// NextOffset returns the next offset that should be consumed for the managed
+	// partition, accompanied by metadata which can be used to reconstruct the state
+	// of the partition consumer when it resumes. NextOffset() will return
+	// `config.Consumer.Offsets.Initial` and an empty metadata string if no offset
+	// was committed for this partition yet.
+	NextOffset(topic string, partition int32) (int64, string, error)
+
 	// MarkMessage marks a message as consumed.
 	MarkMessage(msg *ConsumerMessage, metadata string)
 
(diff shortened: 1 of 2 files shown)


varun06:
that's nice. adding test should be easy for this change.

Danm87:
@varun06 thanks, added a test.

consumer_group_test.go
```diff
@@ -3,8 +3,50 @@ package sarama
 import (
 	"context"
 	"fmt"
+	"testing"
 )
 
+func TestConsumerGroupSessionNextOffset(t *testing.T) {
+	var partition int32 = 13
```
varun06: I like using variable blocks like this, but it's a personal preference.
```suggestion
	var (
	    partition int32 = 13
	    offset int64 = 23
	    meta = "meta"
```

varun06:
@dim can you also have a look?

consumer_group.go
```diff
@@ -433,6 +434,13 @@ type ConsumerGroupSession interface {
 	// allows incrementing the offset. cf MarkOffset for more details.
 	ResetOffset(topic string, partition int32, offset int64, metadata string)
 
+	// NextOffset returns the next offset that should be consumed for the managed
+	// partition, accompanied by metadata which can be used to reconstruct the state
+	// of the partition consumer when it resumes. NextOffset() will return
+	// `config.Consumer.Offsets.Initial` and an empty metadata string if no offset
+	// was committed for this partition yet.
+	NextOffset(topic string, partition int32) (int64, string, error)
```
dim: I don't really see how this is useful.

consumer_group.go
```diff
@@ -543,6 +551,15 @@ func (s *consumerGroupSession) ResetOffset(topic string, partition int32, offset
 	}
 }
 
+func (s *consumerGroupSession) NextOffset(topic string, partition int32) (int64, string, error) {
+	if pom := s.offsets.findPOM(topic, partition); pom != nil {
```
dim: instead of retrieving this information, wouldn't make it more sense to capture the initial offsets at the beginning of the session, see line https://github.com/Shopify/sarama/pull/1271/files#diff-9076b00c0e67e4484d0c04fb247353c7L567

sam-obeid:
In most cases ConsumerGroup is considered a high-level API where you don't need to handle offsets manually.  I am curious, what's your use-case @Danm87 for this new method?

consumer_group.go
danpmx: well, what I'm essentially doing is trying to create an exactly once semantics for uploading  kafka messages to a bucket (inspired heavily by [secor](https://github.com/pinterest/secor)).
In order to do so i'm keeping track of the current offsets, and in case of rebalance, if the same partition is not being managed by the same client, i'm deleting the messages locally.
Otherwise, in order to minimise processing the same message twice or more, I continue from the same location 
In order to do so, I really benefit from knowing which offset in the next one for the current partition

consumer_group.go
danpmx: @dim thanks for the review,
correct me if I'm wrong, this means we need to keep the state for each and every managed partition, which is already kept by the pom

danpmx:
well, what I'm essentially doing is trying to create an exactly once semantics for uploading kafka messages to a bucket (inspired heavily by [secor](https://github.com/pinterest/secor)).
In order to do so, i'm keeping track of the current offsets, and in case of rebalance, if the same partition is not being managed by the same client, i'm deleting the messages locally.
Otherwise, in order to minimise processing the same message twice or more, I continue from the same location
In order to do so, I really benefit from knowing which offset in the next one for the current partition

varun06:
What's the status on this PR?

PR #1273, opened 2019-02-07
Update kafka test version

ebati:
I updated kafka version in `provision.sh` as 1.0.0 is not distributed [here](https://www-us.apache.org/dist/kafka/). Also the suggested version of scala is different for different versions as noted [here](https://kafka.apache.org/downloads). So added new environment variable which is used to find the recommended scala version while downloading kafka.


ebati:
I signed it.

varun06:
You might have to close and reopen again. Once the checks are passed, we can merge it.

bai:
Thanks!

PR #1277, opened 2019-02-08
Set correct Kafka version for DescribeConfigsRequest v1

Mongey:
V1 is only supported on Kafka 1.1.0 and higher.

Difficult to find an official page, other than [this](https://cwiki.apache.org/confluence/display/KAFKA/Kafka+APIs),  where this is documented, but from a describe api versions

Kafka 1.0
```
Key:32, Min: 0, Max: 0
```

Kafka 1.1
```
Key:32, Min: 0, Max: 1
```


bai:
Thanks!

PR #1282, opened 2019-02-16
Go mod files

### LINKED ISSUE #1272: Go module support
CI tests with the latest available packages as it does `go get`. However, since those versions are not recorded, the versions of dependencies this library is tested against is unknown to users of this library.

Is it OK if I send PR for dependency management with go modules.

ebati:
I added `go.sum`&`go.mod` files and exported `GO111MODULES=on` in makefile. Since go1.10 cannot use `go.mod` in `go get` it will download the latest version but go1.11 and later will use `go.mod`. After `go1.12` release i can update `makefile` and travis config to use `go mod verify` etc.   

We can also move `errcheck` dependency to `tools.go` ( as detailed in [here](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) ) instead of separately installing it. Let me know if you prefer `tools.go` approach.

(Fixes  #1272)


varun06:
Go modules are fine and it seems that Sarama already uses semver for release versioning and tags. looks good to me 👍

bai:
Thanks for contributing @ebati 💯

PR #1286, opened 2019-02-18
fixed potential deadlock when a heartbeat request fails

aksdb:
When using a ConsumerGroup I managed to block all but one consumers after Kafka went down (but Zookeeper kept running).

Test setup:
* zookeeper + kafka
* 1 partition
* run a consumer multiple times, restarting as desired causing the group to be rebalanced (just in case)
* kill kafka - one consumer properly leaves the "Consume" method while the others keep waiting

Debugging revealed that the "retries" went into negative numbers, simply because the failed retries were not evaluated at that point in the code. My PR adds this missing check and therefore allows a clean shutdown on a failed kafka connection.

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -657,6 +657,12 @@ func (s *consumerGroupSession) heartbeatLoop() {
 		resp, err := s.parent.heartbeatRequest(coordinator, s.memberID, s.generationID)
 		if err != nil {
 			_ = coordinator.Close()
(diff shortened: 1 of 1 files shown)


aksdb:
I signed the CLA but fail to see how I can rerun the check.

bai:
Thanks for contributing! I re-run the CLA check and all good.

PR #1296, opened 2019-02-28
added a metric to expose consumer batch size

### LINKED ISSUE #1281: Adding consumer batch-size matric to sarama
##### Problem Description

If I have to expose a consumer batch metric from sarama, is this -> https://github.com/Shopify/sarama/blob/master/consumer.go#L519-L545 right place to do it? parseRecords get batch as an argument, can I put len(batch) as consumer-batch-size in metric?

@dim @eapache

varun06:
This is an attempt to fix #1281 

this is my first attempt to dabble in that code, let me know if it works, if not, where can I look.


varun06:
@bai  can you have a look or I need someone else to guide me with this?

varun06:
It will also need some documentation updates, which I will do in later stage.

bai:
This looks good to me; looks like there's a minor conflict.

varun06:
Should be ready now.

PR #1297, opened 2019-03-01
reduce client close bookkeeping

nelz9999:
Instead of carrying around a symbol for `ownClient`, just embed the `Client` passed in by the caller with a wrapper that disables invoking `Close()`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -911,3 +911,18 @@ func (client *client) getConsumerMetadata(consumerGroup string, attemptsRemainin
 	client.resurrectDeadBrokers()
 	return retry(ErrOutOfBrokers)
 }
+
+// nopCloserClient embeds an existing Client, but disables
+// the Close method (yet all other methods pass
+// through unchanged). This is for use in larger structs
+// where it is undesirable to close the client that was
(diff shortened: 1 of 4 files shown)


nelz9999:
(fyi, I've now signed the CLA.)

nelz9999:
There is only 1 fail out of 6 tests, but I don't believe it has anything to do with the code I changed...?

client.go
```diff
@@ -911,3 +911,18 @@ func (client *client) getConsumerMetadata(consumerGroup string, attemptsRemainin
 	client.resurrectDeadBrokers()
 	return retry(ErrOutOfBrokers)
 }
+
+// nopCloserClient embeds an existing Client, but disables
+// the Close method (yet all other methods pass
+// through unchanged). This is for use in larger structs
+// where it is undesirable to close the client that was
+// passed in by the caller.
+type nopCloserClient struct {
```
varun06: really like this pattern, we also do this wrapping pattern in one of our internal library. 👍

bai:
Thank you! 💯

PR #1298, opened 2019-03-01
preallocate message slice in consumer.go and random fixes

varun06:
- Preallocate message slice as it is more efficient
- Apparently in Go, if pointers are in front of struct, it reduces pressure on GC, so I have made that change in consumer file, if everything works fine, I will do that for other types too.

From performance slack channel ->
>So the reason that works, as I understand it, is that for any given block of data, there's a series of bits describing its pointerness. Specifically, two bits per 64-bit word; one indicates whether *this* word is a pointer, one indicates whether *any bit after it* is a pointer.
So the GC scans through from the beginning of the object until there's no more pointers, and checks all the things marked as pointers.
if you have pointers at the end of an object, the gc has to scan through the entire object looking for them. if they're all at the beginning, it stops after the last pointer. 

GC might change in future, but it is a harmless change and if it gives us some better numbers 🤷‍♂️

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -779,7 +779,7 @@ func (bc *brokerConsumer) handleResponses() {
 
 func (bc *brokerConsumer) abort(err error) {
 	bc.consumer.abandonBrokerConsumer(bc)
-	bc.broker.Close() // we don't care about the error this might return, we already have one
+	_ = bc.broker.Close() // we don't care about the error this might return, we already have one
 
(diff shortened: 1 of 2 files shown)


bai:
Oops, looks like it needs a small rebase after https://github.com/Shopify/sarama/pull/1297.

bai:
Neat, thanks! 💯

PR #1300, opened 2019-03-04
Added TLS options with client auth to kafka-console-producer and consumer

andrejvanderzee:
For testing TLS with Kafka ACL I added TLS options for the consumer and producer tools...


andrejvanderzee:
Signed CLA.

bai:
Thanks!

Issue #1284, opened 2019-02-17
The current state of `read_committed`


FrancoisPoinsot:
Even though transaction are not ready yet in sarama, it would still be super useful to be able to read committed messages only. 
Of course such message would typically be produced by java producer to begin with.

Before spending too much time on this, i would like to confirm a few things first.

1) Sarama does not support yet `read_committed` on consumer side. Is it correct?
2) Sarama does not support yet transactions on producer side. Is it correct?
3) `read_committed` is not just about sending the correct flag to the broker on a `fetchRequest`. The client need to filter the results. Is it correct?
4) It is totally doable to support `read_committed` without using transactions on producer side. The magic only happen on consumer side. Is it correct?
5) Is there some on-going work regarding the `read_committed` support?

When looking for information about how to support `read_committed` i stumbled upon 2 docs that i will note here as reference:
- https://cwiki.apache.org/confluence/display/KAFKA/KIP-98+-+Exactly+Once+Delivery+and+Transactional+Messaging
- https://docs.google.com/document/d/11Jqy_GjUGtdXJK94XGsEIK7CP1SnQGdp2eF0wSw9ra8/edit#

However, even after reading those, the work done by the client is still unclear to me. 
I would gladly get my hand on any link that would describe accurately how to implement `read_committed`.

FrancoisPoinsot:
So i figured that first thing i would need to check if i can handle uncommitted transaction correctly, is to have some uncommitted transaction.
So here is a simple producer that output some committed and some uncommitted messages: https://github.com/FrancoisPoinsot/simplest-uncommitted-msg

FrancoisPoinsot:
Using previous custom producer, here is a setup for testing the read_commited:
https://github.com/FrancoisPoinsot/sarama/tree/read-committed/tools/uncomitted-messages

Everything is in the readme, but in short:
- run docker-compose
- wait
- run the go code to consume

With this, i supose next step is to try to actually implement the read_committed.

FrancoisPoinsot:
I think I found a pretty concise description of how client should handle read_committed:
https://github.com/apache/kafka/blob/38a7708a47361929df37ed0f03160bbac5e51e93/clients/src/main/resources/common/message/FetchRequest.json#L58

FrancoisPoinsot:
https://github.com/Shopify/sarama/pull/1307

PR #1307, opened 2019-03-10
Implement ReadCommitted

FrancoisPoinsot:
Here is my contribution to implement the read_committed behavior for the consumers.

This is not full transaction support. You would need to add the ability to produce transaction for that.
This is only the ability to read committed messages only.

Also this is a draft. Unit tests are coming up and I will check a few cases.
But I would like a feedback before spending too much time in the wrong direction.

I mostly copied the java code as best as I could.
For reference I am mostly talking about this part: https://github.com/apache/kafka/blob/38a7708a47361929df37ed0f03160bbac5e51e93/clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java#L1171


FrancoisPoinsot:
oh btw,  fix #1284 

In the issue you will find my setup to make some manual test.

Andy320:
This would be very useful. Is there any plan to add it to master branch?

FrancoisPoinsot:
At this point, the bulk of the effort is done.
Unit test are working.
Manual test is working.

I would change this manual test using docker to a functional test using vagrant to match what is already existing. 
But in both case, I can't remove the reference to the custom java producer I had to use to produce some uncommitted messages. And I am not sure you want to keep such reference in this project.
Tell me what you prefer.

config_test.go
```diff
@@ -272,6 +272,24 @@ func TestProducerConfigValidates(t *testing.T) {
 				cfg.Producer.RequiredAcks = WaitForAll
 			},
 			"Idempotent producer requires Net.MaxOpenRequests to be 1"},
+		{"ReadCommitted Version",
```
fsilberstein: These tests are testing the consumer configuration and they are written in `TestProducerConfigValidates`. IMHO it would be better to create a new test method named `TestConsumerConfigValidates` to test these settings.

fetch_response.go
```diff
@@ -385,6 +385,65 @@ func (r *FetchResponse) AddRecordWithTimestamp(topic string, partition int32, ke
 	batch.addRecord(rec)
 }
 
+// AddRecordBatchWithTimestamp is similar to AddRecordWithTimestamp
+// But instead of append a record a batch of record it append a new batch of record of size 1 to a set of batch
```
fsilberstein:
```suggestion
// But instead of appending .... 
```
Here I'm not sure to fully understand your comment. Could we rephrase it?

consumer.go
```diff
@@ -595,6 +596,16 @@ func (child *partitionConsumer) parseResponse(response *FetchResponse) ([]*Consu
 	child.fetchSize = child.conf.Consumer.Fetch.Default
 	atomic.StoreInt64(&child.highWaterMarkOffset, block.HighWaterMarkOffset)
 
+	abortedProducerIDs := make(map[int64]none, len(block.AbortedTransactions))
```
varun06: nitpicky, but I don't like none type, `struct{}` is more idiomatic. but none has been there from some time and might need a big refactor to get rid of.

consumer.go
```diff
@@ -606,14 +617,60 @@ func (child *partitionConsumer) parseResponse(response *FetchResponse) ([]*Consu
 
 			messages = append(messages, messageSetMessages...)
 		case defaultRecords:
+			for _, abortedTransaction := range abortedTransactions {
+				if abortedTransaction.FirstOffset > records.RecordBatch.LastOffset() {
+					break
+				}
+				// add aborted transaction to abortedProducer list and depile abortedTransactions
+				abortedProducerIDs[abortedTransaction.ProducerID] = none{}
+				abortedTransactions = abortedTransactions[1:]
```
varun06: https://play.golang.org/p/7C8k2ijtvWX

What are you expecting to happen after line 626, it seems that loop still iterate over original slice?

control_record.go
```diff
@@ -0,0 +1,70 @@
...
+// Control records are returned as a record by fetchRequest
+// However unlike "normal" records, they mean nothing application wise.
+// They only serve internal logic for supporting transactions.
+type ControlRecord struct {
+	Version          int16
+	CoordinatorEpoch int32
+	Type             ControlRecordType
+}
+
+func (cr *ControlRecord) decode(key, value packetDecoder) (err error) {
```
varun06: I have been trying to stay away from naked returns, they make code confusing and hide bugs sometimes, if you can remove the naked return that would be great.

control_record.go
```diff
@@ -0,0 +1,70 @@
...
+	{
+		var err error
```
varun06: if you define a naked return, you don't have to define error type again.

consumer.go
FrancoisPoinsot: it is expected that loop l620 iterate over original slice.
Poping the slice is used for the loop l610

Basically how it works:
- Consumer iterate over each batches (l610)
- For each batch, add to the list on aborted producer every transaction that started in a previous offset AND is not yet resolved. (l620)
- Whenever we encounter an abortedTransaction control record: resolve this transaction.
- For a given producer, transactions cannot cross each others.

The whole point of unstacking `abortedTransactions` is just to make sure we add them only once to the list of current (as in for current offset) aborted transactions and never again. 
Even more so  after transaction is resolved and the same producer may have actually  produced a committed batch displayed in the same fetchResponse.

fetch_response.go
FrancoisPoinsot: how is it now?

FrancoisPoinsot:
There, all green

consumer.go
FrancoisPoinsot: comment in code have been updated. I hope it is clear enough now

FrancoisPoinsot:
`TestAsyncProducerRecoveryWithRetriesDisabled` seem to fail intermittently. 

I am not pushing anymore commit until review

PR #1314, opened 2019-03-15
fixed a bug related to error return

varun06:
- Error wasn't returned on line 51, I fixed that bug.
- I also removed naked returns as they are a source of confusion and masks bugs related to error return

@bai please have a look at this before cutting next release.


bai:
Good catch, thank you!

PR #1318, opened 2019-03-17
added offline replicas operation to client for kafka version >= 1.x

drsoares:
Added Offline Replicas Operation to `Client`.

- Added `OfflineReplicas` Operation
- Implemented Operation
- Added unit tests
- Changed the metadata request and response encoders
- Add to add logic to set the version id based on the version present on the `Config`

https://github.com/Shopify/sarama/issues/1311

### DIFF
--- a/metadata_request.go
+++ b/metadata_request.go
@@ -37,15 +37,8 @@ func (r *MetadataRequest) decode(pd packetDecoder, version int16) error {
 	if err != nil {
 		return err
 	}
-	if size < 0 {
-		return nil
-	} else {
-		topicCount := size
-		if topicCount == 0 {
-			goto SKIP
-		}
-
-		r.Topics = make([]string, topicCount)
+	if size > 0 {
+		r.Topics = make([]string, size)
 		for i := range r.Topics {
 			topic, err := pd.getString()
 			if err != nil {
@@ -54,7 +47,6 @@ func (r *MetadataRequest) decode(pd packetDecoder, version int16) error {
 			r.Topics[i] = topic
 		}
 	}
-SKIP:
 	if r.Version > 3 {
 		autoCreation, err := pd.getBool()
 		if err != nil {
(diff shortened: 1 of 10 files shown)


metadata_request.go
```diff
@@ -42,7 +42,7 @@ func (r *MetadataRequest) decode(pd packetDecoder, version int16) error {
 	} else {
 		topicCount := size
 		if topicCount == 0 {
-			return nil
+			goto SKIP
```
varun06: why not return here?

drsoares: If the protocol version is higher than 3 (0.11.0.0), which is the case of version 1.0.0.0 (5) that supports the offline replicas request, the if block on the bottom of the function:
```golang
if r.Version > 3 {
   autoCreation, err := pd.getBool()
   if err != nil {
      return err
   }
   r.AllowAutoTopicCreation = autoCreation
}
```
But obviously I admit this is a potential point of improvement, I can try to refactor this to a more idiomatic way.

varun06: yeah, please have a look, if it's can't be refactored, I am not super against GOTO's

drsoares: done

gbbr:
This breaks the API, so you should probably bump your next major version to `2.x` since this technically isn't `1.x` anymore.

varun06:
Apologies for missing it @gbbr 
I also had to fix an interface after upgrading to this. I am gonna work with @bai and team to handle breaking changes better in future.

drsoares:
Can you please clarify where were the breaking changes introduced? Apologies in advance.

varun06:
OfflineReplicas was a new method in client interface. We have some internal stuff where we implement the interface. So when I ran tests with new version, it was complaining for that method, I implemented it and all fine. But it seems that @gbbr had a breaking build too.

drsoares:
There was however a breaking change introduced on the testing api, however there's nothing stoping you from using those methods on your code as helpers, perhaps they should be moved into some internal package.

gbbr:
You've actually motivated me to write this (very experimental) package over the weekend: https://github.com/gbbr/breakcheck

varun06:
@gbbr that's great, I will have a look and might help with some issues(time permitted).

PR #1330, opened 2019-03-28
Fix typo and comment in balance_strategy.go

jgeiger:
This fixes the comment for the Plan function in balanceStrategy. There was also another typo in a comment.


bai:
Great catch, thanks!

PR #1332, opened 2019-03-29
fixed handle cleanup error

forsaken628:
when cleanup return an error , why handle ‘err’ ( seem always nil ) ？


consumer_group.go
```diff
@@ -614,7 +614,7 @@ func (s *consumerGroupSession) release(withCleanup bool) (err error) {
 	s.releaseOnce.Do(func() {
 		if withCleanup {
 			if e := s.handler.Cleanup(s); e != nil {
-				s.parent.handleError(err, "", -1)
+				s.parent.handleError(e, "", -1)
```
varun06: This is good catch. you have to sign CLA for build to work.

bai:
Thanks!

PR #1334, opened 2019-03-31
Adding note for clarification

matzew:
I did accidently put a goroutine inside the `ConsumeClaim`, but that was wrong.
In the docs it's mentioned that the `ConsumeClaim` itself is called w/in a goroutine.

I felt like adding a note, to avoid others doing the same mistake.


matzew:
@dim Hope you don't mind this

dim:
:+1:

PR #1335, opened 2019-04-01
added phony targets to make file

varun06:
- Added phony targets that improve make performance
- added shell check before running errcheck
- added a lint command to make


bai:
LGTM, feel free to merge when you think it's ready.

PR #1338, opened 2019-04-02
[fix] run multi consumer group instance crash

bailu1901:
[fix] run multi consumer group instance crash


bai:
Thanks! Could you please sign CLA?

bailu1901:
> Thanks! Could you please sign CLA?

Sure.
But oops,when i sign in CLA, I just filled the wrong email address.

PR #1343, opened 2019-04-06
Return ErrRebalanceInProgress errors via the group errors channel

pkedy:
I'm using `github.com/cenkalti/backoff` to perform retries in my consumer group handler.  `backoff` allows for cancelation via context, but I need to be notified by `sarama` that a rebalance is in progress in order to call `cancel`.  I may be missing an existing way to handle this, but if not, I'm thinking that returning the `ErrRebalanceInProgress` error to the group's error channel is simple enough.  To make the change backward compatible I added the `Consumer.Group.Rebalance.SendGroupError` option.  Would this be a reasonable solution?

cc @dim (thanks for adding the consumer group feature!)

### DIFF
--- a/config.go
+++ b/config.go
@@ -246,6 +246,11 @@ type Config struct {
 					// Backoff time between retries during rebalance (default 2s)
 					Backoff time.Duration
 				}
+
+				// If enabled, any ErrRebalanceInProgress errors are returned on the Errors channel (default disabled).
+				// This gives the user the ability to cancel all in progress message handlers so that the consumer can
+				// rejoin the consumer group.
+				SendGroupError bool
 			}
 			Member struct {
 				// Custom metadata to include when joining the group. The user data for all joined members
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -691,7 +691,12 @@ func (s *consumerGroupSession) heartbeatLoop() {
 		switch resp.Err {
 		case ErrNoError:
 			retries = s.parent.config.Metadata.Retry.Max
-		case ErrRebalanceInProgress, ErrUnknownMemberId, ErrIllegalGeneration:
+		case ErrRebalanceInProgress:
+			if s.parent.config.Consumer.Group.Rebalance.SendGroupError {
+				s.parent.errors <- resp.Err
+			}
+			return
+		case ErrUnknownMemberId, ErrIllegalGeneration:
 			return
 		default:
 			s.parent.handleError(err, "", -1)


dim:
@pkedy there is a backoff built into sarama already (see various settings), I wouldn't layer another one on top of that.

pkedy:
@dim Sorry - I didn't describe the use case very well.  Here is a simplified example of what I'm trying to do.  The backoff is used for the processing of the individual messages.  From what I can tell, when a rebalance is in progress, `ConsumeClaim` for each claim needs to return relatively quickly for the consumer to rejoin the group.  Maybe I am still missing something but it looks like the existing consuming backoff features pertain to the rebalance itself or internally reading from the partition (wire protocol).  I'm totally open to alternative approaches to this!  Thanks for taking a look at this.

```
type consumerGroupHandler struct {
	sync.Mutex
	cancel context.CancelFunc
}

func (h *consumerGroupHandler) ConsumeClaim(sess sarama.ConsumerGroupSession, claim sarama.ConsumerGroupClaim) error {
	ctx := sess.Context()

	for msg := range claim.Messages() {
		eb := backoff.NewExponentialBackOff()
		bo := backoff.WithContext(eb, sess.Context())

		err := backoff.RetryNotify(
			func() error {
				// Do something that can fail and need to be retried over a long period of time.
				log.Infof("Message topic:%q partition:%d offset:%d\n", msg.Topic, msg.Partition, msg.Offset)
				return nil
			},
			bo,
			func(err error, delay time.Duration) {
				log.Errorf("Encountered an error: %v", err)
				log.Infof("Backing off for %s...\n", delay)
			},
		)

		if err != nil {
			sess.MarkMessage(msg, "")
		}

		// Return if the context was canceled.
		select {
		case <-ctx.Done():
			return err
		default:
		}
	}
	return nil
}

func (h *consumerGroupHandler) groupErrorHandler(group sarama.ConsumerGroup) {
	go func() {
		for err := range group.Errors() {
			switch err {
			// This PR adds the option for this to be returned via the group errors channel.
			case sarama.ErrRebalanceInProgress:
				h.Lock()
				// This breaks out of the `backoff.RetryNotify` above.
				h.cancel()
				h.Unlock()
			default:
				// Handle other errors as necessary
				return
			}
		}
	}()
}

func (h *consumerGroupHandler) setCancel(cancel context.CancelFunc) {
	h.Lock()
	h.cancel = cancel
	h.Unlock()
}

```

sitano:
@pkedy the publishing of the events to the error channel is not safely implemented. thus you can't reliably rely on whatever is returned from it, especially for control execution flow. because those events can randomly be lost on a way to your code.

BTW, the feature of reading a rebalance event is nice.

FYI, on the implementation of the error channel:

```
c.errors is a buf chan of len 256, but then:

select {
	case c.errors <- err:
	default:
}
```

thus it does not make sense at all in sending the rebalance event to this chan until the a-write-try will be replaced with just-write.

pkedy:
@sitano, @dim - I am considering two approaches here:
1) Create a totally new channel dedicated to rebalance events
2) Internally use `context`/`WithCancel` to cancel any processing inside the user's implementation of `ConsumeClaim` when a rebalance occurs
Thoughts?

dim:
> when a rebalance is in progress, ConsumeClaim for each claim needs to return relatively quickly for the consumer to rejoin the group

You can configure that via `SessionTimeout`, which should take into account your backoff setting.

> Create a totally new channel dedicated to rebalance events

No need a rebalance is triggered when `session.Context().Done()` is firing. If *you* want to cancel a session, just return from `ConsumeClaim`. The first goroutine to exit `ConsumeClaim` will cancel `session.Context()`. 

Finally: 

```
		// Return if the context was canceled.
		select {
		case <-ctx.Done():
			return err
		default:
		}
```

It doesn't hurt, but you don't need to do that, as the `claim.Messages()` channel/loop will exit automatically when `session.Context().Done()`.

sitano:
@pkedy consumer group shutdown current session when the rebalance is in progress. That means it asynchronously closes a message chan and cancels a session context as @dim mentioned. You can't peek a chan for close (actually you can with unsafe). But I bet `session.Context().Done()` must work for you. 

You still can't know whether it was fired by the rebalance event or another type of error, but for canceling a backoff retries it must be enough.

pkedy:
@dim @sitano `session.Context().Done()` works as expected.  Thanks for pointing me that way.  My code was originally using its own context variable instead of pulling it from `session.Context().Done()` so I that sent me down this rabbit hole.  Thanks for your patience!

PR #1344, opened 2019-04-08
fix auto create topic on metadata request when request version greate…

RazOmessi:
When the request version is greater than 3, MetadataRequest will encode the AllowAutoTopicCreation field to the request.

This field is always false so the topic will not create by the broker.


saturn4er:
@bai Have the same problem with latest sarama version. @RazOmessi there's some CI isssues here

RazOmessi:
I signed the CLA.

bai:
Thanks for your contribution!

PR #1353, opened 2019-04-10
bug: V5 MetadataRequest nullable topics array

### LINKED ISSUE #1352: MetadataRequest V5 has wrong encoding for "all topics"
##### Versions

Sarama Version: 2b824357033b9904e3a32bf7fae2efe76f1591fd
Kafka Version: 1.1.0
Go Version: 1.11.8

##### Configuration

```
	cfg := sarama.NewConfig()
	cfg.Net.TLS.Enable = true
	cfg.Version = sarama.V1_1_0_0
```

##### Logs

```
[Sarama] 2019/04/10 22:08:13 Initializing new client
[Sarama] 2019/04/10 22:08:13 client/metadata fetching metadata for all topics from broker kafka-2:9092
[Sarama] 2019/04/10 22:08:19 Connected to broker at kafka-2:9092 (unregistered)
[Sarama] 2019/04/10 22:08:19 client/brokers registered new broker #2 at kafka-2:9092
[Sarama] 2019/04/10 22:08:19 client/brokers registered new broker #1 at kafka-1:9092
len(topics) received 0 topics in all metadata request
```

##### Problem Description

TL;DR if the client config version is at least 1.0.0.0 then the "full" metadata refresh always returns an empty topic list

##### Underlying cause

Since e3ecf7b0882e060c2b0b2375e3682f642373db3a the `tryRefreshMetadata` for a "full" metadata request now appears to use the V5 MetadataRequest if config version is at least 1.0.0.0

https://github.com/Shopify/sarama/blob/2b824357033b9904e3a32bf7fae2efe76f1591fd/client.go#L761-L765

However, this commit appeared to also add an additional if statement to the encoding of the request to send an empty array rather than a null if the provided topics list was an empty slice

https://github.com/Shopify/sarama/blob/2b824357033b9904e3a32bf7fae2efe76f1591fd/metadata_request.go#L13

I'm not sure why this was done as consulting the [protocol guide](https://kafka.apache.org/protocol#The_Messages_Metadata) the V5 request uses the same nullable array for topics as the V1-V4 requests.

> Metadata Request (Version: 5) => [topics] allow_auto_topic_creation 
  topics => STRING
  allow_auto_topic_creation => BOOLEAN

| Field | Description | 
| --- | --- | 
| topics | An array of topics to fetch metadata for. If the topics array is null fetch metadata for all topics. | 
| allow_auto_topic_creation | If this and the broker config auto.create.topics.enable are true, topics that don't exist will be created by the broker. Otherwise, no topics will be created by the broker. |

dnwe:
Consulting the protocol guide the V5 request uses the same nullable
array for topics as the V1-V4 requests. Improve existing unittest to
properly exercise the encoding paths.

Fixes #1352

### DIFF
--- a/metadata_request.go
+++ b/metadata_request.go
@@ -10,7 +10,7 @@ func (r *MetadataRequest) encode(pe packetEncoder) error {
 	if r.Version < 0 || r.Version > 5 {
 		return PacketEncodingError{"invalid or unsupported MetadataRequest version field"}
 	}
-	if r.Version == 0 || r.Version == 5 || len(r.Topics) > 0 {
+	if r.Version == 0 || len(r.Topics) > 0 {
 		err := pe.putArrayLength(len(r.Topics))
 		if err != nil {
 			return err
--- a/metadata_request_test.go
+++ b/metadata_request_test.go
@@ -3,24 +3,63 @@ package sarama
 import "testing"
 
 var (
+	// The v0 metadata request has a non-nullable array of topic names
+	// to request metadata for. An empty array fetches metadata for all topics
+
 	metadataRequestNoTopicsV0 = []byte{
-		0x00, 0x00, 0x00, 0x00}
+		0x00, 0x00, 0x00, 0x00,
+	}
 
 	metadataRequestOneTopicV0 = []byte{
 		0x00, 0x00, 0x00, 0x01,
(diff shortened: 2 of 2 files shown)


bai:
Thank you!

PR #1354, opened 2019-04-10
feat: support SaslHandshakeRequest v1

dnwe:
> If SaslHandshakeRequest version is v0, a series of SASL client and
> server tokens corresponding to the mechanism are sent as opaque packets
> without wrapping the messages with Kafka protocol headers. If
> SaslHandshakeRequest version is v1, the SaslAuthenticate
> request/response are used, where the actual SASL tokens are wrapped in
> the Kafka protocol. The error code in the final message from the broker
> will indicate if authentication succeeded or failed.

References:
  https://kafka.apache.org/protocol#sasl_handshake
  https://kafka.apache.org/protocol#The_Messages_SaslHandshake
  https://kafka.apache.org/protocol#The_Messages_SaslAuthenticate

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -13,7 +13,7 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/rcrowley/go-metrics"
+	metrics "github.com/rcrowley/go-metrics"
 )
 
 // Broker represents a single Kafka broker connection. All operations on this object are entirely concurrency-safe.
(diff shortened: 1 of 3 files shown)


mimaison:
@bai can you take a look?
This is a very interesting change as this allows clients using SASL Plain to get a meaningful message in case of an authentication failure.

bai:
Many thanks! (And thanks for poking me!) 💯

PR #1365, opened 2019-04-29
Custom auth handler interface

mieczkowski:
Hi. I would like to propose a custom handler interface for auth. 

Sarama is missing Kerberos auth method, and PR https://github.com/Shopify/sarama/pull/729 was closed due to external C library requirement. For that I've created custom implementation for kerberos auth (based on #729 PR and my research in librdkafka) that is using (and works!) this custom handler interface:

https://github.com/mieczkowski/sarama-kerberos/blob/master/kerberos.go

What do you think about that? :-)

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -101,7 +101,7 @@ type AccessTokenProvider interface {
 
 // CustomAuthHandler is the interface for custom authorization mechanism
 type CustomAuthHandler interface {
-	// Authorize get broker connection, address of broker and configuration pointer.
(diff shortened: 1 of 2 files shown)


wedi-dev:
+1

mieczkowski:
About CLA: I can not sign CLA because it shows me my old email address. I signed it with my actual email address. I don't know why my old address is required...

mieczkowski:
Closed due to better implementation in #1366

PR #1376, opened 2019-05-21
fetch size overflow

andrewhao888:
When getting a partial trailing message, consumer's fetch size will double.
However, it is possible to reach max int32, then a kafka broker receives a fetch request with negative fetch size.
Looks like this issue has been checked by Kafka itself with ticket: https://issues.apache.org/jira/browse/KAFKA-7656

Otherwise, we should provide a reasonable default Fetch.Max value


consumer.go
```diff
@@ -591,6 +592,10 @@ func (child *partitionConsumer) parseResponse(response *FetchResponse) ([]*Consu
 				child.offset++ // skip this one so we can keep processing future messages
 			} else {
 				child.fetchSize *= 2
+				// int32 overflow
```
varun06: Can you please change the comment to `check int32 overflow` or remove it?

varun06:
Looks good to me, Do you want to sign CLA and add a test too.

consumer.go
andrewhao888: sure, let me update the comment

andrewhao888:
Yes, let me add a test case, as well as CLA

andrewhao888:
sorry @varun06, it is too hard to provide a valid test case without mock broker update. A PartialTrailingRecord will lose trailing record after encoding. I just push a test case in fetch_response_test.go and sign CLA. Is that ok to merge?

varun06:
yeah, sounds good.

bai:
Thanks for your contribution, LGTM.

PR #1393, opened 2019-06-05
Fix acl lint issues

varun06:
There is a ton of lint/vet fixes in this PR. There is still one left, as that one is in a dependency and not in sarama.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,10 @@ before_install:
 install: make install_dependencies
 
 script:
-- make test
+- make lint
 - make vet
 - make errcheck
+- make test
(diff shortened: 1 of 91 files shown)


bai:
🤔Looks like CI failure is legit.

varun06:
Yup, this is where error is coming from, I only changed bunch of names -> https://github.com/Shopify/sarama/pull/1393/files#diff-39b065607d3d0839bdc97f997d6a430cR469

Maybe I changed something else, will try to fix it in next couple days :)

varun06:
No idea why the test is failing, also PR is too big. Gonna close it and then bring these changes in smalls chunks.

PR #1397, opened 2019-06-10
Add Kafka error codes 77-81

MikeSchlosser16:
Adds Kafka server error codes corresponding to https://kafka.apache.org/protocol#protocol_error_codes


bai:
Thank you!

PR #1410, opened 2019-06-25
added SASLVersion in config

prune998:
Sarama force SASL V1 handshake when Kafka protocol is greater than 1.0.
Azure Eventhub is still using SASLV0.

I moved the SASL protocol version to the config, defaulting to V0 (like before), but I also removed the auto V1 when Kafka > 1.0

Refer to https://github.com/Shopify/sarama/issues/1408 for more info.

@varun06 Let me know if this is what you expected or let me know what you prefer, like keeping the old behaviour but adding a config flag to disable the force change to V1 when Kafka > 1.0

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -4,7 +4,6 @@ import (
 	"crypto/tls"
 	"encoding/binary"
 	"fmt"
-	metrics "github.com/rcrowley/go-metrics"
 	"io"
 	"net"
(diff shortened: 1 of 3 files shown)


varun06:
I think it makes sense. @bai what do you think, also do we have anyone else who can have a look?

PR #1414, opened 2019-07-01
[master]: cluster admin from client

bo0rsh201:
Added ...FromClient constructor to ClusterAdmin in the same way as for others in order to reuse client


bo0rsh201:
wrong email, sorry

PR #1415, opened 2019-07-01
cluster admin from client

bo0rsh201:
Added ...FromClient constructor to ClusterAdmin in the same way as for others in order to reuse clients


bo0rsh201:
I did sign CLA, but nothing happens and I have no ability to re-run CLA check. What do I do wrong? :)

varun06:
Looks good. Can you please add a test for `NewAdminFromClient`?

PR #1416, opened 2019-07-01
Sticky partition assignor

skidder:
This PR adds support for the sticky partition assignment strategy defined in Kafka Improvement (KIP) [KIP-54 ](https://cwiki.apache.org/confluence/display/KAFKA/KIP-54+-+Sticky+Partition+Assignment+Strategy), and later in [KIP-341](https://cwiki.apache.org/confluence/display/KAFKA/KIP-341%3A+Update+Sticky+Assignor%27s+User+Data+Protocol). This includes the use of consumer group generation numbers to reconcile conflicting information from consumer group members.

The sticky partition assignment strategy differs from the range and roundrobin strategies by attempting to keep topic partition assignments on the same consumer from the previous assignment. This is accomplished by having the consumer group leader use assignment information stored in the consumer `userdata` to learn about previous assignments.

This assignor aims to be a 1:1 port of the [StickyAssignor](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/StickyAssignor.java) included in the official Java Kafka client, including all unit tests.

## Manual Testing
Manual testing can be performed using the consumer-group example (`examples/consumergroup/main.go`).

### Setup
1. Run a Kafka cluster
1. Create a new topic in the Kafka cluster (e.g. `sticky-assignment-demo`) with multiple partitions (e.g. 12)

### Step 1: Start Consumer 1
Start the first consumer, which will be assigned all 12 topic partitions. Nothing unusual here.
```
→ go run main.go -assignor sticky -brokers localhost:9092 -group sticky-assignment-cg -topics sticky-assignment-demo -verbose true
2019/07/01 10:03:29 Starting a new Sarama consumer
[sarama] 2019/07/01 10:03:29 Initializing new client
[sarama] 2019/07/01 10:03:29 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2019/07/01 10:03:29 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2019/07/01 10:03:29 client/metadata fetching metadata for all topics from broker localhost:9092
[sarama] 2019/07/01 10:03:29 Connected to broker at localhost:9092 (unregistered)
[sarama] 2019/07/01 10:03:29 client/brokers registered new broker #1 at localhost:9092
[sarama] 2019/07/01 10:03:29 Successfully initialized new client
[sarama] 2019/07/01 10:03:29 client/metadata fetching metadata for [sticky-assignment-demo] from broker localhost:9092
[sarama] 2019/07/01 10:03:29 client/coordinator requesting coordinator for consumergroup sticky-assignment-cg from localhost:9092
[sarama] 2019/07/01 10:03:29 client/coordinator coordinator for consumergroup sticky-assignment-cg is #1 (localhost:9092)
[sarama] 2019/07/01 10:03:29 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2019/07/01 10:03:29 Connected to broker at localhost:9092 (registered as #1)
[sarama] 2019/07/01 10:03:29 client/coordinator requesting coordinator for consumergroup sticky-assignment-cg from localhost:9092
[sarama] 2019/07/01 10:03:29 client/coordinator coordinator for consumergroup sticky-assignment-cg is #1 (localhost:9092)
2019/07/01 10:03:29 Sarama consumer up and running!...
[sarama] 2019/07/01 10:03:29 consumer/broker/1 added subscription to sticky-assignment-demo/0
[sarama] 2019/07/01 10:03:30 consumer/broker/1 added subscription to sticky-assignment-demo/11
[sarama] 2019/07/01 10:03:30 consumer/broker/1 added subscription to sticky-assignment-demo/2
[sarama] 2019/07/01 10:03:30 consumer/broker/1 added subscription to sticky-assignment-demo/3
[sarama] 2019/07/01 10:03:30 consumer/broker/1 added subscription to sticky-assignment-demo/4
[sarama] 2019/07/01 10:03:30 consumer/broker/1 added subscription to sticky-assignment-demo/5
[sarama] 2019/07/01 10:03:30 consumer/broker/1 added subscription to sticky-assignment-demo/6
[sarama] 2019/07/01 10:03:30 consumer/broker/1 added subscription to sticky-assignment-demo/7
[sarama] 2019/07/01 10:03:30 consumer/broker/1 added subscription to sticky-assignment-demo/8
[sarama] 2019/07/01 10:03:30 consumer/broker/1 added subscription to sticky-assignment-demo/9
[sarama] 2019/07/01 10:03:30 consumer/broker/1 added subscription to sticky-assignment-demo/10
[sarama] 2019/07/01 10:03:30 consumer/broker/1 added subscription to sticky-assignment-demo/1
```

### Step 2: Start Consumer 2
Start a second consumer, which will obtain 6 of the 12 partitions. Again, nothing unusual yet.

Consumer 2 Startup Output
```
→ go run main.go -assignor sticky -brokers localhost:9092 -group sticky-assignment-cg -topics sticky-assignment-demo -verbose true
2019/07/01 10:05:59 Starting a new Sarama consumer
[sarama] 2019/07/01 10:05:59 Initializing new client
[sarama] 2019/07/01 10:05:59 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2019/07/01 10:05:59 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2019/07/01 10:05:59 client/metadata fetching metadata for all topics from broker localhost:9092
[sarama] 2019/07/01 10:05:59 Connected to broker at localhost:9092 (unregistered)
[sarama] 2019/07/01 10:05:59 client/brokers registered new broker #1 at localhost:9092
[sarama] 2019/07/01 10:05:59 Successfully initialized new client
[sarama] 2019/07/01 10:05:59 client/metadata fetching metadata for [sticky-assignment-demo] from broker localhost:9092
[sarama] 2019/07/01 10:05:59 client/coordinator requesting coordinator for consumergroup sticky-assignment-cg from localhost:9092
[sarama] 2019/07/01 10:05:59 client/coordinator coordinator for consumergroup sticky-assignment-cg is #1 (localhost:9092)
[sarama] 2019/07/01 10:05:59 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2019/07/01 10:05:59 Connected to broker at localhost:9092 (registered as #1)
[sarama] 2019/07/01 10:06:00 client/coordinator requesting coordinator for consumergroup sticky-assignment-cg from localhost:9092
[sarama] 2019/07/01 10:06:00 client/coordinator coordinator for consumergroup sticky-assignment-cg is #1 (localhost:9092)
2019/07/01 10:06:00 Sarama consumer up and running!...
[sarama] 2019/07/01 10:06:00 consumer/broker/1 added subscription to sticky-assignment-demo/9
[sarama] 2019/07/01 10:06:00 consumer/broker/1 added subscription to sticky-assignment-demo/3
[sarama] 2019/07/01 10:06:00 consumer/broker/1 added subscription to sticky-assignment-demo/2
[sarama] 2019/07/01 10:06:00 consumer/broker/1 added subscription to sticky-assignment-demo/1
[sarama] 2019/07/01 10:06:00 consumer/broker/1 added subscription to sticky-assignment-demo/4
[sarama] 2019/07/01 10:06:00 consumer/broker/1 added subscription to sticky-assignment-demo/8
```

Consumer 1 output during partition reassignment:
```
[sarama] 2019/07/01 10:06:00 consumer/broker/1 closed dead subscription to sticky-assignment-demo/11
[sarama] 2019/07/01 10:06:00 consumer/broker/1 closed dead subscription to sticky-assignment-demo/3
[sarama] 2019/07/01 10:06:00 consumer/broker/1 closed dead subscription to sticky-assignment-demo/5
[sarama] 2019/07/01 10:06:00 consumer/broker/1 closed dead subscription to sticky-assignment-demo/7
[sarama] 2019/07/01 10:06:00 consumer/broker/1 closed dead subscription to sticky-assignment-demo/10
[sarama] 2019/07/01 10:06:00 consumer/broker/1 closed dead subscription to sticky-assignment-demo/0
[sarama] 2019/07/01 10:06:00 consumer/broker/1 closed dead subscription to sticky-assignment-demo/4
[sarama] 2019/07/01 10:06:00 consumer/broker/1 closed dead subscription to sticky-assignment-demo/6
[sarama] 2019/07/01 10:06:00 consumer/broker/1 closed dead subscription to sticky-assignment-demo/8
[sarama] 2019/07/01 10:06:00 consumer/broker/1 closed dead subscription to sticky-assignment-demo/9
[sarama] 2019/07/01 10:06:00 consumer/broker/1 closed dead subscription to sticky-assignment-demo/1
[sarama] 2019/07/01 10:06:00 consumer/broker/1 closed dead subscription to sticky-assignment-demo/2
[sarama] 2019/07/01 10:06:00 client/metadata fetching metadata for [sticky-assignment-demo] from broker localhost:9092
[sarama] 2019/07/01 10:06:00 client/coordinator requesting coordinator for consumergroup sticky-assignment-cg from localhost:9092
[sarama] 2019/07/01 10:06:00 client/coordinator coordinator for consumergroup sticky-assignment-cg is #1 (localhost:9092)
[sarama] 2019/07/01 10:06:00 consumer/broker/1 added subscription to sticky-assignment-demo/5
[sarama] 2019/07/01 10:06:00 consumer/broker/1 added subscription to sticky-assignment-demo/11
[sarama] 2019/07/01 10:06:00 consumer/broker/1 added subscription to sticky-assignment-demo/0
[sarama] 2019/07/01 10:06:00 consumer/broker/1 added subscription to sticky-assignment-demo/7
[sarama] 2019/07/01 10:06:00 consumer/broker/1 added subscription to sticky-assignment-demo/6
[sarama] 2019/07/01 10:06:00 consumer/broker/1 added subscription to sticky-assignment-demo/10
```

### Step 3: Start Consumer 3
Start a third consumer, which will obtain 4 of the 12 partitions. At this point we can see the sticky partition assignment take effect, as consumers 1 and 2 retain 4 of their 6 partition claims.

Consumer 3 Startup Output
```
→ go run main.go -assignor sticky -brokers localhost:9092 -group sticky-assignment-cg -topics sticky-assignment-demo -verbose true
2019/07/01 10:08:22 Starting a new Sarama consumer
[sarama] 2019/07/01 10:08:22 Initializing new client
[sarama] 2019/07/01 10:08:22 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2019/07/01 10:08:22 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2019/07/01 10:08:22 client/metadata fetching metadata for all topics from broker localhost:9092
[sarama] 2019/07/01 10:08:22 Connected to broker at localhost:9092 (unregistered)
[sarama] 2019/07/01 10:08:22 client/brokers registered new broker #1 at localhost:9092
[sarama] 2019/07/01 10:08:22 Successfully initialized new client
[sarama] 2019/07/01 10:08:22 client/metadata fetching metadata for [sticky-assignment-demo] from broker localhost:9092
[sarama] 2019/07/01 10:08:22 client/coordinator requesting coordinator for consumergroup sticky-assignment-cg from localhost:9092
[sarama] 2019/07/01 10:08:22 client/coordinator coordinator for consumergroup sticky-assignment-cg is #1 (localhost:9092)
[sarama] 2019/07/01 10:08:22 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[sarama] 2019/07/01 10:08:22 Connected to broker at localhost:9092 (registered as #1)
[sarama] 2019/07/01 10:08:24 client/coordinator requesting coordinator for consumergroup sticky-assignment-cg from localhost:9092
[sarama] 2019/07/01 10:08:24 client/coordinator coordinator for consumergroup sticky-assignment-cg is #1 (localhost:9092)
2019/07/01 10:08:24 Sarama consumer up and running!...
[sarama] 2019/07/01 10:08:24 consumer/broker/1 added subscription to sticky-assignment-demo/8
[sarama] 2019/07/01 10:08:24 consumer/broker/1 added subscription to sticky-assignment-demo/6
[sarama] 2019/07/01 10:08:24 consumer/broker/1 added subscription to sticky-assignment-demo/7
[sarama] 2019/07/01 10:08:24 consumer/broker/1 added subscription to sticky-assignment-demo/4
```

Consumer 1 output during partition reassignment:
```
[sarama] 2019/07/01 10:08:24 consumer/broker/1 closed dead subscription to sticky-assignment-demo/7
[sarama] 2019/07/01 10:08:24 consumer/broker/1 closed dead subscription to sticky-assignment-demo/6
[sarama] 2019/07/01 10:08:24 consumer/broker/1 closed dead subscription to sticky-assignment-demo/10
[sarama] 2019/07/01 10:08:24 consumer/broker/1 closed dead subscription to sticky-assignment-demo/5
[sarama] 2019/07/01 10:08:24 consumer/broker/1 closed dead subscription to sticky-assignment-demo/11
[sarama] 2019/07/01 10:08:24 consumer/broker/1 closed dead subscription to sticky-assignment-demo/0
[sarama] 2019/07/01 10:08:24 client/metadata fetching metadata for [sticky-assignment-demo] from broker localhost:9092
[sarama] 2019/07/01 10:08:24 client/coordinator requesting coordinator for consumergroup sticky-assignment-cg from localhost:9092
[sarama] 2019/07/01 10:08:24 client/coordinator coordinator for consumergroup sticky-assignment-cg is #1 (localhost:9092)
[sarama] 2019/07/01 10:08:24 consumer/broker/1 added subscription to sticky-assignment-demo/11
[sarama] 2019/07/01 10:08:24 consumer/broker/1 added subscription to sticky-assignment-demo/5
[sarama] 2019/07/01 10:08:24 consumer/broker/1 added subscription to sticky-assignment-demo/0
[sarama] 2019/07/01 10:08:24 consumer/broker/1 added subscription to sticky-assignment-demo/10
```

Consumer 2 output during partition reassignment:
```
[sarama] 2019/07/01 10:08:24 consumer/broker/1 closed dead subscription to sticky-assignment-demo/1
[sarama] 2019/07/01 10:08:24 consumer/broker/1 closed dead subscription to sticky-assignment-demo/4
[sarama] 2019/07/01 10:08:24 consumer/broker/1 closed dead subscription to sticky-assignment-demo/8
[sarama] 2019/07/01 10:08:24 consumer/broker/1 closed dead subscription to sticky-assignment-demo/9
[sarama] 2019/07/01 10:08:24 consumer/broker/1 closed dead subscription to sticky-assignment-demo/3
[sarama] 2019/07/01 10:08:24 consumer/broker/1 closed dead subscription to sticky-assignment-demo/2
[sarama] 2019/07/01 10:08:24 client/metadata fetching metadata for [sticky-assignment-demo] from broker localhost:9092
[sarama] 2019/07/01 10:08:24 client/coordinator requesting coordinator for consumergroup sticky-assignment-cg from localhost:9092
[sarama] 2019/07/01 10:08:24 client/coordinator coordinator for consumergroup sticky-assignment-cg is #1 (localhost:9092)
[sarama] 2019/07/01 10:08:24 consumer/broker/1 added subscription to sticky-assignment-demo/1
[sarama] 2019/07/01 10:08:24 consumer/broker/1 added subscription to sticky-assignment-demo/2
[sarama] 2019/07/01 10:08:24 consumer/broker/1 added subscription to sticky-assignment-demo/9
[sarama] 2019/07/01 10:08:24 consumer/broker/1 added subscription to sticky-assignment-demo/3
```

### DIFF
--- a/balance_strategy.go
+++ b/balance_strategy.go
@@ -215,9 +215,9 @@ func (s *stickyBalanceStrategy) Plan(members map[string]ConsumerGroupMemberMetad
 
 	// create a mapping of each partition to its current consumer, where possible
 	currentPartitionConsumers := make(map[topicPartitionAssignment]string, len(currentAssignment))
-	unvisitedPartitions := make(map[topicPartitionAssignment][]string, len(partition2AllPotentialConsumers))
+	unvisitedPartitions := make(map[topicPartitionAssignment]bool, len(partition2AllPotentialConsumers))
 	for partition := range partition2AllPotentialConsumers {
-		unvisitedPartitions[partition] = []string{}
+		unvisitedPartitions[partition] = true
 	}
 	var unassignedPartitions []topicPartitionAssignment
 	for memberID, partitions := range currentAssignment {
--- a/balance_strategy.go
+++ b/balance_strategy.go
@@ -444,7 +444,6 @@ func (s *stickyBalanceStrategy) performReassignments(reassignablePartitions []to
 					modified = true
 					break
 				}
-
 			}
 		}
 		if !modified {
@@ -550,23 +549,6 @@ func filterAssignedPartitions(currentAssignment map[string][]topicPartitionAssig
 	return assignments
 }
 
-func removeValueFromStringSlice(s []string, e string) []string {
-	for i, v := range s {
-		if v == e {
-			s = append(s[:i], s[i+1:]...)
-			break
-		}
-	}
-	return s
-}
-
-func removeIndexFromStringSlice(s []string, i int) []string {
-	if len(s) == 0 {
-		return s
-	}
-	return append(s[:i], s[i+1:]...)
-}
-
 func removeTopicPartitionFromMemberAssignments(assignments []topicPartitionAssignment, topic topicPartitionAssignment) []topicPartitionAssignment {
 	for i, assignment := range assignments {
 		if assignment == topic {
@@ -586,6 +568,11 @@ func memberAssignmentsIncludeTopicPartition(assignments []topicPartitionAssignme
 }
 
 func sortPartitions(currentAssignment map[string][]topicPartitionAssignment, partitionsWithADifferentPreviousAssignment map[topicPartitionAssignment]consumerGenerationPair, isFreshAssignment bool, partition2AllPotentialConsumers map[topicPartitionAssignment][]string, consumer2AllPotentialPartitions map[string][]topicPartitionAssignment) []topicPartitionAssignment {
+	unassignedPartitions := make(map[topicPartitionAssignment]bool, len(partition2AllPotentialConsumers))
+	for partition := range partition2AllPotentialConsumers {
+		unassignedPartitions[partition] = true
+	}
+
 	sortedPartitions := make([]topicPartitionAssignment, 0)
 	if !isFreshAssignment && areSubscriptionsIdentical(partition2AllPotentialConsumers, consumer2AllPotentialPartitions) {
 		// if this is a reassignment and the subscriptions are identical (all consumers can consumer from all topics)
@@ -601,7 +588,6 @@ func sortPartitions(currentAssignment map[string][]topicPartitionAssignment, par
 			pq[i] = &consumerGroupMember{
 				id:          consumerID,
 				assignments: consumerAssignments,
-				index:       i,
 			}
 			i++
 		}
@@ -615,43 +601,36 @@ func sortPartitions(currentAssignment map[string][]topicPartitionAssignment, par
 			member := pq[0]
 
 			// partitions that were assigned to a different consumer last time
-			prevPartitions := make([]topicPartitionAssignment, 0)
-			for partition := range partitionsWithADifferentPreviousAssignment {
-				// from partitions that had a different consumer before, keep only those that are assigned to this consumer now
-				if memberAssignmentsIncludeTopicPartition(member.assignments, partition) {
-					prevPartitions = append(prevPartitions, partition)
+			var prevPartition *topicPartitionAssignment
+			var prevPartitionIndex int
+			for i, partition := range member.assignments {
+				if _, exists := partitionsWithADifferentPreviousAssignment[partition]; exists {
+					prevPartition = &partition
+					prevPartitionIndex = i
+					break
 				}
 			}
 
-			if len(prevPartitions) > 0 {
+			if prevPartition != nil {
 				// if there is a partition on this consumer that was assigned to another consumer before mark it as good options for reassignment
-				partition := prevPartitions[0]
-				prevPartitions = append(prevPartitions[:0], prevPartitions[1:]...)
-				member.assignments = removeTopicPartitionFromMemberAssignments(member.assignments, partition)
-				sortedPartitions = append(sortedPartitions, partition)
-				heap.Fix(&pq, member.index)
+				member.assignments = append(member.assignments[:prevPartitionIndex], member.assignments[prevPartitionIndex+1:]...)
+				sortedPartitions = append(sortedPartitions, *prevPartition)
+				delete(unassignedPartitions, *prevPartition)
+				heap.Fix(&pq, 0)
 			} else if len(member.assignments) > 0 {
 				// otherwise, mark any other one of the current partitions as a reassignment candidate
 				partition := member.assignments[0]
 				member.assignments = append(member.assignments[:0], member.assignments[1:]...)
 				sortedPartitions = append(sortedPartitions, partition)
-				heap.Fix(&pq, member.index)
+				delete(unassignedPartitions, partition)
+				heap.Fix(&pq, 0)
 			} else {
-				heap.Remove(&pq, 0)
+				heap.Pop(&pq)
 			}
 		}
 
-		for partition := range partition2AllPotentialConsumers {
-			found := false
-			for _, p := range sortedPartitions {
-				if partition == p {
-					found = true
-					break
-				}
-			}
-			if !found {
-				sortedPartitions = append(sortedPartitions, partition)
-			}
+		for partition := range unassignedPartitions {
+			sortedPartitions = append(sortedPartitions, partition)
 		}
 	} else {
 		// an ascending sorted set of topic partitions based on how many consumers can potentially use them
@@ -678,12 +657,21 @@ func sortMemberIDsByPartitionAssignments(assignments map[string][]topicPartition
 
 func sortPartitionsByPotentialConsumerAssignments(partition2AllPotentialConsumers map[topicPartitionAssignment][]string) []topicPartitionAssignment {
 	// sort the members by the number of partition assignments in descending order
-	sortedPartionIDs := make([]topicPartitionAssignment, 0, len(partition2AllPotentialConsumers))
+	sortedPartionIDs := make([]topicPartitionAssignment, len(partition2AllPotentialConsumers))
+	i := 0
 	for partition := range partition2AllPotentialConsumers {
-		sortedPartionIDs = append(sortedPartionIDs, partition)
+		sortedPartionIDs[i] = partition
+		i++
 	}
-	sort.SliceStable(sortedPartionIDs, func(i, j int) bool {
-		return len(partition2AllPotentialConsumers[sortedPartionIDs[i]]) > len(partition2AllPotentialConsumers[sortedPartionIDs[j]])
+	sort.Slice(sortedPartionIDs, func(i, j int) bool {
+		if len(partition2AllPotentialConsumers[sortedPartionIDs[i]]) == len(partition2AllPotentialConsumers[sortedPartionIDs[j]]) {
+			ret := strings.Compare(sortedPartionIDs[i].Topic, sortedPartionIDs[j].Topic)
+			if ret == 0 {
+				return sortedPartionIDs[i].Partition < sortedPartionIDs[j].Partition
+			}
+			return ret < 0
+		}
+		return len(partition2AllPotentialConsumers[sortedPartionIDs[i]]) < len(partition2AllPotentialConsumers[sortedPartionIDs[j]])
 	})
 	return sortedPartionIDs
 }
@@ -805,22 +793,18 @@ func prepopulateCurrentAssignments(members map[string]ConsumerGroupMemberMetadat
 		}
 		sort.Sort(sort.Reverse(sort.IntSlice(generations)))
 
-		if len(generations) > 0 {
-			consumer := consumers[generations[0]]
-			currentConsumerAssignments, exists := currentAssignment[consumer]
-			if !exists {
-				currentConsumerAssignments = []topicPartitionAssignment{partition}
-			} else {
-				currentConsumerAssignments = append(currentConsumerAssignments, partition)
-			}
-			currentAssignment[consumer] = currentConsumerAssignments
+		consumer := consumers[generations[0]]
+		if _, exists := currentAssignment[consumer]; !exists {
+			currentAssignment[consumer] = []topicPartitionAssignment{partition}
+		} else {
+			currentAssignment[consumer] = append(currentAssignment[consumer], partition)
+		}
 
-			// check for previous assignment, if any
-			if len(generations) > 1 {
-				prevAssignment[partition] = consumerGenerationPair{
-					MemberID:   consumers[generations[1]],
-					Generation: generations[1],
-				}
+		// check for previous assignment, if any
+		if len(generations) > 1 {
+			prevAssignment[partition] = consumerGenerationPair{
+				MemberID:   consumers[generations[1]],
+				Generation: generations[1],
 			}
 		}
 	}
@@ -1038,7 +1022,6 @@ nextCand:
 }
 
 type consumerGroupMember struct {
-	index       int // the index of the consumer group member on the heap
 	id          string
 	assignments []topicPartitionAssignment
 }
@@ -1059,14 +1042,10 @@ func (pq assignmentPriorityQueue) Less(i, j int) bool {
 
 func (pq assignmentPriorityQueue) Swap(i, j int) {
 	pq[i], pq[j] = pq[j], pq[i]
-	pq[i].index = i
-	pq[j].index = j
 }
 
 func (pq *assignmentPriorityQueue) Push(x interface{}) {
-	n := len(*pq)
 	member := x.(*consumerGroupMember)
-	member.index = n
 	*pq = append(*pq, member)
 }
 
@@ -1074,7 +1053,6 @@ func (pq *assignmentPriorityQueue) Pop() interface{} {
 	old := *pq
 	n := len(old)
 	member := old[n-1]
-	member.index = -1 // for safety
 	*pq = old[0 : n-1]
 	return member
 }
--- a/balance_strategy.go
+++ b/balance_strategy.go
@@ -187,7 +187,7 @@ func (s *stickyBalanceStrategy) Plan(members map[string]ConsumerGroupMemberMetad
 	partition2AllPotentialConsumers := make(map[topicPartitionAssignment][]string)
 	for topic, partitions := range topics {
 		for _, partition := range partitions {
-			partition2AllPotentialConsumers[topicPartitionAssignment{Topic: topic, Partition: partition}] = make([]string, 0)
+			partition2AllPotentialConsumers[topicPartitionAssignment{Topic: topic, Partition: partition}] = []string{}
 		}
 	}
 
--- a/balance_strategy.go
+++ b/balance_strategy.go
@@ -6,6 +6,15 @@ import (
 )
 
 const (
+	// RangeBalanceStrategyName identifies strategies that use the range partition assignment strategy
+	RangeBalanceStrategyName = "range"
+
+	// RoundRobinBalanceStrategyName identifies strategies that use the round-robin partition assignment strategy
+	RoundRobinBalanceStrategyName = "roundrobin"
+
+	// StickyBalanceStrategyName identifies strategies that use the sticky-partition assignment strategy
+	StickyBalanceStrategyName = "sticky"
+
 	defaultGeneration = -1
 )
 
@@ -45,7 +54,7 @@ type BalanceStrategy interface {
 //   M1: {T: [0, 1, 2]}
 //   M2: {T: [3, 4, 5]}
 var BalanceStrategyRange = &balanceStrategy{
-	name: "range",
+	name: RangeBalanceStrategyName,
 	coreFn: func(plan BalanceStrategyPlan, memberIDs []string, topic string, partitions []int32) {
 		step := float64(len(partitions)) / float64(len(memberIDs))
 
@@ -63,7 +72,7 @@ var BalanceStrategyRange = &balanceStrategy{
 //   M1: {T: [0, 2, 4]}
 //   M2: {T: [1, 3, 5]}
 var BalanceStrategyRoundRobin = &balanceStrategy{
-	name: "roundrobin",
+	name: RoundRobinBalanceStrategyName,
 	coreFn: func(plan BalanceStrategyPlan, memberIDs []string, topic string, partitions []int32) {
 		for i, part := range partitions {
 			memberID := memberIDs[i%len(memberIDs)]
@@ -150,7 +159,7 @@ type stickyBalanceStrategy struct {
 }
 
 // Name implements BalanceStrategy.
-func (s *stickyBalanceStrategy) Name() string { return "sticky" }
+func (s *stickyBalanceStrategy) Name() string { return StickyBalanceStrategyName }
 
 // Plan implements BalanceStrategy.
 func (s *stickyBalanceStrategy) Plan(members map[string]ConsumerGroupMemberMetadata, topics map[string][]int32) (BalanceStrategyPlan, error) {
--- a/balance_strategy.go
+++ b/balance_strategy.go
@@ -5,6 +5,10 @@ import (
 	"sort"
 )
 
+const (
+	defaultGeneration = -1
+)
+
 // BalanceStrategyPlan is the results of any BalanceStrategy.Plan attempt.
 // It contains an allocation of topic/partitions by memberID in the form of
 // a `memberID -> topic -> partitions` map.
@@ -68,6 +72,19 @@ var BalanceStrategyRoundRobin = &balanceStrategy{
 	},
 }
 
+// BalanceStrategySticky assigns partitions to members with an attempt to preserve earlier assignments
+// while maintain a balanced partition distribution.
+// Example with topic T with six partitions (0..5) and two members (M1, M2):
+//   M1: {T: [0, 2, 4]}
+//   M2: {T: [1, 3, 5]}
+//
+// On reassignment with an additional consumer, you might get an assignment plan like:
+//   M1: {T: [0, 2]}
+//   M2: {T: [1, 3]}
+//   M3: {T: [4, 5]}
+//
+var BalanceStrategySticky = &stickyBalanceStrategy{}
+
 // --------------------------------------------------------------------
 
 type balanceStrategy struct {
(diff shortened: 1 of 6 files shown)


balance_strategy.go
```diff
@@ -127,3 +144,884 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+	isFreshAssignment := false
+	if len(currentAssignment) == 0 {
+		isFreshAssignment = true
+	}
+
+	// create a mapping of all current topic partitions and the consumers that can be assigned to them
+	partition2AllPotentialConsumers := make(map[topicPartitionAssignment][]string)
+	for topic, partitions := range topics {
+		for _, partition := range partitions {
+			partition2AllPotentialConsumers[topicPartitionAssignment{Topic: topic, Partition: partition}] = make([]string, 0)
```
varun06: Nitpick - if the slice has to be zero length, you can use slice literal `[]string{}`. Do you want it to be a 0 length slice or you can preallocate the slice?

varun06:
That's really nice, lot going on here, give me some time to read and review, looks great on a cursory look 👍

balance_strategy.go
```diff
@@ -127,3 +153,884 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+			opp := len(sortedMemberIDs) - 1 - i
+			sortedMemberIDs[i], sortedMemberIDs[opp] = sortedMemberIDs[opp], sortedMemberIDs[i]
+		}
+		for {
+			// loop until no consumer-group members remain
+			if len(sortedMemberIDs) == 0 {
+				break
+			}
+			updatedMemberIDs := make([]string, 0)
+			for _, memberID := range sortedMemberIDs {
```
twmb: I'm not sure this behavior is the same. The Java code loops on a heap here, continually polling the member with the most partitions remaining, or the lexicographically largest member id.

In Javaland, each loop removes one partition from the remaining partitions for a member, and then pushes that member back. Only if the member has no more partitions does it not get pushed back (missed optimization to not push back if there are no more members...).

Thus, if there were two members, M1 with 1 partition and M2 with 10, M2 would be used first every time. It has more partitions, and when it has one partition left, it has a higher member ID.

This code looks to use M2 once, then M1, and then M2 the remainder of the time (since there is nothing left in M1 and it is gone).

balance_strategy.go
```diff
@@ -127,3 +153,884 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+	for i, s := range sortedCurrentSubscriptions {
+		updatedSubscriptions[i] = s
+	}
+	i := 0
+	for _, memberID := range sortedCurrentSubscriptions {
+		if memberAssignmentsIncludeTopicPartition(consumer2AllPotentialPartitions[memberID], partition) {
+			updatedSubscriptions = removeIndexFromStringSlice(updatedSubscriptions, i)
+			currentAssignment[memberID] = append(currentAssignment[memberID], partition)
+			currentPartitionConsumer[partition] = memberID
+			updatedSubscriptions = append(updatedSubscriptions, memberID)
```
twmb: Same heap problem here. The Java line:
```java
                  currentAssignment.get(consumer).add(partition);
```
changes the sort order of the consumer in `sortedCurrentSubscriptions`, so the first line of these four removes the consumer from the heap and the fourth re-adds it to preserve ordering.

FWIW, this Java code is infuriating to read.

skidder:
@twmb thank you for the review! I agree with your feedback and have addressed your concerns in the commits ranging from 7f26806 to ae8e47f. I introduced a priority queue whose ordering is updated as assignments change. Please let me know if you have any additional suggestions.

balance_strategy.go
```diff
@@ -127,3 +155,926 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+	return sortedMemberIDs
+}
+
+func sortPartitionsByPotentialConsumerAssignments(partition2AllPotentialConsumers map[topicPartitionAssignment][]string) []topicPartitionAssignment {
+	// sort the members by the number of partition assignments in descending order
+	sortedPartionIDs := make([]topicPartitionAssignment, 0, len(partition2AllPotentialConsumers))
+	for partition := range partition2AllPotentialConsumers {
+		sortedPartionIDs = append(sortedPartionIDs, partition)
+	}
+	sort.SliceStable(sortedPartionIDs, func(i, j int) bool {
```
twmb: The Java code here has three levels of comparison:
1) number of consumers, which is what is done here
2) topic name
3) partition
```
 872         public int compare(TopicPartition o1, TopicPartition o2) {
 873             int ret = map.get(o1).size() - map.get(o2).size();
 874             if (ret == 0) {
 875                 ret = o1.topic().compareTo(o2.topic());
 876                 if (ret == 0)
 877                     ret = o1.partition() - o2.partition();
 878             }
 879             return ret;
 880         }
```
This misses the latter two. `SliceStable` is unnecessary once they're added in.

I think that this also sorts from smallest to largest, rather than what you do here of largest to smallest:
```
 545             TreeSet<TopicPartition> sortedAllPartitions = new TreeSet<>(new PartitionComparator(partition2AllPotentialConsumers));
 546             sortedAllPartitions.addAll(partition2AllPotentialConsumers.keySet());
 547 
 548             while (!sortedAllPartitions.isEmpty())
 549                 sortedPartitions.add(sortedAllPartitions.pollFirst());
```
The tree is drained first to last.

FWIW, I'm writing this code up myself and trying to translate the Java code to something that actually makes sense. That's taking a long time. I'm pretty suspicious of some aspects of the Java code. I'm not 100% sure everything the Java code is doing is necessary (for example, I can't think of a way that one map in `areSubscriptionsIdentical` would return true while the other returns false).

However, since this PR is looking to be a direct transliteration, I'll try to vet the accuracy of that transliteration the best I can while I go through the Java and ... try to make sense of it.

balance_strategy.go
```diff
@@ -127,3 +155,926 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+	// current and previous consumers are the last two consumers of each partition in the above sorted map
+	for partition, consumers := range sortedPartitionConsumersByGeneration {
+		// sort consumers by generation in decreasing order
+		var generations []int
+		for generation := range consumers {
+			generations = append(generations, generation)
+		}
+		sort.Sort(sort.Reverse(sort.IntSlice(generations)))
+
+		if len(generations) > 0 {
```
twmb: I'm pretty sure generations must be at least size 1 here, the partition wouldn't be in `sortedPartitionConsumersByGeneration`.

balance_strategy.go
```diff
@@ -127,3 +155,926 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+			consumer := consumers[generations[0]]
+			currentConsumerAssignments, exists := currentAssignment[consumer]
+			if !exists {
+				currentConsumerAssignments = []topicPartitionAssignment{partition}
+			} else {
+				currentConsumerAssignments = append(currentConsumerAssignments, partition)
+			}
+			currentAssignment[consumer] = currentConsumerAssignments
```
twmb: This block can be `currentAssignment[consumer] = append(currentAssignment[consumer], partition)`

balance_strategy.go
twmb: Note that the Java code is exceptionally confusing: if this is not a fresh assignment and all of the consumptions are identical, the sort order is by partitions with the most consumers to the partitions with the least consumers (and some prev consumers stuff).

In this else condition that does the sorting this comment chain is on, the Java code looks to sort by partitions with ascending number of consumers. Basically, the reverse of what it should be, solely because the Java code uses `pollFirst` instead of `pollLast`.

Am I reading that correctly?

twmb: I am pretty confused about the sortPartitions function in general. It operates in two completely different ways: the normal way, which is ordering partitions by how many consumers can consume from them, and the has-a-lot-of-logic-way, which orders by partitions that were on the most loaded consumer.

The two completely different behaviors makes the future usage of the sorting hard to reason about :/

balance_strategy.go
```diff
@@ -127,3 +155,926 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+			for partition := range partitionsWithADifferentPreviousAssignment {
+				// from partitions that had a different consumer before, keep only those that are assigned to this consumer now
+				if memberAssignmentsIncludeTopicPartition(member.assignments, partition) {
+					prevPartitions = append(prevPartitions, partition)
+				}
+			}
+
+			if len(prevPartitions) > 0 {
+				// if there is a partition on this consumer that was assigned to another consumer before mark it as good options for reassignment
+				partition := prevPartitions[0]
```
twmb: Frankly, I'm surprised that the Java code doesn't order `prevPartitions` nor `remainingPartitions`. They aren't yet ordered.

balance_strategy.go
```diff
@@ -127,3 +155,926 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+				prevPartitions = append(prevPartitions[:0], prevPartitions[1:]...)
```
twmb: `prevPartitions = prevPartitions[1:]` to avoid copying the entire array down.

balance_strategy.go
```diff
@@ -127,3 +155,926 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+				partition := member.assignments[0]
+				member.assignments = append(member.assignments[:0], member.assignments[1:]...)
+				sortedPartitions = append(sortedPartitions, partition)
+				heap.Fix(&pq, member.index)
+			} else {
+				heap.Remove(&pq, 0)
+			}
+		}
+
+		for partition := range partition2AllPotentialConsumers {
```
twmb: The Java code here is O(n^2), but we can make this O(n) by cloning partition2AllPotentialConsumers and deleting from it every time a partition is added to `sortedPartitions`. Everything remaining by this line in the clone can just simply be appended.

balance_strategy.go
twmb: However, there is no reason to update prevPartitions here.

balance_strategy.go
```diff
@@ -127,3 +155,926 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+		for {
+			// loop until no consumer-group members remain
+			if pq.Len() == 0 {
+				break
+			}
+			member := pq[0]
+
+			// partitions that were assigned to a different consumer last time
+			prevPartitions := make([]topicPartitionAssignment, 0)
+			for partition := range partitionsWithADifferentPreviousAssignment {
```
twmb: This double loop here can instead range first over member.assignments, looking if the assignment was used before. 
That'd change this to one loop and one map existence check.

Further, if it does exist in the map, we know which index of member.assignments needs to be removed, rather than re-looping through the assignments to remove it.

Optional. I see that this is trying to emulate exactly what the Java code does. Point being is that the Java code does it really poorly.

balance_strategy.go
```diff
@@ -127,3 +155,926 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+				prevPartitions = append(prevPartitions[:0], prevPartitions[1:]...)
+				member.assignments = removeTopicPartitionFromMemberAssignments(member.assignments, partition)
+				sortedPartitions = append(sortedPartitions, partition)
+				heap.Fix(&pq, member.index)
+			} else if len(member.assignments) > 0 {
+				// otherwise, mark any other one of the current partitions as a reassignment candidate
+				partition := member.assignments[0]
+				member.assignments = append(member.assignments[:0], member.assignments[1:]...)
+				sortedPartitions = append(sortedPartitions, partition)
+				heap.Fix(&pq, member.index)
```
twmb: No need to keep member index. The first item in the slice is the largest, which is how you declare `member` above to begin with. So, we know the index is zero here (and just above). Essentially, all uses of member.index can be replaced with 0 and the index field can be removed entirely.

balance_strategy.go
```diff
@@ -127,3 +155,926 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+			} else {
+				heap.Remove(&pq, 0)
```
twmb: `heap.Pop` may be clearer if one is thinking that he first is the largest?

balance_strategy.go
skidder: Great point, but I think we can also break out of the iteration over consumer partition assignments as soon we identify a single assignment that was previously on a different consumer. This eliminates the need for `prevPartitions` to be a slice, and can instead be a pointer that's evaluated if non-nil. Another example of how porting the Java implementation line-by-line exposes some of its inefficiencies.

balance_strategy.go
skidder: I converted `prevPartitions` to a pointer to a `topicPartition` instance since we only need to refer to a single `topicPartition` during each iteration of the loop.

skidder:
@twmb Thank you again! The Java implementation leaves a lot to be desired, and was obviously very challenging to port. Unfortunately, I absolutely must have a Golang implementation of sticky-partition-assignor, so I was left with no choice.

I appreciate your feedback, and would gladly merge in any contributions you have to this PR.

balance_strategy.go
```diff
@@ -127,3 +155,904 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+			currentPartitionConsumer[partition] = memberID
+		}
+	}
+
+	// sort the topic partitions in order of priority for reassignment
+	sortedPartitions := sortPartitions(currentAssignment, prevAssignment, isFreshAssignment, partition2AllPotentialConsumers, consumer2AllPotentialPartitions)
+	unassignedPartitions := deepCopyPartitions(sortedPartitions)
+	for memberID, partitions := range currentAssignment {
+		// if a consumer that existed before (and had some partition assignments) is now removed, remove it from currentAssignment
+		if _, exists := members[memberID]; !exists {
```
twmb: Unless I'm mistaken, it is impossible for the memberID here not to exist in members. `currentAssignment` is built from the member IDs and user datas in the members passed to plan. I can't think of a way that `currentAssignment` would have a member id that is not already in members.

balance_strategy.go
```diff
@@ -127,3 +155,904 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+			var prevPartitionIndex int
+			for i, partition := range member.assignments {
+				if _, exists := partitionsWithADifferentPreviousAssignment[partition]; exists {
+					prevPartition = &partition
+					prevPartitionIndex = i
+					break
+				}
+			}
+
+			if prevPartition != nil {
```
twmb: This code is fine as is, but it could be taken one step further by getting rid of prevPartition entirely and just banking on prevPartitionIndex. If there is no previous partition found, it'll be zero, which is the partition index used in the else case, which would mean that both cases can be merged.

twmb: Kinda something like this (from my in progress)...
```go
348                 usePartition := partitions[useIdx]
349                 partitionsSortedByMovePreference = append(partitionsSortedByMovePreference, usePartition)
350                 delete(unusedPartitions, usePartition)
351 
352                 if useIdx == 0 { // common case
353                         partitions = partitions[1:]
354                 } else { // less common: shift from right to preserve ordering
355                         partitions = append(partitions[:useIdx], partitions[useIdx+1:]...)
356                 }
357                 if len(partitions) > 0 {
358                         consumerWithMostPartitions.partitions = partitions
359                         heap.Fix(consumersByNumPartitions, 0)
360                 } else {
361                         heap.Pop(consumersByNumPartitions)
362                 }
```
^ that also avoids calling `Fix` when there are no partitions, avoiding a needless loop.

balance_strategy.go
```diff
@@ -127,3 +155,904 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+		}
+
+		// add this consumer to currentAssignment (with an empty topic partition assignment) if it does not already exist
+		if _, exists := currentAssignment[memberID]; !exists {
+			currentAssignment[memberID] = make([]topicPartitionAssignment, 0)
+		}
+	}
+
+	// create a mapping of each partition to its current consumer, where possible
+	currentPartitionConsumer := make(map[topicPartitionAssignment]string, len(currentAssignment))
```
twmb: I think the Java code and thus this transliteration from here all the way down to sortedCurrentSubscriptions on line 263 does more than necessary. I think this chunk can be similar to...
```go
167         // Maps all currently consumed partitions to the member that is
168         // currently consuming them.
169         currentPartitionConsumers := make(map[topicPartition]string)
170         // To build a list of unassigned partitions, we visit all partitions
171         // in the current plan and, if they still exist and the prior consumer
172         // no longer wants to consume them, we track it as unassigned.
173         // After, we add all new partitions.
174         unvisitedPartitions := partitions2AllPotentialConsumers.cloneKeys()
175         var unassignedPartitions []topicPartition
176 
177         for member, partitions := range currentPlan {
178                 var keepIdx int
179                 for _, partition := range partitions {
180                         // If this partition no longer exists at all, likely due to the
181                         // topic being deleted, we remove the partition from the member.
182                         if _, exists := partitions2AllPotentialConsumers[partition]; !exists {
183                                 continue
184                         }
185 
186                         delete(unvisitedPartitions, partition)
187                         currentPartitionConsumers[partition] = member
188 
189                         thisMemberIdx := sort.Search(len(members), func(i int) bool { return members[i].ID >= member })
190                         thisMember := members[thisMemberIdx]
191                         if !strsHas(thisMember.Topics, partition.topic) {
192                                 unassignedPartitions = append(unassignedPartitions, partition)
193                                 continue
194                         }
195 
196                         partitions[keepIdx] = partition
197                         keepIdx++
198                 }
199                 currentPlan[member] = partitions
200         }
201         for unvisited := range unvisitedPartitions {
202                 unassignedPartitions = append(unassignedPartitions, unvisited)
203         }
```
Types hopefully obvious.

The main point is that Java builds `unassignedPartitions` from `sortedPartitionConsumers`, which itself is built from of all of the keys in `partitions2AllPotentialConsumers`.

Rather than a bunch of logic to remove things that were added, we can just add as we noticed they are unassigned, with unvisited being everything that nothing was consuming (similar to that change in sortPartitions).

Also, rather than adding the prior assignment to currentPartitionConsumer up front, we can add as we know the partition still exists.

Other than my comment on the conditional below, I believe all of the current code here is correct, though. I'm still just trying to make sense of the Java code. I plan to enter the `balance` function tomorrow.

balance_strategy.go
```diff
@@ -127,3 +155,904 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+		}
+
+		// otherwise (the consumer still exists)
+		updatedPartitions := deepCopyPartitions(partitions)
+		for _, partition := range partitions {
+			if _, exists := partition2AllPotentialConsumers[partition]; !exists {
+				// if this topic partition of this consumer no longer exists remove it from currentAssignment of the consumer
+				updatedPartitions = removeTopicPartitionFromMemberAssignments(updatedPartitions, partition)
+				delete(currentPartitionConsumer, partition)
+			} else if _, exists := topics[partition.Topic]; !exists {
```
twmb: I believe this conditional is different from the Java one, and that this conditional is not correct.

The Java conditional checks whether the member (for the member ID that is currently being iterated over) is _still_ asking for the topic that it did prior:
```java
!subscriptions.get(entry.getKey()).topics().contains(partition.topic())
```
"Does the topic set for the member in this join group request still contain the topic that this partition it was consuming prior is in?" i.e. is it still consuming what it was.

The Go conditional here, by using `topics`, is asking "Does the client itself still know of this partition's topic?"

That is the same as the prior case that checks `partition2AllPotentialConsumers`, since that map is initialized with all topic/partitions that exist in the client:
```go
 187         partition2AllPotentialConsumers := make(map[topicPartitionAssignment][]string)
 188         for topic, partitions := range topics {
 189                 for _, partition := range partitions {
 190                         partition2AllPotentialConsumers[topicPartitionAssignment{Topic: topic, Partition: partition}] = []string{}
 191                 }
 192         }
```

skidder: You're totally right, there was zero test coverage that executed the block associated with this conditional. I've updated the expression to check for the currently-evaluated topic in the set of topics subscribed to by the consumer. I also added some test coverage for this specific condition.

balance_strategy.go
skidder: Thank you for the code suggestion. I adapted it to work with this branch without much difficulty.

Note that I added a slice of consumer partition assignments to keep since the code that you provided doesn't actually remove a partition from the `partitions` slice when it's no longer subscribed to or doesn't exist.

twmb: Hrm I thought my method slid partitions that were kept down, so that in the end after the loop, `partitions` would consist of only partitions to keep. Sort of like what you currently do in `filterAssignedPartitions`.

I do see that I forgot to trim the resulting slice to `keepIdx`, though, so a fix in mine would be `currentPlan[member] = partitions[:keepIdx]`.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+		if len(partition2AllPotentialConsumers[partition]) == 0 {
+			continue
+		}
+		sortedCurrentSubscriptions = assignPartition(partition, sortedCurrentSubscriptions, currentAssignment, consumer2AllPotentialPartitions, currentPartitionConsumer)
+	}
+
+	// narrow down the reassignment scope to only those partitions that can actually be reassigned
+	for partition := range partition2AllPotentialConsumers {
+		if !canTopicPartitionParticipateInReassignment(partition, partition2AllPotentialConsumers) {
+			sortedPartitions = removeTopicPartitionFromMemberAssignments(sortedPartitions, partition)
```
twmb: More fun, if you're up for it. This entire loop can be removed by just not appending fixed partitions to sortedPartitions in the first place while building sorted partitions. All this is doing here is checking if the partition has at least two consumers and if not, removing it from sortedPartitions.

However, again, optional. This loop is pot on accurate to what the Java code does.

twmb: It may be clearer here to keep the fixed stuff. I'm pretty sure fixedAssignments can also be done away with and moved to Balance, but I'm not sure if doing either of these "improvements" would make the code any clearer, so probably best to keep as is.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+		}
+	}
+
+	// narrow down the reassignment scope to only those consumers that are subject to reassignment
+	fixedAssignments := make(map[string][]topicPartitionAssignment)
+	for memberID := range consumer2AllPotentialPartitions {
+		if !canConsumerParticipateInReassignment(memberID, currentAssignment, consumer2AllPotentialPartitions, partition2AllPotentialConsumers) {
+			fixedAssignments[memberID] = currentAssignment[memberID]
+			delete(currentAssignment, memberID)
+			sortedCurrentSubscriptions = sortMemberIDsByPartitionAssignments(currentAssignment)
```
twmb: It'd be better for you to move this sort to after the loop, since inside the loop, it's basically throwing away work every iteration.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+	// narrow down the reassignment scope to only those partitions that can actually be reassigned
+	for partition := range partition2AllPotentialConsumers {
+		if !canTopicPartitionParticipateInReassignment(partition, partition2AllPotentialConsumers) {
+			sortedPartitions = removeTopicPartitionFromMemberAssignments(sortedPartitions, partition)
+		}
+	}
+
+	// narrow down the reassignment scope to only those consumers that are subject to reassignment
+	fixedAssignments := make(map[string][]topicPartitionAssignment)
+	for memberID := range consumer2AllPotentialPartitions {
```
twmb: I know the Java code does it, but there's no reason not to range over `currentAssignment` here, which has all of the same keys as `consumer2AllPotentialPartitions`. Then you have access to the value and can pass it to `canConsumerParticipateInReassignment`, rather than looking it up from `currentAssignment` first thing in the code.

I think?

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+		if !canConsumerParticipateInReassignment(memberID, currentAssignment, consumer2AllPotentialPartitions, partition2AllPotentialConsumers) {
+			fixedAssignments[memberID] = currentAssignment[memberID]
+			delete(currentAssignment, memberID)
+			sortedCurrentSubscriptions = sortMemberIDsByPartitionAssignments(currentAssignment)
+		}
+	}
+
+	// create a deep copy of the current assignment so we can revert to it if we do not get a more balanced assignment later
+	preBalanceAssignment := deepCopyAssignment(currentAssignment)
+	preBalancePartitionConsumers := make(map[topicPartitionAssignment]string, len(currentPartitionConsumer))
```
twmb: I can't see a purpose for this `preBalancePartitionConsumers` map at all. It's populated here, and it _might_ be used to reset `currentPartitionConsumer` below, but `currentPartitionConsumer` is unused after that reset and `Balance` does nothing for it once `balance` returns.

As far as I can tell, the Java code here is just wasting time and objects.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+	sortedCurrentSubscriptions = sortMemberIDsByPartitionAssignments(currentAssignment)
+	min := len(currentAssignment[sortedCurrentSubscriptions[0]])
+	max := len(currentAssignment[sortedCurrentSubscriptions[len(sortedCurrentSubscriptions)-1]])
+	if min >= max-1 {
+		// if minimum and maximum numbers of partitions assigned to consumers differ by at most one return true
+		return true
+	}
+
+	// create a mapping from partitions to the consumer assigned to them
+	allPartitions := make(map[topicPartitionAssignment]string)
```
twmb: I can't think of a reason to not pass `currentPartitionConsumer` to this function and avoid this exact recreation of it.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+	return sortedPartitions
+}
+
+func sortMemberIDsByPartitionAssignments(assignments map[string][]topicPartitionAssignment) []string {
+	// sort the members by the number of partition assignments in ascending order
+	sortedMemberIDs := make([]string, 0, len(assignments))
+	for memberID := range assignments {
+		sortedMemberIDs = append(sortedMemberIDs, memberID)
+	}
+	sort.SliceStable(sortedMemberIDs, func(i, j int) bool {
```
twmb: No reason for SliceStable since member IDs are unique.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+		ret := len(assignments[sortedMemberIDs[i]]) - len(assignments[sortedMemberIDs[j]])
+		if ret == 0 {
+			return sortedMemberIDs[i] < sortedMemberIDs[j]
+		}
+		return len(assignments[sortedMemberIDs[i]]) < len(assignments[sortedMemberIDs[j]])
```
twmb: `return ret < 0`?

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+
+		// skip if this consumer already has all the topic partitions it can get
+		if consumerPartitionCount == len(allSubscriptions[memberID]) {
+			continue
+		}
+
+		// otherwise make sure it cannot get any more
+		potentialTopicPartitions := allSubscriptions[memberID]
+		for _, partition := range potentialTopicPartitions {
+			if !memberAssignmentsIncludeTopicPartition(currentAssignment[memberID], partition) {
```
twmb: I think this should use `allPartitions` (or my suggested `currentPartitionConsumer`). Rather than an O(n) range over all of the member's current partitions, we already have a direct lookup map. I mean, if the conditional here returns false, that's what the next line literally does.

That line could be above this if and then the if could be `if currentConsumer != memberID` or something.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+				otherConsumer := allPartitions[partition]
+				otherConsumerPartitionCount := len(currentAssignment[otherConsumer])
+				if consumerPartitionCount < otherConsumerPartitionCount {
```
twmb: I'm fairly certain the Java code here _should_ check `consumerPartitionCount < otherConsumerPartitionCount - 1`. A delta of one is fine, as was checked at the beginning of this function. Otherwise, moving the partition from the other to this one would create the same balance in reverse.

But,

twmb: Well, I guess it's fine. Even if this results in a partition moving from a member with 3 to a member with 2, the worst that can happen is the loop in `performReassignments` will continue unnecessarily.

If I'm reading it correctly, the one-off imbalance could cause performReassignments to loop across all reassignable partitions no matter what. If a plan has the two fewest partition members A and B that can have partitions 1, 2, and 3, if they look like

A: 1,
B: 2, 3

As a further constraint, they can take no more partitions.

That condition would cause the first conditional in `performReassignments` to always return true.

But, the loop stops if a modification is made, so that's good.

The more I think about it, I think just checking `consumerPartitionCount ... other - 1` would be flawed as well, and in a potentially worse way.

I think what would be an accurate check here would be, in addition to the current check (not my modification), to also compare the sum of both member's partitions against the union of the potential partitions they could consume from.

This would return false on what I suggested above, but would return true if B could consume from a fourth partition but isn't yet.

But, that union suggestion may be overkill, and if this is going for parity with Java, the Java stuff doesn't do that.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+				Logger.Printf("A cycle of length %d was found: %v", len(path)-1, path)
+			}
+		}
+	}
+
+	// for now we want to make sure there is no partition movements of the same topic between a pair of consumers.
+	// the odds of finding a cycle among more than two consumers seem to be very low (according to various randomized
+	// tests with the given sticky algorithm) that it should not worth the added complexity of handling those cases.
+	for _, cycle := range cycles {
+		if len(cycle) == 3 {
```
twmb: Honestly, I can't make sense of what this cycle detection is doing, but it seems convoluted, runtime expensive, and weak. I think real cycle detection should just build a quick directed graph and run the strongly connected components algorithm on it. 

Again, this comment is more about the Java code and what it is even doing, if anything. Like, what is that supercycle stuff above doing? What's the runtime of isSticky? O(n^n^n)?

However, it looks like this isSticky check is a faithful translation of the Java code, so it has that :+1:.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+
+	// determine if we're dealing with a completely fresh assignment, or if there's existing assignment state
+	isFreshAssignment := false
+	if len(currentAssignment) == 0 {
+		isFreshAssignment = true
+	}
+
+	// create a mapping of all current topic partitions and the consumers that can be assigned to them
+	partition2AllPotentialConsumers := make(map[topicPartitionAssignment][]string)
+	for topic, partitions := range topics {
```
twmb: I am fairly certain that this loop is unneeded and causes the code to change the sort order unnecessarily, but the tests work with or without this loop and this is faithful to the Java.

This code sets all partitions in the client as candidates for moving, even if no member is interested in some topics. This causes sortPartitions to unnecessarily fall into the into the second case where it considers some subscriptions are different. I deleted these 5 lines on your branch and ran the tests and everything passes. Even deleting the five lines, the second case is hit the same amount. Most of the tests revolve around consumer group topic subscription disparities.

:) 
:(

balance_strategy.go
twmb: FWIW I tested this union idea in this branch; it kicks in three times during the tests but doesn't change the test results. I do notice that some code in the partitionMovements methods below does some back-and-forth checking. Worst case, Java does a move-once type of scenario when it doesn't need to, and even that may be mitigated by some other duct tape.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+				SrcMemberID: existingPair.SrcMemberID,
+				DstMemberID: newConsumer,
+			})
+		}
+	} else {
+		p.addPartitionMovementRecord(partition, pair)
+	}
+}
+
+func (p *partitionMovements) getTheActualPartitionToBeMoved(partition topicPartitionAssignment, oldConsumer, newConsumer string) topicPartitionAssignment {
```
twmb: It might be worthwhile to document the purpose of this function, which is undocumented in Java as well, but the behavior is pretty confusing.

If this is trying to move a partition from A to B, if we have already moved a partition from B to A, this returns _that_ partition so that the original move is undone. Effectively, this function can counteract some there-and-back again movements. No test yet triggers this. I'd be curious to see something that did.

The partition has to be in the same topic because if it isn't, then maybe the other consumer is no longer interested in the topic.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+func memberAssignmentsIncludeTopicPartition(assignments []topicPartitionAssignment, topic topicPartitionAssignment) bool {
+	for _, assignment := range assignments {
+		if assignment == topic {
+			return true
+		}
+	}
+	return false
+}
+
+func sortPartitions(currentAssignment map[string][]topicPartitionAssignment, partitionsWithADifferentPreviousAssignment map[topicPartitionAssignment]consumerGenerationPair, isFreshAssignment bool, partition2AllPotentialConsumers map[topicPartitionAssignment][]string, consumer2AllPotentialPartitions map[string][]topicPartitionAssignment) []topicPartitionAssignment {
```
twmb: After staring at `performReassignments` for a while, I'm pretty sure this entire function is useless. This entire thing can be deleted and replaced with just returning the keys of partitions2AllPotentialConsumers and the tests pass.

Really, the only benefit of this function is if all group subscriptions are identical. In that case, the return ordering is actually meaningful. `performReassignments` will start from partitions that are on overloaded members, and thus it may find balance sooner. Even the loss of that ordering can be largely mitigated by saving if all subscriptions are identical and then using that in `isBalanced`: if everything is identical, the max imbalance should be one, and the remainder of the function can be skipped.

If there's any disparity of consuming, the return is pretty useless: it sorts by partitions with the least amount of potential consumers. That does not help reassigning.

twmb: For further context, even if all subscriptions are identical, the only aid is in _maybe_ less looping. `performReassignments` _always_ visits all partitions, left to right. If a partition is on an overloaded member, it gets moved. If a move happens, _all_ "sorted" partitions are revisited left to right. After the first round of moves, sortedPartitions is no longer sorted, yet reassigning still happens in the original order.

I'm honestly not quite sure what sorting here buys at all.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+	reversePair := consumerPair{
+		SrcMemberID: newConsumer,
+		DstMemberID: oldConsumer,
+	}
+	if _, exists := partitionMovementsForThisTopic[reversePair]; !exists {
+		return partition
+	}
+	var reversePairPartition topicPartitionAssignment
+	for otherPartition := range partitionMovementsForThisTopic[reversePair] {
+		reversePairPartition = otherPartition
```
twmb: `return otherPairPartition`. No need to iterate through all.

Can panic outside of the loop, which should be unreachable.

twmb:
I've covered most of balance_strategy.go. What exists looks to be a faithful recreation of the Java (although `sortedCurrentSubscriptions` is a red-black tree in Java making for more efficient maintenance). I skimmed the partition movement stuff; that stuff isn't too tricky. I think the Java code took probably the most roundabout way possible to do what it set out to do, so there are some things that can be cleaned up here but that isn't critical considering this intends to be a near 1:1 recreation of the Java.

I don't plan to look at much else, and none of my current comments are blockers, so afaict, lgtm.

balance_strategy_test.go
```diff
@@ -100,3 +105,2198 @@ func TestBalanceStrategyRoundRobin(t *testing.T) {
...
+	verifyFullyBalanced(t, plan3)
+}
+func Test_stickyBalanceStrategy_Plan_AssignmentWithConflictingPreviousGenerations(t *testing.T) {
+	s := &stickyBalanceStrategy{}
+
+	topics := map[string][]int32{"topic1": []int32{0, 1, 2, 3, 4, 5}}
+	members := make(map[string]ConsumerGroupMemberMetadata, 3)
+	members["consumer1"] = ConsumerGroupMemberMetadata{
+		Topics:   []string{"topic1"},
+		UserData:
```
twmb: This can take `testing.TB` and then the duplicate function just below can be eliminated.

balance_strategy.go
twmb: When all can consume the same, it may help in the case of incoming joins with prior plans
```
A -> 1
B-> 2, 3, 4
C -> 5, 6, 7, 8, 9
```
Sorted would say `9, 8, 7, 4, 6, 3, 5, 2, 1` in order of move, which would result in
```
A -> 1, 9, 8
B -> 2, 3, 4
C -> 5, 6, 7
```
and be most sticky,
whereas no sorting would be `1, 2, 3, 4, 5, 6, 7, 8, 9` order of move, resulting in
```
A -> 1, 2, 5
B -> 3, 4, 6
C -> 7, 8, 9
```
being destructive to B when it shouldn't have. I'll see if there's a test case similar to this.

I'm still mixed on how beneficial sorting is for non-equal consuming and will try to think of a scenario there.

twmb: This test does that:
```
                {
                        name: "Three consumers with a disparity coming into harmony.",
                        args: args{
                                members: map[string]ConsumerGroupMemberMetadata{
                                        "c1": ConsumerGroupMemberMetadata{
                                                Topics:   []string{"t1"},
                                                UserData: encodeSubscriberPlanWithGeneration(t, map[string][]int32{"t1": []int32{0}}, 1),
                                        },
                                        "c2": ConsumerGroupMemberMetadata{
                                                Topics:   []string{"t1"},
                                                UserData: encodeSubscriberPlanWithGeneration(t, map[string][]int32{"t1": []int32{1, 2, 3}}, 1),
                                        },   
                                        "c3": ConsumerGroupMemberMetadata{
                                                Topics:   []string{"t1"},
                                                UserData: encodeSubscriberPlanWithGeneration(t, map[string][]int32{"t1": []int32{4, 5, 6, 7, 8}}, 1),
                                        },   
                                },   
                                topics: map[string][]int32{
                                        "t1": []int32{0, 1, 2, 3, 4, 5, 6, 7, 8},
                                },
                        },   
                },
```
And printing the plan with and without sortPartitions shows the problem. Unfortunately, the test code doesn't catch the lack of stickiness without the sorting:
```
=== RUN   Test_stickyBalanceStrategy_Plan/Three_consumers_with_a_disparity_coming_into_harmony.
c1 => map[t1:[0 2 6]]
c2 => map[t1:[1 3 8]]
c3 => map[t1:[7 5 4]]
```
That should fail, since c2 should not have changed.

twmb: As an example of why the second sorting is useless:
```go
                {
                        // A -> 1, can take all
                        // B -> 2, 3, 4
                        // C -> 5, 6, 7, 8, 9
                        //
                        // Bad would be stealing a partition from B:
                        // A -> 1, 2, 5
                        // B -> 3, 4
                        // C -> 6, 7, 8, 9
                        //
                        // Ideal:
                        // A -> 1, 5, 6
                        // B -> 2, 3, 4
                        // C -> 7, 8, 9
                        name: "Bigly disbalancy.",
                        args: args{
                                members: map[string]ConsumerGroupMemberMetadata{
                                        "A": ConsumerGroupMemberMetadata{
                                                Topics: []string{"1", "2", "3", "4", "5", "6", "7", "8", "9"},
                                                UserData: encodeSubscriberPlanWithGeneration(t,
                                                        map[string][]int32{
                                                                "1": []int32{0},
                                                        }, 1),
                                        },
                                        "B": ConsumerGroupMemberMetadata{
                                                Topics: []string{"2", "3", "4"},
                                                UserData: encodeSubscriberPlanWithGeneration(t,
                                                        map[string][]int32{
                                                                "2": []int32{0},
                                                                "3": []int32{0},
                                                                "4": []int32{0},
                                                        }, 1),
                                        },
                                        "C": ConsumerGroupMemberMetadata{
                                                Topics: []string{"5", "6", "7", "8", "9"},
                                                UserData: encodeSubscriberPlanWithGeneration(t,
                                                        map[string][]int32{
                                                                "5": []int32{0},
                                                                "6": []int32{0},
                                                                "7": []int32{0},
                                                                "8": []int32{0},
                                                                "9": []int32{0},
                                                        }, 1),
                                        },   
                                },
                                topics: map[string][]int32{
                                        "1": []int32{0},
                                        "2": []int32{0},
                                        "3": []int32{0},
                                        "4": []int32{0},
                                        "5": []int32{0},
                                        "6": []int32{0},
                                        "7": []int32{0},
                                        "8": []int32{0},
                                        "9": []int32{0},
                                },
                        },
```
That causes a test failure. I think the heap sort is really the only beneficial sort: we should always try to move partitions off of loaded members. I'll try to think of another example.

twmb: Here's another test that fails with either heap strategy:
```
                {
                        // A -> 1, [in all]
                        // B -> 2, [in 2, 3, 4]
                        // D -> 3, 4, 5, 6, 7, 8, 9
                        //
                        // A -> 1, 9, 8, 7
                        // B -> 2,
                        // D -> 3, 4, 5, 6
                        //
                        // A -> 1, 9, 8, 7
                        // B -> 2, 3
                        // D -> 4, 5, 6
                        //
                        // Ideal:
                        // A -> 1, 9, 8
                        // B -> 2, 3, 4
                        // C -> 5, 6, 7
                        name: "Bigly disbalancy.",
                        args: args{
                                members: map[string]ConsumerGroupMemberMetadata{
                                        "A": ConsumerGroupMemberMetadata{
                                                Topics: []string{"1", "2", "3", "4", "5", "6", "7", "8", "9"},
                                                UserData: encodeSubscriberPlanWithGeneration(t,
                                                        map[string][]int32{
                                                                "1": []int32{0},
                                                        }, 1),
                                        },
                                        "B": ConsumerGroupMemberMetadata{
                                                Topics: []string{"2", "3", "4"},
                                                UserData: encodeSubscriberPlanWithGeneration(t,
                                                        map[string][]int32{
                                                                "2": []int32{0},
                                                        }, 1),
                                        },
                                        "C": ConsumerGroupMemberMetadata{
                                                Topics: []string{"3", "4", "5", "6", "7", "8", "9"},
                                                UserData: encodeSubscriberPlanWithGeneration(t,
                                                        map[string][]int32{
                                                                "3": []int32{0},
                                                                "4": []int32{0},
                                                                "5": []int32{0},
                                                                "6": []int32{0},
                                                                "7": []int32{0},
                                                                "8": []int32{0},
                                                                "9": []int32{0},
                                                        }, 1),
                                        },
                                },
                                topics: map[string][]int32{
                                        "1": []int32{0},
                                        "2": []int32{0},
                                        "3": []int32{0},
                                        "4": []int32{0},
                                        "5": []int32{0},
                                        "6": []int32{0},
                                        "7": []int32{0},
                                        "8": []int32{0},
                                        "9": []int32{0},
                                },
                        },
                },
```
I think maybe a better strategy would be to always try reassigning _from_ the consumer with the least partitions, and iterating over all partitions that could be assigned to it. I'll try that out.

balance_strategy.go
```diff
@@ -127,3 +155,899 @@ func balanceStrategyHashValue(vv ...string) uint32 {
...
+// Reassign a specific partition to a new consumer
+func (s *stickyBalanceStrategy) reassignPartition(partition topicPartitionAssignment, currentAssignment map[string][]topicPartitionAssignment, sortedCurrentSubscriptions []string, currentPartitionConsumer map[topicPartitionAssignment]string, newConsumer string) []string {
+	consumer := currentPartitionConsumer[partition]
+	// find the correct partition movement considering the stickiness requirement
+	partitionToBeMoved := s.movements.getTheActualPartitionToBeMoved(partition, consumer, newConsumer)
+	return s.processPartitionMovement(partitionToBeMoved, newConsumer, currentAssignment, sortedCurrentSubscriptions, currentPartitionConsumer)
+}
+
+// Track the movement of a topic partition after assignment
+func (s *stickyBalanceStrategy) processPartitionMovement(partition topicPartitionAssignment, newConsumer string, currentAssignment map[string][]topicPartitionAssignment, sortedCurrentSubscriptions []string, currentPartitionConsumer map[topicPartitionAssignment]string) []string {
```
twmb: The Java code here reorders sortedCurrentSubscriptions, whereas this misses the resorting.

balance_strategy.go
twmb: Alright, I have a scheme that solves all three cases above:

1) always sort by partitions-on-most-loaded-member
2) In `performReassignments`, start from the least loaded member (with sortedCurrentSubscriptions)
3) look for the _first_ partition in `reassignablePartitions` that it can take,
4) if the current consumer of that partition is not itself, and the current consumer is more loaded by more than 1, take the partition

I think this also largely obviates the need for `prevAssignment`.

twmb: After more playing around, I think something like this is ideal:
```go
411         for {
 412                 modified = false
 413                 // reassign all reassignable partitions (starting from the partition with least potential consumers and if needed)
 414                 // until the full list is processed or a balance is achieved
 415                 if isBalanced(currentAssignment, sortedCurrentSubscriptions, consumer2AllPotentialPartitions) {
 416                         break
 417                 }
 418                 for _, leastLoaded := range sortedCurrentSubscriptions {
 419                         potentialPartitions := consumer2AllPotentialPartitions[leastLoaded]
 420                         myConsumptions := len(currentAssignment[leastLoaded])
 421                         var mostConsumptions int
 422                         var mostPartition topicPartitionAssignment
 423                         for _, partition := range potentialPartitions {
 424                                 currentConsumer := currentPartitionConsumer[partition]
 425                                 if currentConsumer == leastLoaded {
 426                                         continue
 427                                 }
 428 
 429                                 currentConsumptions := len(currentAssignment[currentConsumer])
 430                                 if currentConsumptions > myConsumptions {
 431                                         if currentConsumptions > mostConsumptions {
 432                                                 mostConsumptions = currentConsumptions
 433                                                 mostPartition = partition
 434                                         }
 435                                 }
 436                         }
 437                         if mostConsumptions > 0 {
 438                                 sortedCurrentSubscriptions = s.reassignPartitionToNewConsumer(mostPartition, currentAssignment, sortedCurrentSubscriptions, currentPartitionConsumer, con     sumer2AllPotentialPartitions)   
 439                                 reassignmentPerformed = true
 440                                 modified = true
 441                                 break
 442                         }
 443                 }
 444                 if !modified {
 445                         return reassignmentPerformed
 446                 }       
```

I don't expect you to use this, but I do think it provides optimal balancing as well as stickiness. Modulo some bugs, that's the strategy I plan to use in my WIP client.

balance_strategy_test.go
```diff
@@ -100,3 +105,2198 @@ func TestBalanceStrategyRoundRobin(t *testing.T) {
...
+}
+
+func Test_stickyBalanceStrategy_Plan(t *testing.T) {
+	type args struct {
+		members map[string]ConsumerGroupMemberMetadata
+		topics  map[string][]int32
+	}
+	tests := []struct {
+		name string
+		s    *stickyBalanceStrategy
```
twmb: This is unused.

twmb:
AFAICT this LGTM.

I've removed some of my comments about potential ideas that may be better; I've been toying around with those ideas and they're better in some cases but not perfect and they are complicated.

I'm 100% sure that the algorithm used in Java does not accomplish its original stated goal:
```
The sticky partition assignment algorithm described above, as mentioned earlier in this KIP, favors fairness over stickiness (we may call it the fair yet sticky or stickiest fair strategy). Therefore, some partitions may change their consumer towards a fair assignment. This strategy is supposedly more complex to implement due to complex nature of calculating the most balanced assignment.
```
I've devised multiple counterexamples, and while some ideas I'm toying with get some of those edge cases, I can't think of a perfect solution yet.

I have a few remaining comments about potential cleanup, but nothing absolutely critical to address. I've also stared at the `isSticky` function (used only in tests) for a while and I don't think it accomplishes what it intends to do. The main thing that function actually does is check that no two partitions within the same topic swap members, but "stickiness" is much more than that. But, I'm not sure if there even can be a good test for stickiness.

But, as far as an exact recreation of the Java behavior, I think this does it.

balance_strategy.go
twmb: A more complicated one:
```
                {
                        // Start:
                        // A -> 1 2
                        // B -> 3 4
                        // C -> 5
                        // D -> a b c d e
                        // E ->
                        //
                        // A takes all,
                        // B takes 1-5
                        // C takes 3-5
                        // D, E take a-e
                        //
                        // Ideal:
                        // A -> 1 e
                        // B -> 2 3
                        // C -> 4 5
                        // D -> a b
                        // E -> c d
                        name: "Complicated",
                        args: args{
                                members: map[string]ConsumerGroupMemberMetadata{
                                        "A": ConsumerGroupMemberMetadata{
                                                Topics: []string{"1", "2", "3", "4", "5", "a", "b", "c", "d", "e"},
                                                UserData: encodeSubscriberPlanWithGeneration(t,
                                                        map[string][]int32{
                                                                "1": []int32{0},
                                                                "2": []int32{0},
                                                        }, 1),
                                        },
                                        "B": ConsumerGroupMemberMetadata{
                                                Topics: []string{"1", "2", "3", "4", "5"},
                                                UserData: encodeSubscriberPlanWithGeneration(t,
                                                        map[string][]int32{
                                                                "3": []int32{0},
                                                                "4": []int32{0},
                                                        }, 1),
                                        },
                                        "C": ConsumerGroupMemberMetadata{
                                                Topics: []string{"3", "4", "5"},
                                                UserData: encodeSubscriberPlanWithGeneration(t,
                                                        map[string][]int32{
                                                               "4": []int32{0},
                                                        }, 1),
                                        },
                                        "D": ConsumerGroupMemberMetadata{
                                                Topics: []string{"a", "b", "c", "d", "e"},
                                                UserData: encodeSubscriberPlanWithGeneration(t,
                                                        map[string][]int32{
                                                                "a": []int32{0},
                                                                "b": []int32{0},
                                                                "c": []int32{0},
                                                                "d": []int32{0},
                                                                "e": []int32{0},
                                                        }, 1),
                                        },
                                        "E": ConsumerGroupMemberMetadata{
                                                Topics: []string{"a", "b", "c", "d", "e"},
                                                UserData: encodeSubscriberPlanWithGeneration(t,
                                                        map[string][]int32{}, 1),
                                        },
                                },
                                topics: map[string][]int32{
                                        "1": []int32{0},
                                        "2": []int32{0},
                                        "3": []int32{0},
                                        "4": []int32{0},
                                        "5": []int32{0},
                                        "6": []int32{0},
                                        "7": []int32{0},
                                        "8": []int32{0},
                                        "9": []int32{0},
                                        "a": []int32{0},
                                        "b": []int32{0},
                                        "c": []int32{0},
                                        "d": []int32{0},
                                        "e": []int32{0},
                                },
                        },
```

twmb: FWIW, I figured out a way to solve all of these edge cases and implemented it here: https://github.com/twmb/kgo/tree/master/internal/sticky

The key idea is representing the members and partitions they want as a graph with edges, and then using A* search to steal appropriately.

This implementation is faster, more accurate (for the balanced-metric, and I think optimally accurate), uses less garbage, and is cleaner than the Java code.

twmb: After optimizing, the implementation in the above link is >14x faster.

bai:
This is great, thanks for contributing and also thanks for reviewing 👍

PR #1425, opened 2019-07-11
fix deleteRecord  

evazca:
The pr is to fix a bug about delete records. Currently sarama send the request to the controller.  It should be sent to the leader according the code in kafka client.

related with https://github.com/Shopify/sarama/issues/1377

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -374,29 +374,50 @@ func (ca *clusterAdmin) DeleteRecords(topic string, partitionOffsets map[int32]i
 	if topic == "" {
 		return ErrInvalidTopic
 	}
-
-	topics := make(map[string]*DeleteRecordsRequestTopic)
-	topics[topic] = &DeleteRecordsRequestTopic{PartitionOffsets: partitionOffsets}
-	request := &DeleteRecordsRequest{
-		Topics:  topics,
-		Timeout: ca.conf.Admin.Timeout,
+	partitionPerBroker := make(map[*Broker][]int32)
+	for partition := range partitionOffsets {
+		broker, err := ca.client.Leader(topic, partition)
+		if err != nil {
+			return err
+		}
+		if _, ok := partitionPerBroker[broker]; ok {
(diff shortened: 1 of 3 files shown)


errors.go
```diff
@@ -81,6 +81,28 @@ func (err ConfigurationError) Error() string {
 // See https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ErrorCodes
 type KError int16
 
+// MultiError is used to contain multi error.
+type MultiError struct {
```
varun06: What's the reason to add this?

admin.go
```diff
@@ -374,29 +374,50 @@ func (ca *clusterAdmin) DeleteRecords(topic string, partitionOffsets map[int32]i
...
-		Topics:  topics,
-		Timeout: ca.conf.Admin.Timeout,
-	}
-
-	b, err := ca.Controller()
-	if err != nil {
-		return err
+	partitionPerBroker := make(map[*Broker][]int32)
+	for partition := range partitionOffsets {
+		broker, err := ca.client.Leader(topic, partition)
```
varun06: formatting looks strange, can you please run gofmt again?

bai:
Thanks! What @varun06 said; apart from that could you please squash your commits 🙏

admin.go
evazca: gofmt again but nothing changed. Can you please tell me what's the format problem. I'll fix it manually.

errors.go
evazca: Currently the delete record request is sent to leader by leader. Thus we may receive multi error response and probably the errors are not the same.  I use MultiError to contain the array of error.

evazca:
> Thanks! What @varun06 said; apart from that could you please squash your commits 🙏

I have rebased the commit & push again.  But I'm not sure whether it is ok to do it right now and shall I squash the commit again after the code review finished.

PR #1442, opened 2019-07-31
fix brokerProducer goroutine leak

dnwe:
Every create a of a brokerProducer spins off a goroutine for the run()
loop func, but this appeared to just be abandoned if the brokerProducer
was shutdown and would manifest as orphaned goroutine(s) with a long
select time leaking over time:

e.g.,
```
goroutine 6982 [select, 1868 minutes]:
github.com/Shopify/sarama.(*brokerProducer).run(0xc420c4ac00)
	/home/travis/build/org/repo/vendor/src/github.com/Shopify/sarama/async_producer.go:672 +0x258
github.com/Shopify/sarama.(*brokerProducer).(github.com/Shopify/sarama.run)-fm()
	/home/travis/build/org/repo/vendor/src/github.com/Shopify/sarama/async_producer.go:622 +0x2a
github.com/Shopify/sarama.withRecover(0xc4207aa3f0)
	/home/travis/build/org/repo/vendor/src/github.com/Shopify/sarama/utils.go:45 +0x43
created by github.com/Shopify/sarama.(*asyncProducer).newBrokerProducer
	/home/travis/build/org/repo/vendor/src/github.com/Shopify/sarama/async_producer.go:622 +0x1b8
```

Tidied up some chan waits that weren't checking the ok state and also
added an explicit stopchan for the run loop and a unittest to cover the
leak case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,11 +5,11 @@ default: fmt vet errcheck test lint
 # Taken from https://github.com/codecov/example-go#caveat-multiple-files
 .PHONY: test
 test:
-	echo "" > coverage.txt
+	echo "mode: atomic" > coverage.txt
 	for d in `go list ./...`; do \
 		go test -p 1 -v -timeout 6m -race -coverprofile=profile.out -covermode=atomic $$d || exit 1; \
 		if [ -f profile.out ]; then \
-			cat profile.out >> coverage.txt; \
+			tail +2 profile.out >> coverage.txt; \
 			rm profile.out; \
 		fi \
 	done
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ default: fmt vet errcheck test lint
 test:
 	echo "" > coverage.txt
 	for d in `go list ./...`; do \
-		go test -p 1 -v -timeout 240s -race -coverprofile=profile.out -covermode=atomic $$d || exit 1; \
+		go test -p 1 -v -timeout 6m -race -coverprofile=profile.out -covermode=atomic $$d || exit 1; \
 		if [ -f profile.out ]; then \
 			cat profile.out >> coverage.txt; \
 			rm profile.out; \
--- a/async_producer.go
+++ b/async_producer.go
@@ -520,7 +520,6 @@ func (pp *partitionProducer) dispatch() {
 	}()
 
 	for msg := range pp.input {
-
 		if pp.brokerProducer != nil && pp.brokerProducer.abandoned != nil {
 			select {
 			case <-pp.brokerProducer.abandoned:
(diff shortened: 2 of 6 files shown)


bai:
Thanks for doing this. Looks like there's a bunch of legit failures 🤔

dnwe:
@bai yeah I need to analyze that Travis output because it wasn't deterministic (a rebuild had different results to the previous build). I'll update the PR as I work through them

dnwe:
@bai I tracked down the issue (the abandoned channel is intentionally closed to trigger abandonment). I also ended up pulling in github.com/fortytw2/leaktest  as a test-only dependency as it was better than my home-rolled NumGoroutine counting as it does things like status polling and filtering out routines it should ignore etc.

async_producer.go
```diff
@@ -652,6 +651,7 @@ func (p *asyncProducer) newBrokerProducer(broker *Broker) *brokerProducer {
 		input:          input,
 		output:         bridge,
 		responses:      responses,
+		stopchan:       make(chan struct{}),
```
varun06: I wonder if if `context` is better, but reckon that could be way more work.

Makefile
```diff
@@ -5,11 +5,11 @@ default: fmt vet errcheck test lint
 # Taken from https://github.com/codecov/example-go#caveat-multiple-files
 .PHONY: test
 test:
-	echo "" > coverage.txt
+	echo "mode: atomic" > coverage.txt
```
varun06: for my curiosity, why was it required?

async_producer_test.go
```diff
@@ -9,6 +9,9 @@ import (
 	"sync/atomic"
 	"testing"
 	"time"
+
+	"github.com/fortytw2/leaktest"
```
varun06: TIL

Makefile
dnwe: @varun06 I'm glad you asked :) it was described in the commit message, but happy to repeat it here:

> the existing coverage.txt concatenation included multiple `mode:
  atomic` lines which fail to parse correctly in the `go tool cover`
  tooling, so update Makefile so only one modeline exists

async_producer_test.go
dnwe: 😍

async_producer.go
dnwe: No that's a reasonable suggestion, and I don't think it'd be any more work. Code wise the only difference is we'd select on `<--ctx.Done()` rather than `<--stopchan`

Makefile
varun06: 🤦‍♂ I just had my tea :)
Thanks for reiterating.

async_producer.go
varun06: up to you, I am fine either way for now.

dnwe: @varun06 ok, please go ahead and merge for now then if you're happy. I might revisit under a separate PR putting context everywhere

PR #1452, opened 2019-08-06
Set the Kafka version in ListAcls method

symaras:
This PR adds a conditional to set the Kafka version. If the version is at last 2 then it sets request.Version = 1, which enables the functionality of getting the AclResourcePatternType field for the ACLs.


symaras:
Not sure how to resolve these two errors:

1. I signed the CLA already (attached a screenshot).
2. I don't know why tests are failing; running 'go test' locally returns without errors.

Any help would be appreciated :)
CLA_signed

bai:
Thanks!

PR #1453, opened 2019-08-06
Add support for modules

Highstead:
## Changes
- Update golang version in dev.yml
- Enable use of go modules

This should be a NOOP change to any production system and is just a nice to have.


Highstead:
Oops was running on an older version.  Disregard as this change isn't necessary.

PR #1457, opened 2019-08-12
Set request version in ListAcls ClusterAdmin method

agapoff:
I have encountered a problem with unavailability to set the request version other than 0 in DescribeAclsRequest when using ListAcls method. So it is not possible to get use of ResourcePatternTypeFilter in AclFilter or get the pattern types in response.

Setting Version in AclFilter itself has no effect because this value is being rewritten in encode method in acl_describe_request.go.

So this PR contains the obvious fix and allows to request pattern types in ListAcls.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -513,6 +513,10 @@ func (ca *clusterAdmin) ListAcls(filter AclFilter) ([]ResourceAcls, error) {
 
 	request := &DescribeAclsRequest{AclFilter: filter}
 
+	if ca.conf.Version.IsAtLeast(V2_0_0_0) && filter.Version == 1 {
+		request.Version = 1
+	}
+
 	b, err := ca.Controller()
 	if err != nil {
 		return nil, err


bai:
Thanks for your contribution; could you please rebase your PR, looks like there's a conflict.

agapoff:
There is no need anymore - PR #1452 was first to do the trick and I am happy with it.
But I suppose that PR could break the backward compatibility. For example before it in order to list all ACLs one should provide such AclFilter:
```
AclFilter{
        ResourceType: sarama.AclResourceAny,
        Operation:    sarama.AclOperationAny,
        PermissionType: sarama.AclPermissionAny,
}
```
But after the fix this won't be enough for Kafka 2.0+ because 
`ResourcePatternTypeFilter: sarama.AclPatternAny`
will also be required and the code above will return empty set of ACLs.

My PR expects AclFilter to have Version field explicitly set to 1. So could you please decide whether this is really needed. Otherwise I can just close the PR without resolving the conflict.

PR #1458, opened 2019-08-13
Set request version in CreateACL ClusterAdmin method

agapoff:
Right now the request version is always set to 0 when using CreateACL method. And it is therefore ignores the ResourcePatternType value causing created ACLs to be always of LITERAL type.

This PR fixes the issue and allows to create ACLs of any pattern type on brokers with versions 2.0+.


bai:
Thank you!

PR #1461, opened 2019-08-19
Set request version in DeleteACL ClusterAdmin method

agapoff:
One more PR to enable ResourcePatternTypeFilter in the request. For method DeleteACL this time.
This fix will allow deleting ACLs with pattern type other than LITERAL.


bai:
Many thanks.

PR #1465, opened 2019-08-21
removes gofork direct usage as fix is part of go

d1egoaz:
fixes https://github.com/Shopify/sarama/issues/1438

`jcmturner/gokrb5` needs to use the `jcmturner/gofork` in order to have
some workarounds about some stuff not implemented or not working as
expected.

For the specific case of github.com/jcmturner/gofork/encoding/asn1 there
is no reason right now to keep using `jcmturner/gofork` as the issue
golang/go#17321 was solved/closed in Dec 15, 2016 and released in go
1.10

`jcmturner/gokrb5` still uses internally `jcmturner/gofork`, it seems
the `gofork` `asn1` part could be removed and use the standar library,
but that will need some more work and PRs on the `jcmturner/gokrb5`
side.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,3 +50,7 @@ get:
 .PHONY: clean
 clean:
 	go clean ./...
+
+.PHONY: tidy
+tidy:
+	go mod tidy -v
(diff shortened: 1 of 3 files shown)


twmb:
I suspect the reason this broke things is because the fork returned its own asn1.ObjectType, which the stdlib marshalled differently, since it was not a stdlib asn1.ObjectType.

To re-apply this patch, you would need to do `asn1.ObjectType(gssapi.OIN(...`.

PR #1478, opened 2019-09-04
Add support for Go 1.13

bai:
Add support for Go 1.13 and remove Kafka 2.1 from build matrix according to 2 releases + 2 months schedule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-sarama
-======
+# sarama
 
 [![GoDoc](https://godoc.org/github.com/Shopify/sarama?status.svg)](https://godoc.org/github.com/Shopify/sarama)
 [![Build Status](https://travis-ci.org/Shopify/sarama.svg?branch=master)](https://travis-ci.org/Shopify/sarama)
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/Shopify/sarama
 
+go 1.13
```
bai: @varun06 I'm not 100% this is needed (this line added by `go mod`), do you know if it's something that will break 1.11/1.12 compatibility?

According to https://golang.org/doc/go1.13 it shouldn't but 🤔:

> If your code uses modules and your go.mod files specifies a language version, be sure it is set to at least 1.13 to get access to these language changes. You can do this by editing the go.mod file directly, or you can run go mod edit -go=1.13.

README.md
```diff
@@ -16,24 +15,22 @@ Sarama is an MIT-licensed Go client library for [Apache Kafka](https://kafka.apa
...
 You might also want to look at the [Frequently Asked Questions](https://github.com/Shopify/sarama/wiki/Frequently-Asked-Questions).
 
-### Compatibility and API stability
+## Compatibility and API stability
 
 Sarama provides a "2 releases + 2 months" compatibility guarantee: we support
 the two latest stable releases of Kafka and Go, and we provide a two month
 grace period for older releases. This means we currently officially support
-Go 1.11 through 1.12, and Kafka 2.0 through 2.3, although older releases are
+Go 1.11 through 1.13, and Kafka 2.1 through 2.3, although older releases are
```
d1egoaz: is this compatibility matrix important? 
are we even following it?

go.mod
varun06: yeah, I treat it like a ceiling for go features. So it shouldn't impact anything with backward compatibility.

PR #1481, opened 2019-09-05
Improve NewMockListAclsResponse

agapoff:
This PR provides some improvements for NewMockListAclsResponse:
1. It introduces support for ListAcls requests with Version = 1 (for Kafka 2+).
2. It won't fail if ResourceName in request is not defined.
3. It provides a more plausible response by setting some fields like Principal, Host, Operation, PermissionType.


bai:
Thank you! 💯

PR #1494, opened 2019-09-18
modify ofset to offset

llussy:


bai:
Good catch, many thanks!

PR #1502, opened 2019-09-30
Update documentation with Sarama instead of Sarma

philodavies:
Sarama was misspelled as Sarma in one part of the file.


bai:
Many thanks!

Issue #1462, opened 2019-08-19
Add a new partition during the consumer group running，but the cousumer group not rebalance。


leisunstar:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*
Sarama Version: 1.23.1 
Kafka Version: 0.10.1.1
Go Version: go1.12.4

##### Configuration

	config.Net.TLS.Enable = true
	config.Net.TLS.Config = tlsConfig
	config.Consumer.Return.Errors = true

##### Logs

N/A

##### Problem Description

When add new partitions (from 1 to 2), the consumer group, it has no rebalance ,until I start a new consumer or stop a consumer in a same consumer group. How should I get the new partition rebalance message？

PR #1525, opened 2019-10-30
fixed  Add a new partition during the consumer group running，but the cousumer group not rebalance  #1462

antsbean:
fixed Add a new partition during the consumer group running，but the cousumer group not rebalance。


antsbean:
@bai  please handle it,   This problem is more serious.

antsbean:
@bai  Is it possible send new version v1.24.1?

bai:
Done - https://github.com/Shopify/sarama/releases/tag/v1.24.1

PR #1527, opened 2019-10-30
Produce transactions

FrancoisPoinsot:
The goal of this PR is to add the ability to produce transactions from the producer.
It follows https://github.com/Shopify/sarama/issues/1512 discussion.

It is currently a draft because there is still much to do:
- adding unit-tests
- cleanup the mod magic that makes the fork work smoothly
- add `CommitOffset` feature
- fix the `AsyncProducer`'s input problem
- batch the addPartitionToTxn request

But a bit of early feedback is better than a late refactor.

You should be able to run the main.go in `examples/transactions`. This is currently my way to test the changes.
You will get a `Offset's topic has not yet been created` error, the first time you run main.go after reinitializing the broker. 

What would be particularly useful right now, on top of a general feedback, would be a plan for this asyncProducer.
the problem is described there https://github.com/Shopify/sarama/issues/1512#issuecomment-546712840 
tl;dr: I might need to change the `input` channel for a `SendMessageAsync` func to avoid missing some message in the transaction.


FrancoisPoinsot:
@bai @varun06

async_producer.go
```diff
@@ -3,6 +3,7 @@ package sarama
 import (
 	"encoding/binary"
 	"fmt"
+	"github.com/pkg/errors"
```
varun06: any particular reason to use it instead of std lib errors package?

async_producer.go
```diff
@@ -71,15 +88,34 @@ func (t *transactionManager) getAndIncrementSequenceNumber(topic string, partiti
...
+		client:                 client,
+		partitionsInCurrentTxn: topicPartitionSet{},
 	}
 
 	if conf.Producer.Idempotent {
-		initProducerIDResponse, err := client.InitProducerID()
+		if conf.Producer.TransactionalID != nil {
+			txnmgr.transactionalID = *conf.Producer.TransactionalID
+
+			// We need to a "FindCoordinator" request before trying to initiate the producerID
```
varun06:
```suggestion
			// We need to do a  "FindCoordinator" request before trying to initiate the producerID
```

async_producer.go
```diff
@@ -1114,3 +1328,43 @@ func (p *asyncProducer) abandonBrokerConnection(broker *Broker) {
...
+	}
+
+	return p.txnmgr.finishTransaction(commit)
+}
+
+func waitWithTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
+	c := make(chan struct{})
+	go func() {
+		defer close(c)
+		wg.Wait() // yes there is technically a go routine leak here when there is a timeout
```
varun06: you can get a context.Withtimeout() based on your timeout and select on ctx.Done() to remove this go routine leak, right?

FrancoisPoinsot:
I don't think I will ever have the time/energy to finish it.
I will close it for now.

PR #1532, opened 2019-11-02
Add list offset request

### LINKED ISSUE #1522: Improve implementation of OffsetRequest
#### Problem Description

`OffsetRequest` and `OffsetResponse` implement the `ListOffsetRequest` from Kafka's Offset API. You can use them to request (in bulk):

- Topic/Partition High + Low Water marks
- Find an index within a partition for a given timestamp

The implementation of this endpoint is suboptimal though:

- File and struct are called `OffsetRequest` instead of `ListOffsetRequest` as opposed to the official client/documentation
- Requests/Response implementation do not really stick to the original which also leads to problems using it. If someone wants to request the high and low watermarks of a single topic/partition, this will not work as the request and response use a map where the `partitionID` is the key. Thus you are forced to make two separate ListOffsetRequests even though you could put them into one

#### My proposal:

For these reasons I'd like to create a new `ListOffsetRequest` as I can not change the current `OffsetRequest` because this would cause breaking changes. Beside the above thing I'll also add:

- Proper documentation for the properties you can set in the request
- Add Version 0, Version 2, Version 3 and Version 4 of this request so that more brokers and more features are supported.

Can I proceed with the idea / would you accept a PR for this?

weeco:
Fixes #1522

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -333,6 +333,19 @@ func (b *Broker) GetAvailableOffsets(request *OffsetRequest) (*OffsetResponse, e
 	return response, nil
 }
 
+//ListOffsets return an list offset response or erorr
+func (b *Broker) ListOffsets(request *OffsetRequest) (*ListOffsetResponse, error) {
+	response := new(ListOffsetResponse)
+
+	err := b.sendAndReceive(request, response)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return response, nil
+}
+
 //Produce returns a produce response or error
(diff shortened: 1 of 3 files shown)


bai:
Nice PR, look forward to getting this shipped!

weeco:
@bai Have you seen my question in https://github.com/Shopify/sarama/issues/1522 ? I believe that's a blocker

bai:
Oh, I have not, it somehow got lost in my notifications, sorry. Taking a look.

PR #1541, opened 2019-11-17
add String, (Un)MarshalText for acl types.

izolight:
Sometimes it's useful converting the constants of the acl types from/to text.
The text representation is the name without the prefix.
When converting from text to the types the string will be normalized to catch most variants.

### DIFF
--- a/acl_types.go
+++ b/acl_types.go
@@ -48,7 +48,7 @@ func (a *AclOperation) String() string {
 		AclOperationAlterConfigs:    "AlterConfigs",
 		AclOperationIdempotentWrite: "IdempotentWrite",
 	}
-	s, ok := mapping[a]
+	s, ok := mapping[*a]
 	if !ok {
 		s = mapping[AclOperationUnknown]
 	}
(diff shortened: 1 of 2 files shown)


acl_types.go
```diff
@@ -27,6 +32,61 @@ const (
 	AclOperationIdempotentWrite
 )
 
+func (a AclOperation) String() string {
```
bai:
```suggestion
func (a *AclOperation) String() string {
```

izolight: fixed

d1egoaz:
@izolight it seems you'll need to update the `map` declaration `mapping := map[AclOperation]string{`

image

dnwe:
@izolight hiya — just wondering if you think you'll have time to resume work on this PR?

izolight:
I updated it so that the pointers are now dereferenced in the function.

izolight (commented):
I've rebased to the latest master an implemented the proposed changes.
Anything preventing this PR from getting merged now?

bai:
Thanks for contributing 🙏🏼

Issue #1540, opened 2019-11-15
Reading Azure event hubs fails with "kafka: response did not contain all the expected topic/partition blocks"


faec:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka | Go |
|--------|-------|----|
| 1.24.1 | 1.0 | 1.12.4 |

##### Configuration

What configuration values are you using for Sarama and Kafka?
``` go
k := sarama.NewConfig()

k.Version = sarama.V1_0_0_0

k.Consumer.Return.Errors = true
k.Consumer.Offsets.Retry.Max = 5
k.Consumer.Offsets.Initial = sarama.OffsetOldest

k.Net.TLS.Enable = true
k.Net.SASL.Enable = true
k.Net.SASL.User = "$ConnectionString"
k.Net.SASL.Password = connectionStr
```

##### Logs

When filing an issue please provide logs from Sarama and Kafka if at all
possible. You can set `sarama.Logger` to a `log.Logger` to capture Sarama debug
output.

<details><summary>logs: CLICK ME</summary>
<p>

```
[Sarama] 2019/11/15 13:52:56 Initializing new client
[Sarama] 2019/11/15 13:52:56 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[Sarama] 2019/11/15 13:52:56 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
[Sarama] 2019/11/15 13:52:56 client/metadata fetching metadata for all topics from broker filebeat-test.servicebus.windows.net:9093
[Sarama] 2019/11/15 13:52:56 Successful SASL handshake. Available mechanisms: [PLAIN]
[Sarama] 2019/11/15 13:52:56 SASL authentication successful with broker filebeat-test.servicebus.windows.net:9093:4 - [0 0 0 0]
[Sarama] 2019/11/15 13:52:56 Connected to broker at filebeat-test.servicebus.windows.net:9093 (unregistered)
[Sarama] 2019/11/15 13:52:57 client/brokers registered new broker #0 at filebeat-test.servicebus.windows.net:9093
[Sarama] 2019/11/15 13:52:57 Successfully initialized new client
[Sarama] 2019/11/15 13:52:57 client/metadata fetching metadata for [kafka-west] from broker filebeat-test.servicebus.windows.net:9093
[...]

PR #1542, opened 2019-11-19
Workaround for missing offsets in Azure's OffsetFetchResponse

faec:
This is a fix for #1540. It adds an entry in the offset table for (consumer group, partition) pairs in the `OffsetFetchRequest` that are missing from the `OffsetFetchResponse`.

This is for discussion rather than immediate merge; in particular we aren't sure yet whether Azure plans to revert to their previous behavior (returning an explicit offset entry for new consumer groups), in which case this would be superfluous. However, this fix or something like it is a viable workaround in the event that this behavior persists.

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -387,6 +387,23 @@ func (b *Broker) FetchOffset(request *OffsetFetchRequest) (*OffsetFetchResponse,
 		return nil, err
 	}
 
+	// In new consumer group / partition pairs, Microsoft's Azure Event Hubs
+	// communicates the "no error / new offset" state by omitting the requested
+	// entries from the OffsetFetchResponse (in contrast to other implementations
+	// which indicate this by returning an explicit offset of -1). To handle this
+	// case, we check all entries in the request and add an offset to the response
+	// table for any that are missing.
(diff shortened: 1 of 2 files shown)


bai:
Thanks for your contribution. Could you please add a test for this?

faec:
Test added! Sorry for the delay on the CLA -- this will be contributed on behalf of Elastic, we're just trying to figure out who to email to get me put on the official contributor list :-)

o1egl:
@faec any updates about this PR?

bai:
@o1egl according to https://github.com/Shopify/sarama/issues/1540#issuecomment-562259978 this PR shouldn't be relevant anymore. Are you still experiencing the issue?

faec:
Yes, as far as I'm aware everything is working without the fix again... I've procrastinated on closing this in case the context was still relevant to someone but I think it's safe now, if this problem is still happening anywhere please follow up on #1540.

o1egl:
@bai @faec I still get `kafka: response did not contain all the expected topic/partition blocks` error without this fix.

PR #1551, opened 2019-12-02
Fix ConfigResourceType to match the protocol and server.

bradw456:
The struct was erroneously setup identical to the ACL Resources and therefore
(a) had more types than describe and alter config operations actually
supported (they only support Broker and Topic types) and (b) had the
wrong integer values.  See reference for actual integer constants
that the server is expecting.

### DIFF
--- a/config_resource_type.go
+++ b/config_resource_type.go
@@ -3,20 +3,16 @@ package sarama
 //ConfigResourceType is a type for config resource
 type ConfigResourceType int8
 
-// Taken from :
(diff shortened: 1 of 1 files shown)


dnwe:
@bradw456 ah sorry I hadn't seen this PR. I had spotted this same problem and included it in my (now merged) PR https://github.com/Shopify/sarama/pull/1571 which was fixing a related issue with DescribeConfigRequest and BrokerResourceType where I'd noticed it was sending the wrong integer values for the given ResourceTypes and I was getting confusing responses back

d1egoaz:
Thanks @bradw456 
closed by https://github.com/Shopify/sarama/pull/1571

bradw456:
Thanks @dnwe

PR #1556, opened 2019-12-12
Send DescribeConfigsRequest v1 when possible

benweint:
`DescribeConfigsRequest` v1 opts the caller into the v1 response format, described
in [KIP-226](https://cwiki.apache.org/confluence/display/KAFKA/KIP-226+-+Dynamic+Broker+Configuration), which includes various improvements, including replacing the `is_default` field with the more useful `config_source` field, and optionally including information about synonyms.

This change makes `admin.DescribeConfigs` prefer the v1 request/response format if the broker version allows it, in order to get access to these features when possible.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -442,6 +442,10 @@ func (ca *clusterAdmin) DescribeConfig(resource ConfigResource) ([]ConfigEntry, 
 		Resources: resources,
 	}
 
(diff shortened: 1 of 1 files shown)


benweint:
(I'm actually going to re-open this in a slightly different way.)

PR #1560, opened 2019-12-16
add sync pool for gzip 1-9

iyacontrol:
when large number of messages, for gzip(1-9) cost high cpu, because of gc.

### DIFF
--- a/compress.go
+++ b/compress.go
@@ -6,7 +6,7 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/eapache/go-xerial-snappy"
+	snappy "github.com/eapache/go-xerial-snappy"
 	"github.com/pierrec/lz4"
 )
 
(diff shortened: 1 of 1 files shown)


compress.go
```diff
@@ -22,6 +22,87 @@ var (
 			return gzip.NewWriter(nil)
 		},
 	}
+	gzipWriterPoolForCompressionLevel1 = sync.Pool{
+		New: func() interface{} {
+			gz, err := gzip.NewWriterLevel(nil, 1)
+			if err != nil {
+				panic(err)
```
varun06: Not a big fan of panic, what happens if we reach here and panic?

iyacontrol: NewWriterLevel is like NewWriter but specifies the compression level instead of assuming DefaultCompression.

The compression level can be DefaultCompression, NoCompression, HuffmanOnly or any integer value between BestSpeed and BestCompression inclusive. **The error returned will be nil if the level is valid.**

1-9 level is correct.

varun06: I am really not a big fan of using panic. but 👍

PR #1571, opened 2020-01-15
fix: use the broker for any admin on BrokerConfig

dnwe:
When operating on Broker configuration via the Admin API, the request
_must_ be sent to the specific broker that the change applies to, _not_
just (as usual) to the Controller.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -2,7 +2,9 @@ package sarama
 
 import (
 	"errors"
+	"fmt"
 	"math/rand"
+	"strconv"
 	"sync"
 )
 
@@ -226,6 +228,16 @@ func (ca *clusterAdmin) DescribeCluster() (brokers []*Broker, controllerID int32
 	return response.Brokers, response.ControllerID, nil
 }
 
+func (ca *clusterAdmin) findBroker(id int32) (*Broker, error) {
+	brokers := ca.client.Brokers()
+	for _, b := range brokers {
+		if b.ID() == id {
+			return b, nil
+		}
+	}
+	return nil, fmt.Errorf("could not find broker id %d", id)
+}
+
 func (ca *clusterAdmin) findAnyBroker() (*Broker, error) {
 	brokers := ca.client.Brokers()
 	if len(brokers) > 0 {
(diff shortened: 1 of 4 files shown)


admin.go
```diff
@@ -442,11 +454,23 @@ func (ca *clusterAdmin) DescribeConfig(resource ConfigResource) ([]ConfigEntry,
...
+		b   *Broker
+		err error
+	)
+
+	// DescribeConfig of broker must be sent to the broker in question
+	if resource.Type == BrokerResource {
+		id, _ := strconv.Atoi(resource.Name)
+		b, err = ca.findBroker(int32(id))
+	} else {
+		b, err = ca.Controller()
```
mimaison: When describing a non-broker resource, the request can be sent to any broker, so we could use `findAnyBroker()` here.

admin.go
```diff
@@ -479,7 +503,19 @@ func (ca *clusterAdmin) AlterConfig(resourceType ConfigResourceType, name string
...
+		err error
+	)
+
+	// AlterConfig of broker must be sent to the broker in question
+	if resourceType == BrokerResource {
+		id, _ := strconv.Atoi(name)
+		b, err = ca.findBroker(int32(id))
+		_ = b.Open(ca.client.Config())
+	} else {
+		b, err = ca.Controller()
```
mimaison: Similarly to `DescribeConfig`, we don't need to send the request to the controller here, any broker can be used if the resource is not a specific broker and not a broker logger.

config_resource_type.go
```diff
@@ -1,22 +1,18 @@
...
-	//BrokerResource constant type
-	BrokerResource
+	// UnknownResource constant type
+	UnknownResource ConfigResourceType = 0
+	// TopicResource constant type
+	TopicResource ConfigResourceType = 2
+	// BrokerResource constant type
+	BrokerResource ConfigResourceType = 4
+	// BrokerLoggerResource constant type
+	BrokerLoggerResource ConfigResourceType = 8
```
mimaison: This new type of resource was added in Kafka 2.4. Can you add tests to ensure this is handled the same way as BrokerResource when described or altered?

admin.go
```diff
@@ -479,7 +503,19 @@ func (ca *clusterAdmin) AlterConfig(resourceType ConfigResourceType, name string
...
 	}
 
-	b, err := ca.Controller()
+	var (
+		b   *Broker
+		err error
+	)
+
+	// AlterConfig of broker must be sent to the broker in question
+	if resourceType == BrokerResource {
```
mimaison: If the resource is  `BrokerResource` but it has no name (ie. describing its default), then the request can be sent to any broker too. Same for `BrokerLoggerResource`.

admin.go
dnwe: 👍

admin.go
dnwe: 👍

config_resource_type.go
dnwe: 👍

dnwe:
@mimaison thanks for the review — I've incorporate your suggestions into the commit and force pushed it up, it is ready for re-review

bai:
LGTM, thanks for contributing again and thanks @mimaison for reviewing. Let me know if you think it's ready to be merged.

PR #1573, opened 2020-01-15
Add GitHub Actions workflows, use golangci-lint

bai:
1. Use golangci-lint instead of assorted linters, apart from simplifying setup overall it'll allow running linters locally with a simple make lint. We still have a lot of linters disabled but otherwise there're too many failures. Will address those in a follow-up PR.

2. Build with GitHub Actions instead of Travis CI. 

How to read this PR: first commit sets up GitHub Actions and golangci-lint, second commit fixes linter issues. 

Interesting observation: GitHub Actions runs integration tests more than 2x faster compared to Travis.


d1egoaz:
> 2x faster compared to Travis.

🚀 🚢

.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @bai
```
d1egoaz: wondering if this can be the team of sarama maintainers? (if there is such a team ?)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+        go-version: [1.12.x, 1.13.x]
+        platform: [ubuntu-latest]
+
+    env:
+      KAFKA_PEERS: localhost:9091,localhost:9092,localhost:9093,localhost:9094,localhost:9095
+      TOXIPROXY_ADDR: http://localhost:8474
+      KAFKA_INSTALL_ROOT: /home/runner/kafka
+      KAFKA_HOSTNAME: localhost
+      DEBUG: true
+      KAFKA_VERSION: 2.3.0
```
d1egoaz: 🤔 where is going to be defined the different kafka server versions?

.github/CODEOWNERS
bai: There's no such team, but I'll be happy to amend this list — arguably having a list of usernames here is easier than creating a proper GitHub team.

.github/workflows/ci.yml
bai: Updated this PR to include kafka-version in build matrix.

bai:
@d1egoaz this PR is ready for final review, please take a look when you have a chance 🙏

.github/workflows/ci.yml
```diff
@@ -4,12 +4,13 @@ on: push
...
 jobs:
   test:
-    name: Go ${{ matrix.go-version }} on Ubuntu
+    name: Go ${{ matrix.go-version }} with Kafka ${{ matrix.kafka-version }} on Ubuntu
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         go-version: [1.12.x, 1.13.x]
+        kafka-version: [2.3.1, 2.4.0]
```
d1egoaz: do you think it'd be a good idea also to test against `2.2.x` like previously travis did it with `2.2.1`?

PR #1576, opened 2020-01-15
fix: set DescribeConfigRequest Version field

dnwe:
Although fbd8338 had added protocol support for DescribeConfigsRequest
v1 and v2, nothing in the admin client was actually setting the Version
field to utilise these. Set the Version based on the selected
sarama.Config version.


mimaison:
In the future we should use ApiVersions to automatically adjust the versions used by requests instead of relying on the user specifying it.

PR #1586, opened 2020-01-23
fix: retry topic request on ControllerNotAvailable

dnwe:
If an admin request to Create/Delete Topic{Partitions} receives a
ControllerNotAvailable kError back from Kafka, we should expire the
cached controller and refresh it from metadata.

Additionally add a `Retry` section to the `Admin` config similarly to
the Java AdminKafkaClient and consider this case one that is eligible to
be re-tried

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -6,6 +6,7 @@ import (
 	"math/rand"
 	"strconv"
 	"sync"
+	"time"
 )
 
 // ClusterAdmin is the administrative client for Kafka, which supports managing and inspecting topics,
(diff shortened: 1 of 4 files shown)


admin.go
```diff
@@ -373,18 +412,35 @@ func (ca *clusterAdmin) CreatePartitions(topic string, count int32, assignment [
...
+					// if unable to refresh controller just return original err
+					return topicErr
+				}
+				// retriable error
+				time.Sleep(ca.conf.Admin.Retry.Backoff)
+				Logger.Printf(
+					"admin/createpartitions retrying after %dms... (%d attempts remaining)\n",
+					ca.conf.Admin.Retry.Backoff/time.Millisecond, ca.conf.Admin.Retry.Max-i)
+				continue
+			}
```
d1egoaz: wondering if these duplicated codes could be removed and have only a single function to have this specific error handling?

looks the difference is the methods returning different error types and the logged message.

Wondering if some interface/struft could be created to make `DeleteTopicsResponse` return something that has an error with the form of:
```go
type Xxx struct {
	Err    KError
	ErrMsg *string
}
```

similar to the other error types:
``` go
type CreateTopicsResponse struct {
	Version      int16
	ThrottleTime time.Duration
	TopicErrors  map[string]*TopicError
}
type TopicError struct {
	Err    KError
	ErrMsg *string
}

type CreatePartitionsResponse struct {
	ThrottleTime         time.Duration
	TopicPartitionErrors map[string]*TopicPartitionError
}

type TopicPartitionError struct {
	Err    KError
	ErrMsg *string
}

type CreateTopicsResponse struct {
	Version      int16
	ThrottleTime time.Duration
	TopicErrors  map[string]*TopicError
}

type DeleteTopicsResponse struct {
	Version         int16
	ThrottleTime    time.Duration
	TopicErrorCodes map[string]KError
}

KError
```

dnwe: Yeah I'd like to fix that discrepancy in the return type error code, but wasn't sure if that would need a 2.x bump as a change in api

dnwe:
@d1egoaz I had a few more attempts at re-skinning this change to be a bit neater and I've force pushed up what I think is the best of those — if you'd like to have a re-review and merge if you're happy

admin.go
d1egoaz: yeah right, that'll be a breaking change, so we need to wait for 2.x

PR #1589, opened 2020-01-29
fix error message s/CommitInterval/AutoCommit.Interval

d1egoaz:


d1egoaz:
there was a breaking changed added to the API here:
https://github.com/Shopify/sarama/pull/1164/files#r372509658


:D that will make people migrate faster from sarama-cluster to sarama consumer groups 😅

PR #1590, opened 2020-01-29
Adds back Consumer.Offsets.CommitInterval to fix API

d1egoaz:
fix breaking API change added in
https://github.com/Shopify/sarama/pull/1164 https://github.com/Shopify/sarama/pull/1164#pullrequestreview-350270834

Fixes:
- https://github.com/bsm/sarama-cluster/pull/308
- https://github.com/bsm/sarama-cluster/issues/307
- https://github.com/lovoo/goka/issues/211

### DIFF
--- a/config.go
+++ b/config.go
@@ -352,6 +352,11 @@ type Config struct {
 		// offsets. This currently requires the manual use of an OffsetManager
 		// but will eventually be automated.
 		Offsets struct {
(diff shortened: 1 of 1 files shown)


mkevanz:
This change does not directly fix backward compatibility. The default value for CommitInterval was removed as well which requires users to configure this now.

PR #1592, opened 2020-01-30
Allow BalanceStrategy to provide custom assignment data

aldelucca1:
The StickyBalanceStrategy currently provides state information with member assignments. This is achieved through a custom name check in the `syncGroupRequest` of the consumer group. This works well for this case, but falls apart when trying to create additional stateful balance strategies. This update will make it possible to create new stateful balance strategies

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -334,7 +334,7 @@ func (c *consumerGroup) syncGroupRequest(coordinator *Broker, plan BalanceStrate
 	strategy := c.config.Consumer.Group.Rebalance.Strategy
 	for memberID, topics := range plan {
 		assignment := &ConsumerGroupMemberAssignment{Topics: topics}
-		userDataBytes, err := strategy.AssignmentData(plan, memberID, generationID)
+		userDataBytes, err := strategy.AssignmentData(memberID, topics, generationID)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


consumer_group.go
```diff
@@ -331,20 +331,14 @@ func (c *consumerGroup) syncGroupRequest(coordinator *Broker, plan BalanceStrate
...
-		if c.config.Consumer.Group.Rebalance.Strategy.Name() == StickyBalanceStrategyName {
-			userDataBytes, err := encode(&StickyAssignorUserDataV1{
-				Topics:     topics,
-				Generation: generationID,
-			}, nil)
-			if err != nil {
-				return nil, err
-			}
-			assignment.UserData = userDataBytes
+		userDataBytes, err := strategy.AssignmentData(plan, memberID, generationID)
```
d1egoaz: wondering why you'd need to send the plan and member to `AssignmentData` to get the topics:
``` go
	topics, ok := plan[memberID]
	if !ok {
		return nil, nil
	}
```
if that was already obtained in the range above?
Line 335 `for memberID, topics := range plan`

aldelucca1: My thinking here was in the case the state being shipped needed to understand data about the larger plan, but I am happy limiting the scope to just the given member and its assigned topics.

d1egoaz: thanks for the fast response!
I'll 👀 later today

PR #1594, opened 2020-01-31
Sets ConfigEntry.Default flag in addition to the ConfigEntry.Source for Kafka versions > V1_1_0_0

sladkoff:
We were relying on the `ConfigEntry.Default` field to filter out any default config settings when using `admin.DescribeConfig()`.

After updating to the most recent version of sarama, this flag is always set to `false` with Kafka versions > `V1_1_0_0`.

We found out that we can now use the `ConfigEntry.Source` to check for defaults. I don't know it was intended change to "break" the `ConfigEntry.Default` in higher versions - let me know. 

This change sets the `ConfigEntry.Default` flag in addition to the `ConfigEntry.Source` for Kafka versions > `V1_1_0_0`. 

I haven't written any tests yet as I'm not familiar with the code base. I can add some later.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -338,6 +338,15 @@ func (ca *clusterAdmin) ListTopics() (map[string]TopicDetail, error) {
 	describeConfigsReq := &DescribeConfigsRequest{
 		Resources: describeConfigsResources,
 	}
+
+	if ca.conf.Version.IsAtLeast(V1_1_0_0) {
+		describeConfigsReq.Version = 1
+	}
+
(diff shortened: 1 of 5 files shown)


varun06:
@sladkoff you might have to sign CLA.

sladkoff:
@varun06 yeah, I did that already... the check is still in a failed state and I can not re-submit the CLA form because it says that I already signed it... Can you help with that? Maybe reset something?

d1egoaz:
I've re-run the CLA check, it's OK now

sladkoff:
Just letting you know that I've added some tests. Are we good here?

PR #1601, opened 2020-02-05
fix: remove use of testing.T.FailNow() inside goroutine

alrs:
This fixes `TestClientAutoRefreshShutdownRace()` to not fire `t.Fatal()` inside a goroutine.

`t.Fatal()` is a convenience function that calls `t.Log()` and `t.FailNow()`. `t.FailNow()` should not be called from inside a goroutine. From the godoc:

 > FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test.

This replaces the `done` channel with an `error` channel, and tests the error variable outside the goroutine.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -830,13 +830,11 @@ func TestClientAutorefreshShutdownRace(t *testing.T) {
 	// Wait for the background refresh to kick in
 	time.Sleep(110 * time.Millisecond)
 
-	done := make(chan none)
+	errCh := make(chan error, 1)
 	go func() {
 		// Close the client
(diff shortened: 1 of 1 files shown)


alrs:
I'm happy to fill out a CLA, should someone provide a link to one.

bai:
Here we go, it's actually one of the failing checks: https://github.com/Shopify/sarama/pull/1601/checks?check_run_id=428029640 👀

alrs:
@bai CLA is complete.

bai:
Thanks for contributing.

PR #1603, opened 2020-02-06
tests: remove unused testLogger

alrs:
This removes `testLogger{}` and its functions. It is unused today, and it appears to have been unused since its introduction in 8/2019.

@dnwe, am I missing anything here?


dnwe:
@alrs thanks :) this was just remnants of something I'd find useful whilst debugging a goroutine leak

PR #1606, opened 2020-02-06
fix: use modern kafka server.properties in vagrant

dnwe:
Rather than copying and templating an old copy of a Kafka
server.properties, modify the install_cluster.sh to append a small
number of custom values to the end of the Java properties file as
shipped with each Kafka version. This ensures we are otherwise testing
the defaults for each version apart from where we have made
customisations.


dnwe:
Hmm, so it looks like this doesn't work for Kafka <2.4.0 — However, Go Actions seem to just run forever without giving up and there's no user-facing way to cancel them if they're stuck? That's a shame.

PR #1607, opened 2020-02-07
Fix memory leak when Broker.Open and Broker.Close called repeatedly

darklore:
Broker.registeredMetrics is increased in Broker.Open, but isn't cleared in Broker.Close.
So when Open and Close is called repeatedly it causes memory leak.


darklore:
I think I already signed CLA...

PR #1612, opened 2020-02-13
Updated "retrying" log message when BackoffFunc implemented

wblakecaldwell:
The "client/metadata retrying" log message was always reporting
the sleep duration as `config.Metadata.Retry.Backoff`, regardless
of whether `config.Metadata.Retry.BackoffFunc` is implemented.

We now report the actual retry time before sleeping.


wblakecaldwell:
Just signed the CLA - closing PR and re-submitting.

PR #1614, opened 2020-02-13
produce_response.go: Remove Unused Functions

alrs:
This removes two unused functions from the root `sarama` package.


alrs:
I read this failure as a flake, is it a common one?

d1egoaz:
thanks!

PR #1617, opened 2020-02-17
Add support for alter/list partition reassignements APIs

sladkoff:
We added support for API Keys [45 (AlterPartitionReassignments)](https://kafka.apache.org/protocol#The_Messages_AlterPartitionReassignments) and [46 (ListPartitionReassignments)](https://kafka.apache.org/protocol#The_Messages_ListPartitionReassignments).

This enables sarama to update the replication factor of topics for Kafka 2.4.0.0+ (See also https://github.com/Shopify/sarama/issues/1238).

We had to do change quite a lot since these APIs require the use of newer data structures in the body as well as in the header of the requests/responses (mainly compact structures and tagged fields). 

Please let us know what you think.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -478,7 +478,6 @@ func (ca *clusterAdmin) AlterPartitionReassignments(topic string, assignment [][
 		if err != nil {
 			errs = append(errs, err)
 		} else {
-
 			if rsp.ErrorCode > 0 {
 				errs = append(errs, errors.New(rsp.ErrorCode.Error()))
 			}
--- a/admin.go
+++ b/admin.go
@@ -490,7 +490,6 @@ func (ca *clusterAdmin) AlterPartitionReassignments(topic string, assignment [][
 					}
 				}
 			}
-
 		}
 
 		if len(errs) > 0 {
--- a/admin.go
+++ b/admin.go
@@ -470,15 +470,31 @@ func (ca *clusterAdmin) AlterPartitionReassignments(topic string, assignment [][
 			return err
 		}
 
+		errs := make([]error, 0)
+
 		rsp, err := b.AlterPartitionReassignments(request)
 
 		if err != nil {
-			return err
+			errs = append(errs, err)
+		} else {
+
+			if rsp.ErrorCode > 0 {
+				errs = append(errs, errors.New(rsp.ErrorCode.Error()))
+			}
+
+			for topic, topicErrors := range rsp.Errors {
+				for partition, partitionError := range topicErrors {
+					if partitionError.errorCode != ErrNoError {
+						errStr := fmt.Sprintf("[%s-%d]: %s", topic, partition, partitionError.errorCode.Error())
+						errs = append(errs, errors.New(errStr))
+					}
+				}
+			}
+
 		}
 
-		if rsp.ErrorCode > 0 || len(rsp.Errors) > 0 {
-			// TODO pretty error from response
-			return errors.New("Kafka server returned errors during replication-factor assignment")
+		if len(errs) > 0 {
+			return ErrReassignPartitions{MultiError{&errs}}
 		}
 
 		return nil
--- a/admin.go
+++ b/admin.go
@@ -42,12 +42,13 @@ type ClusterAdmin interface {
 	// new partitions. This operation is supported by brokers with version 1.0.0 or higher.
 	CreatePartitions(topic string, count int32, assignment [][]int32, validateOnly bool) error
 
-	// Update the configuration for the specified resources with the default options.
-	// This operation is supported by brokers with version 0.11.0.0 or higher.
-	// The resources with their configs (topic is the only resource type with configs
-	// that can be updated currently Updates are not transactional so they may succeed
-	// for some resources while fail for others. The configs for a particular resource are updated automatically.
+	// Alter the replica assignment for partitions.
+	// This operation is supported by brokers with version 2.4.0.0 or higher.
 	AlterPartitionReassignments(topic string, assignment [][]int32) error
+
+	// Provides info on ongoing partitions replica reassignments.
+	// This operation is supported by brokers with version 2.4.0.0 or higher.
+	ListPartitionReassignments(topics string, partitions []int32) (topicStatus map[string]map[int32]*PartitionReplicaReassignmentsStatus, err error)
 
 	// Delete records whose offset is smaller than the given offset of the corresponding partition.
 	// This operation is supported by brokers with version 0.11.0.0 or higher.
@@ -498,6 +499,33 @@ func (ca *clusterAdmin) AlterPartitionReassignments(topic string, assignment [][
 
 		return nil
 	})
+}
+
+func (ca *clusterAdmin) ListPartitionReassignments(topic string, partitions []int32) (topicStatus map[string]map[int32]*PartitionReplicaReassignmentsStatus, err error) {
+	if topic == "" {
+		return nil, ErrInvalidTopic
+	}
+
+	request := &ListPartitionReassignmentsRequest{
+		TimeoutMs: int32(10000),
+		Version:   int16(0),
+	}
+
+	request.AddBlock(topic, partitions)
+
+	b, err := ca.findAnyBroker()
+	if err != nil {
+		return nil, err
+	}
+	_ = b.Open(ca.client.Config())
+
+	rsp, err := b.ListPartitionReassignments(request)
+
+	if err == nil && rsp != nil {
+		return rsp.TopicStatus, nil
+	} else {
+		return nil, err
+	}
 }
 
 func (ca *clusterAdmin) DeleteRecords(topic string, partitionOffsets map[int32]int64) error {
--- a/admin.go
+++ b/admin.go
@@ -472,13 +472,14 @@ func (ca *clusterAdmin) AlterPartitionReassignments(topic string, assignment [][
 
 		rsp, err := b.AlterPartitionReassignments(request)
 
-		_ = rsp
-
 		if err != nil {
 			return err
 		}
 
-		// TODO error handling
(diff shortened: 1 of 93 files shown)


dnwe:
@sladkoff 🌟 thanks for this, great PR! 

It'll take a little while to review it all as I also need to brush up my knowledge on [KIP-482](https://cwiki.apache.org/confluence/display/KAFKA/KIP-482%3A+The+Kafka+Protocol+should+Support+Optional+Tagged+Fields) and how this fits with the existing [messages](https://github.com/apache/kafka/blob/trunk/clients/src/main/resources/common/message/README.md) too.

bai:
@sladkoff 💯 Do you think you could rebase your branch to make sure it's ✅on CI 🙏

admin.go
```diff
@@ -443,6 +451,82 @@ func (ca *clusterAdmin) CreatePartitions(topic string, count int32, assignment [
 	})
 }
 
+func (ca *clusterAdmin) AlterPartitionReassignments(topic string, assignment [][]int32) error {
+	if topic == "" {
+		return ErrInvalidTopic
+	}
+
+	request := &AlterPartitionReassignmentsRequest{
+		TimeoutMs: int32(10000),
```
dnwe: Where did this 10_000 millisecond default timeout come from? The protocol description suggests 60_000 should be the default?

https://github.com/apache/kafka/blob/2.4.0/clients/src/main/resources/common/message/AlterPartitionReassignmentsRequest.json#L23

admin.go
```diff
@@ -443,6 +451,82 @@ func (ca *clusterAdmin) CreatePartitions(topic string, count int32, assignment [
...
+	})
+}
+
+func (ca *clusterAdmin) ListPartitionReassignments(topic string, partitions []int32) (topicStatus map[string]map[int32]*PartitionReplicaReassignmentsStatus, err error) {
+	if topic == "" {
+		return nil, ErrInvalidTopic
+	}
+
+	request := &ListPartitionReassignmentsRequest{
+		TimeoutMs: int32(10000),
```
dnwe: As above, shouldn't this be 60_000 ?

admin_test.go
```diff
@@ -332,6 +332,139 @@ func TestClusterAdminCreatePartitionsWithoutAuthorization(t *testing.T) {
...
 }
 
+func TestClusterAdminAlterPartitionReassignments(t *testing.T) {
+	seedBroker := NewMockBroker(t, 1)
+	defer seedBroker.Close()
+
+	seedBroker.SetHandlerByMap(map[string]MockResponse{
+		"MetadataRequest": NewMockMetadataResponse(t).
+			SetController(seedBroker.BrokerID()).
+			SetBroker(seedBroker.Addr(), seedBroker.BrokerID()),
```
dnwe: Please could you add a second mockBroker to the test and set that as the controller in the mock metadata response and _not_ the seed broker? That should exercise that the request must be sent to the Controller and not just any broker (or the broker currently connected to)

admin_test.go
```diff
@@ -332,6 +332,139 @@ func TestClusterAdminCreatePartitionsWithoutAuthorization(t *testing.T) {
...
+}
+
+func TestClusterAdminListPartitionReassignments(t *testing.T) {
+	seedBroker := NewMockBroker(t, 1)
+	defer seedBroker.Close()
+
+	seedBroker.SetHandlerByMap(map[string]MockResponse{
+		"MetadataRequest": NewMockMetadataResponse(t).
+			SetController(seedBroker.BrokerID()).
+			SetBroker(seedBroker.Addr(), seedBroker.BrokerID()),
```
dnwe: As above, please use different MockBroker for the Controller

admin.go
```diff
@@ -443,6 +451,82 @@ func (ca *clusterAdmin) CreatePartitions(topic string, count int32, assignment [
...
+		Version:   int16(0),
+	}
+
+	request.AddBlock(topic, partitions)
+
+	b, err := ca.findAnyBroker()
```
dnwe: I thought that ListPartitionReassignments also needed to be sent to the Controller (not just any broker) like the Alter request? The Java Admin client seems to do so [here](https://github.com/apache/kafka/blob/2.4.0/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L3391-L3393)

admin.go
dwi-di: we probably just put something in there :) Using the suggested default makes sense. I added this.

admin.go
dwi-di: yes

admin.go
dwi-di: good catch. I fixed  that

dwi-di:
@dnwe I added commits which address your remarks. Can you have a look? One of the CI jobs failed, but it does not look related to our change, right?

admin.go
```diff
@@ -512,7 +512,7 @@ func (ca *clusterAdmin) ListPartitionReassignments(topic string, partitions []in
 
 	request.AddBlock(topic, partitions)
 
-	b, err := ca.findAnyBroker()
+	b, err := ca.Controller()
```
dnwe: Now this is using the controller you probably want to wrapper the body in a `return ca.retryOnError(isErrNoController, ...)` call like you did for `AlterPartitionReassignments` so it refreshes the cached controller if it is stale

dwi-di: the `retryOnError` function cannot be used when a value needs to be returned. other functions in `admin.go` where a controller is used and a return type is needed also don't use retrying. How should this be handled? should we implement a retry function that returns something like `interface{}`?

dnwe: @dwi-di ah of course — let's just leave this for now and we can always follow-up in another PR

dnwe:
> One of the CI jobs failed, but it does not look related to our change, right?

I'm guessing it was just a slight flake — but it was in the new TestListPartitionReassignmentRequest func:
```
--- FAIL: TestListPartitionReassignmentRequest (0.00s)
##[error]    request_test.go:63: Encoding two blocks failed
        got  [0 0 39 16 3 7 116 111 112 105 99 50 3 0 0 0 1 0 0 0 2 0 6 116 111 112 105 99 2 0 0 0 0 0 0] 
        want [0 0 39 16 3 6 116 111 112 105 99 2 0 0 0 0 0 7 116 111 112 105 99 50 3 0 0 0 1 0 0 0 2 0 0]

```

dnwe:
I was able to re-create with:
```
go test -run '^TestListPartitionReassignmentRequest$' -count=500 -race
```

dnwe:
Ah I can see the problem, the ordering of the two blocks in your check of `testRequest(t, "two blocks", request, listPartitionReassignmentsRequestTwoBlocks)` isn't guaranteed, so sometimes you get "topic" followed by "topic2" in your response, and sometimes you get "topic2" followed by "topic" in your response.

dwi-di:
ah I see, but what do you propose to do here? the only way to fix it is to change the encoder to sort the blocks based on topic name, right? but that does not feel right, because it is not really needed, only for the test...

dnwe:
> ah I see, but what do you propose to do here? the only way to fix it is to change the encoder to sort the blocks based on topic name, right? but that does not feel right, because it is not really needed, only for the test...

Yes I think it just isn't possible to use the existing `testRequest` helper which is what is doing the byte-comparison along the way. In your two-block test you'd probably just test a struct --> protocolEncode --> protocolDecode --> struct roundtrip correctly gives you a pair of structs that are DeepEqual — which is essentially a slimmed down version of the existing helpers

dwi-di:
I modified the flaky test, that should be all then.

dnwe:
Hmm I squash merged this manually from the CLI, but GitHub doesn't seem to have tracked the merge (2d2326efcf8993a79d13a7207ed2fcf4ea1367e2 via 9501120)

sladkoff:
@dnwe Nice! Can we expect to get these changes into a release soon?

random-dwi:
we should probably wait for #1640 before releasing this, otherwise the list partition replica assignement status is not really usable.

sladkoff:
Hi guys! It's been two months now, so I'll allow myself to re-post my question :sweat_smile:  

Could we get an ETA for a new release with this these features?

@dnwe @bai

bai:
That's fair. @dnwe any objections on cutting release today?

dnwe:
@bai yes I was merging a few PRs in yesterday as I think we do need to cut a release with this and also to include @KJTsanaktsidis's fix for the session ID cache exhaustion that was introduced in 1.26 and is a nasty bug.

dnwe:
@bai we might need to check on the github actions builds though, as they've been a bit flakey recently. Not sure if we can do anything to improve them

bai:
I think we can address GitHub Actions issue short-term by reducing build matrix — it's gotten so much flakier with 9 builds instead of 6.

Hopefully when we have that docker-compose setup in place it'll get better 🙏 

Want to do this release yourself or would you prefer me cutting it?

dnwe:
Sure I can give a go to cutting the release. Just a tag in Git and then create a release with a small changelog linking to the PRs?

bai:
@dnwe I usually use `Draft New Release` button on Releases page here: https://github.com/Shopify/sarama/releases.
Once I'm confident (title/tag look what I want them to look) I'm promoting Draft Release with "Publish Release" button.

So basically I'm not doing manual git tag but use GitHub UI to release 🤷‍♀️

dnwe:
@sladkoff released and announced!

https://github.com/Shopify/sarama/releases/tag/v1.26.2

https://twitter.com/oldmanuk/status/1258035278771621888

(apologies for missing shoutouts to those I couldn't immediately find on Twitter)

sladkoff:
@dnwe @bai Awesome :tada: Thanks for the quick reaction!

bai:
Props to Dominic for cutting this one and all of you folks for contributing 🙏

Issue #1616, opened 2020-02-17
Sarama-cluster broke since removal of default CommitInterval config


reenjii:
##### Versions

| Sarama | Kafka | Go |
|--------|-------|----|
| 5885bb2d56ee4de66793d303338cd0aae66099ed | n/a | go1.13.5 |

##### Configuration

All default

##### Logs

<details><summary>logs: CLICK ME</summary>
<p>

```
time.NewTicker(0x0, 0xc0003105c0)
        /usr/local/share/go/src/time/tick.go:23 +0x147
github.com/bsm/sarama-cluster.(*Consumer).cmLoop(0xc000014000, 0xc000650300)
        /home/user/go/pkg/mod/github.com/bsm/sarama-cluster@v2.1.15+incompatible/consumer.go:452 +0x4d
github.com/bsm/sarama-cluster.(*loopTomb).Go.func1(0xc000328f20, 0xc0003de960)
        /home/user/go/pkg/mod/github.com/bsm/sarama-cluster@v2.1.15+incompatible/util.go:73 +0xa9
created by github.com/bsm/sarama-cluster.(*loopTomb).Go
        /home/user/go/pkg/mod/github.com/bsm/sarama-cluster@v2.1.15+incompatible/util.go:69 +0x66
```

</p>
</details>

##### Problem Description

The following commit removes `c.Consumer.Offsets.CommitInterval` default value and the variable.

https://github.com/Shopify/sarama/commit/72a629d7d1ac130c1f1221e66396e0b54dfeef3f#diff-b4bda758a2aef091432646c354b4dc59L426

The following issue rollbacks the variable, but does not rollback the initial value.

https://github.com/Shopify/sarama/pull/1164

Thus, when sarama-cluster creates a `time.NewTicker` with the variable that is now 0, it panics.

https://golang.org/pkg/time/#NewTicker

d1egoaz:
If you are coming from sarama-cluster please see the closing comment on https://github.com/Shopify/sarama/pull/1618: 

tl/dr:
Please update your configuration to use `Consumer.Offsets.AutoCommit` instead of `Consumer.Offsets.CommitInterval`.
The field was added back to Sarama so sarama-cluster users can use the latest sarama version.

rogpeppe:
@d1egoaz That doesn't change the fact that this old code that's been working for ages crashes when the Sarama version is upgraded. It would be nice if that was fixed.

PR #1618, opened 2020-02-17
Add back Consumer.Offsets.CommitInterval default value #1616

reenjii:
https://github.com/Shopify/sarama/issues/1616

### DIFF
--- a/config.go
+++ b/config.go
@@ -457,6 +457,7 @@ func NewConfig() *Config {
 	c.Consumer.Return.Errors = false
 	c.Consumer.Offsets.AutoCommit.Enable = true
 	c.Consumer.Offsets.AutoCommit.Interval = 1 * time.Second
+	c.Consumer.Offsets.CommitInterval = 1 * time.Second
 	c.Consumer.Offsets.Initial = OffsetNewest
 	c.Consumer.Offsets.Retry.Max = 3
 


reenjii:
I signed Shopify's CLA, please re-launch the check. Thanks.

config.go
```diff
@@ -457,6 +457,7 @@ func NewConfig() *Config {
 	c.Consumer.Return.Errors = false
 	c.Consumer.Offsets.AutoCommit.Enable = true
 	c.Consumer.Offsets.AutoCommit.Interval = 1 * time.Second
+	c.Consumer.Offsets.CommitInterval = 1 * time.Second
```
d1egoaz: @reenjii I'm wondering if we need to add back the initial value....
We added this here: https://github.com/Shopify/sarama/pull/1590
image

But if we add back the initial value that would means that by default sarama will *always* log a deprecated message.

Couldn't this be solved by changing the value when creating the sarama-cluster config? as the field was added back

reenjii: Yeah I understand that it would look wierd.
I'll add this default value in our configuration then.

d1egoaz:
closed because of https://github.com/Shopify/sarama/pull/1618#discussion_r380694740

Thanks anyways @reenjii

PR #1619, opened 2020-02-18
tools/kafka-producer-performance: prune unused flag variables 

alrs:
This removes two unused variables that were intended as `flag` options in `tools/kafka-producer-performance`.


bai:
Thanks!

PR #1626, opened 2020-02-21
refresh 3rdparty dependencies

dnwe:
- move to versioned go modules
- bump all to latest releases
- go mod tidy


dnwe:
lz4 will get a basic workout as part of `functional_consumer_test.go`, but it doesn't look like there's any kerberos testing in that suite — we should probably add something to at least do a basic sniff test

Jianli-lin:
Hey. Got this error  `package github.com/pierrec/lz4/v3: cannot find package "github.com/pierrec/lz4/v3" in any of` when go get without go-module.
Can we not add `v3` in import path?

Jianli-lin:
Hey. @dnwe  Got this error when go build.
`/Shopify/sarama/compress.go:10:2: cannot find package "github.com/pierrec/lz4/v3" `
Does this change make it impossible to use this lib without go mod?

dnwe:
Yes you're right I'd forgotten that implication of migrating the deps. We should probably revert this until v2 of Sarama. I'll submit a follow up PR later today

Issue #1577, opened 2020-01-16
Lack of MockJoinGroupResponse in mockresponses.go


liuxin-reg:
master
I haven't found struct MockJoinGroupResponse in mockresponses.go, will we support that?

liuxin-reg:
I have supplement some kind of mock responses to make group consumer work, but not enough to submit a commit. Will try and continue working on it.

carldunham:
@liuxin-reg any luck with this? Happy to help out if you want.

carldunham:
I submitted a PR with a few mocks, but I think consumer group mocking will take a bit more than just that.

liuxin-reg:
@carldunham Thanks for your response and PR. I had add some basic mocks for consumer group responses too, it can works now, but without enough detailed considerations contained. Now, I removed this modification from folder vendor, since I don't think this kind of mock is necessary for normal consumer client.

gunturaf:
Any updates on this one @carldunham @liuxin-reg ?

dnwe:
Fixed by https://github.com/IBM/sarama/pull/1750

PR #1631, opened 2020-02-25
wip: add basic mocks for consumer group responses

carldunham:
Added some response mocks:
- MockJoinGroupResponse
- MockLeaveGroupResponse
- MockSyncGroupResponse
- MockHeartbeatResponse

Wasn't quite sure where to put unit tests for these, if they are needed.

Addresses #1577

### DIFF
--- a/mockresponses.go
+++ b/mockresponses.go
@@ -992,3 +992,154 @@ func (m *MockDeleteGroupsResponse) For(reqBody versionedDecoder) encoder {
 	}
 	return resp
 }
+
+type MockJoinGroupResponse struct {
+	t TestReporter
+
+	ThrottleTime  int32
+	Err           KError
+	GenerationId  int32
+	GroupProtocol string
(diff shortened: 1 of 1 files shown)


mockresponses.go
```diff
@@ -992,3 +992,154 @@ func (m *MockDeleteGroupsResponse) For(reqBody versionedDecoder) encoder {
...
+
+func NewMockJoinGroupResponse(t TestReporter) *MockJoinGroupResponse {
+	return &MockJoinGroupResponse{
+		t: t,
+
+		Members: make(map[string][]byte),
+	}
+}
+
+func (m *MockJoinGroupResponse) For(reqBody versionedDecoder) encoder {
```
yashschandra, line 1016: return type should be encodeWithHeader

gunturaf:
I think this PR is already solved by #1750 ?

PR #1643, opened 2020-03-17
fix `config.net.keepalive`

kzinglzy:
According to https://github.com/golang/go/blob/master/src/net/dial.go#L81

tcp keepalive is **ENABLED** when `KeepAlive` is set to 0 by default

### DIFF
--- a/config.go
+++ b/config.go
@@ -538,8 +538,6 @@ func (c *Config) Validate() error {
 		return ConfigurationError("Net.ReadTimeout must be > 0")
 	case c.Net.WriteTimeout <= 0:
 		return ConfigurationError("Net.WriteTimeout must be > 0")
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -96,8 +96,9 @@ type Config struct {
 			GSSAPI GSSAPIConfig
 		}
 
-		// KeepAlive specifies the keep-alive period for an active network connection.
-		// If zero, keep-alives are disabled. (default is 0: disabled).
+		// KeepAlive specifies the keep-alive period for an active network connection (defaults to 0).
+		// If zero or positive, keep-alives are enable.
```
bai:
```suggestion
		// If zero or positive, keep-alives are enabled.
```

bai:
Thanks for contributing 🙏

Issue #1629, opened 2020-02-24
client access old broker when reduce capacity of the cluster


Stephan14:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka | Go |
|--------|-------|----|
| Version 1.18.0 (2018-09-07) | 0.10.1 |  1.13|

##### Configuration

What configuration values are you using for Sarama and Kafka?

``` 
default configuration
```

##### Logs

When filing an issue please provide logs from Sarama and Kafka if at all
possible. You can set `sarama.Logger` to a `log.Logger` to capture Sarama debug
output.

<details><summary>logs: CLICK ME</summary>
<p>

```
Request was for a topic or partition that does not exist on this broker
```

</p>
</details>

##### Problem Description

When i try to reduce the capacity of cluster and remove data from broker A to broke B, if broker A does not shut down or replace to another cluster, but client still access broker A and client log show above.

dnwe:
Should be fixed by #1645 (thanks @Stephan14 )

PR #1645, opened 2020-03-19
Remove broker(s) which no longer exist in metadata

Stephan14:
This pull request solves the problem: #1629. I find it only have update and add broker when update brokers by latest metadata, however there is no code of remove broker which if offline

### DIFF
--- a/client.go
+++ b/client.go
@@ -562,6 +562,30 @@ func (client *client) RefreshCoordinator(consumerGroup string) error {
 
 // private broker management helpers
 
+func (client *client) updateBroker(brokers []*Broker) {
+	var currentBroker = make(map[int32]*Broker, len(brokers))
+
+	for _, broker := range brokers {
+		currentBroker[broker.ID()] = broker
+		if client.brokers[broker.ID()] == nil { // add new broker
+			client.brokers[broker.ID()] = broker
+			Logger.Printf("client/brokers registered new broker #%d at %s", broker.ID(), broker.Addr())
+		} else if broker.Addr() != client.brokers[broker.ID()].Addr() { // replace broker with new address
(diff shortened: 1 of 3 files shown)


Stephan14:
Hey @d1egoaz, I have found some error as follows:
image
 And run test on my Mac, it show as follows:
image
I think `Got err: kafka: client has run out of available brokers to talk to` is only one log and the code is right, so I delete this log

Stephan14:
Hey @bai @d1egoaz,   I find other PR appear some error on CI, for example https://github.com/Shopify/sarama/pull/1640/checks?check_run_id=504088005,  but it check successful. So how to solve the problem.

d1egoaz:
Thanks @Stephan14 for the new PR, so it seems the changes that you are introducing are indeed changing/breaking some expectation in some tests. We'll need to 👀 what's going on.

Stephan14:
Hey @d1egoaz , how about this problem is going? Can I merge this PR to master? The failures is as follow:
image

Stephan14:
@dnwe , oh, sorry. I have fixed the problem and run `go test` on my mac successfully as follows:
image
But it runs failed in some CI as follows:
image
How to fix it

dnwe:
@Stephan14 thanks for the fix. Looks like go actions had some issue fetching dependencies, but all the re-runs are green. Merging

brenley:
Hello! Just curious if this will be in the next release and when this is slated for? We have observed this issue and are deciding on whether to wait or deviate from release pattern and just pull in master.

dnwe:
@brenley theres a few other open issues and it would be good to get some testing done against the final GA of kafka 2.5 but I'd imagine we'd be able to release a semver bump within a short period of time

PR #1646, opened 2020-03-19
Add DescribeLogDirs to admin client

mimaison:


admin.go
```diff
@@ -878,3 +881,48 @@ func (ca *clusterAdmin) DeleteConsumerGroup(group string) error {
...
+	logDirsMaps := make(chan map[int32][]DescribeLogDirsResponseDirMetadata, len(brokerIds))
+	errors := make(chan error, len(brokerIds))
+	wg := sync.WaitGroup{}
+
+	for _, b := range brokerIds {
+		wg.Add(1)
+		broker, err := ca.findBroker(b)
+		if err != nil {
+			Logger.Printf("Unable to find broker with ID = %v\n", b)
+			continue
```
Sjeanpierre: In the case of a missing broker, will the waitgroup ever finish waiting since a done was never called at this point?

mimaison: In case a broker does not reply, it will block for `config.Net.ReadTimeout` and return with an error `read tcp 127.0.0.1:60810->127.0.0.1:9092: i/o timeout`.

The same happens in `ListConsumerGroups`.

PR #1666, opened 2020-04-15
Bugfix: Allow TLS connections to work over socks proxy.

KevinJCross:
It seems that the code was either proxy or tls but in my case I need tls over the socks proxy so Ive created this PR to fix this issue.
Note there seems to be an issue with the tls test where the server name was never set so Ive fixed that too.

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -154,25 +154,20 @@ func (b *Broker) Open(conf *Config) error {
 	go withRecover(func() {
 		defer b.lock.Unlock()
 
-		dialer := net.Dialer{
-			Timeout:   conf.Net.DialTimeout,
-			KeepAlive: conf.Net.KeepAlive,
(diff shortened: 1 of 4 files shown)


KevinJCross:
@probot-shopify
CLA — Contributor License Agreement (CLA) status  Signed :)

KevinJCross:
@github-actions CI / Go 1.13.x with Kafka 2.4.0 on Ubuntu (pull_request) 
```
+ bin/zookeeper-server-stop.sh
No zookeeper server to stop
##[error]Process completed with exit code 1.
```
hmmm seems like an intermittant problem exists ;)

d1egoaz:
> Note there seems to be an issue with the tls test where the server name was never set so Ive fixed that too.

@KevinJCross did you add the code for this?

PR #1686, opened 2020-05-01
feat: send message with async receive way

xujianhai666:
## MOTIVATION
now, producer use method `sendAndReceive` of Broker to produceMessage, which is blocking model for next send, producer will wait too long when broker slows down, which cause later request waste too much time.

## CHANGE
we split produce and receive method, proudcer send proudcer produceSet one by one, and will push promise to pending channel to wait, so prodcuer will not block on produce, which decrease the cost of next send request. pending channel is consumed one by one too, because the request/response is in order, so next response always arrives later

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ _testmain.go
 
 coverage.txt
 profile.out
+/.idea
+*.diff
+*.iml
+
--- a/async_producer.go
+++ b/async_producer.go
@@ -637,12 +637,20 @@ func (pp *partitionProducer) updateLeader() error {
 	})
 }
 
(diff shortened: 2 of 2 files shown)


xujianhai666:
@bai install dependencies failed, pls check and run again

bai:
Could you please rebase your branch off of latest master? 🙏

bai:
I rebased this PR off of latest master.

bai:
@dnwe Do you think it's something we want to get merged? Asking since this PR's been open for a while.

dnwe (approved):

PR #1687, opened 2020-05-05
Fix auto-commit bug and make a feature of auto-commit

lqiz:
I believe that the sarama-cluster hasn’t designed the auto-commit feature. What we have to do is updating the offset via MarkOffset/MarkMessage after consuming the message.

Here is its reason:

Offset manager is completely independent of the consumer. flushToBroker in main-loop has only flush offset to the broker but it doesn’t update the offset.
It has no relation between the offset updating and the consumer in sarama. Currently, the offset updating depends on the Markoffset after consuming the message.
If auto-commit want to be supported in sarama, it should update the offset when fetching messages. After offset updating, the main-loop will handle commit things left.

So this fix is made via updating the offset.

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -682,6 +682,19 @@ func (s *consumerGroupSession) consume(topic string, partition int32) {
 		s.parent.handleError(err, topic, partition)
 	}
 
+	// update offset when auto commit
+	go func() {
+		for {
+			if !s.parent.config.Consumer.Offsets.AutoCommit.Enable {
+				continue
+			}
+			offsetFetched := claim.PartitionConsumer.OffsetFetched()
+			s.MarkOffset(topic, partition, offsetFetched, "")
+			time.Sleep(time.Second)
+		}
+	}()
+
+
 	// ensure consumer is closed & drained
 	claim.AsyncClose()
 	for _, err := range claim.waitClosed() {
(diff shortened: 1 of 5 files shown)


consumer_group.go
```diff
@@ -682,6 +682,19 @@ func (s *consumerGroupSession) consume(topic string, partition int32) {
...
+		for {
+			if !s.parent.config.Consumer.Offsets.AutoCommit.Enable {
+				continue
+			}
+			offsetFetched := claim.PartitionConsumer.OffsetFetched()
+			s.MarkOffset(topic, partition, offsetFetched, "")
+			time.Sleep(time.Second)
+		}
+	}()
+
```
kamyuentse, line 696: What are you doing?!  You mark all fetched messages consumed every second? And not respect to the closing event (`ctx.Done()` or `parent.closed`), which cause goroutine leak.

luoruiyi: Thanks a lot for you review.
I want to implement the auto-commit feature.  Yes, closing event should be respected.

kamyuentse: IIRC, the auto-commit feature already exists.

luoruiyi: See this thread, https://github.com/Shopify/sarama/issues/1570
The auto-commit feature won't work, and i have add analysis in the desc of this fix.

dnwe:
@lqiz thanks for putting this PR together. I _think_ that the functionality you were looking for has now been covered as part of the merged PR #1699 which correctly disables the auto-commit loop if you choose to disable auto-commits, and provides a `Commit()` func on the consumer group to elect when you'd like to commit.

Feel free to come back if you think there is still some capability missing here though.

PR #1692, opened 2020-05-07
Set ServerName using tls.DialWithDialer approach

d1egoaz:
fixes https://github.com/Shopify/sarama/issues/1691

TBH I'm not sure if this is enough.

There was a big change here:
https://github.com/Shopify/sarama/pull/1666/files#diff-019034382244c82f15c92490660e4a18
`s/tls.DialWithDialer/tls.Client`

and all code that was in `tls.DialWithDialer` is not in `tls.Client`, so no handshake, timeouts, etc.
https://github.com/golang/go/blob/go1.14.2/src/crypto/tls/tls.go#L102

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -165,6 +165,17 @@ func (b *Broker) Open(conf *Config) error {
 
 		if conf.Net.TLS.Enable {
 			Logger.Printf("Using tls")
+
+			// If no ServerName is set, infer the ServerName
+			// from the hostname we're connecting to.
+			if conf.Net.TLS.Config.ServerName == "" {
+				colonPos := strings.LastIndex(b.addr, ":")
+				if colonPos == -1 {
+					colonPos = len(b.addr)
+				}
+				hostname := b.addr[:colonPos]
+				conf.Net.TLS.Config.ServerName = hostname
+			}
 			b.conn = tls.Client(b.conn, conf.Net.TLS.Config)
 		}
 
--- a/broker.go
+++ b/broker.go
@@ -165,18 +165,22 @@ func (b *Broker) Open(conf *Config) error {
 
 		if conf.Net.TLS.Enable {
 			Logger.Printf("Using tls")
-
+			cfg := conf.Net.TLS.Config
+			if cfg == nil {
+				cfg = &tls.Config{}
+			}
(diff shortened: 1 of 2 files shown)


broker.go
```diff
@@ -165,6 +165,17 @@ func (b *Broker) Open(conf *Config) error {
 
 		if conf.Net.TLS.Enable {
 			Logger.Printf("Using tls")
+
+			// If no ServerName is set, infer the ServerName
+			// from the hostname we're connecting to.
+			if conf.Net.TLS.Config.ServerName == "" {
```
dnwe, line 171: That's going to segfault in the default case where the Sarma user has set `conf.Net.TLS.Enable = true` but hasn't change `conf.Net.TLS.Config` from the default nil pointer value

To restore the old behaviour you need to add a:
```go
cfg := conf.Net.TLS.Config
if cfg == nil {
    cfg = &tls.Config{}
    // then proceeed to set ServerName 
}

```

d1egoaz: See my updated comment on the PR.
Not sure if my change + your proposal (which I agree) will be enough

d1egoaz: currently master doesn't do a TLS handshake per `tls.Client`, this makes me nervous :/

d1egoaz: scratch that.
https://pkg.go.dev/crypto/tls?tab=doc#Conn.Handshake

> Handshake runs the client or server handshake protocol if it has not yet been run. Most uses of this package need not call Handshake explicitly: the first Read or Write will call it automatically.

d1egoaz: I just pushed the config change. Still need to figure out how to add a test for this.

The big difference now is that we aren't doing the timeouts/deadlines and handshake checks on this specific point, unsure what would be the effect of that.

d1egoaz: I did modify one test to remove the `ServerName`, that might be good enough for now https://github.com/Shopify/sarama/pull/1692/files#diff-e9534989c7143940469fa20938dbcb71

what do you think @dnwe ?

dnwe: 👍🏻

d1egoaz:
Thanks for being around @dnwe

PR #1693, opened 2020-05-07
testfix: give kafka longer to startup

dnwe:
For some reason we'd ended up with quite a short (10s) limit on waiting
for Kafka to startup and open its listener and since moving from Travis
to GitHub actions, this can periodically take longer causing the build
to abort early. Bump it up to 2m and tweak a couple of other parts of
the boot script.

### DIFF
--- a/vagrant/boot_cluster.sh
+++ b/vagrant/boot_cluster.sh
@@ -4,8 +4,8 @@ set -ex
 
 # Launch and wait for toxiproxy
 ${REPOSITORY_ROOT}/vagrant/run_toxiproxy.sh &
-while ! nc -q 1 localhost 2181 </dev/null; do echo "Waiting"; sleep 1; done
-while ! nc -q 1 localhost 9092 </dev/null; do echo "Waiting"; sleep 1; done
+while ! nc -q 1 localhost 2185 </dev/null; do echo "Waiting"; sleep 1; done
+while ! nc -q 1 localhost 9095 </dev/null; do echo "Waiting"; sleep 1; done
 
 # Launch and wait for Zookeeper
 for i in 1 2 3 4 5; do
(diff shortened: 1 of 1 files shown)


vagrant/boot_cluster.sh
```diff
@@ -4,8 +4,8 @@ set -ex
 
 # Launch and wait for toxiproxy
 ${REPOSITORY_ROOT}/vagrant/run_toxiproxy.sh &
-while ! nc -q 1 localhost 2181 </dev/null; do echo "Waiting"; sleep 1; done
-while ! nc -q 1 localhost 9092 </dev/null; do echo "Waiting"; sleep 1; done
+while ! nc -q 1 localhost 2185 </dev/null; do echo "Waiting"; sleep 1; done
+while ! nc -q 1 localhost 9095 </dev/null; do echo "Waiting"; sleep 1; done
```
d1egoaz, line 8: Were these ports change intended ?

vagrant/boot_cluster.sh
```diff
@@ -21,21 +21,20 @@ for i in 1 2 3 4 5; do
 done
 ps auxww | grep -i kafka
 
-N=10
+N=120
 RC=1
 set +x
-printf "Waiting for Kafka to become available."
+printf "Waiting for kafka5 to become available."
```
d1egoaz, line 27: Is the 5 intended or a typo?

vagrant/boot_cluster.sh
dnwe: Yeah these `nc` polls are waiting for toxiproxy to have completed its startup. However, it starts 5 zookeeper listeners and 5 kafka listeners so it makes sense to poll the highest port numbers (which are created last) for zk-5 and kafka-5 respectively

vagrant/boot_cluster.sh
dnwe: Intended, the `nc` poll in this block is only querying the fifth broker in the cluster, known as "kafka5" in the toxiproxy config and toxi port 909`5` --> kafka port 2909`5`

PR #1699, opened 2020-05-11
Consumer group support for manually comitting offsets

### LINKED ISSUE #1570: commit offset manually when using consumer group
##### Versions

sarama version : Version 1.25.0 
go version : go1.12.7 
Kafka version: 2.2.1

 My use case is: I am reading from a topic, so in case my my consumer fails I should be able to read from my last successful commit rather, If I set in time interval there is chance of message loss. I want to commit manually while using consumer group, I don't want to commit in intervals.
I went through offset_manager.go, but I am unable to do. Kindly help.

wclaeys:
Provide manual Commit method and don't create mainloop if AutoCommit is disabled

Fixes #1570

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -513,6 +513,11 @@ type ConsumerGroupSession interface {
 	// message twice, and your processing should ideally be idempotent.
 	MarkOffset(topic string, partition int32, offset int64, metadata string)
 
+	// Commit the offset to the backend
+	//
+	// Note: calling Commit performs a blocking synchronous operation.
+	Commit()
+
 	// ResetOffset resets to the provided offset, alongside a metadata string that
 	// represents the state of the partition consumer at that point in time. Reset
 	// acts as a counterpart to MarkOffset, the difference being that it allows to
@@ -622,6 +627,10 @@ func (s *consumerGroupSession) MarkOffset(topic string, partition int32, offset 
 	if pom := s.offsets.findPOM(topic, partition); pom != nil {
 		pom.MarkOffset(offset, metadata)
 	}
+}
+
+func (s *consumerGroupSession) Commit() {
+	s.offsets.Commit()
 }
 
 func (s *consumerGroupSession) ResetOffset(topic string, partition int32, offset int64, metadata string) {
--- a/offset_manager_test.go
+++ b/offset_manager_test.go
@@ -169,6 +169,64 @@ func TestNewOffsetManagerOffsetsAutoCommit(t *testing.T) {
 	}
 }
 
+func TestNewOffsetManagerOffsetsManualCommit(t *testing.T) {
+	// Tests to validate configuration when `Consumer.Offsets.AutoCommit.Enable` is false
+	config := NewConfig()
+	config.Consumer.Offsets.AutoCommit.Enable = false
+
+	om, testClient, broker, coordinator := initOffsetManagerWithBackoffFunc(t, 0, nil, config)
+	pom := initPartitionOffsetManager(t, om, coordinator, 5, "original_meta")
+
+	// Wait long enough for the test not to fail..
+	timeout := 50 * config.Consumer.Offsets.AutoCommit.Interval
+
+	ocResponse := new(OffsetCommitResponse)
+	ocResponse.AddError("my_topic", 0, ErrNoError)
+	called := make(chan none)
+	handler := func(req *request) (res encoderWithHeader) {
+		close(called)
+		return ocResponse
+	}
+	coordinator.setHandler(handler)
+
+	// Should not trigger an auto-commit
+	expected := int64(1)
+	pom.ResetOffset(expected, "modified_meta")
+	_, _ = pom.NextOffset()
+
+	select {
+	case <-called:
+		// OffsetManager called on the wire.
+		t.Errorf("Received request when AutoCommit is disabled")
+	case <-time.After(timeout):
+		// Timeout waiting for OffsetManager to call on the wire.
+		// OK
+	}
+
+	// Setup again to test manual commit
+	called = make(chan none)
+
+	om.Commit()
+
+	select {
+	case <-called:
+		// OffsetManager called on the wire.
+		// OK
+	case <-time.After(timeout):
+		// Timeout waiting for OffsetManager to call on the wire.
+		t.Errorf("No request received for after waiting for %v", timeout)
+	}
+
+	// Close up
+	broker.Close()
+	coordinator.Close()
+
+	// !! om must be closed before the pom so pom.release() is called before pom.Close()
+	safeClose(t, om)
+	safeClose(t, pom)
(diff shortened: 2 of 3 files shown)


knoguchi:
I'm just a user but how are you going to return errors?

wclaeys:
In the same manner as how it's now being done in the 'mainLoop' when AutoCommit.Enable is true.

BPScott:
Hi @wclaeys, your CLA check is failing because your commits aren't associated with your Github account. You need to connect your email address with your github account per https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/adding-an-email-address-to-your-github-account.

wclaeys:
@BPScott , thank you for your message. I've updated my profile.

Gypsying:
Is this OK ?
What time will merge the pr?

Morriaty-The-Murderer:
Hi guys, how is this pr going now?

pengfei-xue:
fixed yet?

offset_manager_test.go
```diff
@@ -169,6 +169,76 @@ func TestNewOffsetManagerOffsetsAutoCommit(t *testing.T) {
...
+
+	// Setup again to test manual commit
+	ocResponse = new(OffsetCommitResponse)
+	ocResponse.AddError("my_topic", 0, ErrNoError)
+	called = make(chan none)
+	handler = func(req *request) (res encoderWithHeader) {
+		close(called)
+		return ocResponse
+	}
+	coordinator.setHandler(handler)
```
d1egoaz, line 214: I think this part is not needed, it'll reuse the previous one

offset_manager_test.go
```diff
@@ -169,6 +169,76 @@ func TestNewOffsetManagerOffsetsAutoCommit(t *testing.T) {
...
+
+	// Should trigger a commit
+	go func() {
+		// Defer execution so we can listen to channels
+		time.Sleep(1 * time.Millisecond)
```
d1egoaz, line 219: maybe bump this to 100ms, seems 1ms might hit some edge cases in CI

offset_manager_test.go
wclaeys: That's correct, but we still need to re-create the 'called' channel otherwise we'll have an always green test.

offset_manager_test.go
wclaeys: I was able to remove the defer execution and sleep all together, wasn't actually required for this test case.

d1egoaz:
this looks LGTM, but I'd love to have another 👍 from another Sarama maintainer @bai @dnwe @varun06 @mimaison @edoardocomar

consumer_group.go
```diff
@@ -513,6 +513,9 @@ type ConsumerGroupSession interface {
 	// message twice, and your processing should ideally be idempotent.
 	MarkOffset(topic string, partition int32, offset int64, metadata string)
 
+	// Commit the offset to the backend
+	Commit()
```
dnwe, line 517: Do we want to call out that this is a blocking sync operation, or will that just be expected?

consumer_group.go
```diff
@@ -624,6 +627,10 @@ func (s *consumerGroupSession) MarkOffset(topic string, partition int32, offset
 	}
 }
 
+func (s *consumerGroupSession) Commit() {
+	s.offsets.Commit()
+}
```
dnwe, line 632: Do we need a similar check of the session ctx here, like we have in consume to prevent us from attempting to commit after a rebalance has already started?

```go
// quick exit if rebalance is due
select {
case <-s.ctx.Done():
	return
case <-s.parent.closed:
	return
default:
}
```

wclaeys: @dnwe , that would introduce new behaviour. 
```
        om.flushToBroker()
	om.releasePOMs(false)
```
Above statements were used on the `case <-om.ticker.C` case in the `(om *offsetManager) mainLoop()` function.
Now both (async with 'mainLoop' and sync with 'Commit') use the new 'Commit' function using the exact same contents as above.
```
func (om *offsetManager) Commit() {
	om.flushToBroker()
	om.releasePOMs(false)
}
```

wclaeys:
Can somebody please cancel this new CLA check? Accidentally pushed without a proper e-mail setting (reverted it and did a proper commit). @BPScott maybe? Thank you!

dnwe:
@wclaeys I think the PR is ready, so why don’t you squash your commits now and force push again with your correct email, that should pass the CLA check

dnwe:
@wclaeys you need to rebase your branch on master to get the checks to pass

> This branch is 1 commit ahead, 24 commits behind Shopify:master.

dnwe:
@wclaeys ah don't worry, I found that I had permission to push to your branch myself so I performed the rebase. If the checks both pass green then I will go ahead and merge. Thanks for your contribution on this one, really appreciated 👍

dnwe:
As there's been a lot of interest in this functionality, it would be great if all the people watching this PR and the accompanying issues could test it out with their application code once it's merged, in advance of us cutting any new release version.

Assuming you're using go modules to manage your dependencies, once this has merged, you can update your go.mod to fetch the latest commit from the default branch by doing:
```
go get github.com/Shopify/sarama@HEAD
```

bglmmz:
How about a single consumer?  Could a  single consumer commit offsets manually?

aneeskA:
> As there's been a lot of interest in this functionality, it would be great if all the people watching this PR and the accompanying issues could test it out with their application code once it's merged, in advance of us cutting any new release version.
> 
> Assuming you're using go modules to manage your dependencies, once this has merged, you can update your go.mod to fetch the latest commit from the default branch by doing:
> 
> ```
> go get github.com/Shopify/sarama@HEAD
> ```

@dnwe Is there any timeline for releasing this feature?

alok87:
Please suggest when are you releasing this, this helpful in cases where you want the commit to happen instantly, systems which care about order and duplicates really need it.

jameshalsall:
doesn't this break backwards compatibility with the addition of the new interface method?

d1egoaz:
@jameshalsall yup

I think we could apply this trick if we don't want users to implement directly the interface
``` go
    // A private method to prevent users implementing the
    // interface and so future additions to it will not
    // violate Go 1 compatibility.
    private()
```
from https://blog.golang.org/module-compatibility

what do you think @dnwe ?

this PR was also merged adding a new method to an existing interface https://github.com/Shopify/sarama/pull/1781

jameshalsall:
We implement this interface in our tests so we can use mock versions in our handlers,  I’m not sure preventing people from implementing it is the right option from my perspective.

hipper:
Hey @wclaeys , found a bug that looks like introduced after this change. The problem is when you close the consumer it ignores the AutoCommit flag and commits the offset before it exits. That problem doesn't exists in `1.26.4`

Described more here: https://github.com/Shopify/sarama/issues/1843

PR #1701, opened 2020-05-12
Set server name only for the current broker

d1egoaz:
Fixes https://github.com/Shopify/sarama/issues/1700

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -1427,14 +1427,15 @@ func validServerNameTLS(addr string, cfg *tls.Config) *tls.Config {
 	if cfg == nil {
 		cfg = &tls.Config{}
 	}
-	if cfg.ServerName == "" {
-		c := cfg.Clone()
-		sn, _, err := net.SplitHostPort(addr)
-		if err != nil {
-			Logger.Println(fmt.Errorf("failed to get ServerName from addr %w", err))
-		}
-		c.ServerName = sn
-		cfg = c
+	if cfg.ServerName != "" {
+		return cfg
 	}
-	return cfg
+
+	c := cfg.Clone()
+	sn, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		Logger.Println(fmt.Errorf("failed to get ServerName from addr %w", err))
+	}
+	c.ServerName = sn
+	return c
 }
--- a/broker.go
+++ b/broker.go
@@ -162,29 +162,11 @@ func (b *Broker) Open(conf *Config) error {
 			atomic.StoreInt32(&b.opened, 0)
 			return
 		}
-
 		if conf.Net.TLS.Enable {
-			Logger.Printf("Using tls")
-			cfg := conf.Net.TLS.Config
-			if cfg == nil {
-				cfg = &tls.Config{}
-			}
-			// If no ServerName is set, infer the ServerName
-			// from the hostname we're connecting to.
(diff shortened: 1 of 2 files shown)


broker.go
```diff
@@ -1440,3 +1422,25 @@ func (b *Broker) registerCounter(name string) metrics.Counter {
 	b.registeredMetrics = append(b.registeredMetrics, nameForBroker)
 	return metrics.GetOrRegisterCounter(nameForBroker, b.conf.MetricRegistry)
 }
+
+func validServerNameTLS(addr string, conf *tls.Config) *tls.Config {
+	cfg := conf
```
dselyuzhitskiy, line 1427: @d1egoaz i think here we don't need separate variable and could use conf from args just checked if it nil

dnwe:
@d1egoaz rather than the special splitting code, can't we just use [net.SplitHostPort](https://golang.org/pkg/net/#SplitHostPort) ?

Also, I can't think of a good reason why anyone would have set a single ServerName in their `conf.Net.TLS.Config` to use for all brokers, so I wonder if we should always override whatever is there by forming it with our hostname.

So I wonder if we'd just do something like this:

```go
	var cfg *tls.Config
	if config.Net.TLS.Config == nil {
		cfg = &tls.Config{}
	} else {
		cfg = config.Net.TLS.Config.Clone()
	}

	if cfg.ServerName, _, err = net.SplitHostPort(broker); err != nil {
		return fmt.Errorf("failed to extract SNI got %w", err)
	}

	b.conn = tls.Client(b.conn, cfg)
```

d1egoaz:
> @d1egoaz rather than the special splitting code, can't we just use net.SplitHostPort ?

for sure! thanks!

> Also, I can't think of a good reason why anyone would have set a single ServerName in their conf.Net.TLS.Config to use for all brokers, so I wonder if we should always override whatever is there by forming it with our hostname.

yeah, but I wonder if we would break any other client for doing this, looks like a breaking change.

d1egoaz:
@dnwe @dselyuzhitskiy let me know what you guys think about the latest commits, and if it's approved I'll create a new release today.

I'm testing this branch in one of my apps, and no issues so far

broker.go
```diff
@@ -1440,3 +1422,20 @@ func (b *Broker) registerCounter(name string) metrics.Counter {
...
+		cfg = &tls.Config{}
+	}
+	if cfg.ServerName != "" {
+		return cfg
+	}
+
+	c := cfg.Clone()
+	sn, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		Logger.Println(fmt.Errorf("failed to get ServerName from addr %w", err))
```
dselyuzhitskiy, line 1437: Should we return with error here?

dselyuzhitskiy:
Looks good, but suggestion from @dnwe is also interesting

broker.go
d1egoaz: I should have mentioned this, as this was on purpose.

Currently, this code is called inside a go routine:
https://github.com/Shopify/sarama/blob/master/broker.go#L154

and currently the pattern inside this routine is to log the errors as there is no way to return an error to the caller.

d1egoaz:
@dnwe let us know what do you think about the latest changes.
I'm not sure about overriding the user's configuration, looks like this change might be for v2

dnwe:
@d1egoaz lets go ahead and merge and get another minor version bump out 👍🏻

PR #1707, opened 2020-05-27
Add a function to refresh broker list in client

readygodom:
Add a function to refresh broker list in client.
Fix: https://github.com/danielqsj/kafka_exporter/issues/134#issuecomment-624670818

### DIFF
--- a/client.go
+++ b/client.go
@@ -44,6 +44,10 @@ type Client interface {
 	// topic/partition, as determined by querying the cluster metadata.
 	Leader(topic string, partitionID int32) (*Broker, error)
 
+	// RefreshBrokers refresh broker list of this client, and it removes
+	// brokers went missing.
+	RefreshBrokers()
+
 	// Replicas returns the set of all replica IDs for the given partition.
 	Replicas(topic string, partitionID int32) ([]int32, error)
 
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -429,6 +433,29 @@ func (client *client) Leader(topic string, partitionID int32) (*Broker, error) {
 	return leader, err
 }
 
+func (client *client) RefreshBrokers() {
```
varun06, line 436: There is already RefreshController and Brokers that work on latest metadata. Can you please elaborate on what problem does it solve??

readygodom: Yes, a function already exists. But it does not work on project kafka-exporter with 'client.RefreshMetadata()': https://github.com/danielqsj/kafka_exporter/issues/134#issuecomment-624670818, so I add a
function to refresh brokers specially.

dnwe:
@readygodom kafka_exporter is using github.com/Shopify/sarama v1.22.1 which is missing the fix #1645 from @Stephan14 - Remove broker(s) which no longer exist in metadata

I think if you just bump kafka_exporter to latest Sarama release then your original
problem will be fixed already

readygodom:
We can use new release after fix:https://github.com/Shopify/sarama/pull/1645 to resolve this problem @Shini31 https://github.com/danielqsj/kafka_exporter/issues/134#issuecomment-624670818

PR #1712, opened 2020-05-30
fix metadata retry backoff invalid when get metadata failed

Stephan14:
https://github.com/Shopify/sarama/issues/1711  add global time to store update metadata time and  use it when client retry


dnwe:
@Stephan14 thanks for this PR. The fix looks good to me, but please could you add a small test to cover the basic logic and assert update is run or skipped based on the last update time? That way we won’t regress it in the future

Stephan14:
Ok, I will do it

Stephan14:
Hi, i create new MR to fix the problem. https://github.com/Shopify/sarama/pull/1959

dnwe:
Thanks, will close this PR as superceded by https://github.com/Shopify/sarama/pull/1959

PR #1718, opened 2020-06-09
Fix the error handling in client.InitProducerID()

crivera-fastly:
err was shadowed in InitProducerID() so it would never return a non-nil
error. It now returns ErrOutOfBrokers if there aren't any registered brokers.
If there are registered brokers and the broker calls fail then it returns
the last encoutered error.

This was causing a panic in newTransactionManager() with idempotent
production enabled because nil, nil was being returned if the brokers were
unavailable.


bai:
I rebased this branch off of latest master.

dnwe (approved):

PR #1696, opened 2020-05-11
KIP-392: Allow consumers to fetch from closest replica

dneralla:
[KIP-392: Allow consumers to fetch from closest replica](https://cwiki.apache.org/confluence/display/KAFKA/KIP-392%3A+Allow+consumers+to+fetch+from+closest+replica)

PR #1728, opened 2020-06-19
Allow consumer configuration to specify replica selector

AusIV:
Rather than always using the leader partition, this lets the consumer application specify a function to select which broker they wish to consume from. The function has the signature 

    fn(topic string, partition int32, client Client) (*Broker, error)

Allowing the client to implement their own logic for broker slection. If none is specified, the default behavior is backwards compatible and selects the leader.

This pulls some changes from #1696, but in our use case we want to spread the load across multiple replicas, not pick based on geographic considerations. I didn't think our specific use case made sense for the general library, but it also wasn't possible without changes to Sarama, so I made a generalizable solution. It might be possible to implement the changes for KIP-392 as a ReplicaSelector function, though that may still require some additional code changes to support communications regarding which broker should be used for a given rack.

I've tested that this works with sufficiently high protocol versions.

If you'd like to take a different approach with this, I'm open to helping. If you need more / different tests, let me know.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -144,10 +144,16 @@ func (c *consumer) ConsumePartition(topic string, partition int32, offset int64)
 		return nil, err
 	}
 
-	var leader *Broker
+	var replica *Broker
 	var err error
-	if leader, err = c.client.Leader(child.topic, child.partition); err != nil {
-		return nil, err
+	if c.conf.Consumer.ReplicaSelector == nil {
+		if replica, err = c.client.Leader(child.topic, child.partition); err != nil {
+			return nil, err
+		}
+	} else {
+		if replica, err = c.conf.Consumer.ReplicaSelector(child.topic, child.partition, c.client); err != nil {
+			return nil, err
+		}
 	}
 
 	if err := c.addChild(child); err != nil {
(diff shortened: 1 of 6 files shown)


consumer.go
```diff
@@ -144,10 +144,16 @@ func (c *consumer) ConsumePartition(topic string, partition int32, offset int64)
 		return nil, err
 	}
 
-	var leader *Broker
+	var replica *Broker
 	var err error
-	if leader, err = c.client.Leader(child.topic, child.partition); err != nil {
-		return nil, err
+	if c.conf.Consumer.ReplicaSelector == nil {
+		if replica, err = c.client.Leader(child.topic, child.partition); err != nil {
```
dnwe, line 150: I wonder if it is worth setting up ReplicaSelector in NewConfig to be:
```go
func(topic string, partition int32, client Client) (*Broker, error) {
        return client.Leader(child.topic, child.partition)
}
```
and then you wouldn’t need the nil checking logic to choose how to select which partition to consume from in here

AusIV: Good idea. I'll get that in.

AusIV: I have adjusted per your feedback.

dnwe:
Thanks, can you rebase atop master so the checks pass

> This branch is 2 commits ahead, 10 commits behind Shopify:master.

d1egoaz:
> I didn't think our specific use case made sense for the general library, but it also wasn't possible without changes to Sarama, so I made a generalizable solution. 

🤔 unsure if this code should live in Sarama, as it seems only you have the use case for it, it is not part of any KPI, and it differs from KIP-392: Allow consumers to fetch from closest replica.

Maybe this code should live in your own fork? 🤷

PR #1730, opened 2020-06-22
KIP-42 Add producer and consumer interceptors

d1egoaz:
[KIP-42 Add producer and consumer interceptors](https://cwiki.apache.org/confluence/display/KAFKA/KIP-42%3A+Add+Producer+and+Consumer+Interceptors)

This PR includes:
Producer: `OnSend` but it doesn't implement `onAcknowledgement`
Consumer: `OnConsume` but it doesn't implement `onCommit`

I tried on purpose to not follow the Java design too closely. This PR tries to follow Go idioms to the degree possible.
I didn't add an `onClose` method on the interface, Go has other alternatives for this if needed.

for:
> Since there may be multiple interceptors, the first interceptor will get a record from client passed as the 'record' parameter. The next interceptor in the list will get the record returned by the previous interceptor, and so on. Since interceptors are allowed to mutate records, interceptors may potentially get the record already modified by other interceptors. However, we will state in the javadoc that building a pipeline of mutable interceptors that depend on the output of the previous interceptors is discouraged, because of potential side-effects caused by interceptors potentially failing to mutate the record and throwing and exception. If one of the interceptors in the list throws an exception from onSend(), the exception is caught, logged,and the next interceptor is called with the record returned by the last successful interceptor in the list, or otherwise the client.

specifically the section:
> If one of the interceptors in the list throws an exception from onSend(), the exception is caught, logged,and the next interceptor is called with the record returned by the last successful interceptor in the list, or otherwise the client.

My implementation does the following, as I'm passing the value by reference:
If one of the interceptors in the list throws an exception from onSend(), the exception is caught, logged,and the next interceptor is called ~with the record returned by the last successful interceptor in the list~.

The interceptor method needs to make sure to only modify the object when it's safe to do so, otherwise it can modify the object and then fail, which will make the next interceptor to use a modified value from maybe a non successful interceptor.
For some reason, Sarama `ProducerMessage` contains a channel, which makes hard to clone this object.

Another alternative will be to make a copy of the body, headers, topic, partition data, and if the interceptor call fails, revert the message to this original data, but I'm not sure if it worth the effort, as anyone working with go know that if you modify an object passed as a reference, it'll modify the source object.

Closes https://github.com/Shopify/sarama/issues/1710

### DIFF
--- a/async_producer_test.go
+++ b/async_producer_test.go
@@ -1235,7 +1235,7 @@ type appendInterceptor struct {
 	i int
 }
 
-func (b *appendInterceptor) onSend(msg *ProducerMessage) {
+func (b *appendInterceptor) OnSend(msg *ProducerMessage) {
 	if b.i < 0 {
 		panic("hey, the interceptor has failed")
 	}
@@ -1244,7 +1244,7 @@ func (b *appendInterceptor) onSend(msg *ProducerMessage) {
 	b.i++
 }
 
-func (b *appendInterceptor) onConsume(msg *ConsumerMessage) {
+func (b *appendInterceptor) OnConsume(msg *ConsumerMessage) {
 	if b.i < 0 {
 		panic("hey, the interceptor has failed")
 	}
--- a/async_producer_test.go
+++ b/async_producer_test.go
@@ -1237,7 +1237,7 @@ type appendInterceptor struct {
 
 func (b *appendInterceptor) onSend(msg *ProducerMessage) {
 	if b.i < 0 {
-		panic("hey, the interceptor have failed")
+		panic("hey, the interceptor has failed")
 	}
 	v, _ := msg.Value.Encode()
 	msg.Value = StringEncoder(string(v) + strconv.Itoa(b.i))
@@ -1246,7 +1246,7 @@ func (b *appendInterceptor) onSend(msg *ProducerMessage) {
 
 func (b *appendInterceptor) onConsume(msg *ConsumerMessage) {
 	if b.i < 0 {
-		panic("hey, the interceptor have failed")
+		panic("hey, the interceptor has failed")
 	}
 	msg.Value = []byte(string(msg.Value) + strconv.Itoa(b.i))
 	b.i++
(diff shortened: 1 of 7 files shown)


async_producer_test.go
```diff
@@ -1230,6 +1231,126 @@ func TestBrokerProducerShutdown(t *testing.T) {
...
+		panic("hey, the interceptor have failed")
+	}
+	v, _ := msg.Value.Encode()
+	msg.Value = StringEncoder(string(v) + strconv.Itoa(b.i))
+	b.i++
+}
+
+func (b *appendInterceptor) onConsume(msg *ConsumerMessage) {
+	if b.i < 0 {
+		panic("hey, the interceptor have failed")
```
psycotica0-shopify, line 1249: These panics should never be seen by any user, but either way `"the interceptor has failed"` feels more right to me 😄

d1egoaz: Thanks!

interceptors.go
```diff
@@ -0,0 +1,43 @@
...
+
+// ProducerInterceptor allows you to intercept (and possibly mutate) the records
+// received by the producer before they are published to the Kafka cluster.
+// https://cwiki.apache.org/confluence/display/KAFKA/KIP-42%3A+Add+Producer+and+Consumer+Interceptors#KIP42:AddProducerandConsumerInterceptors-Motivation
+type ProducerInterceptor interface {
+
+	// onSend is called when the producer message is intercepted. Please avoid
+	// modifying the message until it's safe to do so, as this is _not_ a copy
+	// of the message.
+	onSend(*ProducerMessage)
```
dnwe, line 11: These interfaces for the interceptors have only unexported methods. This means code outside the sarama package would be  unable to create types that satisfy the interface?

d1egoaz: 🦅 👁️
yes, intention is to be used outside Sarama, one more reason to not use the package `main` on the tests, as this wasn't highlighted by the test.

I'll export the methods!
Thanks!

dnwe:
@d1egoaz approved, changes LGTM.

It would be good to have something new in the `examples/` tree that makes use of this. Would you be able to look into that under a follow-on PR?

d1egoaz:
@dnwe for sure, I'll add an example soon

PR #1738, opened 2020-06-28
fixed variable names that are named same as some std lib package names

varun06:
This is part of cleaning up some of the Lint and Vet issues. This is first PR in list of many PR's.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ profile.out
 
 simplest-uncommitted-msg-0.1-jar-with-dependencies.jar
 
+.idea
+
+
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,3 @@ coverage.txt
 profile.out
 
 simplest-uncommitted-msg-0.1-jar-with-dependencies.jar
-
-.idea
-
-
--- a/async_producer_test.go
+++ b/async_producer_test.go
@@ -1248,7 +1248,7 @@ func ExampleAsyncProducer_select() {
 	signals := make(chan os.Signal, 1)
 	signal.Notify(signals, os.Interrupt)
 
-	var enqueued, producerError int
+	var enqueued, producerErrors int
 ProducerLoop:
 	for {
 		select {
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -28,3 +28,6 @@ profile.out
 
 simplest-uncommitted-msg-0.1-jar-with-dependencies.jar
 
+.idea
```
d1egoaz, line 31: This can be added in your global gitignore, this is very specific to your dev workflow

.gitignore
```diff
@@ -28,3 +28,6 @@ profile.out
...
+
+
```
d1egoaz, line 33: these empty lines are not needed

async_producer_test.go
```diff
@@ -1248,21 +1248,21 @@ func ExampleAsyncProducer_select() {
 	signals := make(chan os.Signal, 1)
 	signal.Notify(signals, os.Interrupt)
 
-	var enqueued, errors int
+	var enqueued, producerError int
```
d1egoaz, line 1251: I think the plural version might read better `producerErrors`

async_producer_test.go
```diff
@@ -1248,21 +1248,21 @@ func ExampleAsyncProducer_select() {
...
 			log.Println("Failed to produce message", err)
-			errors++
+			producerError++
 		case <-signals:
 			break ProducerLoop
 		}
 	}
 
-	log.Printf("Enqueued: %d; errors: %d\n", enqueued, errors)
+	log.Printf("Enqueued: %d; producerError: %d\n", enqueued, producerError)
```
d1egoaz, line 1265: It doesn't seem necessary to replace errors here, 
this reads better:
```go
log.Printf("Enqueued: %d; errors: %d\n", enqueued, producerErrors)
```

async_producer_test.go
```diff
@@ -1318,5 +1318,5 @@ ProducerLoop:
 
 	wg.Wait()
 
-	log.Printf("Successfully produced: %d; errors: %d\n", successes, errors)
+	log.Printf("Successfully produced: %d; producerError: %d\n", successes, producerError)
```
d1egoaz, line 1321: same comment as above

consumer.go
```diff
@@ -422,13 +422,13 @@ func (child *partitionConsumer) AsyncClose() {
 func (child *partitionConsumer) Close() error {
 	child.AsyncClose()
 
-	var errors ConsumerErrors
+	var consumerErrors ConsumerErrors
```
d1egoaz, line 425: yes, see how the plural version reads better here!

.gitignore
varun06: yeah, I PR'd from a machine that doesn't have all the settings. will fix it now.

async_producer_test.go
varun06: That's IDE refactoring, will fix it

PR #1744, opened 2020-07-07
sarama: Fix isBalanced Function Signature

alrs:
As-is, the first thing `isBalanced()` does is clobber its `sortedCurrentSubscriptions` argument. This changes its signature to ditch that argument, and continue its current behavior of deriving its own `sortedCurrentSubscriptions` via `sortMemberIDsByPartitionAssignments(currentAssignment)`.


alrs:
The failure against 2.5.0 reads as a flake. Can that test be re-run?

d1egoaz:
I've re-run the tests... 🤞

d1egoaz:
Thanks

PR #1753, opened 2020-07-17
add some prints

huija:
Describe the relationship between goroutine.ConsumeClaim and topic.partition more clearly

### DIFF
--- a/examples/consumergroup/main.go
+++ b/examples/consumergroup/main.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"runtime"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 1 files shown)


examples/consumergroup/main.go
```diff
@@ -155,10 +156,17 @@ func (consumer *Consumer) ConsumeClaim(session sarama.ConsumerGroupSession, clai
...
 	}
 
 	return nil
 }
+
+func goroutineID() string {
+	var buf [64]byte
+	n := runtime.Stack(buf[:], false)
+	return strings.Fields(string(buf[:n]))[1]
+}
```
d1egoaz, line 172: I haven't used this before? is this a recommended practice?

it seems go folks doesn't like this:
https://github.com/golang/go/issues/27361#issuecomment-422899200

huija: You are right. It's not a recommended practice to try to get the goroutine id.

This PR may be unnecessary because the comment above has stated that the ConsumeClaim() function would be called in a separate goroutine.

dnwe:
@huija thank you for your contribution! However, as per the [link](https://github.com/golang/go/issues/27361#issuecomment-422899200) from @d1egoaz, there is `go tool trace` for understanding the distribution of work across goroutines and I think it may be confusing for people to see this print statement in the sample code. 

If you believe the changes are still valid then please let us know

PR #1762, opened 2020-07-25
update gokrb5 to v8.4.1

kmhagan:
Hello!

This updates gokrb5 dependency to v8.4.1 which fixes an issue I ran into when a KDC server was unresponsive: https://github.com/jcmturner/gokrb5/pull/399


dnwe:
@bai I’m pretty sure I attempted this upgrade as part of #1626 and had to revert in #1630 because it broke some users. 

Essentially:

> Switching to go module based import paths for dependencies would require every consumer of the client to have also switched. We should wait for Sarama/v2 before doing that

bai:
Oh good point, I missed that one 👍

dnwe:
I’m happy for us to just start prepping a v2 though 😄

kmhagan:
Looking forward to v2! I'll go ahead and close this PR and in the meantime I'll use my branch to handle this scenario

PR #1763, opened 2020-07-26
sarama: remove deprecated tls options from test

alrs:
`tls.Config.BuildNameToCertificate()` is deprecated. `godoc` advises:
>// Deprecated: NameToCertificate only allows associating a single certificate with a given name. Leave that field nil to let the library select the first compatible chain from Certificates.

This removes both uses of it from `doListenerTLSTest()`


d1egoaz:
thanks @alrs

PR #1769, opened 2020-08-05
Add support for Kafka 2.6.0

bai:
Add support for Kafka 2.6.0 🥗

### DIFF
--- a/functional_test.go
+++ b/functional_test.go
@@ -119,7 +119,7 @@ func prepareDockerTestEnvironment(ctx context.Context, env *testEnvironment) err
 	// found here: https://docs.confluent.io/current/installation/versions-interoperability.html
 	var confluentPlatformVersion string
 	switch env.KafkaVersion {
-	case "2.5.0":
+	case "2.6.0":
 		confluentPlatformVersion = "5.5.0"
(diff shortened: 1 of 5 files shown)


functional_test.go
```diff
@@ -119,7 +119,7 @@ func prepareDockerTestEnvironment(ctx context.Context, env *testEnvironment) err
 	// found here: https://docs.confluent.io/current/installation/versions-interoperability.html
 	var confluentPlatformVersion string
 	switch env.KafkaVersion {
```
dnwe, line 121: Need to remember to bump `confluentPlatformVersion` to 5.6 once images from that are available I guess. Probably worth opening a backlog issue so we don’t forget. 

I did also wonder if we should move to [strimzi/kafka](https://hub.docker.com/r/strimzi/kafka) for our CI image

bai: I see that we use a mix of confluent-oss and [wurstmeister/kafka](https://hub.docker.com/r/wurstmeister/kafka/) internally on CI. /cc @d1egoaz

d1egoaz: either should work, 🤷 ideally, we should only use one flavour

PR #1775, opened 2020-08-11
Adds a Producer Interceptor example

d1egoaz:
add some examples for the usage of the new interceptors feature added in https://github.com/Shopify/sarama/pull/1730


examples/README.md
```diff
@@ -8,3 +8,6 @@ In these examples, we use `github.com/Shopify/sarama` as import path. We do this
 
 [http_server](./http_server) is a simple HTTP server uses both the sync producer to produce data as part of the request handling cycle, as well as the async producer to maintain an access log. It also uses the [mocks subpackage](https://godoc.org/github.com/Shopify/sarama/mocks) to test both.
 
+#### Interceptors
+
+Basic example to use a producer interceptor that produces oTel spans and add some headers for each intercepted message.
```
varun06, line 13: What does oTel mean?

varun06: I see, can we call it OpenTelemetry in documentation?

d1egoaz: sure thing!

d1egoaz:
@varun06 I don't _think_ we need to push binaries none of the examples are doing that 🤷 
binaries can also be easily crated as a go.mod is provided

oofpgDLD:
The situation I met is that I want to put the parent span in the context,  then the Kafka Producer interceptor to create the child span from this context. But It seems not supported by getting the context from the OnSend method.

The code looks like:
```
func Caller() {
        ...
        span := tracer.StartSpan(info.FullMethod)
	defer span.Finish()
	ctx = opentracing.ContextWithSpan(ctx, span)
        ....
        Publish(ctx)
        ....
}

func Publish(ctx context.Context) {
      var kafkaPub sarama.SyncProducer
       ...
       m := &sarama.ProducerMessage{
		Key:   sarama.StringEncoder(fromId),
		Topic: appTopic,
		Value: sarama.ByteEncoder(b),
	}
	_, _, _ = kafkaPub.SendMessage(ctx, m)
}

type ProducerInterceptor struct {
}

func (oi *ProducerInterceptor) OnSend(ctx context.Context, msg *sarama.ProducerMessage) {
	span, ctx := opentracing.StartSpanFromContext(ctx, method)
	defer span.Finish()
        ...
}
```

So please add some more information about it or is there any place I could find the right example to solve it?

PR #1781, opened 2020-08-18
Refresh brokers given list of seed brokers

justin-chen:
`RefreshBrokers` takes a list of addresses to be used as seed brokers.
Existing broker connections are closed and the updated list of seed brokers will be used for the next metadata fetch.

`RefreshBrokers` can be used in the case of failovers when the Kafka cluster's DNS would have to be re-resolved

### DIFF
--- a/client.go
+++ b/client.go
@@ -56,6 +56,11 @@ type Client interface {
 	// partition. Offline replicas are replicas which are offline
 	OfflineReplicas(topic string, partitionID int32) ([]int32, error)
 
+	// RefreshBrokers takes a list of addresses to be used as seed brokers.
+	// Existing broker connections are closed and the updated list of seed brokers
+	// will be used for the next metadata fetch.
+	RefreshBrokers(addrs []string) error
+
 	// RefreshMetadata takes a list of topics and queries the cluster to refresh the
 	// available metadata for those topics. If no topics are provided, it will refresh
 	// metadata for all topics.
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -56,6 +56,8 @@ type Client interface {
 	// partition. Offline replicas are replicas which are offline
 	OfflineReplicas(topic string, partitionID int32) ([]int32, error)
 
+	RefreshBrokersWithNewSeedBrokers(seedBrokers []*Broker) error
```
vvuibert, line 59: all the other functions have descriptions so maybe we should add one here too

vvuibert: maybe the function could be called `RefreshBrokers` and the rest could go in the description?

bai (approved):

client.go
```diff
@@ -429,6 +434,30 @@ func (client *client) Leader(topic string, partitionID int32) (*Broker, error) {
...
+		delete(client.brokers, broker.ID())
+	}
+
+	client.seedBrokers = nil
+	client.deadSeeds = nil
+
+	random := rand.New(rand.NewSource(time.Now().UnixNano()))
+	for _, index := range random.Perm(len(addrs)) {
+		client.seedBrokers = append(client.seedBrokers, NewBroker(addrs[index]))
+	}
```
d1egoaz, line 456: it looks these 4 lines are the same when the `Client` is initialized, it might worth moving these to a new method on the client.

client.go
```diff
@@ -56,6 +56,11 @@ type Client interface {
 	// partition. Offline replicas are replicas which are offline
 	OfflineReplicas(topic string, partitionID int32) ([]int32, error)
 
+	// RefreshBrokers takes a list of addresses to be used as seed brokers.
+	// Existing broker connections are closed and the updated list of seed brokers
+	// will be used for the next metadata fetch.
+	RefreshBrokers(addrs []string) error
+
```
d1egoaz, line 63: I'm not sure how Sarama folks are adding new features, but this looks like it's a breaking change if someone is implementing the interface 🤔 

not sure why someone would do that, but the interface is public 🤷 

we could add a trick that I saw in the golang code, to add a private method, so no apps could implement the interface

```
    // A private method to prevent users implementing the
    // interface and so future additions to it will not
    // violate Go 1 compatibility.
    private()
```

found the blog post: https://blog.golang.org/module-compatibility

client.go
justin-chen: https://github.com/Shopify/sarama/pull/1784

client.go
justin-chen: Good idea, i'll add a dummy `private()` method to the interface

justin-chen: https://github.com/Shopify/sarama/pull/1785

d1egoaz: 😞 this wasn't a good idea at the end :/ the initial intention of `private()` is if the user needs to use an interface but don't want other people to implement it, and this is not the case for this interface.

PR #1785, opened 2020-08-19
Add private method to Client interface to prevent implementation

justin-chen:
Related: https://github.com/Shopify/sarama/pull/1781#discussion_r473121038

https://github.com/Shopify/sarama/pull/1781 added an interface method for `Client`.
 
Adding interface methods will break backwards compatibility if there's anyone implementing it.This PR adds a `private()` method to prevent implementation,

### DIFF
--- a/client.go
+++ b/client.go
@@ -93,6 +93,9 @@ type Client interface {
 
 	// Closed returns true if the client has already had Close called on it
 	Closed() bool
+
+	// A private method to prevent users implementing the interface for compatibility
+	private()
 }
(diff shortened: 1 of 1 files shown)


dnwe (approved):

nayeemzen:
Breaking mocks for us too `google-cloud-go` has a [package](https://github.com/googleapis/google-cloud-go-testing) that provides an adapter and interfaces to workaround the problem (feels non-ideal imo).

PR #1788, opened 2020-08-23
feat[balance_strategy]: announcing a new round robin balance strategy

kzinglzy:
There is a different behavior between 
- [sarama's roundrobin balance strategy](https://github.com/Shopify/sarama/blob/master/balance_strategy.go#L80)
	```go
    // pseudo code
	for topic, partitions := range allTopicPartitions {
		i := 0
		for partition := range allPartitions {
			consumer := allconsumers[i % len(allconsumers)]
			consumer.assign(topic, partition)
			i += 1
		}
	}
	```
- [kafka java client's roundrobin balance strategy](https://github.com/apache/kafka/blob/6649b7a91dafcff4f2ed5b5ec1c0bb02b755e7f7/clients/src/main/java/org/apache/kafka/clients/consumer/RoundRobinAssignor.java):
	```go
    // pseudo code
	i := 0
	for partition := range AllSortedPartitions {
		consumer := allconsumers[i % len(allconsumers)]
		consumer.assign(topic, partition)
		i += 1
	}

    ```

In some case, sarama's implementation may incur skewed partitions,
for example:
```go
// balance_strategy_test.go/TestBalanceStrategyRoundRobin:tests
{
    members: map[string][]string{"M": {"T1", "T2", "TT2"}, "M2": {"T1", "T2", "TT2"}, "M3": {"T1", "T2", "TT2"}},
    topics:  map[string][]int32{"T1": {0}, "T2": {0}, "TT2": {0}},
    expected: BalanceStrategyPlan{
        // there are three members and topics, however all the topics and partitions are assign to a single member
	"M": map[string][]int32{"T1": {0}, "T2": {0}, "TT2": {0}},
},
```

This PR provides the "really" roundrobin implementation that same as kafka java client did

### DIFF
--- a/balance_strategy.go
+++ b/balance_strategy.go
@@ -16,9 +16,9 @@ const (
 	// RoundRobinBalanceStrategyName identifies strategies that use the round-robin partition assignment strategy
 	RoundRobinBalanceStrategyName = "roundrobin"
 
-	// EvenRoundRobinBalanceStrategyName identifies strategies that use the round-robin partition assignment strategy base on all topic's partitions,
+	// FairRoundRobinBalanceStrategyName identifies strategies that use the round-robin partition assignment strategy base on all topic's partitions,
 	// which assigns partitions more evenly in comparison to the`RoundRobinBalanceStrategyName` strategy
-	EvenRoundRobinBalanceStrategyName = "even_roundrobin"
+	FairRoundRobinBalanceStrategyName = "fair_roundrobin"
 
 	// StickyBalanceStrategyName identifies strategies that use the sticky-partition assignment strategy
 	StickyBalanceStrategyName = "sticky"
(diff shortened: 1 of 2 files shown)


dnwe (approved):
@kzinglzy I'm happy to approve the changes, although I do wonder if the original implementation of the round robin strategy should just be considered flawed and we should replace it with your correct implementation rather than adding this as a new alternative strategy? Whilst I know it would technically be a potentially unexpected change of behaviour for people, it still seems like the right thing to do.

@bai thoughts? I wonder if we should rename the old strategy to something with "legacy" in the name and then make @kzinglzy's implementation the "RoundRobinBalancer" and just mention the change of behaviour in the next release notes?

kzinglzy:
> @kzinglzy I'm happy to approve the changes, although I do wonder if the original implementation of the round robin strategy should just be considered flawed and we should replace it with your correct implementation rather than adding this as a new alternative strategy? Whilst I know it would technically be a potentially unexpected change of behaviour for people, it still seems like the right thing to do.
> 
> @bai thoughts? I wonder if we should rename the old strategy to something with "legacy" in the name and then make @kzinglzy's implementation the "RoundRobinBalancer" and just mention the change of behaviour in the next release notes?

as mentioned previously, the old implementation of round robin strategy is not so "standard" , in some cases even wrong (e.g. skewed partitions), thus, it may be ok to replace it with a more "standard" implementation.
> so this PR should be considered as a bug fixing :P

dnwe:
Yeah that was my thinking. To my reading our original implementation was just incorrect, I can't think of a good reason why anyone would prefer it and I think we should accept your PR as a replacement that corrects the behaviour to match the original intention.

bai (commented):
I think it'd be OK to push this new strategy as a replacement (essentially a fix) and rename current one to Legacy.

kzinglzy:
@bai  the original implementation has been replaced

> Um... CI failed, looks like a test env issue?

bai:
Looks like a change in GitHub Actions broke CI. I've pushed a fix in https://github.com/Shopify/sarama/pull/1862. Could you please rebase off of master?

kzinglzy:
@bai please take a look, ready for merge

bai (approved):

PR #1792, opened 2020-08-30
bugfix update metadata panic when client is closed

Stephan14:
fix update metadata panic when client is closed, because broker is nil

### DIFF
--- a/client.go
+++ b/client.go
@@ -929,13 +929,12 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int,
 
 // if no fatal error, returns a list of topics that need retrying due to ErrLeaderNotAvailable
 func (client *client) updateMetadata(data *MetadataResponse, allKnownMetaData bool) (retry bool, err error) {
-	if client.Closed() {
-		return
-	}
-
 	client.lock.Lock()
 	defer client.lock.Unlock()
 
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -929,13 +929,12 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int,
 
 // if no fatal error, returns a list of topics that need retrying due to ErrLeaderNotAvailable
 func (client *client) updateMetadata(data *MetadataResponse, allKnownMetaData bool) (retry bool, err error) {
```
varun06, line 931: Do we not want to check for if a client is closed here?

Stephan14: yes, you are right

client.go
```diff
@@ -929,13 +929,12 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int,
...
-	if client.Closed() {
-		return
-	}
-
 	client.lock.Lock()
 	defer client.lock.Unlock()
 
+	if client.brokers == nil {
+		return
+	}
```
dnwe, line 937: I don't understand how the proposed change would prevent a panic? Can you provide backtrace output from when you've hit a problem here? Alternatively can you demonstrate via a unittest what state causes a panic and how the change prevents that?

PR #1794, opened 2020-09-02
INK2-764-library-changes

arora01gunjan:


fetch_response.go
```diff
@@ -3,6 +3,8 @@ package sarama
 import (
 	"sort"
 	"time"
+
+	"bitbucket.org/ffxblue/bp-gateway/pkg/monolog"
```
dackroyd, line 7: Internal project packaging, unrelated to Sarama, Need to remove all internal code references/use

fetch_response.go
```diff
@@ -450,7 +452,15 @@ func (r *FetchResponse) AddRecordWithTimestamp(topic string, partition int32, ke
 		frb.RecordsSet = []*Records{&records}
 	}
 	batch := frb.RecordsSet[0].RecordBatch
-	rec := &Record{Key: kb, Value: vb, OffsetDelta: offset, TimestampDelta: timestamp.Sub(batch.FirstTimestamp)}
```
dackroyd, line 453: Fundamentally the issue here is that it needs to be possible to include headers in the record, not these specific ones?

dackroyd (changes requested):

dackroyd:
It's likely that we can keep these changes within the fork, without a PR back to the main repo for now

dnwe:
From the comments above it sounds like this was an accidentally raised PR containing some code to set a number of default record headers based on a private company project?

Will go ahead and close this PR for now but feel free to come back if you think I'm mistaken :)

PR #1817, opened 2020-10-21
Revert "Add private method to Client interface to prevent implementation"

d1egoaz:
Reverts Shopify/sarama#1785

this is causing more damage that the initial intention to prevent breaking changes.

Some people are using mock libraries and they are getting issues with this change

i.e, when using mockery
```
Cannot use 'mockSaramaClient' (type *mocks.Client) as type sarama.Client Type cannot implement 'sarama.Client' as it has a non-exported method and is defined in a different package
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -93,9 +93,6 @@ type Client interface {
 
 	// Closed returns true if the client has already had Close called on it
 	Closed() bool
-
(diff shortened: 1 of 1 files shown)


nayeemzen (approved):

nayeemzen:
🚢

PR #1819, opened 2020-10-23
update consumer group minimum version

faceair:
Same as this issue https://github.com/Shopify/sarama/issues/1680
I noticed that the consumer group is limited to V0_10_2_0, but the internal implementation is actually [compatible with V0_10_0_0](https://github.com/Shopify/sarama/blob/v1.27.2/consumer_group.go#L309-L312). And I tested that it works fine on V0_10_0_0.

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -96,8 +96,8 @@ func NewConsumerGroupFromClient(groupID string, client Client) (ConsumerGroup, e
 
 func newConsumerGroup(groupID string, client Client) (ConsumerGroup, error) {
 	config := client.Config()
-	if !config.Version.IsAtLeast(V0_10_2_0) {
-		return nil, ConfigurationError("consumer groups require Version to be >= V0_10_2_0")
+	if !config.Version.IsAtLeast(V0_10_0_0) {
+		return nil, ConfigurationError("consumer groups require Version to be >= V0_10_0_0")
 	}
 
 	consumer, err := NewConsumerFromClient(client)


matzew:
I see on https://github.com/Shopify/sarama/issues/1363#issuecomment-486943952

that this was fixed for 0.10.2.0: https://issues.apache.org/jira/browse/KAFKA-3853

Better safe, than sorry?

faceair:
This seems to be a bug in the kafka server. so sarama is restricting the use of versions higher than 0.10.2 to ensure data reliability for users?
Security is of course important, but every version of the kafka server can have bugs. We chose a lower version of the kafka server, and we are aware of the risks.

I used sarama-cluster before sarama, and I have been using kafka server version 0.10.0 for four or five years now, and there are some reasons why I have no way to upgrade kafka server.
I just want to switch from sarama-cluster to sarama.

faceair:
I've been using sarama for over a month now to bypass version restrictions.
@bai  How do I verify this PR feature? I really want this PR to merge into the master branch.

faceair:
@bai Any updates ? Whether you decide to merge or not to merge ?

varun06 (approved):
I don't think it will break anything. Should be safe to merge IMO

bai:
@dnwe @d1egoaz any objections on merging this?

d1egoaz (approved):
🤷 I have mixed feelings about this.

The author for the consumer groups specifically restricted the version to `V0_10_2_0`which was released in February 21, 2017.

And this adds support to use `V0_10_0_0` that was released in  May 22, 2016.

Both versions are **outdated and unsupported.**

It doesn't hurt merging this, but it also don't see the value of merging this for some users that are running very outdated and unsupported kafka versions.

dnwe:
I'm afraid I agree with @d1egoaz, even for 0.10.2.0 whilst the support code is _technically_ still around in the repo, we don't actively do any testing of it and there's no guarantee that we won't break it in the future

The Sarama [support statement](https://github.com/Shopify/sarama/#compatibility-and-api-stability) is:
> Sarama provides a "2 releases + 2 months" compatibility guarantee: we support the two latest stable releases of Kafka and Go, and we provide a two month grace period for older releases. This means we currently officially support Go 1.14 through 1.16, and Kafka 2.6 through 2.7, although older releases are still likely to work.

Additionally I would personally even like us to have the freedom to purge handling of legacy cluster and protocol versions going forward so we can reduce the size of the codebase

bai:
Thanks for contributing! Unfortunately I'm going to have to close this off as stale since we no longer support legacy Kafka.

PR #1832, opened 2020-11-16
Update Godoc link to pkg.go.dev

ilyakaznacheev:
So the godoc.org is outdated now, so maybe it would be more convenient to give a link to a new platform instead.


bai (approved):

bai:
Thanks for contributing. Could you please sign CLA?

ilyakaznacheev:
Just done so, but dunno how to restart the CI without a new commit

PR #1842, opened 2020-11-25
Added fuzzer with oss-fuzz build script

AdamKorcz:
This PR adds a fuzzer for `decodeRequest` along with the build script in case there is interest in setting up continuous fuzzing for Sarama by way of oss-fuzz.

I have the build files available for the oss-fuzz side and I will be happy to integrate Sarama.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,4 @@
+#/bin/bash -eu
+
+compile_go_fuzzer github.com/Shopify/sarama Fuzz fuzz
+ 
--- a/fuzz.go
+++ b/fuzz.go
@@ -1,0 +1,13 @@
+// +build gofuzz
+
+package sarama
+
+import "bytes"
+
+func Fuzz(data []byte) int {
+	_, _, err := decodeRequest(bytes.NewReader(data))
+	if err != nil {
+		return 0
+	}
+	return 1
+}


AdamKorcz:
Does anyone have a solution for the failed CI tests? It sure looks like an issue with the tests and not with the files in the commit

dnwe:
@AdamKorcz I re-based your branch on master (as the test flakiness has been much improved recently. Looks like just one failure on 1.15x + 2.6.1 for the new test you added `--- FAIL: TestFuncConsumerGroupFuzzy (52.18s)`

Please could you take a look?

AdamKorcz:
Thank you for the rebase. The last failing test also looks unrelated to the code committed in this PR.

As a sidenote: The code in this PR should not affect any code in Sarama since the fuzzer is not called by Sarama itself, and a build comment has been added as well.

bai:
This is a flaky test on CI I'm afraid, I've re-run actions and it failed on another version/step.

bai:
👋🏼 Is this still something we want? /cc @dnwe @d1egoaz @vvuibert

dnwe:
@bai I think fuzzing in general would be a useful addition. However, if we want to look at doing some fuzzing in Sarama then I think that as the [official fuzzing proposal](/golang/go/issues/44551) has been accepted (since this PR was originally proposed) we would ideally look at adding an additional build in the GitHub actions to use tip and checkout the `dev.fuzz` branch (as per https://go.dev/blog/fuzz-beta) and use `testing.F` and the `// +build gofuzzbeta` build flags.

@AdamKorcz is that something you'd like to explore under this PR, or would you prefer we write that up as a separate backlog issue?

dnwe:
Looks like the dev.fuzz branch just merged to main and should be available in regular Golang tip now

image

Ref: https://twitter.com/katie_hockman/status/1440082486692773897?s=21

bai:
I'd wait for that then.

Issue #1610, opened 2020-02-12
How to set offset to a specified value in Initialization 


blackfox1983:
value of c.Consumer.Offsets.Initial must be OffsetNewest or OffsetOldest.

But in some cases, I need to comsume message from a specified offset.

i read the code but have not found method to solution my problem.

somebody who knows help me..thanks..

sorry for my poor english :(

dnwe:
@blackfox1983 I think you are looking for this?https://github.com/Shopify/sarama/blob/6fb86a44d5915dcfd42af71a4c99674bb55650ca/consumer.go#L60-L64

d1egoaz:
@blackfox1983 your English is great, don't worry about that.

Regarding the question, one importan thing from @dnwe response, is that this is only supported by the low level consumer and it's not supported by `ConsumerGroups`

Jianli-lin:
Also hope we can support set offset for `ConsumerGroup`

Or we can commit offset manually. (https://github.com/Shopify/sarama/issues/1570)

PR #1860, opened 2020-12-18
Specify initial partition offsets for consumer group

xxxcoltxxx:
This pr adds ability to manage offsets from the application for `ConsumerGroup`.
If the consumer implemented interface `ConsumerManagesOffsets`, application can use specific offsets for any partitions as we want.
Its also resolved issue #1610

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -659,8 +659,13 @@ func (s *consumerGroupSession) consume(topic string, partition int32) {
 
 	// get next offset
 	offset := s.parent.config.Consumer.Offsets.Initial
-	if pom := s.offsets.findPOM(topic, partition); pom != nil {
-		offset, _ = pom.NextOffset()
+
+	if c, ok := s.handler.(ConsumerManagesOffsets); ok {
+		offset = c.GetOffset(topic, partition)
+	} else {
+		if pom := s.offsets.findPOM(topic, partition); pom != nil {
+			offset, _ = pom.NextOffset()
+		}
 	}
 
 	// create new claim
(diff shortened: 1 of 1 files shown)


xxxcoltxxx:
@bai Check the pull request pls. I cannot use the sarama package without this feature

dnwe:
@xxxcoltxxx I don't think we can accept the PR as-is without some further thought and discsussion as any interfaces that we add like this need to be supported forever more

Unless I'm mistaken, it sounds like you're trying to achieve something similar to the Java client's "[Storing Offsets Outside Kafka](https://kafka.apache.org/26/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#rebalancecallback)", so we'd either want to look at supporting that capability more thoroughly or see if you can work within the existing interfaces.

I presume your desired offsets still advance linearly on the partitions? Can you not use the `Setup(session sarama.ConsumerGroupSession) error` function to loop over your desired first offsets for each partition?

e.g.,
```go
wg := &sync.WaitGroup{}
for topic, partitions := range session.Claims() {
		topic := topic
		partition := partition
		wg.Add(1)
		go func() {
			defer wg.Done()

			offset, err := YourCustomGetOffsetHere(topic, partition)
			if err != nil {
				log.Printf("GetOffset for %s-%d got err %v\n", topic, partition, err)
				// TODO: handle err appropriately
				continue
			}
			log.Printf("Marking offset %d on %s-%d\n", offset, topic, partition)
			session.MarkOffset(topic, partition, offset, "")
		}()
	}
}
wg.Wait()
```

xxxcoltxxx:
@dnwe Yes, offsets advance lineary on the partitions, but i cannot use custom offset with high-level consumer.
If i will use my own consuming function, i should handle rebalancing manually, and create the high-level consumer eventually, but with initial offsets from outside storage.

My use-case: we consume to some source (kafka, mysql, etc.) to get batch of messages. Each message handle with application business logic. They transforms or can be ignored. When we collect, for example 200 Mb, we are sending they to some destination (clickhouse, rabbitmq). After sending we are saving offsets to our storage to continue import from previous offset. We cannot commit to kafka for each message until we are save all handled messages to the destination.

PR #1863, opened 2020-12-24
Add support for Kafka 2.7.0 + update lz4 and klauspost/compress dependencies

bai:
This PR adds support for Kafka 2.7.0.


dnwe (approved):
LGTM, although worth calling out that klauspost/compress and pierrec/lz4 were both bumped up minor versions in this PR

PR #1898, opened 2021-03-10
Parsing only known control batches value

mmaslankaprv:
Kafka documentation defines control batches as a single record batches
containing data that shouldn't be passed to applications. The record
contained inside the control batch consist of a key with well defined
format and a value which format is defined by the control batch type.

Control batch key format:
```
ControlRecordKey => version type
  version => INT16
  type => INT16
```

Fixed control batch parsing logic by first parsing the key to check if
control batch is of known type and only then parsing corresponding
record value.

Signed-off-by: Michal Maslanka <michal@vectorized.io>

### DIFF
--- a/control_record.go
+++ b/control_record.go
@@ -23,16 +23,6 @@ type ControlRecord struct {
 
 func (cr *ControlRecord) decode(key, value packetDecoder) error {
 	var err error
-	cr.Version, err = value.getInt16()
-	if err != nil {
-		return err
-	}
-
-	cr.CoordinatorEpoch, err = value.getInt32()
-	if err != nil {
-		return err
-	}
-
 	// There a version for the value part AND the key part. And I have no idea if they are supposed to match or not
 	// Either way, all these version can only be 0 for now
 	cr.Version, err = key.getInt16()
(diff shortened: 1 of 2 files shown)


bai:
Thanks for your contribution! Could you please sign CLA to make this PR green? 🙏🏼

mmaslankaprv:
> Thanks for your contribution! Could you please sign CLA to make this PR green? 🙏🏼

I did sign the CLA, when I try to do it again it says "user with current Github name already signed the CLA". Is there anything I should do to trigger the CLA check again ?

bai:
> I did sign the CLA, when I try to do it again it says "user with current Github name already signed the CLA". Is there anything I should do to trigger the CLA check again ?

Oops, hiccup on our side, my apologies. I'll let Diego review this PR 🙏🏼

control_record.go
```diff
@@ -55,6 +45,18 @@ func (cr *ControlRecord) decode(key, value packetDecoder) error {
 		// UNKNOWN is used to indicate a control type which the client is not aware of and should be ignored
 		cr.Type = ControlRecordUnknown
 	}
+	// we want to parse value only if we are decoding controll record of known type
```
d1egoaz, line 48: s/controll/control

control_record.go
```diff
@@ -55,6 +45,18 @@ func (cr *ControlRecord) decode(key, value packetDecoder) error {
...
+	if cr.Type == ControlRecordAbort || cr.Type == ControlRecordCommit {
```
d1egoaz, line 49: what about:
```go
if cr.Type != ControlRecordUnknown
```

d1egoaz (approved):
@mmaslankaprv  thanks for the contribution.

It looks like we're missing a/some tests in order to check for this issue.

Would be possible to add some test? I'd expect a test to fail before your change, after your change the test should pass and should be part of our continuous test

mmaslankaprv:
> @mmaslankaprv thanks for the contribution.
> 
> It looks like we're missing a/some tests in order to check for this issue.
> 
> Would be possible to add some test? I'd expect a test to fail before your change, after your change the test should pass and should be part of our continuous test

I've added `ControlRecord` decoding tests.

d1egoaz:
Thanks @mmaslankaprv

PR #1925, opened 2021-04-28
Test out CodeQL

bai:
Ref: https://github.com/Shopify/sarama/issues/1924


dnwe:
So here are the initial findings that we would want to fix under this PR before merging and having master scanned:

https://github.com/Shopify/sarama/security/code-scanning?query=ref%3Arefs%2Fpull%2F1925%2Fmerge+tool%3ACodeQL

PR #1926, opened 2021-04-28
fix: correct initial CodeQL findings

dnwe:
- Incorrect conversion between integer types
- Size computation for allocation may overflow


bai (approved):
You're fast, thank you! :shipit:

PR #1929, opened 2021-04-29
Handle isolation level in Offset(Request|Response) and require stable offset in FetchOffset(Request|Response)

celrenheit:
This PR enables OffsetRequest and to require read committed isolation.
It adds RequireStable field to FetchOffsetRequest, this requires the use of the new compact protocol present in version 6+.

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -376,6 +376,7 @@ func (b *Broker) CommitOffset(request *OffsetCommitRequest) (*OffsetCommitRespon
 //FetchOffset returns an offset fetch response or error
 func (b *Broker) FetchOffset(request *OffsetFetchRequest) (*OffsetFetchResponse, error) {
 	response := new(OffsetFetchResponse)
+	response.Version = request.Version // needed to handle the two header versions
 
 	err := b.sendAndReceive(request, response)
 	if err != nil {
--- a/errors.go
+++ b/errors.go
@@ -208,6 +208,12 @@ const (
 	ErrPreferredLeaderNotAvailable        KError = 80
 	ErrGroupMaxSizeReached                KError = 81
 	ErrFencedInstancedId                  KError = 82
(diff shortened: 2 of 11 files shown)


broker.go
```diff
@@ -376,6 +376,7 @@ func (b *Broker) CommitOffset(request *OffsetCommitRequest) (*OffsetCommitRespon
 //FetchOffset returns an offset fetch response or error
 func (b *Broker) FetchOffset(request *OffsetFetchRequest) (*OffsetFetchResponse, error) {
 	response := new(OffsetFetchResponse)
+	response.Version = request.Version // needed to handle the two header versions
```
celrenheit, line 379: not sure about how to handle this, this is needed in order to give sarama the correct header version in the [Response](https://github.com/Shopify/sarama/pull/1929/files#diff-4a14f4d736e178d87744550232d8206b39c185eac6f021f88078383dc344c281R230-R232)

celrenheit: this was set in order to be able to decode the header which is decode before calling  `func (r *OffsetFetchResponse) decode(pd packetDecoder, version int16)` .
It seems that it needs the header version 2 for v6+ of `OffsetFetchResponse`, not sure why, I wasn't able to find a doc about this.

dnwe: Yeah the headerVersion relates to the flexibleVersions support and the rules are a little murky, but if you look at the Java protocol generator source you can find them:
https://github.com/apache/kafka/blob/9dbf2226cdae4a30e44340a552613446c5053d87/generator/src/main/java/org/apache/kafka/message/ApiMessageTypeGenerator.java#L364-L379

I think yours is the first Response type that has needed support for both flexible and non-flexible headerVersions. I think your idea here is fine

dnwe (approved):
Approved. Thanks again!

Note: the status checks may be a little flaky in the functional test atm and may need a re-run or two before we can merge

PR #1939, opened 2021-05-07
fix(test): use Sarama transactional producer

### LINKED ISSUE #1733: Revisit publishing of uncommited messages in functional_test
One of the pieces of #1695 was to use a small jar for publishing uncommited messages as part of the functional test:

https://github.com/Shopify/sarama/blob/221ed1a25150fa98d750f2e2f33f58ff018238b4/functional_test.go#L343-L345

What's missing from https://github.com/Shopify/sarama/pull/1152 to allow us to do this in-house?

/cc @KJTsanaktsidis

dnwe:
One of the pieces of #1695 was to use a small jar for publishing
uncommited messages as part of the functional test. Replacing that with
the native Sarama transactional producer

Fixes #1733

Co-authored-by: @KJTsanaktsidis

### DIFF
--- a/functional_consumer_test.go
+++ b/functional_consumer_test.go
@@ -8,6 +8,7 @@ import (
 	"math"
 	"os"
 	"sort"
+	"strconv"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


bai:
👋🏼 Is this still something we want?

bai:
Cool, thanks thank makes sense. I've been going through some PRs here to clean things up a bit and was curious about this one.

KJTsanaktsidis:
I don't really use Sarama much these days anymore (I've moved to a Ruby team) but even so I'm still delighted to see the back of that jar 😂

PR #1949, opened 2021-05-20
fix: explicitly use uint64 for payload length

### LINKED ISSUE #1943: Can't build in 32-bit
##### Versions

| Sarama | Kafka | Go |
|--------|-------|----|
| v1.29.0 | * | v1.16.4 |

##### Problem Description

Sarama fails to build in 32-bit mode.

```
github.com/Shopify/sarama/gssapi_kerberos.go:62:10: constant 4294967295 overflows int
```

This is the code from gassapi_kerberos.go:
```go
func (krbAuth *GSSAPIKerberosAuth) writePackage(broker *Broker, payload []byte) (int, error) {
	length := len(payload)
	size := length + 4 // 4 byte length header + payload
	if size > math.MaxUint32 { // <<< math.MaxUint32 overflows `int` in 32-bit
		return 0, errors.New("payload too large, will overflow uint32")
	}
```

You could change the line setting length to be: `length := uint64(len(payload))`

dnwe:
Tested with:
```
docker run --platform=linux/i386 --rm -v "$PWD":/usr/src -w /usr/src golang:1.16 go build -v
```

Fixes #1943

### DIFF
--- a/gssapi_kerberos.go
+++ b/gssapi_kerberos.go
@@ -57,7 +57,7 @@ type KerberosClient interface {
 
 // writePackage appends length in big endian before the payload, and sends it to kafka
 func (krbAuth *GSSAPIKerberosAuth) writePackage(broker *Broker, payload []byte) (int, error) {
-	length := len(payload)
+	length := uint64(len(payload))
 	size := length + 4 // 4 byte length header + payload
 	if size > math.MaxUint32 {
 		return 0, errors.New("payload too large, will overflow uint32")


gssapi_kerberos.go
```diff
@@ -57,7 +57,7 @@ type KerberosClient interface {
 
 // writePackage appends length in big endian before the payload, and sends it to kafka
 func (krbAuth *GSSAPIKerberosAuth) writePackage(broker *Broker, payload []byte) (int, error) {
-	length := len(payload)
+	length := uint64(len(payload))
 	size := length + 4 // 4 byte length header + payload
 	if size > math.MaxUint32 {
```
aaronbee, line 62: You might want to change this to `size > math.MaxInt32` rather than `MaxUint32`. Line 65 would panic on a 32-bit system if size is > `math.MaxInt32`. But, you are kinda asking for it if you are trying to write > 2GB of data all at once.

PR #1950, opened 2021-05-21
Make size a uint to build on GOARCH=386

evantorrie:
This is failing to build with `GOARCH=386` on a linux box.

Explicitly forcing size to a `uint` fixes this problem.


evantorrie:
superseded by #1949

PR #1962, opened 2021-06-07
fix(consumer):  call interceptors when MaxProcessingTime expire

hanxiaolin:
Missing call for interception when MaxProcessingTime timed out and we process the remaining messages


dnwe:
@d1egoaz please could you take a look and confirm you're happy with the interceptor changes?

d1egoaz (approved):

PR #1964, opened 2021-06-10
Add DelegationToken ResourceType

ajanikow:
According to current ResourceTypes in https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/resource/ResourceType.java#L66 Sarama is missing DelegationToken definition.

### DIFF
--- a/acl_types.go
+++ b/acl_types.go
@@ -141,6 +141,7 @@ const (
 	AclResourceGroup
 	AclResourceCluster
 	AclResourceTransactionalID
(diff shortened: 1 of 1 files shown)


ajanikow:
Hello!

CLA Signed, but do not see any option to rerun build.

Best Regards,
Adam.

dnwe (approved):
@ajanikow many thanks for your first contribution — this is a good spot, approved and merged

Issue #1969, opened 2021-06-20
kafka-producer-performance with throughput fail


KerryJava:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka | Go |
|--------|-------|----|
| v1.29.0 |  0.10.2.0 |  go1.15.7 linux/amd64 |

##### Configuration

What configuration values are you using for Sarama and Kafka?
``` go
```
./kafka-producer-performance  -brokers 172.18.8.143:9092,172.18.8.142:9092,172.18.8.141:9092 \
  -topic test \
  -throughput 500 \
  -version  "0.10.2.0" \
  -message-load 10000 \
  -message-size 5000 \
  -required-acks -1
##### Logs

  panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x740d46]

goroutine 10 [running]:
github.com/Shopify/sarama.(*produceSet).wouldOverflow(0xc00024c1b0, 0xc0000c0a00, 0x7fff00000008)
        /root/go/pkg/mod/github.com/!shopify/sarama@v1.29.0/produce_set.go:242 +0x206
github.com/Shopify/sarama.(*brokerProducer).run(0xc00381e0e0)
        /root/go/pkg/mod/github.com/!shopify/sarama@v1.29.0/async_producer.go:782 +0x4c5
github.com/Shopify/sarama.withRecover(0xc0000a0a00)
        /root/go/pkg/mod/github.com/!shopify/sarama@v1.29.0/utils.go:43 +0x49
created by github.com/Shopify/sarama.(*asyncProducer).newBrokerProducer
        /root/go/pkg/mod/github.com/!shopify/sarama@v1.29.0/async_producer.go:690 +0x20c
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x740d46]

goroutine 6 [running]:
github.com/Shopify/sarama.(*produceSet).wouldOverflow(0xc0000a3260, 0xc0000c0a00, 0x7fff0000000a)
        /root/go/pkg/mod/github.com/!shopify/sarama@v1.29.0/produce_set.go:242 +0x206
github.com/Shopify/sarama.(*brokerProducer).run(0xc00381e000)
[...]

PR #1971, opened 2021-06-20
fix  kafka-producer-performance throughput panic

KerryJava:
#1969  fix  kafka-producer-performance tool panic


dnwe (approved):
Thanks — the fix looks good to me. It's not obvious what the original code intended — it seemed to attempt to duplicate the message (from the original generated set) _throughput_ times before sleeping which is unusual. I think the new version which just keeps adding unique messages from the set and pausing whenever throughput in total have been put is more reasonable

PR #1990, opened 2021-07-27
fix: correctly pass ValidateOnly through to CreatePartitionsRequest

### LINKED ISSUE #1955: ValidateOnly when set true in CreatePartitions() is creating topic
##### Versions
Sarama Version : 1.29.0
Kafka Version : 2.7.0
go Version : 1.16

##### Config for creating ClusterAdmin:
config.version = sarama.V2_7_0_0

##### Problem Description
When argument validateOnly is set to true in the CreatePartitions function here (https://github.com/Shopify/sarama/blob/master/admin.go#L43)

This is supposed to just validate the possibility of creating partitions right? But, it is creating partitions of the topic. 
(Correct me if my understanding of validateOnly is incorrect)

doxsch:
fix #1955

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -443,6 +443,7 @@ func (ca *clusterAdmin) CreatePartitions(topic string, count int32, assignment [
 	request := &CreatePartitionsRequest{
 		TopicPartitions: topicPartitions,
 		Timeout:         ca.conf.Admin.Timeout,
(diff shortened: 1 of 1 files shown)


dnwe:
@doxsch thanks for the PR, this is a good fix

Apologies for the delay, it’s Summer vacation season 🌞

PR #2001, opened 2021-08-30
docs: inform AsyncProducer Close pitfalls

HurSungYun:
Not closing AsyncProducer before process shutting down seems a common mistake when using it.

I believe it is worth to inform users that messages can be lost if they omit to call `Close()` when using AsyncProducer.


HurSungYun:
I've signed CLA :)

bai:
Thanks for your contribution! 🙏🏼

PR #2013, opened 2021-09-09
fix(test): disable encoded response/request check when map contains multiple elements

### LINKED ISSUE #2012: test: TestDeleteOffsetsResponse is brittle and fails periodically
The new `TestDeleteOffsetsResponse` looks like it can fail periodically due to a timing issue:

```
% go test -run 'TestDeleteOffsetsResponse' -v -count=10
=== RUN   TestDeleteOffsetsResponse
--- PASS: TestDeleteOffsetsResponse (0.00s)
=== RUN   TestDeleteOffsetsResponse
--- PASS: TestDeleteOffsetsResponse (0.00s)
=== RUN   TestDeleteOffsetsResponse
--- PASS: TestDeleteOffsetsResponse (0.00s)
=== RUN   TestDeleteOffsetsResponse
--- PASS: TestDeleteOffsetsResponse (0.00s)
=== RUN   TestDeleteOffsetsResponse
--- PASS: TestDeleteOffsetsResponse (0.00s)
=== RUN   TestDeleteOffsetsResponse
--- PASS: TestDeleteOffsetsResponse (0.00s)
=== RUN   TestDeleteOffsetsResponse
    request_test.go:96: Encoding no error failed
        got  [0 0 0 0 0 0 0 0 0 1 0 3 98 97 114 0 0 0 2 0 0 0 7 0 0 0 0 0 6 0 0]
        want [0 0 0 0 0 0 0 0 0 1 0 3 98 97 114 0 0 0 2 0 0 0 6 0 0 0 0 0 7 0 0]
--- FAIL: TestDeleteOffsetsResponse (0.00s)
=== RUN   TestDeleteOffsetsResponse
--- PASS: TestDeleteOffsetsResponse (0.00s)
=== RUN   TestDeleteOffsetsResponse
--- PASS: TestDeleteOffsetsResponse (0.00s)
=== RUN   TestDeleteOffsetsResponse
--- PASS: TestDeleteOffsetsResponse (0.00s)
FAIL
exit status 1
FAIL    github.com/Shopify/sarama       0.361s
```

Needs investigating if we can make it more robust.

cc @faillefer

faillefer:
Fix #2012 
Disable encoded response/request check when map contains multiple elements. In that situation the encoded/request response varies because of unpredictable map traversal order

### DIFF
--- a/delete_offsets_request_test.go
+++ b/delete_offsets_request_test.go
@@ -7,27 +7,6 @@ var (
 		0, 3, 'f', 'o', 'o', // group name: foo
 		0, 0, 0, 0, // 0 partition
 	}
-
-	doubleDeleteOffsetsRequest = []byte{
-		0, 3, 'f', 'o', 'o', // group name: foo
-		0, 0, 0, 1, // 1 topic
-		0, 3, 'b', 'a', 'r', // topic name: bar
-		0, 0, 0, 2, // 2 partitions
-		0, 0, 0, 6, // partition 6
-		0, 0, 0, 7, // partition 7
-	}
-
-	tripleDeleteOffsetsRequest = []byte{
-		0, 3, 'f', 'o', 'o', // group name: foo
-		0, 0, 0, 2, // 2 topics
-		0, 3, 'b', 'a', 'r', // topic name: bar
-		0, 0, 0, 2, // 2 partitions
-		0, 0, 0, 6, // partition 6
-		0, 0, 0, 7, // partition 7
-		0, 3, 'b', 'a', 'z', // topic name: baz
-		0, 0, 0, 1, // 1 partition
-		0, 0, 0, 0, // partition 0
-	}
 )
 
 func TestDeleteOffsetsRequest(t *testing.T) {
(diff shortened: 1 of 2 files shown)


dnwe:
Thanks for looking at this. The change here does fix the test, although if we do want the result to always be reproducible you'd want to sort the parent topic names as well as that is also a map isn't it?

However, it does also raise the question of: "do we want to be sorting/ordering the protocol responses to make testing easier, when the clients themselves don't care?" — I wonder if we should change our comparison method in our tests

faillefer:
Since iteration order on map is randomized the encoded response is variable and cannot be checked against a fixed byte array.
Should we disable this check like in  [offset_commit_response_test.go](https://github.com/Shopify/sarama/blob/master/offset_commit_response_test.go) ?

faillefer:
If we do as in [offset_commit_response_test.go](https://github.com/Shopify/sarama/blob/master/offset_commit_response_test.go) the result would be :

https://github.com/faillefer/sarama/commit/d3b16f1e5786c745889d13464d6201158d1adb09

dnwe:
@faillefer yeah I think that’s the best option for now

dnwe (approved):

PR #2016, opened 2021-09-13
chore: replace deprecated Go calls

dnwe:
- always perform time.Time comparison with the Equal func
  see https://pkg.go.dev/time#Time.Equal
- replace deprecated ioutil usage with corresponding io.|os. calls


bai (approved):
Awesomeness!

PR #2020, opened 2021-09-13
fix(test): remove testLogger from TrackLeader test

### LINKED ISSUE #2011: test: TestConsumeMessagesTrackLeader is brittle and periodically fails (and then passes on a re-run)
Needs investigating

dnwe:
As it turns out, this setup was periodically causing the test to fail
due to a data race occurring on the `Logger` global when the defer ran
(!)

Fixes #2011


bai:
Thanks! 🙌🏼

PR #2025, opened 2021-09-15
feat: support ApiVersionsRequest V3 protocol

dnwe:
Add support for the ApiVersionsRequest V3 protocol which includes
ClientSoftwareName and ClientSoftwareVersion as specified by KIP-511


bai (approved):
😍

PR #2034, opened 2021-09-20
Add support for kafka 3.0.0

bai:
Sonic fast: https://github.com/apache/kafka/releases/tag/3.0.0


dnwe:
Aside: should we have constants for V2_7_0_1 and V2_8_0_1 ?

bai:
Great question: do we want to add constants for all minor versions? This is the hack I'm doing in Burrow to get around this, for instance: https://github.com/linkedin/Burrow/blob/master/core/internal/helpers/sarama.go#L26-L70

dnwe:
If/when we move to using ApiVersionsResponse results to gate protocol, I guess we wouldn't necessarily have these constants anymore

dnwe (approved):

PR #2051, opened 2021-10-14
Expose the TLS connection state of a broker connection

seveas:
I'm adding kafka support to a piece of monitoring software that does protocol-level checks but also does tls verification beyond what net/tls does. For this to work the software needs access to the tls connection state. [net/smtp](https://pkg.go.dev/net/smtp#Client.TLSConnectionState) and [go-ldap](https://pkg.go.dev/github.com/go-ldap/ldap#Conn.TLSConnectionState) expose this in the same way.

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -241,6 +241,24 @@ func (b *Broker) Connected() (bool, error) {
 	return b.conn != nil, b.connErr
 }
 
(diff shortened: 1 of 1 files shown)


bai:
Thanks for contributing! This change looks good to me - thoughts @dnwe?

dnwe (approved):

PR #2054, opened 2021-10-29
logger: fix debug logs' formatting directives

utrack:
Default DebugLogger had broken Printf directives -
for DebugLogger.Printf(fmt,A,B,C) it called Logger.Printf(fmt,[A B C]),
which resulted in logs like
```
client/coordinator coordinator for consumergroup [topicname
%!s(int32=1) host:9094] is #%!d(MISSING) (%!s(MISSING))
```

This commit fixes value passing for Printf.


dnwe (approved):
Thanks!

Issue #2053, opened 2021-10-22
Sarama has stuck on the batch with zero records length.


pachmu:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka | Go |
|--------|-------|----|
| Version 1.29.1 | AWS Managed Kafka | 1.17 |

##### Configuration

What configuration values are you using for Sarama and Kafka?
``` go
                 cfg.Net.MaxOpenRequests = 1
		// consume all messages
		cfg.Consumer.IsolationLevel = sarama.ReadUncommitted
		// require producer idempotency
		cfg.Producer.Idempotent = true
		// require consistency on Kafka side before considering the message produced
		cfg.Producer.RequiredAcks = sarama.WaitForAll
		// specify network keep-alive to reduce 'broken pipe' errors on non-KMS environments
		cfg.Net.KeepAlive = time.Second * 30
		// increase processing time from 1s (default) to 5s
		cfg.Consumer.MaxProcessingTime = time.Second * 5
```

##### Logs

When filing an issue please provide logs from Sarama and Kafka if at all
possible. You can set `sarama.Logger` to a `log.Logger` to capture Sarama debug
output.

<details><summary>logs: CLICK ME</summary>
<p>

```
2021/10/22 10:38:26.384689 lvl=INF component=kafka records batch &{5805968 49 2 snappy 0 false false 0 2021-09-30 11:45:34.617 +0200 CEST 2021-09-30 11:45:34.617 +0200 CEST 82609 0 1563 [0xc00030e150] false false [] 1326}
2021/10/22 10:38:26.384705 lvl=INF component=kafka records batch &{5806083 49 2 snappy 0 false false 0 2021-09-30 12:00:01.188 +0200 CEST 2021-09-30 12:00:01.188 +0200 CEST 82609 0 1678 [0xc00030e1c0] false false [] 733}
2021/10/22 10:38:26.384728 lvl=INF component=kafka records batch &{5806313 49 2 snappy 0 false false 0 2021-09-30 12:29:55.542 +0200 CEST 2021-09-30 12:29:55.542 +0200 CEST 82609 0 1908 [0xc00030e230] false false [] 10546}
[...]

PR #2057, opened 2021-11-02
fix: stuck on the batch with zero records length

pachmu:
#2053 
Sometimes in rare cases of corruption or other reason, Kafka can return batches with zero records even if a high watermark was not reached. In this case, sarama just tries to reread this batch again and again and is stuck. There is no resolution in this case except skipping this batch.
librdkafka can handle these cases skipping with a warning.


dnwe (approved):
Thanks, this looks like a great catch

bai (approved):

PR #2063, opened 2021-11-05
use ioutil package to support go1.16

ZQHcode:


dnwe:
@ZQHcode whilst we obviously could do this, and on the face of it it seems like a small change, as I mentioned on https://github.com/Shopify/sarama/issues/2062#issuecomment-962046996 — k8s.io/client-go and others have similarly moved towards a minimum of Go 1.16 in recent months and we only test and maintain Sarama on the most recent two major releases of Go. Unfortunately we don’t have a large enough pool of maintainers to cover more. Go 1.15 and earlier no longer receive security updates and are unsupported by the Go team so I’d strongly recommend that you update your toolchain

PR #2067, opened 2021-11-18
fix: zstd decoder does not appropriately closed.

3AceShowHand:
Our service use sarama as kafka produce, when we try to bump to v1.30.0, our CI found potential goroutine leak.

Please take the following as reference:
https://github.com/pingcap/ticdc/pull/3508
https://github.com/klauspost/compress/issues/451

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ coverage.txt
 profile.out
 
 simplest-uncommitted-msg-0.1-jar-with-dependencies.jar
+
+# ide
+.idea/*
--- a/consumer.go
+++ b/consumer.go
@@ -119,6 +119,7 @@ func newConsumer(client Client) (Consumer, error) {
 }
 
 func (c *consumer) Close() error {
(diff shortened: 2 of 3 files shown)


3AceShowHand:
@dnwe 
@bai 

PTAL

bai:
👋🏼 Thanks for your contribution! Could you please sign CLA?

3AceShowHand:
I should have already signed CLA, but it failed.

3AceShowHand:
so how about closing all decoders when closing the consumer ?

dnwe:
But if a user has more than one consumer you can’t close the decoder until all of them have finished with it. So you’d have to refcount. 

A goroutine leak is when a blocked goroutine is left alive and can no longer be used or accessed. In this case the zstd decoder goroutines are intentionally kept around for re-use and don’t grow in number. Unless your app is never going to use Sarama again during its runtime there’s no real need to shutdown the decoder goroutines

Issue #2071, opened 2021-11-26
regression: enabling rackawareness causes severe throughput drops


lizthegrey:
##### Versions

| Sarama | Kafka | Go |
|--------|-------|----|
| v1.29.0 | v3.0.0 (Confluent 7) |  v1.17.1 |

Regression has been bisected to https://github.com/Shopify/sarama/commit/1aac8e5424614233887c5600e8ab4258c89c6b80

See https://github.com/Shopify/sarama/issues/1927#issuecomment-920294156 for another report than just mine.

##### Configuration

Pertinent config variables:
``` go
	c.conf = conf
	if (c.LD != nil && c.LD.BoolVariationCtx(
		ctx,
		launchdarkly.FlagKafkaRackAwareFollowerFetch,
		types.UserAndTeam{User: nil, Team: nil})) {
		c.rackID = os.Getenv("AZ")
	}
```

```
broker.rack=<%= node["ec2"]["availability_zone"] %>
replica.selector.class=<%= node["kafka"]["replica"]["selector"]["class"] %>
```

```
default['kafka']['replica']['selector']['class'] = "org.apache.kafka.common.replica.RackAwareReplicaSelector"
```

##### Problem Description

When AZ is not populated (due to a bug on our side, hah), or when FlagKafkaRackAwareFollowerFetch is false, things behave normally. But throughput drops by 75% if the flag is set and the Sarama library is at or past 1aac... https://share.getcloudapp.com/nOu54vNq
There is a corresponding lag in timestamps between producer and consumer: https://share.getcloudapp.com/DOu6LBAQ

lizthegrey:
@dnwe opened as per your request

dnwe:
Thank you!

lizthegrey:
anything I can do to assist here? rolling back the commit might help, but also would break that fix so...

dnwe:
@lizthegrey apologies, I've actually been off sick this week so hadn't had a chance to code up a fix for this yet, but I do aim to look at it soon.

[...]

PR #2076, opened 2021-12-01
fix: only update preferredReadReplica if valid

dnwe:
FetchResponse from a follower will _not_ contain a PreferredReadReplica. It seems like the partitionConsumer would overwrite the previously assigned value from the leader with -1 which would then trigger the "reconnect to the current leader" changes from #1936 causing a flip-flop effect.

Contributes-to: #2071

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -132,16 +132,17 @@ func (c *consumer) Partitions(topic string) ([]int32, error) {
 
 func (c *consumer) ConsumePartition(topic string, partition int32, offset int64) (PartitionConsumer, error) {
 	child := &partitionConsumer{
-		consumer:  c,
-		conf:      c.conf,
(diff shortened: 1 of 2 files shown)


bai (approved):

lizthegrey:
Fix confirmed working on a canary. Deploying with dep pinned to this branch across our environments, will report back.

lizthegrey:
Deployed fleetwide and rack awareness turned back on. It looks buttery smooth, thank you!

lizthegrey (approved):

dnwe:
Perfect. Ok let’s go ahead and merge this so we can cut a release with the fix and mark the previous one as bad. 

I’ll revisit this with some test coverage in a follow on PR

PR #2086, opened 2021-12-07
feat: add default SCRAMClientGeneratorFunc

liangxj8:
Make the client generator in the example used as the default `SCRAMClientGeneratorFunc` automatically when the user set the `SASL.Mechanism` as `sarama.SASLTypeSCRAMSHA256` or `sarama.SASLTypeSCRAMSHA512` in `config.go`.

When I am using the config, I think the client should be automatically generated according to my configuration but not telling me `kafka: invalid configuration (A SCRAMClientGeneratorFunc function must be provided to Net.SASL.SCRAMClientGeneratorFunc)`. When I'm stepping into the imported source code, it is a little bit hard for me to target the useful `GeneratorFunc` because Golang import without the examples.

But when I tried to edit the `config.go`, I thought the best way to add the default `GeneratorFunc` is in the `Validate` function, it isn't a config generator. This is a point that I think may need to be discussed.


liangxj8:
CLA signed.

liangxj8:
> Hi.
> 
> Any chances to move this PR forward?

Merged upstream. It seems @bai is not active now, maybe @dnwe can help for review ?

Tr1monster:
Thank you @liangxj8 this worked perfektly form my project and saved me a hole lot of time!

Issue #2095, opened 2021-12-29
Kafka 2.8.1 error messages are not Nullable strings anymore ?


agriffaut:
##### Versions
Working
| Sarama | Kafka | Go |
|--------|-------|----|
| 1.30.1 | 2.7.0 | 1.16 |

Fail
| Sarama | Kafka | Go |
|--------|-------|----|
| 1.30.1 | 2.8.1 | 1.16 |

##### Logs
<details><summary>logs: CLICK ME</summary>
<p>
Error detected while describing quotas
</p>
</details>

##### Problem Description
For some reason, Kafka 2.8.1 doesn't return a nullable string anymore for DescribeClientQuotasResponse.ErrorMsg on that [API (key: 48)](https://kafka.apache.org/protocol#The_Messages_DescribeClientQuotas).
When no error occurs, a pointer to an empty string is returned.
In the packetDecoder, the getNullableString function reads a 0 in Kafka 2.8.1 rather than -1 in Kafka 2.7.0.

I'll submit a PR to handle this one, **but maybe other APIs are impacted too**

dnwe:
Fixed by https://github.com/IBM/sarama/pull/2096

PR #2096, opened 2021-12-29
fix: ignore non-nil but empty error strings in Describe/Alter client quotas responses

agriffaut:
That PR resolve issue #2095, it has been tested against Kafka 2.7.0 and Kafka 2.8.1

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -1125,7 +1125,7 @@ func (ca *clusterAdmin) DescribeClientQuotas(components []QuotaFilterComponent, 
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


dnwe:
Interesting. We should check the broker source to see when/why this changed. This also suggests we need DescribeClientQuotas coverage in the functional_test suite too

agriffaut:
Hi @dnwe i added some integration tests which covers describe / alter / delete of client quotas, thanks for your feedbacks !

dnwe (approved):

PR #2101, opened 2022-01-06
Fix HighWaterMarkOffset and messages offset in mocks.Consumer

### LINKED ISSUE #2050: `HighWaterMarkOffset()` behaves differently for `mocks.Consumer`
> Huh, ok. Unfortunately, consuming a message is the only way I know of to fetch the HighWatermark. If you find another, I'm happy to include it.
>
>_Originally posted by @eapache in https://github.com/Shopify/sarama/issues/1134#issuecomment-406704004_

I encountered the same issue where `HighWaterMarkOffset()` returns `0` if no messages have been consumed yet. I understand that there is no way to know the `HighWaterMarkOffset` before the first message has been consumed. However, my problem was that I had tests in place that did not cover that case because `mocks.Consumer` always returns the "true" `HighWaterMarkOffset`. 

I'd expect `mocks.Consumer` to also return `0` if no message has been consumed yet. What do you think about that? Otherwise it might be helpful to have that case documented in `HighWaterMarkOffset()`.

dkotvan:
The  behavior of the PartitionConsumer is the HighWaterMarkOffset()
method to only return zero until the first message is consumed but the
mocks.PartitionConsumer is updating the HighWaterMarkOffset() every time
the YieldMessage is called.

This commit changes the PartitionConsumer behavior to be the same as the
real PartitionConsumer.

Fix #2050

### DIFF
--- a/mocks/consumer.go
+++ b/mocks/consumer.go
@@ -156,12 +156,13 @@ func (c *Consumer) ExpectConsumePartition(topic string, partition int32, offset 
 
 	if c.partitionConsumers[topic][partition] == nil {
 		c.partitionConsumers[topic][partition] = &PartitionConsumer{
-			t:         c.t,
-			topic:     topic,
-			partition: partition,
-			offset:    offset,
-			messages:  make(chan *sarama.ConsumerMessage, c.config.ChannelBufferSize),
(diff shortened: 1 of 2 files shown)


dkotvan:
My bad I didn't saw that the behavior of the Consumer was fixed by #2082

PR #2106, opened 2022-01-11
fix: Added broker connection check in metadata refresh loop. Closes broken broker connections

### LINKED ISSUE #1796: ClusterAdmin api gets error broken pipe after about 10 min of inactivity.
Sarama version: 1.27
Kafka version: 1.1.0
I have a Rest application which uses sarama ClusterAdmin to manage topic. Once I initialize the clusteradmin client, in about 10 minutes of inactivity, it starts getting error - broken pipe.
Below is the test code that can generate this error.

```
package main

import (
	"fmt"
	. "github.com/Shopify/sarama"
	"time"
)

func main() {
	fmt.Println("Main started") 
      config := NewConfig()
	config.Version = sarama.V1_1_0_0
	adm, err := sarama.NewClusterAdmin([]string{"172.16.0.28:9092"}, config)
	if err != nil {
		log.Println(err.Error())
	}
        
        topicMetadata, err := adm.DescribeTopics([]string{"test1"})              //works
	fmt.Println(fmt.Sprintf("Describe1.., metadata %+v", topicMetadata))
	if err!=nil {
		fmt.Println(err.Error())
	}

        //sleep for 15 min
        time.Sleep(time.Minute*15)
        topicMetadata, err = adm.DescribeTopics([]string{"test1"})               //fails
	fmt.Println(fmt.Sprintf("Describe2.., metadata %+v", topicMetadata))
	if err!=nil {
		fmt.Println(err.Error())
	}	
}
   
```
Error message is:
Describe1.., metadata [0xc00017e240]

Describe2.., metadata []
write tcp 172.16.0.1:51662->172.16.0.28:9092: write: broken pipe

doxsch:
I have implemented a workaround here. The solution is not optimal but makes the admin client usable for long term executions.

During the metadata request all brokers, which have an open connection, are checked by an ApiVersionRequest if it still works.
If this is not the case, the connection is closed and reopened with the next request.

This way with the refresh interval < 10min or better said < min(connections.max.idle.ms, keepAlive kernel settings) we can achieve that the connections are no longer closed by inactivity and if this should happen anyway, the cached connection will be closed.

In v2 we should take into account that a TCP connection can also be closed remotely.

closes #1796

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/rcrowley/go-metrics"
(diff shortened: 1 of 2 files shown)


VipulZopSmart:
Still experiencing this error in the application, is it got fixed? @doxsch @look

mpavlikWandera:
@bai @dnwe could you pls look at the review?

mpavlikWandera:
Bump ... pretty please have a look at this @bai @dnwe

rnkaufman:
@doxsch @dnwe Hi! Is it possible to get this PR reopened and reprioritized? We are running into this bug as well. Thank you so much 🙇🏻

PR #2107, opened 2022-01-11
fix: skip over KIP-482 tagged fields

dnwe:
The decoder would error out with "non-empty tagged fields are not supported yet" if it encountered a non-zero length for the tagged fields array. Handle these more graciously (as per the KIP) by skipping over them without deserializing.

This fixes an issue with the v3 ApiVersionsResponse on Kafka 2.7.0 and newer which now return the `SupportedFeatures` tagged field on the response (KAFKA-10028).


bai:
❤️

Issue #1608, opened 2020-02-10
Rebalancing issues with slow consumers


pavius:
## Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka | Go |
|--------|-------|----|
| a20d26723487482b4f1af8b8537657a5ba72ae37 (1.25.0)  | * | 1.12 |

## Problem Description

We've identified three issues that affect rebalancing when the consumer takes tens of seconds or minutes to process an event. 

### 1: Consumers must deplete Messages() before they can be shut down
Sarama asks for the following:

```
// Please note, that once a rebalance is triggered, sessions must be completed within
// Config.Consumer.Group.Rebalance.Timeout. This means that ConsumeClaim() functions must exit
// as quickly as possible to allow time for Cleanup() and the final offset commit. If the timeout
// is exceeded, the consumer will be removed from the group by Kafka, which will cause offset
// commit failures.
```
This is a reasonable request since the Sarama waits for all `ConsumeClaim()`s to return before re-establishing a session and joining the new generation of the consumer group. However, it communicates this by `close()`ing the buffered `Messages` channel - not something the consumer can  act upon immediately, and even worse so without first reading all the messages in the channel.

If it takes 3 minutes to handle a single message (which is a viable scenario) the consumers have to set their `ChannelBufferSize` to 1 and their `Rebalance.Timeout` to 6 minutes + some slack (since the consumer will at worst have to handle the message it is handling and the one in the channel. 

#### Proposed solution
Add:

```
// StopConsuming forces an immediate stop of consumption, in addition to closing the Messages() channel
StopConsuming() <-chan struct{}
```

To `ConsumerGroupClaim`. A consumer can then run their message handling in a background go routine and wait for completion of handling *or* `StopConsuming`. They can then decide whether to terminate processing immediately, wait a grace period or complete the current processing of the event (*without* having to deplete the Messages channel to understand that it should exit).

### 2: The first few fetches from Kafka may only fetch data from one or two partitions, starving the rest for a very long time (depending on message size / processing time)
Once a member joins the consumer groups and receives its partitions, they are fed into the "subscription manager" from different go routines. The subscription manager then performs batching and executes a fetch for all the partitions. However, it seems like the batching logic in `subscriptionManager` is faulty, perhaps assuming that `case:` order prioritizes which `case` should be handled when all are signaled which is not the case, according to the go docs (https://golang.org/ref/spec#Select_statements):
```
[...]

PR #2110, opened 2022-01-12
fix: ensure heartbeats only stop after cleanup

dnwe:
Sarama would immediately stop heartbeating on rebalance. If the session teardown took longer than the session timeout, it would get evicted from the group.

This can be fixed by ensuring that heartbeats only stop _after_ the session has completed its cleanup process.

Contributes-to: #1608

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -507,7 +507,9 @@ func (c *consumerGroup) loopCheckPartitionNumbers(topics []string, session *cons
 		select {
 		case <-pause.C:
 		case <-session.ctx.Done():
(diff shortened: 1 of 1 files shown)


dnwe:
@pavius similarly, I extracted this change out of your original issue and fork so we can test and review it in isolation — thanks for your work on this, please could you sign the CLA?

bai:
@pavius 👋🏼 Friendly bump 🙏🏼 We're about to cut a new release - do you think you could sign CLA to unblock us? I'd appreciate it so much. ❤️

pavius:
Signed :)

bai (approved):

PR #2119, opened 2022-01-24
Add Kafka 3.1.0 version number

bai:
Add support for Kafka 3.1.0


dnwe:
@bai I wonder if we should name the PR/commit “add Kafka 3.1.0 version number” rather than “add support”, unless you’ve run through the FV test against a 3.1 cluster already?

bai:
> @bai I wonder if we should name the PR/commit “add Kafka 3.1.0 version number” rather than “add support”, unless you’ve run through the FV test against a 3.1 cluster already?

Done.

Siyu-silvia:
Hi, may I ask whether Kafka 3.1.0 is supported now? Thanks!

PR #2122, opened 2022-01-31
fix(test): make it simpler to re-use toxiproxy

dnwe:
Sometimes it is convenient to re-use the docker containers between runs of an FV test. The existing support for supplying a pre-existing TOXIPROXY_ADDR was a little flakey. Tidy up functional_test.go so that it will getOrCreate the toxiproxy routes to the kafka brokers. Then you can just manually `docker-compose up -d` before running a given test yourself and prevent the teardown at the end.

This is particularly useful for people who have migrated away from docker desktop to limavm and want to use `lima nerdctl compose` locally.

### DIFF
--- a/functional_test.go
+++ b/functional_test.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"io"
 	"log"
-	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


bai (approved):
👍🏼

PR #2124, opened 2022-01-31
Add Kafka 3.1.0 to CI matrix, migrate to bitnami kafka image

bai:
Migrate from cp-kafka proprietary kafka docker image to bitnami one. 

As you could see in the diff, we've had quite some hackery around handling the fact that Confluent Platform versioning scheme is very off with Kafka one. 
Not only that, Confluent Platform lags behind quite noticeably — we add workarounds to pretend we're really testing against the latest Kafka version, while we're not.

To mitigate these annoyances, in this PR I'm migrating to a more actively maintained Bitnami Kafka docker image that seems to be pretty up to date with mainstream Kafka releases. Thanks Bitnami folks!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,8 @@ jobs:
     - name: Lint
       uses: golangci/golangci-lint-action@v2
       if: ${{ matrix.go-version == '1.17.x' }}
+      env:
+        GOFLAGS: -tags=functional
       with:
         version: v1.44.0
 


dnwe:
@bai I’m +1 for moving away from proprietary images so we can get closed to main releases. However, should we just build our own images rather than rely on bitnami ones?

bai:
Oh wow, didn't think it's going to work just like that. @dnwe could you please take a look at this PR once again? 🙏🏼

docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: '3.7'
 services:
   zookeeper-1:
-    image: 'confluentinc/cp-zookeeper:${CONFLUENT_PLATFORM_VERSION:-7.0.1}'
+    image: 'confluentinc/cp-zookeeper:7.0.1'
```
bai, line 4: I don't expect Zookeeper version requirement to change as often as Kafka versions chan ge, so oped to hardcode Zookeeper version instead of adding an extra configuration parameter.

dnwe:
@bai good work — I wonder if it's worth keeping 2.8.1 around in the CI matrix for a bit longer as well? Whilst I know our official support statement is N-1, I think as 3.1.0 has only just released we probably want a bit of coverage on the 2.x series
Unless you think 6 (3 x 1.16, 3 x 1.17) runs of the test suite is too much?

Aside: I wonder if we should split up the matrix so that we run the non-functional unittests with Go 1.16.x and 1.17.x and then run the functional tests against the Kafka versions matrix with just 1.17.x

bai:
@dnwe — thanks for feedback. I've updated the build matrix to reflect it, I think it's a great suggestion. Let me know what you think.

PR #2133, opened 2022-02-08
Fix deadlock when closing Broker in brokerProducer

slaunay:
# Bugfix

A regression was introduced in #2094 (Sarama `1.31.0`) and can result in a deadlock inside the `AsyncProducer` (and indirectly the `SyncProducer`) when handling `Produce` response error (typically network errors, see #2129).
The root cause is an `AsyncProduce` callback used to pipeline `Produce` requests must not close the `Broker` directly.
Indeed, closing the `Broker` blocks till the `responseReceiver` goroutine is done yet the `responseReceiver` goroutine might be blocked invoking a callback.
In this specific case, we are not directly closing the `Broker` in the callback but from:
- the `brokerProducer` `run` goroutine when calling `handleError` after receiving an error producing
- while the `brokerProducer` `bridge` goroutine can be blocked if trying to send another response to the `run` goroutine
- and the `broker` `responseReceiver` goroutine is blocked invoking the callback defined in the `bridge` goroutine

I believe this happens when more than one `Produce` request is in flight to a given broker and we are getting a missing or malformed response which is why it was not detected by the current test suite.
We have been using `Broker.AsyncProduce` and callbacks without issues in production ourselves but this is because we do not rely on the `AsyncProducer` and also made sure not to close the `Broker` in a blocking path.

A simple fix would be to close the broker asynchronously when we receive a response error:
```diff
--- a/async_producer.go
+++ b/async_producer.go
@@ -1028,7 +1028,7 @@ func (bp *brokerProducer) handleError(sent *produceSet, err error) {
        default:
                Logger.Printf("producer/broker/%d state change to [closing] because %s\n", bp.broker.ID(), err)
                bp.parent.abandonBrokerConnection(bp.broker)
-               _ = bp.broker.Close()
+               safeAsyncClose(bp.broker)
                bp.closing = err
                sent.eachPartition(func(topic string, partition int32, pSet *partitionSet) {
                        bp.parent.retryMessages(pSet.msgs, err)
```
But that approach might leak such a goroutine once the `brokerProducer` is abandonned.
So the proposed fix consists of:
- adding a new goroutine to the `brokerProducer` to receive a request to close the `Broker`
- signal such goroutine once when we want to close the `Broker` (e.g. network error)
- wait for it to be done (whether a closing request was sent or not) when we shutdown the `brokerProducer`

## Changes

- add unit test to reproduce the deadlock by simulating a network error
- document possible deadlock when closing the `Broker` from an `AsyncProduce` callback when handling a response error
[...]

PR #2144, opened 2022-02-13
feat(test): add an fvt for broker deadlock

dnwe:
Just a simple test to kill broker connectivity whilst the producer is
in-use and confirm that it is able to successfully recover.


dnwe:
rebased on main now that #2133 has been merged and re-running tests

PR #2156, opened 2022-02-22
Implement resolve_canonical_bootstrap_servers_only

gebn:
This PR allows specifying the equivalent of `client.dns.lookup=resolve_canonical_bootstrap_servers_only`, described in [KIP-235][]. The motivation is identical to the one in the proposal: the lack of this feature effectively mandates providing individual canonical broker hostnames when using GSSAPI, rather than a single alias which resolves to the same underlying IPs. This increases the likelihood of errors, and commits application owners to checking for new and removed brokers on an ongoing basis.

[KIP-235]: https://cwiki.apache.org/confluence/display/KAFKA/KIP-235%3A+Add+DNS+alias+support+for+secured+connection

The Java client implementation for this can be found [here](https://github.com/apache/kafka/blob/3.1.0/clients/src/main/java/org/apache/kafka/clients/ClientUtils.java#L62-L71).

The config could be designed as a `type DNSLookupStrategy int8` enum, with `UseAllDNSIPs` and `ResolveCanonicalBootstrapServersOnly` values, however I went for the `bool` as there is no precedent for mirroring the Java API, and this seemed much simpler to understand.

I was torn about adding `context.Context` to `resolveCanonicalNames()`'s signature. To be consistent with existing code, I've kept this internal. It is not fully supported anyway due to `Net.Proxy.Dialer` being a `proxy.Dialer` rather than `proxy.ContextDialer`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1,13 +1,18 @@
 package sarama
 
 import (
+	"context"
 	"errors"
 	"math"
 	"math/rand"
+	"net"
 	"sort"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
+
+	"golang.org/x/net/proxy"
 )
 
 // Client is a generic Kafka client. It manages connections to one or more Kafka brokers.
@@ -189,6 +194,14 @@ func NewClient(addrs []string, conf *Config) (Client, error) {
 		cachedPartitionsResults: make(map[string][maxPartitionIndex][]int32),
 		coordinators:            make(map[string]int32),
 		transactionCoordinators: make(map[string]int32),
+	}
+
+	if conf.Net.ResolveCanonicalBootstrapServers {
+		var err error
+		addrs, err = client.resolveCanonicalNames(addrs)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	client.randomizeSeedBrokers(addrs)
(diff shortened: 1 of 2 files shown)


gebn:
@dnwe, could you approve the workflows so I can fix any issues? Thanks

gebn:
@dnwe I've rebased and added `Signed-off-by`. Still really keen to add this feature :)

gebn:
The [Test with Kafka 3.1.2](https://github.com/IBM/sarama/actions/runs/5536853969/jobs/10195431297#logs) failure was due to a timeout only 8 tests from the end:

|  STATUS | ELAPSED | PACKAGE | COVER | PASS | FAIL | SKIP  |
|-|-|-|-|-|-|-|
| FAIL   | 960.07s | github.com/Shopify/sarama |  -   | 423  |  8   |  1    |
|  PASS   |  0.03s  | github.com/Shopify/sarama/examples/http_server | 22.5% |  3   |  0   |  0    |
|  PASS   |  0.04s  | github.com/Shopify/sarama/mocks                | 82.1% |  36  |  0   |  0    |

 There are no changes here that would affect performance once connected to brokers, so I think that's a false negative. I've rebased, can you re-approve the workflow to give it another go?

gebn:
Thanks! Please let me know if I need to do anything else - I'll hold off rebasing the source branch, as it'll only wait for the checks to be re-approved.

dnwe (approved):
LGTM, it would be good if you could also add a unitest around the functionality but I’m happy for you to do that in a follow up PR

PR #2164, opened 2022-02-25
fix: return underlying sasl error message

dnwe:
The SASL Authentication response message from Kafka has an additional field which can contain a description of why the authentication failed. Currently we drop this in Sarama and just return a generic message based on the kError code.

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -1167,7 +1167,7 @@ func (b *Broker) sendAndReceiveSASLHandshake(saslType SASLMechanism, version int
 		return res.Err
 	}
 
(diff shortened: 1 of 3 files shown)


dnwe:
Added a commit to condense the default output a little which now gives:

```
kafka: client has run out of available brokers to talk to (Is your cluster reachable?): 3 errors occurred:
        * kafka server: SASL Authentication failed.: Authentication failed, invalid credentials
        * kafka server: SASL Authentication failed.: Authentication failed, invalid credentials
        * kafka server: SASL Authentication failed.: Authentication failed, invalid credentials
```

Issue #2150, opened 2022-02-16
"empty" ProducerMessage may lead to the producer hang


hxiaodon:
> @slaunay
with my local laptop(mac OS 10.15.7), I think I produced the hang problem again :). I suppose that your fix do work and cover many scenarios which make hang problem's occurring more and more harder

Thanks for providing those details.
I am not sure if this is linked to the current issue (deadlock regression) but something is not right indeed and we should probably create another issue for that particular scenario.

Here are some things I found:
- the `brokerProducer` is blocked sending a success:
```
3 @ 0x1038d16 0x1006925 0x10064dd 0x1301574 0x130038a 0x1326142 0x12ff72f 0x12ff65f 0x12fe85f 0x132e9de 0x10688a1
#	0x1301573	github.com/Shopify/sarama.(*asyncProducer).returnSuccesses+0xb3		/Users/tiger/go/pkg/mod/github.com/slaunay/sarama@v1.31.2-0.20220211051606-f1bc44e541ee/async_producer.go:1155
#	0x1300389	github.com/Shopify/sarama.(*brokerProducer).handleSuccess.func1+0x469	/Users/tiger/go/pkg/mod/github.com/slaunay/sarama@v1.31.2-0.20220211051606-f1bc44e541ee/async_producer.go:974
#	0x1326141	github.com/Shopify/sarama.(*produceSet).eachPartition+0x101		/Users/tiger/go/pkg/mod/github.com/slaunay/sarama@v1.31.2-0.20220211051606-f1bc44e541ee/produce_set.go:211
#	0x12ff72e	github.com/Shopify/sarama.(*brokerProducer).handleSuccess+0x8e		/Users/tiger/go/pkg/mod/github.com/slaunay/sarama@v1.31.2-0.20220211051606-f1bc44e541ee/async_producer.go:950
#	0x12ff65e	github.com/Shopify/sarama.(*brokerProducer).handleResponse+0x3e		/Users/tiger/go/pkg/mod/github.com/slaunay/sarama@v1.31.2-0.20220211051606-f1bc44e541ee/async_producer.go:938
#	0x12fe85e	github.com/Shopify/sarama.(*brokerProducer).run+0x1be			/Users/tiger/go/pkg/mod/github.com/slaunay/sarama@v1.31.2-0.20220211051606-f1bc44e541ee/async_producer.go:872
[...]

PR #2182, opened 2022-03-16
producer: ensure that the management message (fin) is never "leaked"

niamster:
Since async producer now support multiple inflight messages
thanks to https://github.com/Shopify/sarama/pull/1686 and https://github.com/Shopify/sarama/pull/2094, it now may "leak"
the "fin" internal management message to Kafka (and to the client)
when broker producer is reconnecting to Kafka broker and retries
multiple inflight messages at the same time.

Steps to reproduce: start async producer and restart Kafka broker (potentially multiple times). 
This trigger leader update and retires current broker producer.
There's a chance that newly created broker producer is used to retry an
inflight message so it is also marked as "dying" but it does not contain
any explicit errors.
This results in a "fin" message to be sent to Kafka and reported as successful
operation to the client.

Versions affected: `v1.31.0` and above.

Sarama logs of the event (with the patch applied):
```
[Sarama] 2022/03/16 09:54:31 producer/broker/10001 starting up
[Sarama] 2022/03/16 09:54:31 producer/broker/10001 state change to [open] on my-topic/26
[Sarama] 2022/03/16 09:54:31 producer/leader/my-topic/26 selected broker 10001
[Sarama] 2022/03/16 09:54:31 producer/leader/my-topic/26 state change to [flushing-1]
[Sarama] 2022/03/16 09:54:31 producer/leader/my-topic/26 state change to [normal]
[Sarama] 2022/03/16 09:54:31 producer/leader/my-topic/26 state change to [retrying-1]
[Sarama] 2022/03/16 09:54:31 producer/broker/10001 state change to [dying-1] on my-topic/26
[Sarama] 2022/03/16 09:54:31 producer/leader/my-topic/26 abandoning broker 10001
[Sarama] 2022/03/16 09:54:31 producer/broker/10001 input chan closed
[Sarama] 2022/03/16 09:54:33 producer/broker/10001 state change to [open] on my-topic/26
[Sarama] 2022/03/16 09:54:33 producer/leader/my-topic/26 selected broker 10001
[Sarama] 2022/03/16 09:54:33 producer/leader/my-topic/26 state change to [flushing-1]
[Sarama] 2022/03/16 09:54:33 producer/leader/my-topic/26 state change to [normal]
```

Kafka version: `2.7.1`
Go version: `1.17.7`

Kafka producer config:
```
cfg.Producer.Retry.Backoff = 2 * time.Second
cfg.Producer.Return.Errors = true
cfg.Producer.Flush.Messages = 1000
cfg.Producer.Flush.Frequency = 100 * time.Millisecond
cfg.ChannelBufferSize = 1024
```

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -840,6 +840,15 @@ func (bp *brokerProducer) run() {
 				continue
 			}
 
+			if msg.flags&fin == fin {
+				// New broker producer that was caught up by the retry loop
+				bp.parent.retryMessage(msg, ErrShuttingDown)
+				delete(bp.currentRetries[msg.Topic], msg.Partition)
+				Logger.Printf("producer/broker/%d state change to [dying-%d] on %s/%d\n",
+					bp.broker.ID(), msg.retries, msg.Topic, msg.Partition)
+				continue
+			}
+
 			if bp.buffer.wouldOverflow(msg) {
 				Logger.Printf("producer/broker/%d maximum request accumulated, waiting for space\n", bp.broker.ID())
 				if err := bp.waitForSpace(msg, false); err != nil {
--- a/async_producer.go
+++ b/async_producer.go
@@ -843,7 +843,6 @@ func (bp *brokerProducer) run() {
 			if msg.flags&fin == fin {
 				// New broker producer that was caught up by the retry loop
 				bp.parent.retryMessage(msg, ErrShuttingDown)
-				delete(bp.currentRetries[msg.Topic], msg.Partition)
 				Logger.Printf("producer/broker/%d state change to [dying-%d] on %s/%d\n",
 					bp.broker.ID(), msg.retries, msg.Topic, msg.Partition)
 				continue
--- a/async_producer.go
+++ b/async_producer.go
@@ -843,7 +843,7 @@ func (bp *brokerProducer) run() {
 			if msg.flags&fin == fin {
 				// New broker producer that was caught up by the retry loop
 				bp.parent.retryMessage(msg, ErrShuttingDown)
-				Logger.Printf("producer/broker/%d state change to [dying-%d] on %s/%d\n",
+				DebugLogger.Printf("producer/broker/%d state change to [dying-%d] on %s/%d\n",
 					bp.broker.ID(), msg.retries, msg.Topic, msg.Partition)
 				continue
 			}
--- a/async_producer_test.go
+++ b/async_producer_test.go
@@ -824,7 +824,7 @@ func TestAsyncProducerBrokerRestart(t *testing.T) {
 
 	wg.Wait()
 
-	expectResultsWithTimeout(t, producer, 40, 00, 10*time.Second)
+	expectResultsWithTimeout(t, producer, 40, 0, 10*time.Second)
 
 	seedBroker.Close()
 	leader.Close()
--- a/async_producer_test.go
+++ b/async_producer_test.go
@@ -731,7 +731,8 @@ func TestAsyncProducerMultipleRetriesWithConcurrentRequests(t *testing.T) {
 }
 
 func TestAsyncProducerBrokerRestart(t *testing.T) {
-	Logger = log.New(os.Stdout, "[sarama] ", log.LstdFlags)
+	// Logger = log.New(os.Stdout, "[sarama] ", log.LstdFlags)
+
 	seedBroker := NewMockBroker(t, 1)
 	leader := NewMockBroker(t, 2)
 
--- a/async_producer_test.go
+++ b/async_producer_test.go
@@ -17,31 +17,67 @@ import (
 
 const TestMessage = "ABC THE MESSAGE"
 
-func closeProducer(t *testing.T, p AsyncProducer) {
+func closeProducerWithTimeout(t *testing.T, p AsyncProducer, timeout time.Duration) {
 	var wg sync.WaitGroup
 	p.AsyncClose()
 
+	closer := make(chan struct{})
+	timer := time.AfterFunc(timeout, func() {
+		t.Error("timeout")
+		close(closer)
+	})
+	defer timer.Stop()
+
 	wg.Add(2)
 	go func() {
-		for range p.Successes() {
-			t.Error("Unexpected message on Successes()")
+		defer wg.Done()
+		for {
+			select {
+			case <-closer:
+				return
+			case _, ok := <-p.Successes():
+				if !ok {
+					return
+				}
+				t.Error("Unexpected message on Successes()")
+			}
 		}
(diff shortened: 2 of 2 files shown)


slaunay:
@niamster Thanks for the fix,  any chance you can include a unit test to reproduce that tricky race condition?

I think the linked issue is #2150, if you add `fixes #2150` in your commit message somewhere that issue should get closed once the PR is merged.

async_producer.go
```diff
@@ -840,6 +840,15 @@ func (bp *brokerProducer) run() {
 				continue
 			}
 
+			if msg.flags&fin == fin {
+				// New broker producer that was caught up by the retry loop
+				bp.parent.retryMessage(msg, ErrShuttingDown)
+				delete(bp.currentRetries[msg.Topic], msg.Partition)
```
slaunay, line 846: I'm not sure those 3 lines are necessary because the `bp.currentRetries` should be "empty" for that particular topic partition because `bp.needsRetry(msg)` returned `nil` above:
https://github.com/Shopify/sarama/blob/bad67e5b089437bc73f8034a95017e77be71e8b0/async_producer.go#L903-L909

niamster: This is true that `bp.currentRetries[msg.Topic]` is `nil` in this case (`bp.currentRetries[msg.Topic][msg.Partition]` is set to `nil` in the "syn" case above). 
I decided to `delete` the key to ensure that the key does not exist to avoid confusion "key exist but nil" vs. "key does not exist" to mimic the other block of code above related to handling `fin` message.

TBH I'm not entirely sure about the purpose of the following code in handling `syn` case:
```
				if bp.currentRetries[msg.Topic] == nil {
					bp.currentRetries[msg.Topic] = make(map[int32]error)
				}
				bp.currentRetries[msg.Topic][msg.Partition] = nil
```

I don't see where it might be useful (for now).

niamster:
> @niamster Thanks for the fix, any chance you can include a unit test to reproduce that tricky race condition?
> 
> I think the linked issue is #2150, if you add `fixes #2150` in your commit message somewhere that issue should get closed once the PR is merged.

Thanks @slaunay for the review! I will try to add a test, this is a little bit tricky to reproduce robustly though.

As for https://github.com/Shopify/sarama/issues/2150 - TBH I'm not 100% sure this is exactly the same issue. I was looking into it before submitting this PR but was not confident that this fix also fixes that one.
Maybe @hxiaodon can confirm that this patch fixes their issue as well? If this patch fixes their problem I'm happy to add `fixes #2150` some commit :)

async_producer_test.go
```diff
@@ -17,31 +17,67 @@ import (
...
+			}
 		}
-		wg.Done()
 	}()
 	wg.Wait()
 }
 
-func expectResults(t *testing.T, p AsyncProducer, successes, errors int) {
+func closeProducer(t *testing.T, p AsyncProducer) {
+	closeProducerWithTimeout(t, p, time.Hour)
```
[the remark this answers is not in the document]
niamster: I've put `1h` to preserve current behaviour. I think `5mn` makes more sense than `10mn` if we want the test fail before the whole test suite times out. I will change it to `5mn` in next commit.

async_producer.go
slaunay: > TBH I'm not entirely sure about the purpose of the following code in handling `syn` case:
> 
> ```
> 				if bp.currentRetries[msg.Topic] == nil {
> 					bp.currentRetries[msg.Topic] = make(map[int32]error)
> 				}
> 				bp.currentRetries[msg.Topic][msg.Partition] = nil
> ```
> 
> I don't see where it might be useful (for now).

I believe when a `syn` message is seen the `brokerProducer` should:
- initialize `currentRetries` to prevent a panic when accessing `bp.currentRetries[topic][partition]` (typically on first use of the `brokerProducer`).
- clear a previous retries "state" for such topic partition.

niamster: Thanks @slaunay for the reply!
I don't see how this could lead to any panics 🤔
I've given another thought and I think it does not really matter (as far as I can tell) whether the retries "metadata" structure is cleared or not (I was trying to imagine what if the code is changed in the future so this branch is executed even if `bp.currentRetries[msg.Topic][msg.Partition]` is not `nil`). So I will update the code to minimise the changes.

dnwe (commented):
Thanks @niamster — these changes look good to me.

@slaunay did you have any further questions/suggestions before I merge?

async_producer_test.go
```diff
@@ -693,6 +730,112 @@ func TestAsyncProducerMultipleRetriesWithConcurrentRequests(t *testing.T) {
 	closeProducer(t, producer)
 }
 
+func TestAsyncProducerBrokerRestart(t *testing.T) {
+	// Logger = log.New(os.Stdout, "[sarama] ", log.LstdFlags)
+
+	seedBroker := NewMockBroker(t, 1)
+	leader := NewMockBroker(t, 2)
+
+	var leaderLock sync.Mutex
```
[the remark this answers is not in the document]
slaunay: Got it, I thought like some other unit tests the same address was reused with `leader = NewMockBrokerAddr(t, 2, leaderAddr)` to stop/start but moving to a different port works too.

async_producer_test.go
```diff
@@ -693,6 +730,112 @@ func TestAsyncProducerMultipleRetriesWithConcurrentRequests(t *testing.T) {
...
+
+		time.Sleep(50 * time.Millisecond)
+
+		prodSuccess := new(ProduceResponse)
+		prodSuccess.AddTopicPartition("my_topic", 0, ErrNotLeaderForPartition)
+		return prodSuccess
+	})
+
+	config := NewTestConfig()
+	config.Producer.Retry.Backoff = 250 * time.Millisecond
```
slaunay, line 782: I believe this is the same as the default and could be removed but nothing wrong in being explicit.

niamster: It's now set to [100ms](https://github.com/Shopify/sarama/blob/f07b7b87556741de750337199ccd02aa2a996fe2/config.go#L479).
Also I prefer to have an explicit value to avoid any surprises in the future.

slaunay:
Thanks a lot @niamster!

@dnwe The fix looks good to me.

Do you think a `retract` is beneficial to users and if so should we add it in this PR to reduce the work for the next release?

niamster:
> Do you think a retract is beneficial to users and if so should we add it in this PR to reduce the work for the next release?

Yes, I think so. Because this regression causes data corruption (unexpected data in kafka). I believe it should be a different PR thought.

dnwe (approved):
@niamster thanks again! and thanks @slaunay for the detailed review and guidance on the fix

PR #2186, opened 2022-03-21
fix: report more unrecoverable errors during SASL authentication

faec:
Several types of unrecoverable connection / authentication failure are not recognized as unrecoverable, and result in retrying all brokers and falling back on "client has run out of brokers to talk to" after they all fail. The errors that can cause this include unsupported SASL mechanisms, TLS versions, and ciphers, as well as invalid broker certificates and bad connection protocols. This PR adds additional error types and recognizes them as unrecoverable, so they are reported immediately to the client's error channel instead of only within Sarama's debug output.

This is based on a [similar PR](https://github.com/elastic/sarama/pull/15) in Elastic's vendored Sarama, based on confusion we were seeing with the "run out of brokers to talk to" message.

### DIFF
--- a/client.go
+++ b/client.go
@@ -929,11 +929,28 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int,
 				Logger.Println("client is not authorized to access this topic. The topics were: ", topics)
 				return err
 			}
+
+			if errors.Is(err, ErrUnsupportedSASLMechanism) {
+				Logger.Println("requested SASL mechanism is not supported by the broker")
+				return err
+			}
+
 			// else remove that broker and try again
 			Logger.Printf("client/metadata got error from broker %d while fetching metadata: %v\n", broker.ID(), err)
 			_ = broker.Close()
 			client.deregisterBroker(broker)
 		} else {
+			if errors.Is(err, ErrSASLHandshakeReadEOF) ||
+				errors.Is(err, ErrSASLHandshakeSendEOF) ||
+				errors.Is(err, ErrFetchMetadataEOF) ||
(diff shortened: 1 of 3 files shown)


faec:
Oh, that test was my mistake! That must have been leftover in my local directory and got re-added when I made the upstream patch.

dnwe (approved):

client.go
```diff
@@ -918,11 +918,28 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int,
...
+				errors.Is(err, ErrSASLHandshakeSendEOF) ||
+				errors.Is(err, ErrFetchMetadataEOF) ||
+				errors.Is(err, ErrBadTLSHandshake) {
+				// These errors are typically unrecoverable connection /
+				// authentication failures, so we return them
+				// directly here to avoid falling back on the less useful
+				// "client has run out of brokers" error after retrying.
+				return err
+			}
+
```
dnwe, line 942: Having looked at the test failures, and thought about this PR some more — do we still want to keep this "treat as unrecoverable" case here? Theoretically our client may have just been unluckly and it's connection with a broker happened to be disconnected (e.g., due to a restart) whilst it was refreshing metadata from that one — we don't want to hard fail there do we?

Now that Sarama will wrap all errors are return them (rather than just a "run out of brokers" message) perhaps the per-broker specific messages will be sufficient to help with diagnosing any issue?

dnwe (changes requested):

client.go
dnwe: Yes so https://github.com/Shopify/sarama/pull/2131 made that available in wrapped errors

PR #2191, opened 2022-03-22
feat: add batched CreateACLs func to ClusterAdmin

nkostoulas:
Add new method to the `ClusterAdmin` interface to allow batch creation
of `ACL`s. Creating `ACL`s one at a time seems to be very slow and for
applications that need to create a lot of these, it takes time (and
might cause downtime).

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -80,6 +80,7 @@ type ClusterAdmin interface {
 	// This operation is not transactional so it may succeed or fail.
 	// If you attempt to add an ACL that duplicates an existing ACL, no error will be raised, but
 	// no changes will be made. This operation is supported by brokers with version 0.11.0.0 or higher.
+	// Deprecated: Use CreateACLs instead.
 	CreateACL(resource Resource, acl Acl) error
 
 	// Creates access control lists (ACLs) which are bound to specific resources.


admin.go
```diff
@@ -76,11 +76,17 @@ type ClusterAdmin interface {
 	// The configs for a particular resource are updated automatically.
 	IncrementalAlterConfig(resourceType ConfigResourceType, name string, entries map[string]IncrementalAlterConfigsEntry, validateOnly bool) error
 
+	// Creates an access control list (ACL) which is bound to a specific resource.
+	// This operation is not transactional so it may succeed or fail.
+	// If you attempt to add an ACL that duplicates an existing ACL, no error will be raised, but
+	// no changes will be made. This operation is supported by brokers with version 0.11.0.0 or higher.
+	CreateACL(resource Resource, acl Acl) error
```
dnwe, line 83: Should we mark the old singular API as deprecated so we can remove it in a future major version bump?

```suggestion
	// no changes will be made. This operation is supported by brokers with version 0.11.0.0 or higher.
	// Deprecated: Use CreateACLs instead.
	CreateACL(resource Resource, acl Acl) error
```

nkostoulas: added!

PR #2202, opened 2022-03-31
Experiment: looking for goroutine leaks triggered by tests

k-wall:
Test failures on another PR caused me to notice some goroute leaks when running the tests.  I fixed some by way of #2198.  Three still remain which have a less obvious root cause:

To reproduce, with this PR do:

```
GOFLAGS="-tags=goleak" go test -c -o tests
for test in $(go test -list . | grep -E "^(Test|Example)"); do ./tests -test.run "^$test\$" &>/dev/null && echo -n "." || echo -e "\n$test failed"; done
```

gives:
```
........................................................
TestRefreshMetaDataWithDifferentController failed
................................
TestAsyncProducerIdempotentRetryCheckBatch failed
..............................................................
TestClientRefreshBrokers failed
................................................
```

Running the test individually gets the details:

./tests -test.run ^TestAsyncProducerIdempotentRetryCheckBatch$

```
kwall@Oslo sarama % ./tests -test.run ^TestRefreshMetaDataWithDifferentController$
PASS
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 11 in state chan receive, with github.com/Shopify/sarama.(*Broker).responseReceiver on top of the stack:
goroutine 11 [chan receive]:
github.com/Shopify/sarama.(*Broker).responseReceiver(0xc0000b4e00)
	/Users/kwall/src/sarama/broker.go:1047 +0x74
github.com/Shopify/sarama.withRecover(0x0)
	/Users/kwall/src/sarama/utils.go:43 +0x3e
created by github.com/Shopify/sarama.(*Broker).Open.func1
	/Users/kwall/src/sarama/broker.go:245 +0xb65
]
kwall@Oslo sarama % ./tests -test.run ^TestAsyncProducerIdempotentRetryCheckBatch$
PASS
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 38 in state select, with github.com/Shopify/sarama.(*brokerProducer).run on top of the stack:
goroutine 38 [select]:
github.com/Shopify/sarama.(*brokerProducer).run(0xc0002824e0)
	/Users/kwall/src/sarama/async_producer.go:807 +0x19a
github.com/Shopify/sarama.withRecover(0x0)
	/Users/kwall/src/sarama/utils.go:43 +0x3e
created by github.com/Shopify/sarama.(*asyncProducer).newBrokerProducer
	/Users/kwall/src/sarama/async_producer.go:692 +0x25b

[...]

PR #2203, opened 2022-04-01
fix: prevent RefreshBrokers leaking old brokers 

k-wall:
#2202 exposes a goroutine test in the client's `RefreshBrokers` API. 

On analysis I see that `RefreshBrokers` implementations leak the original seed brokers and the test itself was leaking a client.  This PR resolves both issues.


dnwe (approved):
LGTM thanks!

PR #2204, opened 2022-04-01
fix: prevent RefreshController leaking controller

k-wall:
As exposed by #2202, the `RefreshController` API leaks the broker that was in the controller role.


dnwe (commented):
LGTM thanks!

PR #2205, opened 2022-04-05
fix: prevent metrics leak when authenticate fails 

### LINKED ISSUE #1960: Memory leak when broker authenticate failed
```
		b.incomingByteRate = metrics.GetOrRegisterMeter("incoming-byte-rate", conf.MetricRegistry)
		b.requestRate = metrics.GetOrRegisterMeter("request-rate", conf.MetricRegistry)
		b.requestSize = getOrRegisterHistogram("request-size", conf.MetricRegistry)
		b.requestLatency = getOrRegisterHistogram("request-latency-in-ms", conf.MetricRegistry)
		b.outgoingByteRate = metrics.GetOrRegisterMeter("outgoing-byte-rate", conf.MetricRegistry)
		b.responseRate = metrics.GetOrRegisterMeter("response-rate", conf.MetricRegistry)
		b.responseSize = getOrRegisterHistogram("response-size", conf.MetricRegistry)
		b.requestsInFlight = metrics.GetOrRegisterCounter("requests-in-flight", conf.MetricRegistry)
		// Do not gather metrics for seeded broker (only used during bootstrap) because they share
		// the same id (-1) and are already exposed through the global metrics above
		if b.id >= 0 {
			b.registerMetrics()
		}

		if conf.Net.SASL.Enable {
			b.connErr = b.authenticateViaSASL()

			if b.connErr != nil {
				err = b.conn.Close()
				if err == nil {
					Logger.Printf("Closed connection to broker %s\n", b.addr)
				} else {
					Logger.Printf("Error while closing connection to broker %s: %s\n", b.addr, err)
				}
				b.conn = nil
				atomic.StoreInt32(&b.opened, 0)
				return
			}
		}
```

Broker register metrics name to `registeredMetrics` and then authenticate to broker by socket, when authenticate failed many times, the `registeredMetrics` is slice and will append a metrics name many times

Stephan14:
Broker register metrics name to `registeredMetrics` and then
authenticate to broker by socket, when authenticate failed many times,
the `registeredMetrics` is slice and will append a metrics name many
times

Fixes #1960

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -32,7 +32,7 @@ type Broker struct {
 	responses     chan *responsePromise
 	done          chan bool
 
-	registeredMetrics []string
+	registeredMetrics map[string]struct{}
 
 	incomingByteRate       metrics.Meter
 	requestRate            metrics.Meter
@@ -1683,7 +1683,7 @@ func (b *Broker) registerMetrics() {
 }
 
 func (b *Broker) unregisterMetrics() {
-	for _, name := range b.registeredMetrics {
+	for name := range b.registeredMetrics {
 		b.conf.MetricRegistry.Unregister(name)
 	}
 	b.registeredMetrics = nil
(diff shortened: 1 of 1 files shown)


dnwe (approved):
Thanks good catch! The changes to fix this look good to me

PR #2217, opened 2022-04-14
feature: basic support for context

aybabtme:
👋 I'm just messing around here to try and find a way to have contexts supported by code that I work with that uses sarama. Is this something that would be of interest? If so, I can look at writing tests and iterating and so on.

The idea being that you can set `Context` on a message and it will be best-effort supported by the producers. In the sync producer case, this makes a lot of mostly obvious sense: _"I'm willing to wait this long to send a message, otherwise give up"_. 

In the async producer case, the use case is a bit less obvious, since the assumption is that a message will be sent _sometime_ in the future. Yet, since the `ProducerMessage.Context` is a field and not a method argument, and therefor opt-in, messages that don't set a context wouldn't see any difference. A user of the async producer would have to set the field to signify that they want to have the message sort-of respect the `ctx.Done()` lifetime. Why would someone do this? Perhaps they're willing to send a message async for up to 5s, after which the message would be untimely and its usefulness gone.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -1,6 +1,7 @@
 package sarama
 
 import (
+	"context"
 	"encoding/binary"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


zhangyueshan:
I need this!

PR #2218, opened 2022-04-15
Backport: Fix a potential data race on a global variable

pior:
Re-apply the same fix as https://github.com/Shopify/sarama/pull/2171 on v1.30.1

### DIFF
--- a/version.go
+++ b/version.go
@@ -1,11 +1,17 @@
 package sarama
 
-import "runtime/debug"
+import (
+	"runtime/debug"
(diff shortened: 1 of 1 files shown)


dnwe:
@pior I'm not sure I understand? you've branched from the v1.30.1 tag and added the version.go fix that was previously merged to main — I guess this branch is for your own usage?

I'm not sure what the pull request is intended for though — were you hoping that we would make a 1.30.2 based on your branch?

Issue #2233, opened 2022-05-23
KIP-368 implementation connection disconnects during re-auth or OOM


k-wall:
##### Versions

Sarama: 1.33.0

##### Configuration

Any sarama based application connecting to Kafka brokers with SASL connection re-auth enabled.  The Issue was detected testing out the [Strimzi Canary 0.3.0 RC](https://lists.cncf.io/g/cncf-strimzi-users/topic/rc1_of_strimzi_canary_0_3_0/91119990?p=,,,20,0,0,0::recentpostdate/sticky,,,20,2,0,91119990,previd%3D1653065206495820530,nextid%3D1643721642661594726&previd=1653065206495820530&nextid=1643721642661594726)

##### Logs

```
[Sarama] 2022/05/23 08:36:42 Completed pre-auth SASL handshake. Available mechanisms: [SCRAM-SHA-512 PLAIN]
[Sarama] 2022/05/23 08:36:42 Session expiration in 10000 ms and session re-authentication on or after 8680 ms
[Sarama] 2022/05/23 08:36:42 Connected to broker at localhost:9092 (registered as #0)
[Sarama] 2022/05/23 08:36:42 Closed connection to broker localhost:9092
```

##### Problem Description

Issue is with the KIP-368 implementation added by PR #2197.    The defect means that sometimes the re-auth fails to function correctly.  The client application may suffer unexpected disconnections, or worse suffer an OOM condition.  The problem worsens under load.

The issue is with the new implementation.  It fails to account for the the I/O model of sarama correctly. Its calling of `authenticateViaSASL` results in input contention between the receiver Go routine and the itself.  One my steal bytes from the other, resulting in unexpected connection state.  The OOM is a result of incorrect bytes being interpreted as payload lengths.

k-wall:
Apologies for the defect, I intend to open an PR soon.

PR #2234, opened 2022-05-23
fix: KIP-368 use receiver goroutine to process all sasl v1 responses

k-wall:
Fixes defect (#2233) in kip-368 implementation where existing incoming wire traffic would get confused with re-authentication traffic leading to unexpected disconnects and OOM errors.

The approach taken uses the existing receiver routine to process all SASL v1+ responses for both authentication and re-authentication.   For SASL v0, the existing behaviour is maintained.

### DIFF
--- a/response_header.go
+++ b/response_header.go
@@ -2,11 +2,6 @@ package sarama
 
 import "fmt"
 
-const (
-	responseLengthSize = 4
-	correlationIDSize  = 4
-)
-
 type responseHeader struct {
 	length        int32
 	correlationID int32


ppatierno:
@dnwe do you think it is possible to review and get this fix merged soon? The issue makes the re-auth feature not usable.
Our next Strimzi canary release 0.3.0 was heavily based on this feature but now we are blocked waiting for the fix.
Any chance to have a Sarama patched release 1.33.1 out?

dnwe (approved):
LGTM

dnwe:
@ppatierno no problem, I actually pushed it out as Version 1.34.0 because I wanted to include the KIP-345 (static membership) support to allow people to test that too

broker.go
```diff
@@ -239,12 +244,27 @@ func (b *Broker) Open(conf *Config) error {
...
+		if conf.Net.SASL.Enable && !useSaslV0 {
+			b.connErr = b.authenticateViaSASLv1()
+			if b.connErr != nil {
+				close(b.responses)
+				err = b.conn.Close()
+				if err == nil {
+					DebugLogger.Printf("Closed connection to broker %s\n", b.addr)
+				} else {
+					Logger.Printf("Error while closing connection to broker %s: %s\n", b.addr, err)
+				}
```
pierDipi, line 257: Hi @k-wall, I'm wondering if there is there a reason why `b.connErr` is not logged here too? https://github.com/IBM/sarama/issues/2994

PR #2248, opened 2022-06-07
fix(protocol): tidyup DescribeGroupsResponse

dnwe:
As we currently expose this directly in the response from admin.go it
makes sense to document the struct fields using the text included in the
kafka json protocol description. Whilst checking the protocol
implementation against the specs, I worked these in.


aiquestion:
@dnwe yes, that's much better. thanks :-)

PR #2252, opened 2022-06-16
feat: add option to propagate OffsetOutOfRange error

dkolistratova:
When consuming a partition using a consumer group, the code
handles ErrOffsetOutOfRange errors by resetting to the "initial"
position, as specified by user (i.e. either oldest or newest available
offset). This, however, can be very dangerous. Say a consumer has
consumed up to offset 100 on replica A but replica B has only
replicated up to offset 99 due to temporary under-replication. During
a rebalance, sarama can end up with an offset out-of-range error if it
fetches partition metadata from replica B since the desired offset of
100 is greater than the newest offset of 99. The sarama consumer would
reset the offset in this case, which can cause reprocessing of old
data, especially if the initial offset is configured as "oldest".

This commit adds a config flag to disable this automatic reset. In the
above case, the consumer will be able to proceed normally after the
data replicates.

### DIFF
--- a/config.go
+++ b/config.go
@@ -296,8 +296,17 @@ type Config struct {
 				// coordinator for the group.
 				UserData []byte
 			}
+
 			// support KIP-345
 			InstanceId string
+
+			// If true, consumer offsets will be automatically reset to configured Initial value
+			// if the fetched consumer offset is out of range of available offsets. Out of range
+			// can happen if the data has been deleted from the server, or during situations of
+			// under-replication where a replica does not have all the data yet. It can be
+			// dangerous to reset the offset automatically, particularly in the latter case. Defaults
+			// to true to maintain existing behavior.
+			ResetInvalidOffsets bool
 		}
 
 		Retry struct {
@@ -499,6 +508,7 @@ func NewConfig() *Config {
 	c.Consumer.Group.Rebalance.Timeout = 60 * time.Second
 	c.Consumer.Group.Rebalance.Retry.Max = 4
 	c.Consumer.Group.Rebalance.Retry.Backoff = 2 * time.Second
+	c.Consumer.Group.ResetInvalidOffsets = false
 
 	c.ClientID = defaultClientID
 	c.ChannelBufferSize = 256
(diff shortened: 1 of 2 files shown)


dkolistratova:
recreation of https://github.com/Shopify/sarama/pull/1183 because it is still actual and needed. Nobody was against this change, but it ended up in a stalled state and closed.

dnwe:
@dkolistratova hmm, whilst it may or may not be desirable, it seems like we match the behaviour of the Java kafka-client at the moment? 

Isn't this PR adding functionality similar to unmerged https://github.com/apache/kafka/pull/9006 that came out of [KAFKA-3370](https://issues.apache.org/jira/browse/KAFKA-3370)?

dkolistratova:
@dnwe we've coped with the CLA, can it be reviewed/discussed? 
I looked closer into the KAFKA-3370 and yes, it has connection with the current issue, it is about the line " 2) needs to be avoid (i.e. people need to be notified upon offsets out-of-range rather than silently offset reset"

maximkosov:
@dnwe Looks like unmerged https://github.com/apache/kafka/pull/9006 will reset offsets to nearest value. But it also might be not what we need. What is this PR is adding is more similar to OffsetResetStrategy.None

dnwe (approved):

config.go
```diff
@@ -499,6 +508,7 @@ func NewConfig() *Config {
 	c.Consumer.Group.Rebalance.Timeout = 60 * time.Second
 	c.Consumer.Group.Rebalance.Retry.Max = 4
 	c.Consumer.Group.Rebalance.Retry.Backoff = 2 * time.Second
+	c.Consumer.Group.ResetInvalidOffsets = false
```
dethi, line 511: Comment says "default to true" but it's not reflected here.

yakenji: This parameter is so dangerous. It should be true by default.
If this parameter is false, the resetting of offsets is invalid, which can cause an infinite loop of closing and adding subscriptions.

PR #2255, opened 2022-06-19
 fix: fix metadata retry backoff invalid when get metadata failed

Stephan14:
fixes https://github.com/Shopify/sarama/issues/1711


Stephan14:
fixes https://github.com/Shopify/sarama/issues/1711

PR #2267, opened 2022-07-14
updated golang.org/x/crypto having CVE-2022-27191 ssh vulnerability

mohit-angelone:
Addresses issue: https://github.com/Shopify/sarama/issues/2265


dnwe:
Sarama doesn’t pull x/crypto as a direct dependency. We get it as an indirect dep via https://github.com/jcmturner/gokrb5 and so as a library we can’t control its version. We either need gokrb5 to tag a new release with a suitable bump in x/crypto or you as a consumer of us can add a replace directive in your application go.mod to force a new version to be used

PR #2268, opened 2022-07-14
Remove duplicates from topics

Smriti-OSS:
This PR is intended to fix the issues observed in https://github.com/Shopify/sarama/issues/2264.

Summary of the issue: Passing duplicate topics into the Consume function leads to uneven partition division between different consumers of the consumer group.

References: Java client handles this issue by always discarding duplicates - as it [collapses the collection of topics into a set](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L974).


Smriti-OSS:
@dnwe Could you please review the issue and PR.

PR #2317, opened 2022-08-11
fix(consumer): avoid pushing unrelated responses to paused children

pkoutsovasilis:
This PR addresses [this issue](https://github.com/Shopify/sarama/issues/2316).

Signed-off-by: Panos Koutsovasilis <koutsobill@hotmail.com>

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -944,6 +944,16 @@ func (bc *brokerConsumer) subscriptionConsumer() {
 
 		bc.acks.Add(len(bc.subscriptions))
 		for child := range bc.subscriptions {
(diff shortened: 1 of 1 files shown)


dnwe:
@pkoutsovasilis thanks — the change looks good, although I wonder if we need to add a functional test that exercises the Pause and Resume functionality as it feels like we have a gap in testing there

Is that something you might be available to look into?

dnwe:
@pkoutsovasilis happy for it to be handled separately — I'll go ahead and review and merge this fix. Thanks again!

PR #2327, opened 2022-09-06
Validate the `Config` when creating a mock producer/consumer

joewreschnig:
Normally this happens during creation of the `Client`, but for mock
interfaces there is no `Client`.


dnwe (approved):
LGTM, thanks for this

PR #2328, opened 2022-09-07
feat: support key in MockFetchResponse.

Skandalik:
While working with `MockBroker`, I found out that this mock doesn't support key in the message for `FetchRequest`. In my case having key mocked is essential as in the project I'm working on, we rely on the parsed key from Kafka message.
I added the support and added the test for this behavior.

### DIFF
--- a/consumer_test.go
+++ b/consumer_test.go
@@ -1,6 +1,7 @@
 package sarama
 
 import (
+	"bytes"
 	"errors"
 	"log"
 	"os"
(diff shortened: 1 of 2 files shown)


Skandalik:
I signed the CLA, but I can't rerun the GHA check.

Skandalik:
@dnwe Could you please review the PR? I don't know if I should create an issue too.

Issue #1999, opened 2021-08-19
Not handling `RefreshFrequency` being 0 correctly.


Jason5Lee:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka       | Go      |
|--------  |------------|--------|
| 1.26.1   | 2.13-2.8.0 | 1.16.7 |

##### Configuration

What configuration values are you using for Sarama and Kafka?
``` go
saramaConfig = sarama.NewConfig()
saramaConfig.Consumer.Return.Errors = true
saramaConfig.ChannelBufferSize = defaultBufferSize
saramaConfig.Net.MaxOpenRequests = defaultConnectionNum
saramaConfig.Metadata.RefreshFrequency = 0 // disable metadata fetching;
saramaConfig.Version = sarama.V1_1_1_0
```

##### Logs

<details><summary>logs: CLICK ME</summary>
<p>

```
2021/08/19 20:53:02 Initializing new client
2021/08/19 20:53:02 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
2021/08/19 20:53:02 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
2021/08/19 20:53:02 client/metadata fetching metadata for all topics from broker localhost:9092
2021/08/19 20:53:02 Connected to broker at localhost:9092 (unregistered)
2021/08/19 20:53:02 client/brokers registered new broker #0 at <internal domain>:9092
2021/08/19 20:53:02 Successfully initialized new client
2021/08/19 20:53:02 Initializing new client
2021/08/19 20:53:02 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
2021/08/19 20:53:02 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
2021/08/19 20:53:02 client/metadata fetching metadata for all topics from broker localhost:9092
2021/08/19 20:53:02 Connected to broker at localhost:9092 (unregistered)
2021/08/19 20:53:02 client/brokers registered new broker #0 at <internal domain>:9092
2021/08/19 20:53:02 Successfully initialized new client
2021/08/19 20:53:02 client/metadata fetching metadata for [rebuild] from broker localhost:9092
2021/08/19 20:53:02 client/coordinator requesting coordinator for consumergroup test from localhost:9092
2021/08/19 20:53:02 client/coordinator coordinator for consumergroup test is #0 (<internal domain>:9092)
[...]

PR #2329, opened 2022-09-07
fix: avoid panic when Metadata.RefreshFrequency is 0

Jacob-bzx:
In #1578 , `RefreshFrequency` is used to initialize a ticker to periodically check whether to trigger a rebalance.
However, once it is set to `0`, it would cause panic when calling `pause := time.NewTicker(c.config.Metadata.RefreshFrequency)`, as shown in #1999.

When `RefreshFrequency` is `0`, there is no `backgroundMetadataUpdater` to periodically refresh metadata. Thus, there is no need for `loopCheckPartitionNumbers` as well.


Jacob-bzx:
hi master @dnwe, I noticed one FVT failed and I found the error log was `*** Test killed with quit: ran too long (16m0s).`
I believe it's irrelevant to this commit. Could you help retry it? Thanks a lot!

dnwe (approved):
Thanks, good catch!

PR #1581, opened 2020-01-16
Fix deadlock in consumer group handleError

matthewloring:
The deadlock can happen if the heartbeat request runs out of retries while the consumer group session is being released. `consumerGroup.Consume` will acquire the consumer group lock and then call `consumerGroupSession.release` which will block waiting for the channel `hbDead` to be closed. If the `heartbeatLoop` then runs out of retries and calls `consumerGroup.handleError` the `handleError` call will block on the lock preventing `heartbeatLoop` from closing `hbDead`.

It is safe to remove the lock acquisition in `handleError` because the channel operations in the critical section are already thread safe.

I also moved the lock/unlock from `Close` into `leave` to more accurately reflect the resources being protected.

Example stack traces for the deadlock:
```
goroutine 1 [semacquire, 2 minutes]:
sync.runtime_SemacquireMutex(0xc00131aa74, 0xc0010b5500, 0x1)
        /usr/local/Cellar/go/1.13.3/libexec/src/runtime/sema.go:71 +0x47
sync.(*Mutex).lockSlow(0xc00131aa70)
        /usr/local/Cellar/go/1.13.3/libexec/src/sync/mutex.go:138 +0xfc
sync.(*Mutex).Lock(...)
        /usr/local/Cellar/go/1.13.3/libexec/src/sync/mutex.go:81
github.com/Shopify/sarama.(*consumerGroup).Close.func1()
        /Users/<redacted>/go/pkg/mod/github.com/!shopify/sarama@v1.25.0/consumer_group.go:123 +0x1f0
sync.(*Once).doSlow(0xc00131aa80, 0xc0045fb7c8)
        /usr/local/Cellar/go/1.13.3/libexec/src/sync/once.go:66 +0xe3
sync.(*Once).Do(...)
        /usr/local/Cellar/go/1.13.3/libexec/src/sync/once.go:57
github.com/Shopify/sarama.(*consumerGroup).Close(0xc00131aa20, 0x0, 0x0)
        /Users/<redacted>/go/pkg/mod/github.com/!shopify/sarama@v1.25.0/consumer_group.go:120 +0x7d
...

goroutine 468 [chan receive, 10 minutes]:
github.com/Shopify/sarama.(*consumerGroupSession).release.func1()
        /Users/<redacted>/go/pkg/mod/github.com/!shopify/sarama@v1.25.0/consumer_group.go:719 +0x97
sync.(*Once).doSlow(0xc0012f2664, 0xc0018f6d38)
[...]

PR #2331, opened 2022-09-09
fix: race condition(may panic) when closing consumer group

Jacob-bzx:
# Background

We encounter a data race issue when closing consumer group.
##### Versions

| Sarama | Kafka | Go |
|--------|-------|----|
| 1.36.0 | 2.8.1  | 1.17.7 |
##### Logs

<p>

```

WARNING: DATA RACE
Write at 0x00c000f448b0 by goroutine 750:
  runtime.closechan()
      /Users/xxxxxxxxx/.gvm/gos/go1.17.7/src/runtime/chan.go:355 +0x0
  github.com/Shopify/sarama.(*consumerGroup).Close.func1.1()
      /Users/xxxxxxxxx/go/pkg/mod/github.com/!shopify/sarama@v1.36.0/consumer_group.go:161 +0x44

Previous read at 0x00c000f448b0 by goroutine 1238:
  runtime.chansend()
      /Users/xxxxxxxxx/.gvm/gos/go1.17.7/src/runtime/chan.go:158 +0x0
  github.com/Shopify/sarama.(*consumerGroup).handleError()
      /Users/xxxxxxxxx/go/pkg/mod/github.com/!shopify/sarama@v1.36.0/consumer_group.go:568 +0x22d
  github.com/Shopify/sarama.newConsumerGroupSession.func1()
      /Users/xxxxxxxxx/go/pkg/mod/github.com/!shopify/sarama@v1.36.0/consumer_group.go:724 +0xb7
  github.com/Shopify/sarama.newConsumerGroupSession·dwrap·79()
      /Users/xxxxxxxxx/go/pkg/mod/github.com/!shopify/sarama@v1.36.0/consumer_group.go:726 +0x63

```
</p>
</details>

# Investigation Results
After investigation, we find the root cause.
* `handleError` is used to collect errors from `heartBeatLoop`, `partitionConsumer`, `offsetManager`, etc.
Some goroutines are also spawned for error collecting, including [example1](https://github.com/Shopify/sarama/blob/v1.36.0/consumer_group.go#L807-L812), [example1](https://github.com/Shopify/sarama/blob/v1.36.0/consumer_group.go#L1003-L1007), [example3](https://github.com/Shopify/sarama/blob/v1.36.0/consumer_group.go#L721-L726) ....
* If `c.config.Consumer.Return.Errors=true`, those errors will be sent to a collective channel `c.errors`.
* The order of `c.errors <- err ` and `close(c.errors)` is non-deterministic and could even cause panic theoretically. A possible panic circumstance is shown in the below flow -- after checking the consumer group is not closed, the goroutine switches. And `close(c.errors)` is called. When goroutine switches back, `c.errors <- err` could cause panic as an error is sent to a closed channel. 

image(here is a screenshot)

# Solution
To prevent deadlock issue raised in #1581, I just create another dedicated lock instead of reusing the current `c.lock` one.

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -86,9 +86,10 @@ type consumerGroup struct {
 	memberID        string
 	errors          chan error
 
-	lock      sync.Mutex
-	closed    chan none
-	closeOnce sync.Once
+	lock       sync.Mutex
+	errorsLock sync.RWMutex
+	closed     chan none
+	closeOnce  sync.Once
 
 	userData []byte
 }
@@ -156,10 +157,13 @@ func (c *consumerGroup) Close() (err error) {
 			err = e
 		}
 
-		// drain errors
 		go func() {
+			c.errorsLock.Lock()
+			defer c.errorsLock.Unlock()
 			close(c.errors)
 		}()
+
+		// drain errors
 		for e := range c.errors {
 			err = e
 		}
@@ -557,6 +561,8 @@ func (c *consumerGroup) handleError(err error, topic string, partition int32) {
 		return
 	}
 
+	c.errorsLock.RLock()
+	defer c.errorsLock.RUnlock()
 	select {
 	case <-c.closed:
 		// consumer is closed


dnwe (approved):
Thanks!

PR #2352, opened 2022-09-30
fix: support existing deprecated Rebalance.Strategy field usage

spongecaptain:
see https://github.com/Shopify/sarama/issues/2351

### DIFF
--- a/config.go
+++ b/config.go
@@ -794,12 +794,6 @@ func (c *Config) Validate() error {
 		return ConfigurationError("ReadCommitted requires Version >= V0_11_0_0")
 	}
 
-	// backward compatibility on Rebalance Strategy
-	if c.Consumer.Group.Rebalance.Strategy != nil {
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -794,6 +794,12 @@ func (c *Config) Validate() error {
 		return ConfigurationError("ReadCommitted requires Version >= V0_11_0_0")
 	}
 
+	// backward compatibility on Rebalance Strategy
```
Jacob-bzx, line 797: The code works, but I'm not sure if it's good to modify the config in `Validation` method. 

And btw there is no need to clear `c.Consumer.Group.Rebalance.GroupStrategies` here.
Because `c.Consumer.Group.Rebalance.Strategy` takes precedence of `c.Consumer.Group.Rebalance.GroupStrategies` in sarama (
https://github.com/Shopify/sarama/blob/v1.37.0/consumer_group.go#L443-L453 & https://github.com/Shopify/sarama/blob/v1.37.0/consumer_group.go#L333-L340)

dnwe:
Thanks!

PR #2371, opened 2022-10-27
feat: add metadata-age metric

auntan:


auntan:
Better not merge now, it records all topics in the cluster

PR #2378, opened 2022-11-06
fix(producer): return errors for every message in retryBatch to avoid producer hang forever

cch123:
This fixes https://github.com/Shopify/sarama/issues/2377


cch123:
CLA signed

dnwe (approved):
Thanks for investigating this issue and submitting the fix, this change looks good to me

dnwe:
@cch123 do you think you might be able to add a functional test that covers this scenario?

cch123:
> @cch123 do you think you might be able to add a functional test that covers this scenario?

let me think about it.. I think the backoff param also didn't work, and need some time to find out why...

cch123:
@dnwe PTAL

dnwe:
Thanks!

PR #2387, opened 2022-11-11
fix: add nil check while producer is retrying

hsweif:
The function `newHighWatermark` send fin message to the input channel; it might panic when the `pp.brokerProducer` is `nil`. I added a simple check to fix it. Please help double-check if there are any unwanted side effects when you have time. Thanks a lot!

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -613,6 +613,18 @@ func (pp *partitionProducer) backoff(retries int) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


hsweif:
I put it in the draft stage first as I'm still learning the codebase, will update and make it ready when I confirm there's no side effect.

dnwe (approved):

dnwe:
Thank you for your contribution, approved and merged

PR #2423, opened 2023-01-19
fix: remove NewConsumerFromClient method call in newConsumerGroup method

Lumotheninja:
fix the issue of consumergroup not closing properly when calling consumergroup.close()

currently, the NewConsumerGroup method calls the NewConsumerFromClient method which creates a nopclient, and thus this client cannot be closed properly

Screenshot 2023-01-19 at 5 28 15 PM
Screenshot 2023-01-19 at 5 24 18 PM
Screenshot 2023-01-19 at 5 24 35 PM

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -126,14 +126,14 @@ func newConsumerGroup(groupID string, client Client) (ConsumerGroup, error) {
 		return nil, ConfigurationError("consumer groups require Version to be >= V0_10_2_0")
 	}
 
-	c, err := newConsumer(client)
+	consumer, err := newConsumer(client)
 	if err != nil {
 		return nil, err
 	}
 
 	cg := &consumerGroup{
 		client:         client,
-		consumer:       c,
+		consumer:       consumer,
 		config:         config,
 		groupID:        groupID,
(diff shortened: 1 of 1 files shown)


consumer_group.go
```diff
@@ -126,14 +126,17 @@ func newConsumerGroup(groupID string, client Client) (ConsumerGroup, error) {
...
 
-	consumer, err := NewConsumerFromClient(client)
-	if err != nil {
-		return nil, err
+	c := &consumer{
+		client:          client,
+		conf:            client.Config(),
+		children:        make(map[string]map[int32]*partitionConsumer),
+		brokerConsumers: make(map[*Broker]*brokerConsumer),
+		metricRegistry:  newCleanupRegistry(client.Config().MetricRegistry),
```
dnwe, line 134: Thanks for catching this problem. Was there a specific reason why you in-lined the consumer struct initialisation rather than calling `newConsumer` as-is? We're just in one big `sarama` package so it doesn't matter that its not exported

```suggestion
	c, err := newConsumer(client)
	if err != nil {
		return nil, err
	}
```

Lumotheninja:
I think I am unable to change the email of past commiters, let me open a new PR

Issue #2444, opened 2023-03-02
Sarama reports error for non-existent topic even though topic-auto-create option is on


thangnt294:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka | Go |
|--------|-------|----|
| 3d819b2 | 2.3.1 | 1.20.1 |

##### Configuration

What configuration values are you using for Sarama and Kafka?
``` go
```

##### Logs

When filing an issue please provide logs from Sarama and Kafka if at all
possible. You can set `sarama.Logger` to a `log.Logger` to capture Sarama debug
output.

<details><summary>logs: CLICK ME</summary>
<p>

```
Error:          Received unexpected error:
                                kafka server: Request was for a topic or partition that does not exist on this broker
```

</p>
</details>

##### Problem Description
Sarama reports error when sending a message to a non-existent topic, even though topic-auto-create is enabled. This causes confusion since topic-auto-create is the user's intended behavior, so no error should be reported if topic auto creation is on. We should only report error when this option is off.

##### Solution
I've created a pull request for this. Please review.

dnwe:
@thangnt294 the issue is that whilst conf.Metadata.AllowAutoTopicCreation lets the broker know that the client permits it to auto-create the topic if it doesn't exist — the ability to auto-create is still governed by the server-side broker configuration. So if the broker has disabled auto-creation and we don't return ErrUnknownTopicOrPartition then Sarama will just appear to hang

PR #2445, opened 2023-03-02
not returning non-existent topic error when auto create topic is enabled

thangnt294:
this PR is intended to resolve the issue #2444


dnwe:
@thangnt294 the issue is that whilst `conf.Metadata.AllowAutoTopicCreation` lets the broker know that the client _permits_ it to auto-create the topic if it doesn't exist — the ability to auto-create is still governed by the server-side broker configuration. So if the broker has disabled auto-creation and we don't return ErrUnknownTopicOrPartition then Sarama will just appear to hang

PR #2447, opened 2023-03-03
Fix HighWaterMarkOffset of mocks partition consumer

gr8web:
Currently the message offset and the HighWaterMarkOffset differs by 2 and was not part of the test as it was changed in https://github.com/Shopify/sarama/pull/2078.


gr8web:
Changed HighWaterMarkOffset values in `TestConsumerHandlesExpectationsPausingResuming` which was reported as failed.

Issue #2421, opened 2023-01-18
Feature: Support for Kerberos authentication using a credentials cache.


mrogaski:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka | Go |
|--------|-------|----|
| 1.38.0 | any | any |

##### Problem Description

We would like to use Sarama in on platforms where Kerberos credential caches are exposed, but no password credentials or keytabs are available.

This feature would require a new GSSAPI auth type (e.g. KRB_CCACHE) that would result in `NewKerberosClient` returning a Kerberos client produced by the [NewFromCCache](https://pkg.go.dev/github.com/jcmturner/gokrb5/v8@v8.4.3/client#NewFromCCache) function.

mrogaski:
We're still interested in having this feature for use cases where a pure-Go client is needed.

dnwe:
Implemented by https://github.com/IBM/sarama/pull/2457

PR #2448, opened 2023-03-04
Added support for Kerberos authentication with a credentials cache.

mrogaski:
Implemented the support for Kerberos authentication using a credentials cache, as described in #2421. 
This adds a KRB5_CCACHE_AUTH value for GSSAPIConfig.AuthType.


mrogaski:
Will reopen to see if the CLA workflow completes successfully.

Issue #2443, opened 2023-03-01
Error when using DescribeConsumerGroups with cluster version 2.3.X


faillefer:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka | Go |
|--------|-------|----|
| 3.8 | 2.3.1 | 1.19.4 |

##### Configuration

What configuration values are you using for Sarama and Kafka?
``` go
```

##### Logs

```
EOF
```

##### Problem Description

When requesting for consumer groups on a kafka cluster with version 2.3.1, sarama systematically returns an error (EOF). On the server side I see those errors : 

```
Exception while processing request from  <MASKED> (kafka.network.Processor)
org.apache.kafka.common.errors.InvalidRequestException: Error getting request for apiKey: DESCRIBE_GROUPS, apiVersion: 4, connectionId: <MASKED>  listenerName: ListenerName(SASL_SSL), principal: User:<MASKED> 
Caused by: java.lang.IllegalArgumentException: Invalid version for API key DESCRIBE_GROUPS: 4
	at org.apache.kafka.common.protocol.ApiKeys.schemaFor(ApiKeys.java:319)
```
Sarama tries to use version 4 of DESCRIBE_GROUPS if the version of the server is >= to 2.3.0
```go
# From admin.go
		if ca.conf.Version.IsAtLeast(V2_3_0_0) {
			describeReq.Version = 4
		}
```
I think this version 4 has been introduced by KIP-345 and is only available from version 2.4.0

PR #2451, opened 2023-03-06
#2443 Use version 4 of DescribeGroupsRequest only if kafka broker version is > 2.4.0.0

faillefer:
Fixes issue #2443


dnwe:
@faillefer thanks, we'd also need to update this check [here](https://github.com/Shopify/sarama/blob/1015b4f7fe96e65eba69ccc5d51d687e05cc5792/describe_groups_request.go#L45-L51) which seems to be completely wrong at the moment.

dnwe (approved):
Thanks, LGTM

PR #2192, opened 2022-03-22
add support for aws mks iam

sethpollack:
fixes https://github.com/Shopify/sarama/issues/1985

usage would look like:

```
cfg.Net.SASL.Mechanism = sarama.SASLMechanism(sarama.SASLTypeAWSMSKIAM)
cfg.Net.SASL.AWSMSKIAM = sarama.AWSMSKIAMConfig{
    Region: "us-east-1",
}

```

PR #2465, opened 2023-04-17
Add support for IAM MSK auth

bpaquet:
This is an updated version of https://github.com/Shopify/sarama/pull/2192 from @sethpollack


dnwe:
@bpaquet as mentioned in the discussion on #2192 — it's a bit tricky to approve and merge this as-is 

> I wasn't aware that AWS had implemented their own custom SASL mechanism for this, rather than adopting one of the existing standards. Whilst I can understand that people using AWS would like this support, ideally we wouldn't want to a) introduce a direct dependency on aws-sdk-go-v2 for everyone or b) have service-specific configuration parameters etc. — this is also quite hard for us to regression test without provisioning a live service
> 
> I notice that [other kafka clients](https://github.com/segmentio/kafka-go/issues/661) had this support added using a pluggable "opt-in" sasl provider architecture. I wonder how much work it would be to achieve a similar result with Sarama? i.e., keeping the dependencies and configuration in a separate module path that people can opt-in to using by pulling it in as a dependency

The suggestion on the other PR was to look at making SCRAMClientGeneratorFunc flexible enough that the IAM MSK auth could be a pluggable addition for users that need it in their applications

bpaquet:
Thx for the feedback :)

I do not want to use the franz-go library, which does not handle all authent scheme. I will have a look on the opt-in solution.

longit644:
Hi everyone, IMHO, we could simply remove this validation:
https://github.com/Shopify/sarama/blob/9127f1c432c08b9a766fdb0931bfd0da80dddde0/config.go#L675-L678
to allow users to use any custom `SASL` mechanism. This way, users can implement their own `AWS_MSK_IAM` authentication mechanism or use one from another library. Sarama could also provide an example like the way we provided for [XDG SCRAM](https://github.com/Shopify/sarama/tree/main/examples/sasl_scram_client).

dnwe:
I think the nuance for AWS_MSK_IAM is that [SCRAMClientGeneratorFunc](https://github.com/Shopify/sarama/blob/9127f1c432c08b9a766fdb0931bfd0da80dddde0/config.go#L90-L92) doesn't currently pass along the broker address as a parameter and it seems that AWS expect the specific broker host address (minus the port) to be used for the HTTP auth request and to also be included in the json payload

It would probably be simplest to assume that current usage of SCRAMClientGeneratorFunc is low enough that just bumping the version to /v2 and adding the param (breaking compile compat) would be OK. Otherwise we'd have to add a new config name and deprecate the old one if we wanted to stay v1

bpaquet:
> longit644



> Hi, everyone, I'm trying to create a client using this code and I'm getting the following error:
> 
> ```
> [sarama] 2023/04/19 10:47:27 client/metadata fetching metadata for all topics from broker b-1.mycluster.kafka.us-east-1.amazonaws.com:9098
> [sarama] 2023/04/19 10:47:27 Error while performing SASL handshake b-1.mycluster.kafka.us-east-1.amazonaws.com:9098
> [sarama] 2023/04/19 10:47:27 Closed connection to broker b-1.mycluster.kafka.us-east-1.amazonaws.com:9098
> [sarama] 2023/04/19 10:47:27 client/metadata got error from broker -1 while fetching metadata: unexpected EOF
> ```
> 
> My code looks like this: `client, err := sarama.NewClient(strings.Split(kafkaConfig.ConnectionString, ","), config)` while in the config I set these parameters:
> 
> ```
> conf.Net.SASL.Enable = true
> conf.Net.SASL.Mechanism = IAMMechanism
> conf.Net.SASL.AWSMSKIAM = AWSMSKIAMConfig{Region: "us-east-1"}
> ```
> 
> Do you have any suggestions on how to solve this?

You can find a working version here: https://github.com/birdayz/kaf/compare/master...bpaquet:kaf:msk_iam_2. Are you sure your AWS creds are properly set? What is the result of `aws sts get-caller-identity`?

bokimilinkovic:
Hi, is there any update on this, 
Is it going to be merged, and is IAM authentication  possible with this kafka client ?

bpaquet:
The current plan on my side is to finish and merge this PR
<https://github.com/Shopify/sarama/pull/2466>, which will allow to provide an
external mechanism for MSK <https://github.com/bpaquet/awsIamMskForSarama>

jadhub:
Can anyone give an ETA? Very interested.

PR #2472, opened 2023-05-14
feat: return KError instead of errors in AlterConfigs and DescribeConfig

zhuliquan:
return KError instead of errors in AlterConfigs and DescribeConfig, It's help for check error type, like grpc error code.


dnwe (approved):
Thanks LGTM, apidiff shows the error interface is still satisfied so it should be safe to merge this without a major version bump 👍🏻

PR #2475, opened 2023-05-19
feat: add new error for MockDeleteTopicsResponse

javiercri:
Allow set Error for Delete Topic Mock Response
With this we can set up tests for error while deleting topics


javiercri:
I have signed the CLA!

dnwe (approved):

PR #2479, opened 2023-06-21
fix(example): check if msg channel is closed

ioanzicu:


dnwe:
@ioanzicu as per https://github.com/IBM/sarama/wiki/Frequently-Asked-Questions#why-am-i-getting-a-nil-message-from-the-sarama-consumer it should only be the case that you receive nil if the channel has been closed due to some error in the consumer

So we probably want to update the example from:

case message := <-claim.Messages():

to

case message, ok := <-claim.Messages():
    if !ok { … }
 
 pattern

ioanzicu:
Looks like I have to fix this MR :)

dnwe:
No worries, I've rebased your branch for you and the checks are running now

dnwe (approved):

PR #2480, opened 2023-06-28
feat: explore topic metadata as public method

harveyyue:
Explore topic metadata as public method instead of admin client


harveyyue:
Done with cla signed

PR #2482, opened 2023-06-29
Allow to implement a custom SASL mechanism and deprecate the previous mechanism

bpaquet:
@dnwe This is a follow up of https://github.com/Shopify/sarama/pull/2466

In this PR, I'm replacing the previous mechanisms `SCRAM-SHA-256` and `SCRAM-SHA-512` by a generic function `SCRAMClientExternal` which be used to implement custom auth mechanism, like AWS IAM MSK.

This is a breaking change in the config API, as we discussed in the previous PR.

`SCRAM-SHA-256` and `SCRAM-SHA-512` can be implemented with the new API (see `examples/sasl_scram_client/main.go`).


tooptoop4:
🦗🌵

dtjm:
@bpaquet I've added a couple of commits to fix the lint breakage as well as merge in the latest from the main branch if you would like to pull them into your branch: https://github.com/bpaquet/sarama/pull/1

bpaquet:
This PR seems to be useless now, because of https://github.com/aws/aws-msk-iam-sasl-signer-go :)

PR #2484, opened 2023-07-04
Use buffer pools for decompression

ronanh:
While profiling my application, I've noticed that Sarama spends a significant amount of time growing slices when it's decompressing messages:

growSlice

This patch uses an intermediate buffer from a  pool for decompression, to avoid these allocations.
When the final size is know, the result buffer can be allocated and copied.

After the patch:

withpool2

### DIFF
--- a/decompress.go
+++ b/decompress.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"compress/gzip"
 	"fmt"
-	"io"
 	"sync"
 
 	snappy "github.com/eapache/go-xerial-snappy"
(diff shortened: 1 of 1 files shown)


hindessm:
Looks very good to me:
```
before:
BenchmarkDecompression/lz4_2048_compressable-4          578866    9368 ns/op    8533 B/op   11 allocs/op
BenchmarkDecompression/lz4_2048_lessCompressable-4      830926    6521 ns/op    8506 B/op   11 allocs/op
BenchmarkDecompression/lz4_2048_compressable-16         804277    7391 ns/op    8544 B/op   11 allocs/op
BenchmarkDecompression/lz4_2048_lessCompressable-16    1221556    4904 ns/op    8538 B/op   11 allocs/op

after:
BenchmarkDecompression/lz4_2048_compressable-4          792148    7512 ns/op    2373 B/op    7 allocs/op
BenchmarkDecompression/lz4_2048_lessCompressable-4     1302768    4614 ns/op    2395 B/op    7 allocs/op
BenchmarkDecompression/lz4_2048_compressable-16         920294    5854 ns/op    2364 B/op    7 allocs/op
BenchmarkDecompression/lz4_2048_lessCompressable-16    1544532    3909 ns/op    2420 B/op    7 allocs/op
```

PR #2453, opened 2023-03-10
fix: prevent data race in balance strategy

napallday:
For every `BalanceStrategy`, it may be shared across different consumer groups in one instance. Considering every built-in balance strategy has several fields in the struct as shown in the below snippet, data race issues may happen when several consumer group rebalances simultaneously.
```
type balanceStrategy struct {
	coreFn func(plan BalanceStrategyPlan, memberIDs []string, topic string, partitions []int32)
	name   string
}
type stickyBalanceStrategy struct {
	movements partitionMovements
}
```

Even worse, I think this issue could cause
- panic
- some topics/partitions are not subscribed by the corresponding consumers, because the corresponding fields in the struct may be modified by other consumer groups

To verify this issue, a unit test was written, which fails when -race detection is enabled. 
```
func Test_stickyBalanceStrategy_Plan_data_race(t *testing.T) {
	for i := 0; i < 1000; i++ {
		go func(bs BalanceStrategy) {
			members := map[string]ConsumerGroupMemberMetadata{
				"m1": {
					Version: 3,
					Topics:  []string{"topic"},
				},
			}
			topics := map[string][]int32{
				"topic": {0, 1, 2},
			}
			_, _ = bs.Plan(members, topics)
		}(BalanceStrategySticky)
	}
}
```

To avoid this problem, it may be best not to use a singleton for each and every of BalanceStrategy. Instead, newly-added functions like `NewBalanceStrategyRange() BalanceStrategy ` are recommended so that every time an exclusive variable is assigned to each consumer group.

Fixes #2422

PR #2487, opened 2023-07-10
add lock for reassigning partition to new consumer

yitian108:
When I try to test peer to peer consume using sarama library, some codes snippet like this:

```go
func TestP2P(topic string) {
	if err := NewConsumer(topic, P2PConsumer1); err != nil {
		// some logs
	}

	if err := NewConsumer(topic，P2PConsumer2); err != nil {
		// some logs
	}
	
	stop := false
	count := 0
	go func() {
		for {
			if !stop {
				count++
				msg := "test p2p msg: " + strconv.Itoa(count)
				NewProducer(topic，[]byte(msg))
			}
			time.sleep(time.Second)			
		}
	}()

	select {}
}
```

Sometimes, running this code will report the panic 'concurrent map writes', and the location of panic points to the method 'addPartitionMovementRecord' in balance_strategy.go, so I add the lock for concurrently running the multiple consumers access the same new partition.

Please check it.

lockMap


yitian108:
hello @dnwe, I've modify some codes that only lock one map in the partitionMovements, and the commit of the branch lock-for-adding-partition is 9913274 , hopefully, the newer could be pass all checks.

dnwe:
@yitian108 are you using NewBalanceStrategySticky for each of your consumers rather than the deprecated shared global of BalanceStrategySticky? They should already have a unique partitionMovements instance if so:

https://github.com/IBM/sarama/blob/fb814087d8d76fe9f48fcb415f8118543a52e064/balance_strategy.go#L107-L112

yitian108:
Hi @dnwe , I used the following api for each consumer:

```go
config.Consumer.Group.Rebalance.Strategy = sarama.BalanceStrategySticky
```

Is this global of BalanceStrategySticky? If that, How should I use the NewBalanceStrategySticky at the starting of config kafka?

ps: I am using the samara version 1.37, which new version using the NewBalanceSTrategySticky?

dnwe:
It was `fix: prevent data race in balance strategy` by @napallday from https://github.com/IBM/sarama/pull/2453 included in the latest https://github.com/IBM/sarama/releases/tag/v1.40.0

yitian108:
Thanks @dnwe and @napallday, I'll try to use NewBalanceStrategyRange() instead to test the case.

yitian108:
Hi @dnwe @napallday, I try to use NewBalanceStrategySticky() as below
```go
	config := sarama.NewConfig()
	config.Consumer.Group.Rebalance.Strategy = sarama.NewBalanceStrategySticky()
```
However, the IDE GoLand prompts "'Strategy' is deprecated", it seems that I used the wrong way to achive the object, do you have another good way to implement the requirement?

napallday:
hi @yitian108, it's expected. From Sarama v1.37.0(more precisely PR #2339 ), `config.Consumer.Group.Rebalance.Strategy` is deprecated and `config.Consumer.Group.Rebalance.GroupStrategies` is preferred for supporting specifying multiple balance strategies.

yitian108:
Thanks @napallday , to my understand, I should  use like this `config.Consumer.Group.Rebalance.GroupStrategies = []sarama.BalanceStrategy{sarama.NewBalanceStrategySticky(), sarama.NewBalanceStrategyRoundRobin()}`, and if only one strategy would be applied, it looks like `config.Consumer.Group.Rebalance.GroupStrategies = []sarama.BalanceStrategy{sarama.NewBalanceStrategySticky()}`, right?

napallday:
correct! @yitian108

PR #2520, opened 2023-07-23
Add some retry logic to more admin client functions

hindessm:
Contrary to the discussion [here](https://github.com/IBM/sarama/pull/1617#discussion_r385132244), you can use the retryOnError function when you are returning results by using a closure to capture the results. This PR wraps a few more functions with retry logic.

I stopped before doing any more in part because I was distracted by the inconsistent use of 'Acls', 'ACLs', etc in variables and function names.  :-) (Something else for v2 perhaps).

I also wondered about the fact that the retryOnError function currently only checks for isNotControllerError *and* many of the functions that it wraps also check the error since they need to refresh the controller. Perhaps it would be better to make simpler retryOnError function like:

```
func (ca *clusterAdmin) retryOnError(fn func() (bool, error)) error {
  ...
  isRetryable, err := fn()
  ...
}
```

so that the functions can check the error once and return `true, err`. (Or `false, err` for all other cases.)

Or perhaps there is a better way? Suggestions welcome.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -556,13 +556,20 @@ func (ca *clusterAdmin) ListPartitionReassignments(topic string, partitions []in
 
 	request.AddBlock(topic, partitions)
 
-	b, err := ca.Controller()
-	if err != nil {
-		return nil, err
-	}
-	_ = b.Open(ca.client.Config())
+	var rsp *ListPartitionReassignmentsResponse
+	err = ca.retryOnError(isErrNoController, func() error {
+		b, err := ca.Controller()
+		if err != nil {
+			return err
+		}
+		_ = b.Open(ca.client.Config())
 
-	rsp, err := b.ListPartitionReassignments(request)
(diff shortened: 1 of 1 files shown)


dnwe (approved):
Changes look good to me, but we should probably refresh our memories and re-review which of these requests:

1. absolutely need to be sent to the controller
2. can be sent to any broker but are ultimately proxied to the controller (so we may as well send them to the controller directly if it’s trivial to find, but could fallback to anyBroker 
3. can be sent to any broker and are directly answered by any broker

dnwe:
Somewhat relatedly, in kraft mode there is no longer the concept of a dedicated controller node and they try and force you to sending requests across any broker https://github.com/IBM/sarama/issues/2521#issuecomment-1649151606

dnwe:
Still, currently the code does fetch the controller so using the existing common retry mechanism makes sense

dnwe (approved):

PR #2528, opened 2023-07-28
fix: correct unsupported version check

hindessm:
Version 8 is unsupported since the existing code doesn't support the fields IncludeClusterAuthorizedOperations and IncludeTopicAuthorizedOperations required by that version.


dnwe (approved):

dnwe:
Raised https://github.com/IBM/sarama/issues/2529 for the flakey test and re-running

PR #2533, opened 2023-07-29
extend throttling metric scope

hindessm:
This is PR is refactoring the current throttling metric support with a view to implementing throttling.

Currently, there were three different throttle time field implementations:
1. a `ThrottleTimeMs int32` field - for example, on `MetadataResponse`,
2. a `ThrottleTime int32` field - for example, on `LeaveGroupResponse`, and
3. a `ThrottleTime time.Duration` field - for example, on `ProduceResponse`.

It would be nice to make these more consistent but, since they are public fields, that would need to be a v2 change.
Therefore, in order to make progress now, this PR implements a common, private `throttleTime() time.Duration` method on all of the responses covered by the above three field types. The `updateThrottleMetric` implementation is then refactored to apply more generally to responses that have the new common method (via a trivial new interface `throttleSupport`).

This should make it possible to implement throttling support  - i.e. delaying future requests in accordance with KIP-219.

Two typos are also fixed - one related and one randomly spotted while in `broker.go`.

### DIFF
--- a/acl_create_response.go
+++ b/acl_create_response.go
@@ -63,6 +63,10 @@ func (c *CreateAclsResponse) requiredVersion() KafkaVersion {
 	return V0_11_0_0
 }
 
+func (r *CreateAclsResponse) throttleTime() time.Duration {
+	return r.ThrottleTime
+}
+
 // AclCreationResponse is an acl creation response type
 type AclCreationResponse struct {
 	Err    KError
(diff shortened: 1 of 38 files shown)


dnwe (approved):
LGTM

dnwe:
Raised https://github.com/IBM/sarama/issues/2534 for flake and rerunning

PR #2538, opened 2023-08-01
feat: add isValidVersion to protocol types

dnwe:
The intention here is to add a common `isValidVersion` to all the protocol types so that we can check that a request version value is supported by the protocol encoder/decoder implementation before sending it (hence avoiding malformed requests on the broker side).

Additionally in this PR fixup the existing `requiredVersion` checks where they were incorrect and/or add missing values, also ensuring the request and response implementations were in agreement.

Finally add some testing in this area and in a few places start using newer version numbers where the config version value allows.

### DIFF
--- a/consumer_metadata_response.go
+++ b/consumer_metadata_response.go
@@ -54,7 +54,7 @@ func (r *ConsumerMetadataResponse) encode(pe packetEncoder) error {
 	}
 
 	tmp := &FindCoordinatorResponse{
-		Version:     0,
+		Version:     r.Version,
 		Err:         r.Err,
 		Coordinator: r.Coordinator,
 	}
@@ -79,9 +79,16 @@ func (r *ConsumerMetadataResponse) headerVersion() int16 {
 }
 
 func (r *ConsumerMetadataResponse) isValidVersion() bool {
-	return r.Version == 0
+	return r.Version >= 0 && r.Version <= 2
 }
 
 func (r *ConsumerMetadataResponse) requiredVersion() KafkaVersion {
-	return V0_8_2_0
+	switch r.Version {
+	case 2:
+		return V2_0_0_0
+	case 1:
+		return V0_11_0_0
+	default:
+		return V0_8_2_0
+	}
 }
--- a/consumer_metadata_response.go
+++ b/consumer_metadata_response.go
@@ -7,6 +7,7 @@ import (
 
 // ConsumerMetadataResponse holds the response for a consumer group meta data requests
 type ConsumerMetadataResponse struct {
+	Version         int16
 	Err             KError
 	Coordinator     *Broker
 	CoordinatorID   int32  // deprecated: use Coordinator.ID()
@@ -70,11 +71,15 @@ func (r *ConsumerMetadataResponse) key() int16 {
 }
 
 func (r *ConsumerMetadataResponse) version() int16 {
-	return 0
+	return r.Version
 }
 
 func (r *ConsumerMetadataResponse) headerVersion() int16 {
 	return 0
+}
+
+func (r *ConsumerMetadataResponse) isValidVersion() bool {
+	return r.Version == 0
 }
 
 func (r *ConsumerMetadataResponse) requiredVersion() KafkaVersion {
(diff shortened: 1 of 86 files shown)


offset_request.go
```diff
@@ -137,6 +137,10 @@ func (r *OffsetRequest) headerVersion() int16 {
 	return 1
 }
 
+func (r *OffsetRequest) isValidVersion() bool {
+	return r.Version <= 2
```
hindessm: > Version 3, 4, and 5 are the same as version 2

```suggestion
	return r.Version <= 5
```

hindessm (commented):
Not sure if all my comments made it but I can't see them in my refreshed review page now.

create_topics_request.go
```diff
@@ -83,6 +83,10 @@ func (r *CreateTopicsRequest) headerVersion() int16 {
 	return 1
 }
 
+func (c *CreateTopicsRequest) isValidVersion() bool {
+	return c.Version >= 0 && c.Version <= 2
```
hindessm, line 87: Version 3 seems to only be a server behaviour change?

```suggestion
	return c.Version >= 0 && c.Version <= 3
```

hindessm: Actually I think version 4 only changes semantics not fields and we would support it?

delete_topics_request.go
```diff
@@ -42,6 +42,10 @@ func (d *DeleteTopicsRequest) headerVersion() int16 {
 	return 1
 }
 
+func (d *DeleteTopicsRequest) isValidVersion() bool {
+	return d.Version >= 0 && d.Version <= 1
```
hindessm, line 46: Versions 0, 1, 2, and 3 are the same and the responses only include semantic and error response differences?

```suggestion
	return d.Version >= 0 && d.Version <= 3
```

hindessm (commented):
When adding Version fields, we should probably fix the version functions to return them?

create_topics_response.go
```diff
@@ -74,6 +74,10 @@ func (c *CreateTopicsResponse) headerVersion() int16 {
 	return 0
 }
 
+func (c *CreateTopicsResponse) isValidVersion() bool {
+	return c.Version >= 0 && c.Version <= 2
```
hindessm, line 78: I think we support version 3 which simply makes the NumPartitions and ReplicationFactor in the request optional - with -1 which the code would allow.

delete_records_request.go
```diff
@@ -74,13 +75,17 @@ func (d *DeleteRecordsRequest) key() int16 {
...
-	return 0
+	return d.Version
 }
 
 func (d *DeleteRecordsRequest) headerVersion() int16 {
 	return 1
 }
 
+func (d *DeleteRecordsRequest) isValidVersion() bool {
+	return d.Version == 0
```
hindessm, line 86: Version 1 is just response throttling behaviour change so I think we support that too.

describe_log_dirs_request.go
```diff
@@ -82,6 +82,10 @@ func (r *DescribeLogDirsRequest) headerVersion() int16 {
 	return 1
 }
 
+func (r *DescribeLogDirsRequest) isValidVersion() bool {
+	return r.Version == 0
```
hindessm, line 86: Version 1 is just response throttling behaviour change so I think we support that too.

consumer_metadata_response.go
```diff
@@ -70,13 +71,17 @@ func (r *ConsumerMetadataResponse) key() int16 {
...
 func (r *ConsumerMetadataResponse) version() int16 {
-	return 0
+	return r.Version
 }
 
 func (r *ConsumerMetadataResponse) headerVersion() int16 {
 	return 0
 }
 
+func (r *ConsumerMetadataResponse) isValidVersion() bool {
```
hindessm, line 81: This should just call `FindCoordinatorRequest.isValidVersion()` ?

produce_response.go
```diff
@@ -175,6 +175,10 @@ func (r *ProduceResponse) headerVersion() int16 {
 	return 0
 }
 
+func (r *ProduceResponse) isValidVersion() bool {
+	return r.Version == 0
```
hindessm, line 179: This doesn't match the request which permits up to version 7

hindessm (approved):
LGTM

PR #2539, opened 2023-08-01
fix(consumer): loopCheckPartitionNumbers may not cause rebalance after adding new partitions

### LINKED ISSUE #2460: loopCheckPartitionNumbers function some problem
image
When loopCheckPartitionNumbers function go to line 627,the partition number has changed，so the consumer no longer rebalance,should oldTopicToPartitionNum get when the session was established instead of  in the goroutinue?

napallday:
fix #2460

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -79,12 +79,13 @@ type ConsumerGroup interface {
 type consumerGroup struct {
 	client Client
 
-	config          *Config
-	consumer        Consumer
-	groupID         string
-	groupInstanceId *string
-	memberID        string
-	errors          chan error
+	config               *Config
+	consumer             Consumer
+	groupID              string
+	groupInstanceId      *string
+	memberID             string
+	errors               chan error
+	startTopicPartitions map[string][]int32
 
 	lock       sync.Mutex
 	errorsLock sync.RWMutex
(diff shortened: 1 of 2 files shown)


napallday:
I know there is a data race in the newly-added test file
```
func (h *loopCheckerHandler) Setup(s ConsumerGroupSession) error {
	if len(s.Claims()["my-topic"]) == 1 {
		if err := h.group.client.RefreshMetadata("my-topic"); err != nil {
			h.Error(err)
		}
		// change RefreshFrequency so that loopCheckPartitionNumbers won't return directly
***		h.group.client.(*nopCloserClient).Client.(*client).conf.Metadata.RefreshFrequency = 5 * time.Millisecond
		go h.group.loopCheckPartitionNumbers([]string{"my-topic"}, s.(*consumerGroupSession))
	} else {
		h.cancel()
	}
	return nil
}
```

The added test will fail in the old versions without a race detector, but I haven't managed to write a test to reproduce the issue without a data race. 

If it's difficult to resolve the data race, how about we just remove the unit test if the changes look good to you? @dnwe @hindessm

napallday:
I found another critical bug in our company, which is related to this issue.
I'll open another MR to fix both of them.

PR #2545, opened 2023-08-02
Fix printing of final metrics

prestona:
For short test runs `kafka-producer-performance` may not print any metrics at all. This is fixed by moving the final call to `printMetrics` just before closing the producer. Calling `printMetrics` after the producer is closed does nothing as the producer metrics are unregistered as part of the close.


hindessm (approved):
LGTM.

PR #2546, opened 2023-08-03
fix(producer): use newer ProduceReq as appropriate

dnwe:


produce_set.go
```diff
@@ -141,8 +141,13 @@ func (ps *produceSet) buildRequest() *ProduceRequest {
 			req.TransactionalID = &ps.parent.conf.Producer.Transaction.ID
 		}
 	}
-
-	if ps.parent.conf.Producer.Compression == CompressionZSTD && ps.parent.conf.Version.IsAtLeast(V2_1_0_0) {
+	if ps.parent.conf.Version.IsAtLeast(V1_0_0_0) {
+		req.Version = 5
+	}
```
hindessm, line 146: slightly pedantic but these could be `else` cases.

hindessm (approved):

PR #2548, opened 2023-08-03
fix(proto): ensure req+resp requiredVersion match

dnwe:
Add a test to ensure that the requiredVersion of the request and response pairs matches. Accordingly fix DescribeConfigs which had a V1_0_0_0 instead of V1_1_0_0 bug for V1 in the Response


hindessm (approved):
LGTM

PR #2551, opened 2023-08-03
chore(proto): permit DeleteGroupsRequest V1

dnwe:
This is identical in format to V0, but just influences the broker's response behaviour


hindessm (approved):
LGTM

PR #2553, opened 2023-08-03
fix(proto): correct JoinGroup usage for wider version range

dnwe:


consumer_group.go
```diff
@@ -433,7 +433,24 @@ func (c *consumerGroup) joinGroupRequest(coordinator *Broker, topics []string) (
...
+	if c.config.Version.IsAtLeast(V0_11_0_0) {
+		req.Version = 2
+	}
+	if c.config.Version.IsAtLeast(V0_11_0_0) {
+		req.Version = 2
+	}
+	if c.config.Version.IsAtLeast(V2_0_0_0) {
+		req.Version = 3
+	}
+	// XXX: protocol states "Starting from version 4, the client needs to issue a
```
hindessm, line 445: the `if` below this comment sets `req.Version = 5` so this must work or already be broken?

hindessm (approved):

consumer_group.go
dnwe: Yes, although I’ve not personally tested the V5 and static membership stuff so I didn’t want to introduce any breakage until I’d tested it

PR #2556, opened 2023-08-03
fix(proto): use full range of MetadataRequest

dnwe:
Even though most of these are identical, we may as well match up correctly.

### DIFF
--- a/metadata_request.go
+++ b/metadata_request.go
@@ -17,6 +17,10 @@ func NewMetadataRequest(version KafkaVersion, topics []string) *MetadataRequest 
 		m.Version = 6
 	} else if version.IsAtLeast(V1_0_0_0) {
 		m.Version = 5
+	} else if version.IsAtLeast(V0_11_0_0) {
+		m.Version = 4
+	} else if version.IsAtLeast(V0_10_1_0) {
+		m.Version = 2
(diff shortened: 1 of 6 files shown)


metadata_request.go
```diff
@@ -17,6 +17,10 @@ func NewMetadataRequest(version KafkaVersion, topics []string) *MetadataRequest
 		m.Version = 6
 	} else if version.IsAtLeast(V1_0_0_0) {
 		m.Version = 5
+	} else if version.IsAtLeast(V0_11_0_0) {
```
hindessm, line 20: DYI: I prefer these `else if` highest first which makes me dislike those `if {}; if {}` lowest first additions I queried yesterday even more.

hindessm (approved):

metadata_request.go
dnwe: @hindessm I agree it's nicer and it works here because there's no mandatory additional fields, just the version number being set.

However, In some of those other cases it (e.g.,) needs a mandatory field included from V4 upward, so an additive approach works better.

One additional thing I _did_ like about Metadata was that it puts the version choice logic alongside the proto definition rather than within client.go / broker.go / consumer.go etc.

Issue #2430, opened 2023-02-03
Crc32 Partitioning Scheme Incompatible with librdkafka


csm8118:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka | Go |
|--------|-------|----|
| v1.29.0 |  |1.15 |

### Problem Description
There is a difference between how sarama partitions when using the Crc32 hash and what is done in librdkafka (and ruby-kafka).    It looks like this issue was reported some time ago but nothing came of it (https://github.com/Shopify/sarama/issues/1213).  The main issue is that librdkafka/ruby-kafka treat the hashed value from Crc32 as an unsigned value, whereas sarama casts it to a signed value.  However, when using other hashing algorithms, the signed cast is necessary.

- librdkafka
  - [rdkafka_msg.c](https://github.com/confluentinc/librdkafka/blob/a83cadf5eab3a43f7f0d3dd09f5b1a3e9e88312f/src/rdkafka_msg.c#L1049)
    - `return rd_crc32(key, keylen) % partition_cnt;`
- ruby-kafka
  -  [crc32_hash.rb](https://github.com/zendesk/ruby-kafka/blob/62a0cf9c44b0c05a4f98c4ec073ae5f04f4194e0/lib/kafka/crc32_hash.rb) and [partitioner.rb](https://github.com/zendesk/ruby-kafka/blob/62a0cf9c44b0c05a4f98c4ec073ae5f04f4194e0/lib/kafka/partitioner.rb#L36)
     - `@digest.hash(key) % partition_count` where `@digest.hash` is defined as `Zlib.crc32(value)`
- sarama
  -  [partitioner.go](https://github.com/Shopify/sarama/blob/main/partitioner.go#L203)
    - `partition = int32(p.hasher.Sum32()) % numPartitions`
 
Here are some sample inputs that show the difference:
```ruby
#ruby: key of "SheetJS", num partitions: 100 -> partition 26
 DEV [8] pry(main)> Zlib.crc32('SheetJS') % 100
=> 26
```

```go
//go: key of "SheetJS", num partitions: 100 -> partition 70
sc := sarama.NewConfig()
sc.Producer.Partitioner = sarama.NewCustomHashPartitioner(crc32.NewIEEE)
partitioner := sc.Producer.Partitioner("test")

msg := &sarama.ProducerMessage{
	Key: sarama.StringEncoder("SheetJS"),
}
[...]

Issue #1213, opened 2018-11-14
Partition difinite compatibility with librdkafka


DeathKR:
##### Versions

Sarama Version: All
Kafka Version: All
Go Version: All

##### Problem Description

We are using Sarama for golang and Confluentinc/confluent-kafka-dotnet (bindings to librdkafka) for C#. We see a diference in partition definition. Librdkafka use crc32, i wrote CustomHashPartitioner, code:

`
	crc32q := crc32.MakeTable(crc32.IEEE)

	hasher:= func() hash.Hash32 {
		return crc32.New(crc32q)
	}

	dict := map[string]int{
		"593b1335-381e-4281-b339-6c85cf06266a": 5,
		"53aaab45-67b3-468a-a9a7-b74c46868bd1": 0,
		"f63d51ca-e9c9-4703-987b-8d83c1efdf20": 7,
		"d4279299-34c6-40a4-8796-ac6ece9385cb": 8,
		"66b91fdb-f1a0-458b-a8c6-cde77465c4f3": 4,
		"d4eeeb1b-8c45-4250-9c2c-131b38208ee8": 5,
	}

	for key,v := range dict {
		partitioner := sarama.NewCustomHashPartitioner(hasher)("topic")
		msg := sarama.ProducerMessage{}
		msg.Key = sarama.StringEncoder(key)

		crc := crc32.Checksum([]byte(key), crc32q)
		resu := (crc ) % 9

		choice, _ := partitioner.Partition(&msg, 9)
		fmt.Println("[key]:",key, "[librdkafka]:", v, "[sarama]: ", choice, "[checksum % num]:", resu, crc)
	}
`

And out:

`
[key]: 66b91fdb-f1a0-458b-a8c6-cde77465c4f3 [librdkafka]: 4 [sarama]:  4 [checksum % num]: 4 280646581
[key]: d4eeeb1b-8c45-4250-9c2c-131b38208ee8 [librdkafka]: 5 [sarama]:  8 [checksum % num]: 5 2542580951
[key]: 593b1335-381e-4281-b339-6c85cf06266a [librdkafka]: 5 [sarama]:  5 [checksum % num]: 5 482296334
[key]: 53aaab45-67b3-468a-a9a7-b74c46868bd1 [librdkafka]: 0 [sarama]:  0 [checksum % num]: 0 567591138
[...]

PR #2560, opened 2023-08-04
bug: implement unsigned modulus for partitioning with crc32 hashing

csm8118:
See https://github.com/IBM/sarama/issues/2430 for a discussion of the issue.  Relevant items from that issue are copied here for visibility:

### Problem Description
There is a difference between how sarama partitions when using the Crc32 hash and what is done in librdkafka (and ruby-kafka). It looks like this issue was reported some time ago but nothing came of it (https://github.com/IBM/sarama/issues/1213). The main issue is that librdkafka/ruby-kafka treat the hashed value from Crc32 as an unsigned value, whereas sarama casts it to a signed value. However, when using other hashing algorithms, the signed cast is necessary.

### DIFF
--- a/partitioner.go
+++ b/partitioner.go
@@ -2,6 +2,7 @@ package sarama
 
 import (
 	"hash"
+	"hash/crc32"
 	"hash/fnv"
 	"math/rand"
 	"time"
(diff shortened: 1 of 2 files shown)


dnwe (approved):
Thanks, LGTM 👍🏻

PR #2563, opened 2023-08-05
fix(consumer): cannot automatically fetch newly-added partitions unless restart

### LINKED ISSUE #2460: loopCheckPartitionNumbers function some problem

### LINKED ISSUE #2562: consumer: consumer group cannot automatically fetch newly-added partitions
##### Versions

| Sarama | Kafka | Go |
|--------|-------|----|
| v1.40.1  |  |  |

##### Configuration

``` go
```

##### Logs

When filing an issue please provide logs from Sarama and Kafka if at all
possible. You can set `sarama.Logger` to a `log.Logger` to capture Sarama debug
output.

<p>

```
```

</p>
</details>

##### Problem Description
The fix #1525 hasn't completely fixed the problem.

And the issue can be reproduced easily(A functional test is provided in the fix #2563):
1. Set `config.Metadata.RefreshFrequency = 10 * time.Second`(to trigger the checking logic in `loopCheckPartitionNumbers` more frequently)
1. Two consumer `c1` and `c2` subscribe to `t1p0` and `t2p0` **respectively** in the same consumer group.
2. Let's say if `c1` is the leader, after adding a new partition in `t2p1`,  the new partition cannot be fetched until restart.

The root cause is that:
1. Although `loopCheckPartitionNumbers` in `c2` will detect the new partition `t2p1` and try to send `JoinGroupRequest` to the server, the server won't trigger a rebalance because `c2` is not the leader, and will just return the previous assignment to `c2`.
2. Although `c1` is the leader, goroutine  `loopCheckPartitionNumbers`  still won't exit to trigger a rebalance, because it will only check the **topics it subscribes** to instead of **all the subscribed topics** in the consumer group.

napallday:
fix #2562 

A functional test is provided as well, which will fail in the old versions. (if u simply copy the new test to the older version, plz also remember to copy [this line](https://github.com/IBM/sarama/pull/2563/files#diff-6dc37cfd96e85af03c7f31398c791c686048539779e71fee06011dcd3d866fafR409))

btw, this also fixes #2460

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -210,11 +210,6 @@ func (c *consumerGroup) Consume(ctx context.Context, topics []string, handler Co
 		return err
 	}
 
-	// loop check topic partition numbers changed
-	// will trigger rebalance when any topic partitions number had changed
-	// avoid Consume function called again that will generate more than loopCheckPartitionNumbers coroutine
-	go c.loopCheckPartitionNumbers(topics, sess)
-
 	// Wait for session exit signal
 	<-sess.ctx.Done()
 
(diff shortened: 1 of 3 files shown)


dnwe (approved):
Thanks! Changes and explanation look good to me

PR #2599, opened 2023-08-15
chore: use go install for installing sarama tools

vigith:
install using `go get` is deprecated. https://go.dev/doc/go-get-install-deprecation


dnwe (approved):
Thanks. Tweaked to add @latest but otherwise LGTM

Issue #2359, opened 2022-10-07
consider klauspost/compress for gzip?


shermanCRL:
##### Versions

| Sarama | Kafka | Go |
|--------|-------|----|
| 610514edec1825240d59b62e4d7f1aba4b1fa000 |  |1.19.1  |

##### Problem Description

Not a problem, but wondering if Sarama has considered using klauspost/compress for gzip. Looking around, Sarama appears still to use the stdlib gzip, though has klauspost for other algos. klauspost claims between a 1.5x-7x improvement, and appears to use the Go stdlib tests.

I swapped it in on [this branch](https://github.com/Shopify/sarama/compare/main...shermanCRL:sarama:klauspost-gzip), and got the following two errors:

```
--- FAIL: TestMessageEncoding (0.01s)
    request_test.go:19: Encoding empty gzip failed
        got  [196 46 92 177 0 1 255 255 255 255 0 0 0 20 31 139 8 0 0 9 110 136 0 255 3 0 0 0 0 0 0 0 0 0] 
        want [132 99 80 148 0 1 255 255 255 255 0 0 0 23 31 139 8 0 0 0 0 0 0 255 1 0 0 255 255 0 0 0 0 0 0 0 0]
--- FAIL: TestRecordBatchEncoding (0.00s)
    request_test.go:19: Encoding gzipped record failed
        got  [0 0 0 0 0 0 0 0 0 0 0 95 0 0 0 0 2 231 74 206 165 0 1 0 0 0 0 0 0 1 88 141 205 89 56 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1 31 139 8 0 0 9 110 136 0 255 0 21 0 234 255 40 0 10 0 8 1 2 3 4 6 5 6 7 2 6 8 9 10 4 11 12 3 0 173 201 88 103 21 0 0 0] 
        want [0 0 0 0 0 0 0 0 0 0 0 94 0 0 0 0 2 159 236 182 189 0 1 0 0 0 0 0 0 1 88 141 205 89 56 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1 31 139 8 0 0 0 0 0 0 255 210 96 224 98 224 96 100 98 102 97 99 101 99 103 98 227 224 228 98 225 230 1 4 0 0 255 255 173 201 88 103 21 0 0 0]
```

[...]

PR #2600, opened 2023-08-16
feat(gzip): switch to klauspost/compress gzip

dnwe:
As suggested in #2359, use klauspost/compress for our gzip implementation as we're already using the module for other compression algorithms and benchmarks show 1.5x-7x improvement

### DIFF
--- a/compress.go
+++ b/compress.go
@@ -2,11 +2,11 @@ package sarama
 
 import (
 	"bytes"
-	"compress/gzip"
 	"fmt"
 	"sync"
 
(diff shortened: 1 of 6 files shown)


klauspost:
@dnwe If possible you shouldn't do testing against compressed output. There is sure to be changes to the output - which would mean an upgrade would break your tests again.

dnwe:
@klauspost yes I'm not really sure what the intention of that unittest was tbh, but for now I just kept it compatible

Issue #1858, opened 2020-12-16
Support for KIP-429: Kafka Consumer Incremental Rebalance Protocol


chickensoup:
Kafka supports consumer incremental cooperative rebalance protocal starting from version 2.4.0, and I'm wondering is there any plan for Sarama client to support this protocol?

[KIP-429](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol)

ilyakaznacheev:
I just want to bump this issue from closing. Cooperative rebalance support would be very useful to work with different deployment kinds.

lizthegrey:
Definitely would also help our issues at Honeycomb where we've had to stop and start all workers at once rather than rolling restart to update

past-urbanist:
Now Apache Kafka 3.0 has used `CooperativeStickyAssignor` as default assignor instead of `RangeAssignor`. It seems Sarama has fallen far behind it. It is much more efficient in `CooperativeStickyAssignor`.

ilyakaznacheev:
Still waiting for it!

gofish:
Incremental rebalance is describe by [KIP-429](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol).

dnwe:
Thanks to @napallday we have a work-in-progress implementation of the incremental cooperate rebalance protocol. If you'd like to test out the changes they are available under PR https://github.com/IBM/sarama/pull/2608

dberardo-com:
any ETA for this feature ?

alicejgibbons:
Is this still be actively worked on? @dnwe would be great to have an update 🙏

dnwe:
I think @napallday ran out of free time to finish off the PR and hasn’t been able to return to it. I’ll try to resurrect what was started and finish it off this summer

rahulpoddar-fyndna:
support for CooperativeStickyAssignor strategy would help  in reducing rebalance time and resource optimization. Having this support will be helpful in critical applications transactions.

olitomlinson:
+1 for this!

iutx:
+1 for this!

dlopezby93:
+1!

dnwe:
FYI I have been gradually working through this in a series of independent PRs

PR #2608, opened 2023-08-20
feat(consumer): incremental cooperative balance strategy (KIP-429)

napallday:
## Incremental Cooperative Protocol POC Proposal

I'm excited to present this Proof of Concept (POC) version that supports the incremental cooperative protocol as outlined in [KIP-429](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol). The proposed feature addresses a long-standing community demand (#1858). I've conducted several manual tests, and the results have been promising. (Can run [examples](https://github.com/IBM/sarama/pull/2608/files#diff-afb249c8d04161977383a3baaadd0a55adb6411984eeb5d291e4e8b574d8fc57) for a try)

### Major Changes

The existing design in Sarama tightly couples with the EAGER protocol, particularly evident in `ConsumerGroupSession`. In order to accommodate the new `COOPERATIVE` rebalance protocol, I've introduced the `ConsumerGroupHandlerV2` interface and a new method called `ConsumeV2`. 

### Seeking Community Guidance

I'm eager to align this work with the community's future direction. Two possible paths are under consideration:

1. **Major Version Update Plan:** If Sarama intends to increment the major version while introducing support for the `COOPERATIVE` rebalance protocol, it would allow us to streamline our implementation. We could eliminate `ConsumerGroupSession`-related coded from versions `v2.x.x`, focusing solely on the new approach.

2. **Maintaining Compatibility:** Alternatively, if a major version update isn't planned, I'll take steps to enhance future maintainability. This involves refactoring common logic within the `Consume` and `ConsumerV2` methods, paving the way for smoother maintenance.

### Upcoming Work

In the pipeline, I plan to incorporate the following enhancements:

- [ ] Integration of metrics for the new protocol.
- [ ] Implementation of unit tests, functional tests, and benchmark tests to ensure reliability.
- [x] Validation of `Pause` and `Resume` functionality within the new protocol.
- [x] Verification for customized cooperative balance strategy implementation.

I'm looking forward to the community's insights and feedback as we work towards a more efficient and feature-rich Sarama.


prestona:
@napallday - I'm exited to see that cooperative re-balancing is making its way into Sarama. I think there's a few places where we could use it. I've had a quick play with the consumergroup_cooperative example and here's what I found:
1. Multiple co-operative Sarama consumers in the same group work as expected. I tried starting / stopping individual consumers and each time other consumers in the group were revoked/assigned partitions appropriately without a complete re-assignment of all of the topic's partitions. Sometimes `Setup(...)` is invoked with an empty map of newly assigned partitions, I think corresponding to re-balances where the consumer's set of partitions is unchanged. This appears to be consistent with how the Java client behaves, so I'm not suggesting changing this behavior.
2. Upgrading Sarama consumers from sticky -> co-operative sticky via rolling restarts worked as expected. E.g. all consumers started out "sticky", then were upgraded to "sticky, sticky-cooperative" , then finally upgraded to "sticky-cooperative" only.
3. Co-existence with a Java `partition.assignment.strategy=org.apache.kafka.clients.consumer.CooperativeStickyAssignor` consumer in the same group has mixed results. If a Java consumer was leading the group and then a Sarama consumer joins - this works as expected. If a Sarama consumer was leading the group and a Java consumer joins then the Sarama leader panics:
    ```
     2023/08/21 20:49:59 Error from consumer: kafka: insufficient data to decode packet, more bytes expected
    panic: Error from consumer: kafka: insufficient data to decode packet, more bytes expected

     goroutine 11 [running]:
     log.Panicf({0x100e23868?, 0x100f5b3d8?}, {0x140000affa0?, 0x14000198af0?, 0x1?})
	     /usr/local/go/src/log/log.go:395 +0x68
    main.main.func1()
	    /tmp/sarama/examples/consumergroup_cooperative/main.go:119 +0x114
    created by main.main
	    /tmp/sarama/examples/consumergroup_cooperative/main.go:110 +0x440
    ```
    I can grab more debug, if it would be helpful. (This was using Kafka 3.3.1, a Kafka 3.1.2 client, and the consumergroups_cooperative example).

Regarding your _Seeking Community Guidance_ question. IMO it would be great to add cooperative re-balancing support as soon as practical - without necessarily waiting for Sarama v2. For example, Sarama v1 could be extended with  the `ConsumerV2` method, which would then be re-factored into a more elegant single consumer approach in Sarama v2.

dnwe:
That seemed to work well for me. I ran sarama, rdkafka and (Java) kafka-console-consumer in the same cooperative balance group and ran some chaos testing restarting each of them randomly and each member was able to join/rejoin/lead without causing issues for the others

In the middle of lots of runs I did trigger one protocol decode failure in `deserializeTopicPartitionAssignment` where it seemed like UserData for one member was 4 bytes (0xff,0xff,0xff,0xff) but it's not clear what caused that. We may just wish to guard the deserializeTopicPartitionAssignment call by ignoring "InsufficientData" err and continuing in the loop

napallday:
Big thanks to @prestona and @dnwe for the thorough testing and fix! 😄

> In the middle of lots of runs I did trigger one protocol decode failure in deserializeTopicPartitionAssignment where it seemed like UserData for one member was 4 bytes (0xff,0xff,0xff,0xff) but it's not clear what caused that. We may just wish to guard the deserializeTopicPartitionAssignment call by ignoring "InsufficientData" err and continuing in the loop

@dnwe Regarding this issue, could you help provide information about the combination of Kafka clients used during testing? I personally conducted tests using Sarama, without direct interaction with `rdkafka` or `Java clients`. So far, I have not come across this particular issue.

dnwe:
@napallday no worries, so I set KAFKA_VERSION=3.3.2 and started up docker-compose as used in the FV with the toxiproxy routes configured and created a six partition topic called `cooperative-topic`. I setup a sarama producer to continuously write some records to the topic.

Then in three different terminals I ran:
```
# sarama
./consumergroup_cooperative -version 3.3.2 -brokers 127.0.0.1:29091 -group sarama-coop -topics cooperative-topic

# librdkafka
./kcat -b 127.0.0.1:29091 -G sarama-coop -X partition.assignment.strategy=cooperative-sticky cooperative-topic

# java
./kafka-console-consumer.sh --bootstrap-server 127.0.0.1:29091 --topic cooperative-topic --group sarama-coop --consumer-property partition.assignment.strategy=org.apache.kafka.clients.consumer.CooperativeStickyAssignor
```

And stopped and started each of them at various points so they all had a turn leading and joining the group

napallday:
The panic will be reproduced following these steps:

Step 1. Run the Sarama consumer (Leader).
Step 2. Run the Java consumer.

This is due to the following improvements of cooperative-sticky strategy implementations within Java. Specifically, the first 4 bytes represent the `generationID` (where 4 bytes of `0xff` actually translate to -1). This can be observed in the code changes [here](https://github.com/apache/kafka/pull/10985/files#diff-daf969a8fe592b2c4c5a64a6ea524992b2831aff0cd0c9f1c069b04586269e2bR92-R93).

I will address this issue at a later time.

napallday:
Hi @prestona @dnwe,

FYI, I've submitted a commit to fix the issue mentioned above. 🛠️

After extensive testing with different Kafka clients, including `librdkafka` (v3.3.2), `java client` (v3.5.0 & v3.3.2), and `sarama` (v3.3.2), I believe the panic issues have been successfully resolved. 

However, during testing with `java client` (v2.8.1), I observed an occurrence of `java.lang.IllegalStateException` in this client when the cooperative assignment was being validated by it. It's worth noting that this behavior appears to be specific to the older versions of the Java client. Since this java client is the leader who did the assignments, I believe that the underlying cause might be rooted in these outdated Java client versions rather than sarama.

dnwe:
Interesting. It took me a few attempts, but yes I also managed to trigger that IllegalStateException with a Java 2.8.1 consumer:

```[2023-08-23 21:03:16,182] ERROR [Consumer clientId=consumer-sarama-coop-1, groupId=sarama-coop] With the COOPERATIVE protocol, owned partitions cannot be reassigned to other members; however the assignor has reassigned partitions [cooperative-topic-2] which are still owned by some members (org.apache.kafka.clients.consumer.internals.ConsumerCoordinator)
[2023-08-23 21:03:16,189] ERROR Error processing message, terminating consumer process:  (kafka.tools.ConsoleConsumer$)
java.lang.IllegalStateException: Assignor supporting the COOPERATIVE protocol violates its requirements
        at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.validateCooperativeAssignment(ConsumerCoordinator.java:668)
        at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.performAssignment(ConsumerCoordinator.java:592)
        at org.apache.kafka.clients.consumer.internals.AbstractCoordinator.onJoinLeader(AbstractCoordinator.java:693)
        at org.apache.kafka.clients.consumer.internals.AbstractCoordinator.access$1000(AbstractCoordinator.java:111)
        at org.apache.kafka.clients.consumer.internals.AbstractCoordinator$JoinGroupResponseHandler.handle(AbstractCoordinator.java:599)
        at org.apache.kafka.clients.consumer.internals.AbstractCoordinator$JoinGroupResponseHandler.handle(AbstractCoordinator.java:562)
        at org.apache.kafka.clients.consumer.internals.AbstractCoordinator$CoordinatorResponseHandler.onSuccess(AbstractCoordinator.java:1182)
        at org.apache.kafka.clients.consumer.internals.AbstractCoordinator$CoordinatorResponseHandler.onSuccess(AbstractCoordinator.java:1157)
        at org.apache.kafka.clients.consumer.internals.RequestFuture$1.onSuccess(RequestFuture.java:206)
        at org.apache.kafka.clients.consumer.internals.RequestFuture.fireSuccess(RequestFuture.java:169)
        at org.apache.kafka.clients.consumer.internals.RequestFuture.complete(RequestFuture.java:129)
        at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient$RequestFutureCompletionHandler.fireCompletion(ConsumerNetworkClient.java:602)
        at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.firePendingCompletedRequests(ConsumerNetworkClient.java:412)
        at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.poll(ConsumerNetworkClient.java:297)
        at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.poll(ConsumerNetworkClient.java:236)
        at org.apache.kafka.clients.consumer.KafkaConsumer.pollForFetches(KafkaConsumer.java:1296)
        at org.apache.kafka.clients.consumer.KafkaConsumer.poll(KafkaConsumer.java:1237)
        at org.apache.kafka.clients.consumer.KafkaConsumer.poll(KafkaConsumer.java:1210)
        at kafka.tools.ConsoleConsumer$ConsumerWrapper.receive(ConsoleConsumer.scala:443)
        at kafka.tools.ConsoleConsumer$.process(ConsoleConsumer.scala:102)
        at kafka.tools.ConsoleConsumer$.run(ConsoleConsumer.scala:76)
        at kafka.tools.ConsoleConsumer$.main(ConsoleConsumer.scala:53)
        at kafka.tools.ConsoleConsumer.main(ConsoleConsumer.scala)
```

dnwe:
@napallday OK I looked over [apache/kafka](/apache/kafka) and I see that they skipped this validation code in the kafka-client under [KAFKA-13406](https://issues.apache.org/jira/browse/KAFKA-13406) and PR [#11439](https://github.com/apache/kafka/pull/11439) due to issues with the verification and that was the Java client behaviour from 2.8.2 onward

napallday:
Cool! 🎉 Huge thanks to @dnwe for the invetigation.

The explanations provided in the Jira ticket are crystal clear.

By the way, just like this fix in [apache/kafka](https://github.com/apache/kafka), the official Java clients have some improvements in sticky and cooperative-sticky assignors since the initial introduction of the incremental cooperative balance protocol. In the future, we can integrate these enhancements gradually.

dnwe:
@napallday were you planning to do the follow up work around metrics, pause/resume and additional FV tests under this PR?

I suppose the alternative would be to merge this as-is, but report the feature as experimental in the changelog so people can test it out in their staging environments

napallday:
@dnwe Thanks for the reminder.

For the `pause/resume` feature, I've verified that the behaviour is consistent with Java Kafka.
- For **EAGLE** protocol: all paused status will be dropped when rebalance
- For **COOPERATIVE** protocol: paused status of **revoked partitions** will be dropped when rebalance

Regarding the metrics and FV tests, I may add it at a later time due to some personal matters. Therefore, I concur with releasing it as an experimental feature initially.

dnwe:
@wwwjfy 👋🏻 thanks for getting in touch

The issue with merging to main at the moment is that the PR branch introduces new interfaces of ConsumeV2 and ConsumerGroupHandlerV2, which we'd then be locked into supporting and ideally we want to be exposing cooperative balance strategy as just a config change for users, rather than requiring them to rewrite their app code to make use of it.

Similarly, if we merge to main without more FV coverage, it might not be obvious to users (unless they read notes on GitHub Releases) that it is considered experimental.

Would it help your testing if I pull the commits into a branch on the main repository rather than them being only on a fork?
i.e.,., if I make a `cooperative-balance-strategy` branch with these commits then you can `go get github.com/IBM/sarama@cooperative-balance-strategy`, start calling the new interfaces and experiment with the code and provide feedback?

lifepuzzlefun:
I hope the current ConsumerGroupHandler can still use the cooperative balance stragtegy. If user want to receive the revoke and assign partition, they can provide their own OnAssignPartition and OnRevokePartition method. and the default handler will be provide if user use origin ConsumerGroupHandler without change current code.

wwwjfy:
Thanks for the quick response :)

It does make sense, if possible, we maintain a single interface than two. (Sorry I didn't read the change and the comments in this PR)

I'm not sure what's the way forward. Using single interfaces will probably break backward compatibility I guess, which means we'll wait until a major version upgrade as well as the consolidation work before that.

For testing, to me alone, I have my forked version anyway, though for others, what you mentioned may be useful :D

Still evaluating using sarama or confluent sdk which already has this feature. will have feedback if I get anything useful

dnwe:
Yes I think the ideal scenario would be if we can rework this PR to keep the Sarama consumer interface consistent, but expose an opt-in similarly to the Java client via an equivalent to the `partition.assignment.strategy` config variable

napallday:
I feel it's hard to keep using the `ConsumerGroupHandler` interface since it's highly coupled with `eager` rebalance protocol - e.g. `ConsumerGroupSession` as the function parameter, is generated in every new generation.

As the new implementation `ConsumerGroupHandlerV2` also works for the previous one `ConsumerGroupHandler`, may I know if it's possible to have another major version of Sarama in the future to include this PR?

I can help add fv tests later but it seems to be a huge work...

aiquestion:
if we want to keep `ConsumerGroupHandler`, i think the possible way is to : 
* call `Setup(ConsumerGroupSession)` / `Cleanup(ConsumerGroupSession) ` when new assignment receives / a new rebalance tirggered.
* change `ConsumerGroupSession`'s `Claims/MemberId/GenerationId` functions to be thread safe.

so if user don't care about assignment change in their previous code(or they only do some log print), only enable cooperative in config will be enough.
But user will still need to be aware of the API behaviour change when enable cooperative rebalance.

offset_manager.go
```diff
@@ -100,13 +132,81 @@ func (om *offsetManager) ManagePartition(topic string, partition int32) (Partiti
...
+	for topic, partitions := range topicPartitions {
+		for _, partition := range partitions {
+			wg.Add(1)
+			go func(topic string, partition int32) {
+				defer wg.Done()
+
+				om.pomsLock.RLock()
+				pom := om.poms[topic][partition]
+				om.pomsLock.RUnlock()
+				err := pom.Close()
```
wwwjfy, line 156: I met panic during local test.

```go
panic: send on closed channel

goroutine 67 [running]:
github.com/IBM/sarama.(*partitionOffsetManager).handleError(0x1400013a300, {0x104c21ac8, 0x104fb5e30})
    /Users/wwwjfy/Develop/streamsdk/tmp/vendor/github.com/IBM/sarama/offset_manager.go:748 +0xac
github.com/IBM/sarama.(*offsetManager).handleResponse(0x140004961a0, 0x14000494008, 0x140003414a0, 0x1400039bde0)
    /Users/wwwjfy/Develop/streamsdk/tmp/vendor/github.com/IBM/sarama/offset_manager.go:491 +0x338
github.com/IBM/sarama.(*offsetManager).flushToBroker(0x140004961a0)
    /Users/wwwjfy/Develop/streamsdk/tmp/vendor/github.com/IBM/sarama/offset_manager.go:381 +0x80
github.com/IBM/sarama.(*offsetManager).Commit(0x140004961a0)
    /Users/wwwjfy/Develop/streamsdk/tmp/vendor/github.com/IBM/sarama/offset_manager.go:357 +0x20
github.com/IBM/sarama.(*offsetManager).mainLoop(0x140004961a0)
    /Users/wwwjfy/Develop/streamsdk/tmp/vendor/github.com/IBM/sarama/offset_manager.go:341 +0x88
github.com/IBM/sarama.withRecover(0x0?)
    /Users/wwwjfy/Develop/streamsdk/tmp/vendor/github.com/IBM/sarama/utils.go:43 +0x40
created by github.com/IBM/sarama.newOffsetManagerFromClient in goroutine 23
    /Users/wwwjfy/Develop/streamsdk/tmp/vendor/github.com/IBM/sarama/offset_manager.go:113 +0x230
exit status 2
```

It happened during a partition is revoked from this consumer.

I tried to track how this happens, and I think this is the case:

goroutine 1:
1.1. consumer is informed a partition is revoked
1.2. it comes here to do cleanup, and this line leads to `errors` channel closure at line 755
1.3. later this pom is removed from `om.poms` at line 180

goroutine 2:

2.1 in the `mainLoop`, `om.Commit` is triggered every second by default
2.2 it constructs requests based on `om.poms` (line 440)
2.3 it gets response, and if it's an error `ErrIllegalGeneration`, send to `pom.errors` (line 490)

In a race condition, 2.3 happens between 1.2 and 1.3, a panic will happen.

consumer_group.go
```diff
@@ -220,6 +312,269 @@ func (c *consumerGroup) Consume(ctx context.Context, topics []string, handler Co
...
+	case ErrNotCoordinatorForConsumer, ErrRebalanceInProgress, ErrOffsetsLoadInProgress:
+		// retry after backoff
+		return c.retryJoinGroup(ctx, topics, retries, join.Err)
+	case ErrMemberIdRequired:
+		// from JoinGroupRequest v4 onwards (due to KIP-394) if the client starts
+		// with an empty member id, it needs to get the assigned id from the
+		// response and send another join request with that id to actually join the
+		// group
+		c.memberID = join.MemberId
+		return c.retryJoinGroup(ctx, topics, retries+1 /*keep retry time*/, join.Err)
```
wwwjfy, line 462: Similar to the accepted pull request #2759, should this be `return c.joinGroup(ctx, topics, retries)`?

offset_manager.go
wwwjfy: propose change:

```diff
iff --git offset_manager.go offset_manager.go
index 6c40a88..4cfe3f5 100644
--- offset_manager.go
+++ offset_manager.go
@@ -150,10 +150,15 @@ func (om *offsetManager) RemovePartitions(topicPartitions map[string][]int32) er
                        go func(topic string, partition int32) {
                                defer wg.Done()
 
-                               om.pomsLock.RLock()
+                               om.pomsLock.Lock()
                                pom := om.poms[topic][partition]
-                               om.pomsLock.RUnlock()
                                err := pom.Close()
+                               delete(om.poms[topic], partition)
+                               if len(om.poms[topic]) == 0 {
+                                       delete(om.poms, topic)
+                               }
+
+                               om.pomsLock.Unlock()
                                if err != nil {
                                        errsLock.Lock()
                                        var consumerErrs ConsumerErrors
@@ -174,17 +179,6 @@ func (om *offsetManager) RemovePartitions(topicPartitions map[string][]int32) er
                }
        }
 
-       om.pomsLock.Lock()
-       for topic, partitions := range topicPartitions {
-               for _, partition := range partitions {
-                       delete(om.poms[topic], partition)
-                       if len(om.poms[topic]) == 0 {
-                               delete(om.poms, topic)
-                       }
-               }
-       }
-
-       om.pomsLock.Unlock()
        if len(errs) > 0 {
                return errs
        }
```

napallday: Thanks for troubleshooting. I think this propose change works, and I also move the logic of last offset flush ahead.

dberardo-com:
any ETA for this feature ?

dailixian:
Hi, Team, 
any updates on this KIP? our engineer team is also keeping eye of this feature support there in newer version of release.

examples/consumergroup/main.go
```diff
@@ -84,9 +84,6 @@ func main() {
 		config.Consumer.Offsets.Initial = sarama.OffsetOldest
 	}
 
-	/**
-	 * Setup a new Sarama consumer group
-	 */
```
puellanivis, line 89: This change seems outside of the scope of the PR.

balance_strategy.go
```diff
@@ -40,6 +43,45 @@ func (p BalanceStrategyPlan) Add(memberID, topic string, partitions ...int32) {
 
 // --------------------------------------------------------------------
 
+type RebalanceProtocol int
+
+const (
+	EAGER RebalanceProtocol = iota
+	COOPERATIVE
```
puellanivis, line 50: In Go, one does not ALLCAPS constants.

balance_strategy.go
```diff
@@ -40,6 +43,45 @@ func (p BalanceStrategyPlan) Add(memberID, topic string, partitions ...int32) {
...
+)
+
+func (p RebalanceProtocol) String() string {
+	switch p {
+	case EAGER:
+		return "EAGER"
+	case COOPERATIVE:
+		return "COOPERATIVE"
+	default:
+		return "UNKNOWN"
```
puellanivis, line 60: Usually, I recommend including the value of `p` in the unknown case, so that if it shows up anywhere, we can know what value it has.

balance_strategy.go
```diff
@@ -40,6 +43,45 @@ func (p BalanceStrategyPlan) Add(memberID, topic string, partitions ...int32) {
...
+}
+
+type RebalanceProtocolSlice []RebalanceProtocol
+
+func (p RebalanceProtocolSlice) Len() int           { return len(p) }
+func (p RebalanceProtocolSlice) Less(i, j int) bool { return p[i] < p[j] }
+func (p RebalanceProtocolSlice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+
+func (p RebalanceProtocolSlice) retainAll(p2 RebalanceProtocolSlice) RebalanceProtocolSlice {
+	var result RebalanceProtocolSlice
```
puellanivis, line 71: Move the declaration of variables to the latest point reasonable. We don’t need to define it here, because it’s not used until the second `for` loop.

Moving it there also allows us to also do `result := make(RebalanceProtocolSlice, len(set))` to pre-allocate the precise length necessary to hold the slice.

balance_strategy.go
```diff
@@ -40,6 +43,45 @@ func (p BalanceStrategyPlan) Add(memberID, topic string, partitions ...int32) {
...
+	default:
+		return "UNKNOWN"
+	}
+}
+
+type RebalanceProtocolSlice []RebalanceProtocol
+
+func (p RebalanceProtocolSlice) Len() int           { return len(p) }
+func (p RebalanceProtocolSlice) Less(i, j int) bool { return p[i] < p[j] }
+func (p RebalanceProtocolSlice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
```
puellanivis, line 68: Do we even sort this slice at all?

If not, why are we defining sort receiver methods on it?

consumer_group.go
```diff
@@ -135,18 +182,48 @@ func newConsumerGroup(groupID string, client Client) (ConsumerGroup, error) {
...
+		supportedProtocols = config.Consumer.Group.Rebalance.GroupStrategies[0].SupportedProtocols()
+		for _, strategy := range config.Consumer.Group.Rebalance.GroupStrategies {
+			supportedProtocols = supportedProtocols.retainAll(strategy.SupportedProtocols())
+		}
+	}
+	if len(supportedProtocols) == 0 {
+		return nil, ConfigurationError("no common rebalance protocol found")
+	}
+	sort.Sort(supportedProtocols)
+	cg.protocol = supportedProtocols[len(supportedProtocols)-1]
```
puellanivis, line 218: It seems that this is just an implementation of `max` over the slice of `supportedProtocols`, except that it runs in `O(n log n)` time instead of `O(n)` time. Rather than sorting a slice and throwing away everything but the max value, just find the max value directly:
```go
cg.protocol = supportedProtocols[0]
for _, protocol := range supportedProtocols[1:] {
	cg.protocol = max(cg.protocol, protocol)
} 
```

balance_strategy.go
```diff
@@ -343,6 +404,75 @@ func (s *stickyBalanceStrategy) balance(currentAssignment map[string][]topicPart
...
+	previousAssignmentSet := make(map[topicPartitionAssignment]string)
+	for prevMember, metadata := range members {
+		for _, ownedTopicPartitions := range metadata.OwnedPartitions {
+			for _, partition := range ownedTopicPartitions.Partitions {
+				previousAssignmentSet[topicPartitionAssignment{Topic: ownedTopicPartitions.Topic, Partition: partition}] = prevMember
+			}
+		}
+	}
+
+	for currMember, assignment := range assignments {
```
puellanivis, line 448: [lint] Typically `current` is abbreviated to `cur`.

balance_strategy.go
```diff
@@ -343,6 +404,75 @@ func (s *stickyBalanceStrategy) balance(currentAssignment map[string][]topicPart
...
+					newAssignments[memberID][topic] = append(newAssignments[memberID][topic], partition)
+				}
+			}
+		}
+	}
+	return newAssignments
+}
+
+func computePartitionsTransferringOwnership(members map[string]ConsumerGroupMemberMetadata, assignments BalanceStrategyPlan) map[topicPartitionAssignment]bool {
+	partitionsTransferringOwnership := make(map[topicPartitionAssignment]bool)
```
puellanivis, line 438: This isn’t needed until the second for loop. Moving it there means that it doesn’t clutter confusion about what’s being modified in this for loop.

sticky_assignor_user_data.go
```diff
@@ -5,7 +5,7 @@ type topicPartitionAssignment struct {
 	Partition int32
 }
 
-type StickyAssignorUserData interface {
+type MemberData interface {
```
puellanivis, line 8: This change is a braking change. Do not rename identifiers that are already exported.

balance_strategy.go
```diff
@@ -619,7 +753,7 @@ func assignPartition(partition topicPartitionAssignment, sortedCurrentSubscripti
 }
 
 // Deserialize topic partition assignment data to aid with creation of a sticky assignment.
-func deserializeTopicPartitionAssignment(userDataBytes []byte) (StickyAssignorUserData, error) {
+func deserializeStickyUserData(userDataBytes []byte) (MemberData, error) {
```
puellanivis, line 756: I don’t know why this function is changing name, without anything else changing.

It seems outside of the scope of necessary changes.

balance_strategy_test.go
```diff
@@ -271,18 +271,22 @@ func Test_deserializeTopicPartitionAssignment(t *testing.T) {
...
 			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("deserializeTopicPartitionAssignment() = %v, want %v", got, tt.want)
+				t.Errorf("deserializeStickyUserData() = %v, want %v", got, tt.want)
 			}
 		})
 	}
 }
 
+func Test_deserializeCooperativeStickyUserData(t *testing.T) {
+	// todo: add unit test
```
puellanivis, line 287:
```suggestion
	// TODO: add unit test
```
Many editors will special highlight the allcaps version, so that it’s prevalent. And even without special syntax highlighting, ALLCAPs stands out more.

balance_strategy.go
```diff
@@ -832,35 +997,46 @@ func areSubscriptionsIdentical(partition2AllPotentialConsumers map[topicPartitio
...
+	var err error
+	var memberData MemberData
 	// for each partition we create a sorted map of its consumers by generation
 	sortedPartitionConsumersByGeneration := make(map[topicPartitionAssignment]map[int]string)
 	for memberID, meta := range members {
-		consumerUserData, err := deserializeTopicPartitionAssignment(meta.UserData)
+		if len(meta.UserData) == 0 {
+			continue
+		}
+		if s.parent != nil {
```
puellanivis, line 1012: I’m unclear why this needs to be: a field in a struct, and as part of a receiver method.

Similar to above, couldn’t there just be a `cooperative bool` function argument?

balance_strategy.go
```diff
@@ -161,8 +213,13 @@ func (s *balanceStrategy) AssignmentData(memberID string, topics map[string][]in
 	return nil, nil
 }
 
+func (s *balanceStrategy) SupportedProtocols() RebalanceProtocolSlice {
+	return []RebalanceProtocol{EAGER}
+}
+
 type stickyBalanceStrategy struct {
 	movements partitionMovements
+	parent    *cooperativeStickyBalanceStrategy
```
puellanivis, line 222: I’m not sure why we need a `parent` here? What does this design get us that just putting a `cooperative bool` does not accomplish?

consumer_group.go
```diff
@@ -107,6 +152,8 @@ func NewConsumerGroup(addrs []string, groupID string, config *Config) (ConsumerG
 	if err != nil {
 		_ = client.Close()
 	}
+
+	// start heartbeat loop
```
puellanivis, line 156: I don’t see any heartbeat loop getting started here.

consumer_group.go
```diff
@@ -45,6 +45,37 @@ type ConsumerGroup interface {
...
+	//    ConsumeV2 will be called again to trigger a new rebalance so that leader can re-assigned the revoked partitions
+	//    to other consumers.
+	//
+	// The difference between COOPERATIVE and EAGER rebalance protocol is that,
+	// when a rebalance happens, EAGER rebalance protocol will revoke all the partitions in the current generation,
+	// no matter whether they will be assigned to the same consumer later or not.
+	//
+	// Please note, once ctx is done, `ConsumeClaim` loops must exit as quickly as possible.
+	// Otherwise, it will be kicked out of the next generation and cause offset commit failures.
+	ConsumeV2(ctx context.Context, topics []string, handler ConsumerGroupHandlerV2) error
```
puellanivis, line 77: I’m not convinced that we need a whole new `ConsumerGroupHandlerV2` to accomplish the goals of this feature?

Plus, because this is adding a function to an exported interface, this would be a breaking change anyways, since any non-us implementations of this interface would now suddenly not implement this interface. (It’s more likely than you think.)

consumer_group.go
```diff
@@ -220,6 +312,269 @@ func (c *consumerGroup) Consume(ctx context.Context, topics []string, handler Co
...
+		}
+	}
+
+	newAssignedPartitions := diffAssignment(claims, c.ownedPartitions)
+	if c.protocol == COOPERATIVE {
+		revokedPartitions := diffAssignment(c.ownedPartitions, claims)
+		Logger.Printf("updating consumer(group:%s, member:%s, generation:%d, isLeader:%v)\n"+
+			"*** All Assignments: %v\n"+
+			"*** New Partitions: %v\n"+
+			"*** Revoked Partitions: %v\n",
```
puellanivis, line 387: Logging with newlines can be problematic on many production environments that expect logs to be single lines.

consumer_group.go
```diff
@@ -220,6 +312,269 @@ func (c *consumerGroup) Consume(ctx context.Context, topics []string, handler Co
...
+	return syncGroupResponse, nil
+}
+
+func (c *consumerGroup) retryJoinGroup(ctx context.Context, topics []string, retries int, previousErr error) (*SyncGroupResponse, error) {
+	if retries <= 0 {
+		return nil, previousErr
+	}
+
+	nextRetryTimer := time.NewTimer(c.config.Consumer.Group.Rebalance.Retry.Backoff)
+	defer nextRetryTimer.Stop()
```
puellanivis, line 547: The concerns about using `time.After` no longer apply. If one is never resetting a timer, then `<-time.After(c.config.Consumer.Group.Rebalance.Retry.Backoff)` is fine.

> Before Go 1.23, this documentation warned that the underlying [Timer](https://pkg.go.dev/time@go1.26.2#Timer) would not be recovered by the garbage collector until the timer fired, and that if efficiency was a concern, code should use NewTimer instead and call [Timer.Stop](https://pkg.go.dev/time@go1.26.2#Timer.Stop) if the timer is no longer needed. As of Go 1.23, the garbage collector can recover unreferenced, unstopped timers. There is no reason to prefer NewTimer when After will do.

consumer_group.go
```diff
@@ -220,6 +312,269 @@ func (c *consumerGroup) Consume(ctx context.Context, topics []string, handler Co
...
+// member, and then in the next rebalance it can finally reassign those partitions not owned by anyone to consumers.
+func validateCooperativeAssignment(previousAssignment map[string]ConsumerGroupMemberMetadata, currentAssignment BalanceStrategyPlan) error {
+	set := computePartitionsTransferringOwnership(previousAssignment, currentAssignment)
+	if len(set) > 0 {
+		var topicPartitions []string
+		for k := range set {
+			topicPartitions = append(topicPartitions, fmt.Sprintf("%s/%d", k.Topic, k.Partition))
+		}
+		return fmt.Errorf("in the customized cooperative rebalance strategy, "+
+			"topic-partitions %v should be revoked before reassigning them to other consumers", topicPartitions)
```
puellanivis, line 573: I feel like there’s too much context here. By breaking up the line, it becomes impossible to search for `"strategy, topic-partitons"` won from reading through any logs . Keeping the string to one line is ideal: `"topic/partitions %v should be revoked before reassigning them"` should probably be sufficient.

And more ideally, put the maximum amount of constant string up front so that the text becomes even more greppable: `"topic/partitions should be revoked before reassigning them: %v"`

consumer_group.go
```diff
@@ -425,15 +781,14 @@ func (c *consumerGroup) newSession(ctx context.Context, topics []string, handler
...
-	session, err := newConsumerGroupSession(ctx, c, claims, join.MemberId, join.GenerationId, handler)
+	session, err := newConsumerGroupSession(ctx, c, claims, handler)
 	if err != nil {
 		return nil, err
 	}
 
 	// only the leader needs to check whether there are newly-added partitions in order to trigger a rebalance
 	if join.LeaderId == join.MemberId {
-		go c.loopCheckPartitionNumbers(allSubscribedTopicPartitions, allSubscribedTopics, session)
+		go c.loopCheckPartitionNumbers(session.ctx, session.cancel, allSubscribedTopicPartitions, allSubscribedTopics)
```
puellanivis, line 791: Why do we need to switch from passing one `session` to passing two fields of that very struct?

consumer_group.go
```diff
@@ -744,6 +1105,125 @@ func (c *consumerGroup) topicToPartitionNumbers(topics []string) (map[string]int
 	return topicToPartitionNum, nil
 }
 
+func (c *consumerGroup) revokedPartitions(revokedPartitions map[string][]int32) error {
+	Logger.Printf("consumer(group:%s, member:%s, generation:%d, isLeader:%v) revoking partitions: %v\n",
```
puellanivis, line 1109: [FYI, and probably applies to a lot more lines] `%t` is the formatting verb for booleans.

consumer_group.go
```diff
@@ -744,6 +1105,125 @@ func (c *consumerGroup) topicToPartitionNumbers(topics []string) (map[string]int
...
+
+	// close revoked partition consumers
+	c.removeClaims(revokedPartitions)
+
+	c.handlerV2.Cleanup(c.offsetManager, revokedPartitions)
+
+	// close partition offset managers for revoked partitions
+	if err := c.offsetManager.RemovePartitions(revokedPartitions); err != nil {
+		Logger.Printf("error when removing partition offset managers for %v, err: %v", revokedPartitions, err)
+		return err
```
puellanivis, line 1120: This should be better with `fmt.Errorf` rather than forcing a log line on everyone.

consumer_group.go
```diff
@@ -744,6 +1105,125 @@ func (c *consumerGroup) topicToPartitionNumbers(topics []string) (map[string]int
...
+	}
+	return nil
+}
+
+func (c *consumerGroup) revokedOwnedPartitions() {
+	if len(c.ownedPartitions) > 0 {
+		err := c.revokedPartitions(c.ownedPartitions)
+		if err != nil {
+			Logger.Printf("error revoking owned partitions: %v", err)
```
puellanivis, line 1129: See? We’re already logging the error here. Why do we need to log it in `revokedPartitions` as well?

consumer_group.go
```diff
@@ -744,6 +1105,125 @@ func (c *consumerGroup) topicToPartitionNumbers(topics []string) (map[string]int
...
+	// close partition offset managers for revoked partitions
+	if err := c.offsetManager.RemovePartitions(revokedPartitions); err != nil {
+		Logger.Printf("error when removing partition offset managers for %v, err: %v", revokedPartitions, err)
+		return err
+	}
+	return nil
+}
+
+func (c *consumerGroup) revokedOwnedPartitions() {
+	if len(c.ownedPartitions) > 0 {
```
puellanivis, line 1126: Invert this condition so that the normal code path is unindented:
```go
if len(c.ownedPartitions) < 1 {
	return
}
```

consumer_group.go
```diff
@@ -744,6 +1105,125 @@ func (c *consumerGroup) topicToPartitionNumbers(topics []string) (map[string]int
...
+	Logger.Printf("consumer(group:%s, member:%s, generation:%d, isLeader:%v) starting new assigned partitions: %v",
+		c.groupID, c.memberID, c.generationID, c.isLeader, newAssignedPartitions)
+
+	// create partition offset managers for each new assigned partitions
+	for topic, partitions := range newAssignedPartitions {
+		for _, partition := range partitions {
+			pom, err := c.offsetManager.ManagePartition(topic, partition)
+			if err != nil {
+				Logger.Printf("unable to create partition offset manager for %s/%d, err: %v", topic, partition, err)
+				return err
```
puellanivis, line 1145: Again, `fmt.Errorf` can add this sort of context directly into the error, and not requiring debugging logs.

consumer_group.go
```diff
@@ -744,6 +1105,125 @@ func (c *consumerGroup) topicToPartitionNumbers(topics []string) (map[string]int
...
+			}
+
+			// handle POM errors
+			go func(topic string, partition int32) {
+				for err := range pom.Errors() {
+					c.handleError(err, topic, partition)
+				}
+			}(topic, partition)
```
puellanivis, line 1153: `topic` and `partition` are already available via closure. Putting them as function parameters to an anonymous function like this only serves to add inertia, because now the types of `topic` and `partition` are baked into the code here, and if their types change, we have to change the type information here as well, and this also provides no real additional safety. (We’re after all already passing them as arguments out to `handleError`.

Additionally, the code reading path goes from the top of the `go func` to the bottom of the function call to see what variables are put into the arguments, and then we can come back to the top of the function now with that data. An unusual reading path.

Unless the goroutine needs its own copy of the variables (like before `loopvar` experiment was rolled out) there’s really not a good reason to put variables we already have available into new variables just so that they’re in different variables from the first variables that we would have had anyways.

And if there is a need for the goroutine to need its own copy, then this is better accomplished with:
```go
topic, partition := topic, partition
```
Here, we still repeat the variable twice, but the types are all inferred, and code reading order remains top to bottow.

consumer_group.go
```diff
@@ -744,6 +1105,125 @@ func (c *consumerGroup) topicToPartitionNumbers(topics []string) (map[string]int
...
+	var errsLock sync.Mutex
+
+	var wg sync.WaitGroup
+	// create partition consumers for each new assigned partitions
+	for topic, partitions := range newAssignedPartitions {
+		for _, partition := range partitions {
+			wg.Add(1)
+			// get next offset
+			go func(topic string, partition int32) {
+				defer wg.Done()
```
puellanivis, line 1169: We’re using go 1.25 https://github.com/IBM/sarama/blob/main/go.mod#L3

This means we have [sync.WaitGroup.Go](https://pkg.go.dev/sync@go1.26.2#WaitGroup.Go)

The complaints about duplicatging `topic` and `partition` go extra here. The values assigned to them are now a whole screen if not two below right here, which is where the information about what their value is is most important.

consumer_group.go
```diff
@@ -744,6 +1105,125 @@ func (c *consumerGroup) topicToPartitionNumbers(topics []string) (map[string]int
...
+					c.handleError(err, topic, partition)
+				}
+			}(topic, partition)
+		}
+	}
+
+	c.handlerV2.Setup(c.offsetManager, newAssignedPartitions)
+
+	var errs ConsumerErrors
+	var errsLock sync.Mutex
```
puellanivis, line 1160: I feel like we could use a `chan error` here instead. With a `for err := range errch { errs = append(errs, err) }` collector in the primary goroutine.

Rather than communicating by sharing memory, we can share memory by communicating. —Rob Pike, paraphrased

consumer_group.go
```diff
@@ -744,6 +1105,125 @@ func (c *consumerGroup) topicToPartitionNumbers(topics []string) (map[string]int
...
+					errsLock.Unlock()
+					return
+				}
+
+				// handle errors
+				go func(pc *partitionClaim) {
+					for err := range pc.claim.Errors() {
+						c.handleError(err, pc.claim.topic, pc.claim.partition)
+					}
+				}(pc)
```
puellanivis, line 1206: The passing of `pc` here through anonymous function parameters is even less justified than a value variable, since it is a pointer, `pc := pc` gives the same pointer to the same data anyways.

consumer_group.go
```diff
@@ -744,6 +1105,125 @@ func (c *consumerGroup) topicToPartitionNumbers(topics []string) (map[string]int
...
+
+				c.wg.Add(1)
+				go func(pc *partitionClaim) {
+					defer c.wg.Done()
```
puellanivis, line 1210:
```suggestion
				c.wg.Go(func() {
```

consumer_group.go
```diff
@@ -744,6 +1105,125 @@ func (c *consumerGroup) topicToPartitionNumbers(topics []string) (map[string]int
...
+				}(pc)
+			}(topic, partition)
+		}
+	}
+	wg.Wait()
+
+	if len(errs) > 0 {
+		return errs
+	}
+	return nil
```
puellanivis, line 1224:
```suggestion
	return errors.Join(errs...)
```

consumer_group.go
```diff
@@ -1139,6 +1645,233 @@ func newConsumerGroupClaim(sess *consumerGroupSession, topic string, partition i
...
+
+	topicClaims, ok := c.claims[topic]
+	if !ok {
+		return nil
+	}
+	pc, ok := topicClaims[partition]
+	if !ok {
+		return nil
+	}
+	return pc
```
puellanivis, line 1716: [unnecessary guarding] Keys that do not exist return zero values. Nil maps act like empty maps when read, thus this code is equivalent to:
```suggestion
	return c.claims[topic][partition]
```

consumer_group.go
```diff
@@ -1139,6 +1645,233 @@ func newConsumerGroupClaim(sess *consumerGroupSession, topic string, partition i
...
+
+	coordinator, err := c.client.Coordinator(c.groupID)
+	if err != nil {
+		return nil, err
+	}
+
+	return coordinator, nil
+}
+
+func (c *consumerGroup) startHeartbeatLoop(cancelFunc context.CancelFunc) (chan none, chan none) {
```
puellanivis, line 1758: Rather than repeating `chan none`, and providing no information about which return value is which, name them, so that it is documented which goes where:
```go
func (c *consumerGroup) startHeartbeatLoop(cancelFunc context.CancelFunc) (hbDying, hbDone chan none) {
```

Actually, considering that `hbDone` should be receive only:

```suggestion
func (c *consumerGroup) startHeartbeatLoop(cancelFunc context.CancelFunc) (hbDying chan none, hbDone <-chan none) {
```
Now, callers cannot close `hbDone` on us.

cooperative_sticky_assignor_user_data.go
```diff
@@ -0,0 +1,31 @@
...
+	RackID                string
+}
+
+func (m *CooperativeStickyMemberData) partitions() []topicPartitionAssignment {
+	return m.PartitionsAssignments
+}
+func (m *CooperativeStickyMemberData) hasGeneration() bool { return true }
+func (m *CooperativeStickyMemberData) generation() int     { return int(m.Generation) }
+
+var _ MemberData = (*CooperativeStickyMemberData)(nil)
```
puellanivis, line 31: This can be put in a `_test.go` file, since it’s a test.

metrics_test.go
puellanivis, line 1: These changes are outside of the scope of this PR.

offset_manager.go
```diff
@@ -14,6 +16,35 @@ type OffsetManager interface {
 	// topic/partition.
 	ManagePartition(topic string, partition int32) (PartitionOffsetManager, error)
 
+	RemovePartitions(partitions map[string][]int32) error
```
puellanivis, line 19: Adding methods to an exported interface is a breaking change.

Convention is that one uses an extension interface to add new methods:
```go
type FooBarOffsetManager interface{
	OffsetManager

	Foo()
	Bar()
}
```

offset_manager.go
```diff
@@ -14,6 +16,35 @@ type OffsetManager interface {
...
+
+	// ResetOffset resets to the provided offset, alongside a metadata string that
+	// represents the state of the partition consumer at that point in time. Reset
+	// acts as a counterpart to MarkOffset, the difference being that it allows to
+	// reset an offset to an earlier or smaller value, where MarkOffset only
+	// allows incrementing the offset. cf MarkOffset for more details.
+	ResetOffset(topic string, partition int32, offset int64, metadata string)
+
+	// MarkMessage marks a message as consumed.
+	MarkMessage(msg *ConsumerMessage, metadata string)
```
puellanivis, line 46: So, this is just now repeating `ConsumerGroupSession`? Why didn’t we just keep the session around, and call these functions on that, and `ManagePartitions` and `RemovePartitions` etc on an extension `OffsetManager`?

We are allowed to pass more than just one interface in a function call.

offset_manager.go
```diff
@@ -35,6 +66,7 @@ type offsetManager struct {
 	memberID        string
 	groupInstanceId *string
 	generation      int32
+	updateLock      sync.RWMutex
```
puellanivis, line 69: Convention is to group protected variables in their own paragraph, and put the lock that protects them at the top:
```go
	otherVariables…

	updateLock sync.RWMutex
	memberID   string
	generation int32

	moreVariables…
```

offset_manager.go
```diff
@@ -100,13 +132,75 @@ func (om *offsetManager) ManagePartition(topic string, partition int32) (Partiti
...
 	}
 
 	topicManagers[partition] = pom
 	return pom, nil
 }
 
+func (om *offsetManager) RemovePartitions(topicPartitions map[string][]int32) error {
+	// flush one last time
+	if om.conf.Consumer.Offsets.AutoCommit.Enable {
+		for attempt := 0; attempt <= om.conf.Consumer.Offsets.Retry.Max; attempt++ {
```
puellanivis, line 145:
```suggestion
		for range om.conf.Consumer.Offsets.Retry.Max {
```

offset_manager.go
```diff
@@ -100,13 +132,75 @@ func (om *offsetManager) ManagePartition(topic string, partition int32) (Partiti
...
+			om.flushToBroker()
+		}
+	}
+
+	var errs ConsumerErrors
+	var errsLock sync.Mutex
```
puellanivis, line 151:
```go
errch := make(chan *ConsumerError)
for {
	wg.Go(func() {
		if err := foo(); err != nil {
			errch <- err
			return
		}
	})
}
go func() {
	wg.Wait()
	close(errch)
}()
var errs ConsumerErrors
for err := range err {
	errs = append(errs, err)
}
```

offset_manager.go
```diff
@@ -605,13 +718,14 @@ func (pom *partitionOffsetManager) AsyncClose() {
...
+	var consumerErrors ConsumerErrors
 	for err := range pom.errors {
-		errors = append(errors, err)
+		consumerErrors = append(consumerErrors, err)
 	}
 
-	if len(errors) > 0 {
-		return errors
+	if len(consumerErrors) > 0 {
+		return consumerErrors
```
puellanivis, line 728: Renaming this variable is out of scope for this PR.

offset_manager_test.go
```diff
@@ -532,6 +532,9 @@ func TestPartitionOffsetManagerCommitErr(t *testing.T) {
 
 	pom.MarkOffset(100, "modified_meta")
 
+	// flaky test: wait for sending commit requests
+	time.Sleep(10 * time.Millisecond)
+
```
puellanivis, line 537: Do not “fix” flaky tests with a `time.Sleep`.

Either address the race condition, or live with the flakiness. Please do not make tests run unconditionally longer for everyone.

utils.go
```diff
@@ -264,6 +264,7 @@ var (
 	DefaultVersion = V2_1_0_0
 
 	// reduced set of protocol versions to matrix test
+	// nolint:unused
```
puellanivis, line 267: This change is outside of the scope of this PR.

utils.go
```diff
@@ -313,3 +314,40 @@ func (v KafkaVersion) String() string {
 
 	return fmt.Sprintf("%d.%d.%d", v.version[0], v.version[1], v.version[2])
 }
+
+func diffAssignment(map1 map[string][]int32, map2 map[string][]int32) map[string][]int32 {
+	set := make(map[string]map[int32]bool)
```
puellanivis, line 319: Rather than use a map of maps, and allocate a ton of background maps, let’s use a single map with a struct key:
```go
type topicPart struct {
	topic string
	part  int32
}
set := make(map[topicPart]bool)
for topic, partitions := range map2 {
		set[topicPart{ topic, parititon }] = true
	}
}
```

balance_strategy.go
```diff
@@ -19,6 +19,8 @@ const (
 	// StickyBalanceStrategyName identifies strategies that use the sticky-partition assignment strategy
 	StickyBalanceStrategyName = "sticky"
 
+	CooperativeStickyBalanceStrategyName = "cooperative-sticky"
```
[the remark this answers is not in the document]
puellanivis: For situations like this, there should be a comment at the top of the block explaining the purpose of all the values.

The names themselves are otherwise sufficiently self-explanatory.

Example-wise: we don’t have a comment on every `http.MethodXY` constant, because they’re exactly what they say in the name.

dnwe:
Closing this PR down as it has been superseded by other work now

Issue #1261, opened 2019-01-23
renaming ErrXxx constants is an API changing change


nsd20463:
FYI commit 04e4286cc194cd39150d98dcaf7e399601febf5f breaks [my] code because it removes several public ErrXxx names.

Perhaps this is deliberate, though strange in a tagged "1.something" repo. Or it's an accident and you meant to keep the old names around too.

bai:
cc: @nrvnrvn @varun06

bai:
Yep, this is a valid degradation.

varun06:
ouch, can we please create a release and mention the breaking api change part. Will be more careful in future for these changes.

bai:
I'm revering the PR in https://github.com/Shopify/sarama/pull/1262 on 🍏. Sorry for the trouble and thanks for reporting!

nsd20463:
Thanks.

Issue #543, opened 2015-09-26
Differentiate between client and server version of ErrMessageSizeTooLarge


wvanbergen:
Right now, the Kafka broker can return `ErrMessageSizeTooLarge` if a message is deemed to large based on the broker configuration, but we also return it client-side in the producer based on the sarama configuration: https://github.com/Shopify/sarama/blob/53f2a54a5cfd662728648b20eb26e4a8e19b4887/async_producer.go#L255

For debugging purposes, it would be nice to be able to differentiate between the client and server-side version.

cc @andremedeiros @kvs @eapache

eapache:
There are a lot of cases where we return kafka server errors directly:
- https://github.com/Shopify/sarama/blob/53f2a54a5cfd662728648b20eb26e4a8e19b4887/consumer.go#L366
- https://github.com/Shopify/sarama/blob/53f2a54a5cfd662728648b20eb26e4a8e19b4887/async_producer.go#L255
- https://github.com/Shopify/sarama/blob/53f2a54a5cfd662728648b20eb26e4a8e19b4887/client.go#L225
- https://github.com/Shopify/sarama/blob/53f2a54a5cfd662728648b20eb26e4a8e19b4887/client.go#L253
- https://github.com/Shopify/sarama/blob/53f2a54a5cfd662728648b20eb26e4a8e19b4887/client.go#L275
- https://github.com/Shopify/sarama/blob/53f2a54a5cfd662728648b20eb26e4a8e19b4887/client.go#L312
- https://github.com/Shopify/sarama/blob/53f2a54a5cfd662728648b20eb26e4a8e19b4887/client.go#L351
- https://github.com/Shopify/sarama/blob/53f2a54a5cfd662728648b20eb26e4a8e19b4887/client.go#L506
- https://github.com/Shopify/sarama/blob/53f2a54a5cfd662728648b20eb26e4a8e19b4887/client.go#L513
- https://github.com/Shopify/sarama/blob/53f2a54a5cfd662728648b20eb26e4a8e19b4887/client.go#L540

At the time it seemed unnecessary to define a new local error when the kafka-provide value had the exact right meaning, but I guess this could lead to confusion...

wvanbergen:
This case is a bit more confusing because it is determined by two configuration settings that live in different places that could be different from each other. This made debugging this error a lot harder.

[...]

PR #2628, opened 2023-08-29
fix: make clear that error is configuration issue not server error

### LINKED ISSUE #1225: consumer: batch record number >2*math.MaxUint16 leads to error decoding packet: invalid array length
##### Versions

Sarama Version:
1.18
Kafka Version:
2.0.0
Go Version:
1.11
##### Configuration

sarama
```
    sarama.MaxResponseSize = 1 << 30
    sarama.MaxRequestSize = 1 << 30
    config.Producer.MaxMessageBytes = 1 << 30
    config.Producer.Partitioner = sarama.NewManualPartitioner
    config.Producer.RequiredAcks = sarama.WaitForAll
```

##### Logs

When filing an issue please provide logs from Sarama and Kafka if at all
possible. You can set `sarama.Logger` to a `log.Logger` to capture Sarama debug
output.

sarama
maybe cause by here https://github.com/Shopify/sarama/blob/master/real_decoder.go#L88
```
error decoding packet        : invalid array length
```

##### Problem Description
while consume, meet the `invalid array length` error.

after change, the producer(set config.Producer.Flush.MaxMessages = math.MaxUint16), the consumer would'not meet the invalid array length.

 i print the tmp at maybe cause by here https://github.com/Shopify/sarama/blob/master/real_decoder.go#L88, it will not bigger than the MaxMessages config in producer

i think the problem is using the default `Flush` config, the batch record number may bigger than 2 
 x math.MaxUint16, and when consuming, sarama  treat it invalid array length if record number > 2 x math.MaxUint16.

if kafka has no limit for batch records num, can we remove the length limit?
or better to change the default config of Flush to limit MaxMessages

hindessm:
Since we already return a `ConfigurationError` on line 449 it should be reasonable to return one here rather than returning an error that suggests that the server has rejected the message.

Fixes #1225

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -449,8 +449,10 @@ func (p *asyncProducer) dispatcher() {
 			p.returnError(msg, ConfigurationError("Producing headers requires Kafka at least v0.11"))
 			continue
 		}
-		if msg.ByteSize(version) > p.conf.Producer.MaxMessageBytes {
-			p.returnError(msg, ErrMessageSizeTooLarge)
+
+		size := msg.ByteSize(version)
+		if size > p.conf.Producer.MaxMessageBytes {
+			p.returnError(msg, ConfigurationError(fmt.Sprintf("Attempt to produce message larger than configured Producer.MaxMessageBytes: %d > %d", size, p.conf.Producer.MaxMessageBytes)))
 			continue
 		}
 


dnwe (approved):

ae-govau:
FWIW this change broke some of our running code. We had code that would:

```golang
if err == sarama.ErrMessageSizeTooLarge {
    // log a message, bump a metric, but don't retry
}
```

ie for our use-case, we want to ignore messages related to a message being too large. It's OK for us to throw these out. We used to check a size before sending into Sarama, but then we had some cases where our payload was under the limit, but once headers etc were added, it was over, so we switched to checking this error code. 

Noting that https://github.com/IBM/sarama/issues/543 reports that previous error code has been returned for at least 8 years, this seems a reasonably significant change... (and a good example of [Hyrum's Law](https://nordicapis.com/what-does-hyrums-law-mean-for-api-design/) )

It's now not clear to me how I should check for size related issues.

ae-govau:
Interestingly the last PR which touched that line of code was 4 years ago in 2019 where https://github.com/IBM/sarama/pull/1262 reverts a change a few days earlier (https://github.com/IBM/sarama/pull/1218) that renamed a bunch of error variables because:

> renaming ErrXxx constants is an API changing change #1261

(that change is arguably less impactful than this one because it could be detected/corrected at compile time whereas we only caught this at runtime)

puellanivis:
This change also caused a critical at my company. We were originally skipping messages on the `ErrMessageSizeTooLarge` as well. When combining this breaking change of functionality with a different-language client having a different and larger default `message.max.bytes` value, it left us with a message consumed that we could not then republish.

As mentioned the new error is also extremely error prone to check for. We’ve had to resort to:
```
var confErr sarama.ConfigurationError

if errors.As(err, &confErr) && strings.HasPrefix("Attempt to produce message larger than configured Producer.MaxMessageBytes", string(confErr)) {
	// handle skipping the message here
}
```

This code is inherently fragile, and could break if the error text changes in any way.

With this sort of behavior change, it really should have involved a breaking change removing the `ErrMessageSizeTooLarge` symbol, which is now unused in the package itself. At least then, users would have at least been aware of the breaking change of behavior, and not left with a landmine to eventually step on? I know that’s also considered bad-form in Golang, but it might have just been the least bad option in this case.

Issue #2470, opened 2023-05-05
SASL connection to Eventhubs using kafka fails 1.38.0 onwards


DeepanshuA:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka | Go |
|1.38.0|2.0.0|1.20.1|
|  |  |  |

##### Configuration

What configuration values are you using for Sarama and Kafka?
``` go
```

##### Logs

When filing an issue please provide logs from Sarama and Kafka if at all
possible. You can set `sarama.Logger` to a `log.Logger` to capture Sarama debug
output.

<details><summary>logs: CLICK ME</summary>
<p>

```
time="2023-05-05T20:13:49.447569+05:30" level=debug msg="client/metadata fetching metadata for all topics from broker sasl-issue.servicebus.windows.net:9093\n" app_id=orderapp component="kafkabind (bindings.kafka/v1)" instance=LAPTOP-J62MA7E9 scope=dapr.contrib type=log ver=unknown
time="2023-05-05T20:13:50.4581451+05:30" level=debug msg="Completed pre-auth SASL handshake. Available mechanisms: [PLAIN OAUTHBEARER]" app_id=orderapp component="kafkabind (bindings.kafka/v1)" instance=LAPTOP-J62MA7E9 scope=dapr.contrib type=log ver=unknown
time="2023-05-05T20:13:50.6997856+05:30" level=debug msg="SASL authentication successful with broker sasl-issue.servicebus.windows.net:9093:4 - [0 0 0 0]\n" app_id=orderapp component="kafkabind (bindings.kafka/v1)" instance=LAPTOP-J62MA7E9 scope=dapr.contrib type=log ver=unknown
time="2023-05-05T20:13:50.6999026+05:30" level=debug msg="Connected to broker at sasl-issue.servicebus.windows.net:9093 (unregistered)\n" app_id=orderapp component="kafkabind (bindings.kafka/v1)" instance=LAPTOP-J62MA7E9 scope=dapr.contrib type=log ver=unknown
time="2023-05-05T20:13:50.9402314+05:30" level=debug msg="client/metadata got error from broker -1 while fetching metadata: read tcp 172.21.220.249:41466->13.66.149.8:9093: read: connection reset by peer\n" app_id=orderapp component="kafkabind (bindings.kafka/v1)" instance=LAPTOP-J62MA7E9 scope=dapr.contrib type=log ver=unknown
[...]

PR #2633, opened 2023-08-30
fix(client): force Event Hubs to use V1_0_0_0

dnwe:
The problem with supporting Azure Event Hubs is that it isn't Apache Kafka under the covers, it's an intermediate proxy that supports a subset of the Kafka APIs at various versions and then maps them onto Event Hubs protocol(s) at the backend. As as result Sarama's current mechanism of specifying the KAFKA_VERSION to determine what protocol versions to support and use doesn't really work properly with Event Hubs because it supports an unusual set of protocols and even defines minimum versions for ProduceRequest (v3) and FetchRequest (v4). For some reason EventHubs is very behind on FetchRequest and MetadataRequest so the max configuration you can use in Sarama is V1_0_0_0 and the minimum is V0_11_0_0.

As we have recently bumped the default Version to V2_1_0_0 and support a wider range of protocol versions, we're more likely to see issues raised by Event Hubs users unless they have already pinned their version correctly. To try and prevent this, attempt to detect use with Event Hubs by inspecting the bootstrap broker addresses and overriding the Version.

Contributes-to: #2470

### DIFF
--- a/client.go
+++ b/client.go
@@ -183,6 +183,13 @@ func NewClient(addrs []string, conf *Config) (Client, error) {
 		return nil, ConfigurationError("You must provide at least one broker address")
 	}
 
+	if strings.Contains(addrs[0], ".servicebus.windows.net") {
+		if conf.Version.IsAtLeast(V1_1_0_0) || !conf.Version.IsAtLeast(V0_11_0_0) {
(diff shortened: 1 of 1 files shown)


hindessm (approved):
LGTM

Issue #2417, opened 2023-01-12
How can I identify a commit operation is successful?


ZzIris:
hello! it's i first time to learn to use sarama. I find that there is not a error return when i call sarama.ConsumerGroupSession.Commit,  i should use sarama.ConsumerGroup.Errors() if i want to get a error, but i don't know the error which i get from sarama.ConsumerGroup.Errors() is really from commit operation or other operation,  how can i identify a error is a commit error in another word? please give me example, thanks!!!

ZzIris:
@dnwe for example, in some case, my application encountering a network error for a long time, and the application will recrod the offset have consumed but fail to commit when i restart the application at this moment to prevent process again，the application will nerver know which offset is consumed but fail to commit if i can't match the error from commit operation~ how can i do? what key word of the error that I can get from sarama.ConsumerGroup.Errors()? i try to find a definition about commit error, but just find sarama.ErrTxnOffsetCommit、sarama.ErrUnstableOffsetCommit、sarama.ErrInvalidCommitOffsetSize，the commit errors are not always related to them in my opinion.

joewreschnig:
Would it be acceptable (in terms of API stability) to also have `Commit` return the error from the offset manager (as well as forwarding it through the handlers)? All existing code should be compatible, but it would trigger some popular linters.

dnwe:
@joewreschnig thanks, I pull in your changes and rebased them under PR #2635 so we can have a think about the API compatibility changes

PR #2635, opened 2023-08-31
feat(consumer): return err from offsets commit

dnwe:
This changes the `ConsumerGroupSession` and `OffsetManager` interface in a way that’s compatible for users of these interfaces, but incompatible for implementers.

Contributes-to: IBM/sarama#2417


dnwe:
As noted that even our own internals fail due the errcheck linter on offsetManager.Commit, it probably makes sense to defer this to `v2`

Issue #2637, opened 2023-08-31
Incorrect metadata refresh when the seed broker used for refresh is stale and moved to another Kafka cluster


HaoSunUber:
## Description

Hi, my name is Hao Sun from the Uber Kafka team. We plan to report an issue with the Sarama metadata refresh.

## Context about the Uber instance lifecycle management:

At Uber, our Kafka broker instance will be automatically replaced when there is a host hardware issue. Afterward, this replaced instance with the same IP address will sometimes be reused in another Kafka cluster and never belong to the original Kafka cluster.

## Stale broker Metadata Refresh Issue
In our current Sarama [metadata refresh logic](https://github.com/IBM/sarama/blob/f3c419405d597f94ea995e229ebd58bd1fab61e1/client.go#L1004), we will use the first available broker in the seed broker list to fetch the metadata. This seed broker list never gets changed after client initialization. 

1. Suppose we pass [broker1, broker2] in the seed broker list. Everything looks good after the Sarama client initialization. 
2. At some point, the broker2 got replaced internally due to a hardware issue. The Sarama client still looks good because it uses broker1 to fetch metadata, although broker2 still exists in the seed broker list.
3. Later, broker1 has some issues and becomes unavailable. According to the [tryRefreshMetadata](https://github.com/IBM/sarama/blob/f3c419405d597f94ea995e229ebd58bd1fab61e1/client.go#L1004) method, broker1 will be removed to dead broker list, and the Sarama client will move the second broker(broker2) to fetch metadata by using anyBroker method. Because the broker2 is online but in a different cluster, the client think this broker is healthy and fetch metadata from it. 
 
Unfortunately, the metadata from broker2 isn't from the right cluster. All cached metadata info in the client is messed up. When we use this client to produce/consume, we will encounter the topic partition error. The more dangerous case is if two Kafka cluster has topics with the same name, we are more likely to produce/consume the wrong topics from the wrong cluster. 

## Proposed fix:

[...]

PR #2645, opened 2023-09-07
fix: use least loaded broker to refresh metadata

HaoSunUber:
Seed brokers never change after client initialization. If the first seed
broker became stale (still online, but moved to other Kafka cluster),
Sarama client may use this stale broker to get the wrong metadata. To
avoid using the stale broker to do metadata refresh, we will choose the
least loaded broker in the cached broker list which is the similar to
how the Java client implementation works:

https://github.com/apache/kafka/blob/7483991a/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java#L671-L736

Contributes-to: https://github.com/IBM/sarama/issues/2637

### DIFF
--- a/client.go
+++ b/client.go
@@ -260,7 +260,7 @@ func (client *client) Broker(brokerID int32) (*Broker, error) {
 func (client *client) InitProducerID() (*InitProducerIDResponse, error) {
 	// FIXME: this InitProducerID seems to only be called from client_test.go (TestInitProducerIDConnectionRefused) and has been superceded by transaction_manager.go?
 	brokerErrors := make([]error, 0)
-	for broker := client.anyBroker(); broker != nil; broker = client.anyBroker() {
+	for broker := client.LeastLoadedBroker(); broker != nil; broker = client.LeastLoadedBroker() {
 		request := &InitProducerIDRequest{}
 
 		if client.conf.Version.IsAtLeast(V2_7_0_0) {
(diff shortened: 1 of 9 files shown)


functional_test.go
```diff
@@ -228,7 +228,7 @@ mainLoop:
 			}
 			for _, broker := range brokers {
 				err := broker.Open(client.Config())
-				if err != nil {
+				if err != nil && err != ErrAlreadyConnected {
```
HaoSunUber, line 231: Updated the functional tests to skip the ErrAlreadyConnected error because this change will use one broker from the broker list to refresh metadata, which is already connected when reaching the following validation.

HaoSunUber:
Updated the functional tests to skip the ErrAlreadyConnected error, because my change will use one broker in the broker cache to do metadata refresh, which is already connected before the following validation.

HaoSunUber:
@dnwe Could you please review my change? Thanks

client.go
```diff
@@ -763,22 +763,21 @@ func (client *client) registerBroker(broker *Broker) {
...
 
+	_, ok := client.brokers[broker.ID()]
+	if ok {
+		Logger.Printf("client/brokers deregistered broker #%d at %s", broker.ID(), broker.Addr())
+		delete(client.brokers, broker.ID())
+		return
+	}
 	if len(client.seedBrokers) > 0 && broker == client.seedBrokers[0] {
 		client.deadSeeds = append(client.deadSeeds, broker)
 		client.seedBrokers = client.seedBrokers[1:]
```
dnwe, line 780: Currently we only seem to deregister the broker from the seedBrokers list if it's the first element in the list — is that still the desired behaviour?

dnwe (approved):
Changes look good to me. One minor query on the existing deregisterBroker behaviour for the seedBrokers list, but happy to approve and merge as-is

client.go
HaoSunUber: Yes, I think it is expected. seed brokers only shuffle during client initialization or hard RefreshBrokers. After that, the cached broker list is empty, and the client will use the first seed broker to fetch metadata. deregisterBroker method deregisters the first seed broker will apply that moment when this first seed broker is unavailable.

HaoSunUber: Do you know when we have a new release including this change so that our side can use it? @dnwe

dnwe: Released as https://github.com/IBM/sarama/releases/tag/v1.41.2

PR #2646, opened 2023-09-07
perf: Alloc records in batch

ronanh:
This change aims to reduce pressure on GC by grouping allocations of records in a record batch.


dnwe (approved):
Thanks for the contribution! LGTM

PR #2668, opened 2023-10-10
Fix typo in balance strategy range documentation

ppizarro:


dnwe:
@ppizarro thanks! This was actually also spotted by @maksadbek and fixed under https://github.com/IBM/sarama/pull/2657, I just hadn't gotten around to reviewing and merging yet. Should now be fixed on main. Thanks to both for catching this problem 🙏🏻

PR #2672, opened 2023-10-13
fix(client): ignore empty Metadata responses when refreshing

### LINKED ISSUE #2664: Metadata refresh should skip updates when metadata response is empty
##### Description

After my [PR](https://github.com/IBM/sarama/commit/98ec384372ecbb9d86036c6f210f840c45dbfa70), the Sarama client will always pick up a least loaded broker from cached brokers to fetch metadata. Then the metadata response will update our cached brokers ([updateBroker](https://github.com/IBM/sarama/blob/main/client.go#L723-L745)). 

But if the chosen broker returns a response with empty brokers, updateBroker will clear our cached brokers into empty. The client will fall back to the seed broker to refresh metadata in the next metadata refresh because cached brokers are empty. This is what we want to avoid using the seed broker after bootstrap due to the stale metadata issue ([issues](https://github.com/IBM/sarama/issues/2637)).

Technically, the broker should not return an empty broker in the response. But we do come across this situation where the broker returns empty brokers without any error and emptys our cached brokers. It looks like a bug in [Kafka](https://issues.apache.org/jira/browse/KAFKA-7924). The Java client skips the empty response to update the metadata cache ([code](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java#L1149)). 

In Sarama, we should also implement the same logic to skip the update when the response is empty and move to the next candidate broker. 

I can make a PR if you think it is reasonable.

HaoSunUber:
We should skip the metadata refresh if the startup phase broker returns empty brokers in metadata response. The Java client skips the empty response to update the metadata cache (https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java#L1149) and we should make a feature parity in Sarama too

Fixes #2664

### DIFF
--- a/client.go
+++ b/client.go
@@ -1035,6 +1035,13 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int,
 		var kerror KError
 		var packetEncodingError PacketEncodingError
 		if err == nil {
+			// When talking to the startup phase of a broker, it is possible to receive an empty metadata set. We should remove that broker and try next broker (https://issues.apache.org/jira/browse/KAFKA-7924).
+			if len(response.Brokers) == 0 {
+				Logger.Println("client/metadata receiving empty brokers from the metadata response when requesting the broker #%d at %s", broker.ID(), broker.addr)
+				_ = broker.Close()
+				client.deregisterBroker(broker)
+				continue
+			}
 			allKnownMetaData := len(topics) == 0
 			// valid response, use it
 			shouldRetry, err := client.updateMetadata(response, allKnownMetaData)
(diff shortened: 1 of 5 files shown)


HaoSunUber:
@dnwe  Could you please take a look when getting a chance? Thanks

dnwe (approved):
@HaoSunUber thanks for finding and fixing this bug! Changes look good to me ⭐

HaoSunUber:
@dnwe  would you mind creating a new release including this change so that we can use it? Thanks so much

PR #2678, opened 2023-10-18
fix: data race in Broker.AsyncProduce

lzakharov:
Closes #2673

PR #2693, opened 2023-10-26
Asynchronously close brokers during a RefreshBrokers

bmassemin:
In production, we noticed that calling RefreshBrokers could take some time before closing the channels, which could cause a bottleneck on the producer.

This PR enables asynchronous closure of brokers (fire and forget) to make the client available as quickly as possible.

We have tested this modification in production, and the results have been satisfactory.

### DIFF
--- a/client.go
+++ b/client.go
@@ -519,16 +519,16 @@ func (client *client) RefreshBrokers(addrs []string) error {
 	defer client.lock.Unlock()
 
 	for _, broker := range client.brokers {
-		_ = broker.Close()
-		delete(client.brokers, broker.ID())
+		safeAsyncClose(broker)
 	}
+	client.brokers = make(map[int32]*Broker)
 
 	for _, broker := range client.seedBrokers {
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -519,16 +519,16 @@ func (client *client) RefreshBrokers(addrs []string) error {
 	defer client.lock.Unlock()
 
 	for _, broker := range client.brokers {
-		_ = broker.Close()
-		delete(client.brokers, broker.ID())
+		safeAsyncClose(broker)
 	}
+	client.brokers = make(map[int32]*Broker)
```
bmassemin, line 524: Instead of removing the brokers one by one, we just instantiate an new `map`

bmassemin:
Also, https://github.com/IBM/sarama/pull/2678 will help to resolve the situation we observe in prod.

dnwe (approved):
Thanks, this change looks reasonable to me

Issue #2382, opened 2022-11-09
Data Race with Incorrect Username or Password


PatrickDuncan:
I tracked down the root cause to this PR that was merged in 1.34.0 https://github.com/Shopify/sarama/pull/2234/files

##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka | Go |
|--------|-------|----|
| v1.37.2  |  2.13-2.7.1  | go version go1.18.2 darwin/amd64 |

##### Configuration

What configuration values are you using for Sarama and Kafka?
``` go
config.Producer.Flush.Bytes = 0
config.Producer.Flush.Frequency = 5 * time.Second
config.Producer.Flush.MaxMessages = 0
config.Producer.Idempotent = true
config.Producer.RequiredAcks = sarama.WaitForAll
config.Producer.Retry.Max = 3
config.Producer.Retry.Backoff = 100 * time.Millisecond
config.Producer.Return.Successes = true
config.Producer.Return.Errors = true
```

##### Logs

```
kafka server: SASL Authentication failed: Authentication failed: Invalid username or password
```

```
==================
WARNING: DATA RACE
Write at 0x00c00012e768 by goroutine 74:
  github.com/Shopify/sarama.(*Broker).Open.func1()
      /go/pkg/mod/github.com/!shopify/sarama@v1.37.2/broker.go:248 +0x1a44
  github.com/Shopify/sarama.withRecover()
      /go/pkg/mod/github.com/!shopify/sarama@v1.37.2/utils.go:43 +0x48
  github.com/Shopify/sarama.(*Broker).Open.func2()
      /go/pkg/mod/github.com/!shopify/sarama@v1.37.2/broker.go:178 +0x39

Previous read at 0x00c00012e768 by goroutine 72:
  github.com/Shopify/sarama.(*Broker).responseReceiver()
      /go/pkg/mod/github.com/!shopify/sarama@v1.37.2/broker.go:1160 +0x555
  github.com/Shopify/sarama.(*Broker).responseReceiver-fm()
      <autogenerated>:1 +0x39
  github.com/Shopify/sarama.withRecover()
      /go/pkg/mod/github.com/!shopify/sarama@v1.37.2/utils.go:43 +0x48
  github.com/Shopify/sarama.(*Broker).Open.func1.2()
      /go/pkg/mod/github.com/!shopify/sarama@v1.37.2/broker.go:251 +0x39

[...]

PR #2698, opened 2023-10-31
Fix data race on Broker.done channel

prestona:
The underlying case was not waiting for the goroutine running the `responseReceiver()` method to fully complete if SASL authentication failed. This created a window where a further call to `Broker.Open()` could overwrite the `Broker.done` channel value while the goroutine still running `responseReceiver()` was trying to close the same channel.

Fixes: #2382


dnwe (approved):
Excellent job tracking this one down!

Issue #2677, opened 2023-10-18
Offsets are reset on Azure EventHubs when a consumer starts


mfamador:
##### Description

Starting from version `1.41.0`, every time a consumer group starts or leaves, all running consumers from the same consumer group consume all messages all over again. Seems like the offsets are being reset and aren't commited again.
I couldn't reproduce the issue on a regular kafka, this only occurs apparently using Azure Event Hubs, so kafka version 1.0.0.
Reverting back to 1.40.1 everything works as expected.

##### Versions

| Sarama | Kafka | Go  |
| ------ | ----- | --- |
|  1.41.0,1.41.1, 1.41.2, 1.41.3     |   1.0.0    |  1.21.3   |

##### Configuration

```go
config := sarama.NewConfig()
config.Net.TLS.Enable = true
config.Net.SASL.Enable = true
config.Net.SASL.User = "$ConnectionString"
config.Net.SASL.Password = connectionString
config.Consumer.Offsets.Initial = sarama.OffsetOldest
config.Consumer.Offsets.AutoCommit.Enable = true
config.Consumer.Offsets.AutoCommit.Interval = time.Second
config.Version = sarama.V1_0_0_0
brokerAddresses := []string{brokerAddress}
consumerGroup := "foo"
consumer, err := sarama.NewConsumerGroup(brokerAddresses, consumerGroup, config)
```

##### Logs

<details><summary>logs: CLICK ME</summary>
<p>

```

Message received: key=, value={"a":"35fb0d14-a2d8-43a6-b5ad-2bd29dfd4741"}, parition=0, offset=0
Message received: key=, value={"a":"ce7ed018-70d3-43fb-9fed-e5fa45e93228"}, parition=0, offset=1
Message received: key=, value={"a":"3d9ccaff-1011-4ff8-bb02-46490f626c13"}, parition=0, offset=2
Message received: key=, value={"a":"7767f7a4-fbfa-43d5-ab0b-05f399095361"}, parition=0, offset=3
Message received: key=, value={"a":"a9540bf2-0be2-406b-8fba-a6c1fd3f6185"}, parition=0, offset=4
Message received: key=, value={"a":"edc0f104-31a1-4ee7-b82a-6e6fc982585a"}, parition=0, offset=5
Message received: key=, value={"a":"150fb310-28ff-42a7-8e63-bedd00419882"}, parition=0, offset=6
[...]

PR #2700, opened 2023-11-02
Fix default retention time value in offset commit

prestona:
The retention time field of the offset commit request uses -1 to mean "use the broker's default". Sarama uses the value 0 in the `Config.Consumer.Offsets.Retenton` field to mean "use the broker's default". Ensure that Sarama's default (0) is correctly mapped to the broker's default (-1).

Fixes: #2677


dnwe (approved):
Thanks for fixing this one up 👍🏻

dnwe:
Fix released in [Version 1.42.0](https://github.com/IBM/sarama/releases/tag/v1.42.0)

PR #2705, opened 2023-11-02
fix: make fetchInitialOffset use correct protocol

### LINKED ISSUE #2694: `OffsetFetchResponse` and `OffsetCommitRequest` `leaderEpoch` is not working as expected
##### Description
When using `PartitionOffsetManager` to reset consumer group offsets, we noticed that other consumers kept getting `offsetEpoch` = 0 in `FetchRequest`s

I think the culprit is that we are not setting
https://github.com/IBM/sarama/blob/main/offset_fetch_response.go#L7
to -1 by default

in `newPartitionOffsetManager` we invoke `fetchInitialOffset` which populates the `pom.leaderEpoch`. However, we use Version 1 for the `OffsetFetchRequest`: https://github.com/IBM/sarama/blob/main/offset_manager.go#L157

Since we don't set the `leaderEpoch` for Version < 5 ([ref](https://github.com/IBM/sarama/blob/main/offset_fetch_response.go#L21)), `PartitionOffsetManager` is always sets leaderEpoch = 0, which when doing a resetOffset and commit later on, will get committed with the incorrect `leaderEpoch` ([ref](https://github.com/IBM/sarama/blob/main/offset_manager.go#L333))

##### Versions

| Sarama | Kafka | Go  |
| ------ | ----- | --- |
| 1.41.1   | 2.7.1      |  1.21   |

##### Configuration

n/a

dnwe:
The `fetchInitialOffset` func in offsetManager was hardcoded to send a Version 1 OffsetFetchRequest rather than sending the appropriate version based on the config KafkaVersion. As discussed in #2694 this meant that LeaderEpoch was always being decoded as the default value '0' (because it was only returned in Version >= 5 OffsetFetchRequest). However, other areas of the offsetManager code were sending the newer protocol versions, so for example the OffsetCommitRequest would include a leader epoch value of 0 rather than an accurate one.

Correct this bug by sending the correct protocol version in fetchInitialOffset and also ensure we default to `-1` when we decode an OffsetFetchResponse of a Version < 5

Fixes #2694

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -1018,35 +1018,7 @@ func (ca *clusterAdmin) ListConsumerGroupOffsets(group string, topicPartitions m
 		return nil, err
 	}
 
-	request := &OffsetFetchRequest{
-		ConsumerGroup: group,
-		partitions:    topicPartitions,
-	}
-
-	if ca.conf.Version.IsAtLeast(V2_5_0_0) {
-		// Version 7 is adding the require stable flag.
-		request.Version = 7
-	} else if ca.conf.Version.IsAtLeast(V2_4_0_0) {
-		// Version 6 is the first flexible version.
-		request.Version = 6
-	} else if ca.conf.Version.IsAtLeast(V2_1_0_0) {
-		// Version 3, 4, and 5 are the same as version 2.
-		request.Version = 5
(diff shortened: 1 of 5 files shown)


hindessm (approved):
LGTM

prestona (approved):
LGTM

PR #2706, opened 2023-11-03
fix(config): relax ClientID validation after 1.0.0

### LINKED ISSUE #2697: ClientID regex is too restrictive.
##### Description
The sarama client allows to set the client.id in the (producer) configuration. It is not uncommon to use symbols like '|' in the client ID string. The regex does not allow this, however. See https://github.com/IBM/sarama/blob/82f0e48b0e2b6dfefc3088fe5e45195f6cc461d8/config.go#L18

dnwe:
The original validation regex was based on the one that existed in the Java clients pre-1.0.0 but KIP-190 removed the client-side validation because instead the brokers were updated to be able to cope with any clientID and sanitize it before using it in metrics etc.

We can do similar in Sarama and only do client-side validation when the user has specified a version number older than 1.0.0

Fixes #2697

### DIFF
--- a/config.go
+++ b/config.go
@@ -15,7 +15,9 @@ import (
 
 const defaultClientID = "sarama"
 
-var validID = regexp.MustCompile(`\A[A-Za-z0-9._-]+\z`)
+// validClientID specifies the permitted characters for a client.id when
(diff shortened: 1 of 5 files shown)


hindessm (approved):
LGTM

PR #2717, opened 2023-11-09
fix: prevent ConsumerGroup.Close infinitely locking

maqdev:
If context isn't cancelled, ConsumerGroup.Close can indefinitely lock on waiting for `c.lock` inside `c.leave()` which is holden by the `Consume` method.


maqdev:
@dnwe tests should pass now, there was an issue with the fix previously which I corrected

dnwe (approved):

PR #2720, opened 2023-11-10
fix(examples): housekeeping of code and deps

dnwe:
- standardise `-version` flag for controlling Version field in sarama.Config and default to sarama.DefaultVersion
- bump to mod directive of go1.18 to match parent module
- bump otel deps to latest and update interceptor code to match API changes


hindessm (approved):
LGTM

PR #2725, opened 2023-11-19
fix: Fixed data race in consumer group example

Kaushal28:
In the [consumer loop](https://github.com/IBM/sarama/blob/main/examples/consumergroup/main.go#L105), if any error occurs (other than `sarama.ErrClosedConsumerGroup`), then it'll retry calling `client.Consume` since it's infinite `for` loop. It will also assign a new ready channel at the end of the iteration. That is the race condition. Since it's in goroutine and the parent function will already be waiting on the different channel now (`<-consumer.ready`), which will keep waiting forever even if the `Setup()` closes the channel (because both the channels are now different).

Let me know if this makes sense. Please refer to the comments that I have added in the PR code changes below for more clarity.

I can see the same issue in other examples as well. But fixed this one for now. If this fix makes sense, I can also fix the other occurrences as well.

Here is the explanation of fix:

In this example, since we are only waiting at a single place on the `ready` channel, we'll close the `ready` channel only once (`sync.Once` is required since `Setup` can be called multiple times in events of rebalance).

Here is a similar fix that was required after following this buggy example: https://github.com/open-telemetry/opentelemetry-collector/pull/1696/files, which was required in my project as well because of the issue that I explained above reagarding infinite waiting in case of error.

Thanks.


examples/consumergroup/main.go
```diff
@@ -116,7 +116,6 @@ func main() {
 			if ctx.Err() != nil {
 				return
 			}
-			consumer.ready = make(chan bool)
```
Kaushal28, line 119: If we assign a new channel here in goroutine, the `Setup` will only close this new channel and the `<-consumer.ready` on line 123 below will be waiting on the older channel. So even if the new channel is closed, the wait will never end.

examples/consumergroup/main.go
```diff
@@ -162,13 +161,16 @@ func toggleConsumptionFlow(client sarama.ConsumerGroup, isPaused *bool) {
...
-	ready chan bool
+	ready       chan bool
+	readyCloser sync.Once
 }
 
 // Setup is run at the beginning of a new session, before ConsumeClaim
 func (consumer *Consumer) Setup(sarama.ConsumerGroupSession) error {
 	// Mark the consumer as ready
-	close(consumer.ready)
+	consumer.readyCloser.Do(func() {
```
Kaushal28, line 171: Only close the channel once since only the `main()` is waiting on the `ready` channel. No need to close again in cases of rebalance since no one is waiting after first close.

PR #2759, opened 2023-12-29
fix: don't waste time for backoff on member id required error

lzakharov:
This PR optimizes new session initialization in case of client with empty member id. There is no need to backoff before sending another join request with assigned id.


dnwe (approved):
Thanks LGTM

PR #2795, opened 2024-02-09
feat: update go directive to 1.19

dnwe:
Bumping minimum Go version to 1.19, aligning with compress and other dependencies which now require 1.19

```
go get go@1.19
go mod tidy
go fix ./...
```

Also bump GitHub workflows to use latest 1.22 by default, and use 1.21 (N-1) and 1.19 as the additional unittest runs for stdlib compatibility.


hindessm (approved):
LGTM

Issue #2801, opened 2024-02-14
Interceptor logger usage is invalid


puellanivis:
##### Description

In [interceptors.go](https://github.com/IBM/sarama/blob/v1.42.2/interceptors.go#L28) there is a `Logger.Printf` which uses `%s` for the “could be anything interceptor”, as well as `%w` assuming from `fmt.Errorf` that this is the formatting directive that should be used for `error`s, except that: a) `recover` doesn’t return an `error` it returns an `any` and b) even when `recover` returns an error, the `%w` verb is only allowed in `Errorf`. This code even fails `go vet`:
```
./interceptor.go:28:4: (*log.Logger).Printf format %w has arg r of wrong type interface{}
```

It still fails `go vet` even with `r.(error)`:
```
./interceptor.go:28:4: (*log.Logger).Printf does not support error-wrapping directive %w
```

This means we can end up with a log message similar to this:
```
[Sarama] 2009/11/10 23:00:00 Error when calling producer interceptor: {foo %!s(int64=42) %!s(complex128=(0+0i))}, %!w(runtime.errorString=invalid memory address or nil pointer dereference)
```

The fix is simply to change both of these directives to `%v`.

##### Versions

| Sarama | Kafka | Go  |
| ------ | ----- | --- |
|  v1.42.2 | irrelevant | any  |

##### Configuration

irrelevant

##### Logs

<details><summary>logs: CLICK ME</summary>
<p>

```
[Sarama] 2009/11/10 23:00:00 Error when calling producer interceptor: {foo %!s(int64=42) %!s(complex128=(0+0i))}, %!w(runtime.errorString=invalid memory address or nil pointer dereference)
```

</p>
</details>

PR #2802, opened 2024-02-14
Use %v formatting words and remove unnecessary newline

puellanivis:
Fixes https://github.com/IBM/sarama/issues/2801


dnwe (approved):
Thanks, good catch. Change LGTM

PR #2812, opened 2024-02-22
chore(ci): disable checkout credentials persist

dnwe:
We don't need actions/checkout to persist credentials for any submodules or other fetches. Whilst it is a read-only token, there's no need to leave it around on disk after the checkout has completed.

https://github.com/actions/checkout/blob/v4.1.1/action.yml#L48-L50


hindessm (approved):
LGTM

Issue #2630, opened 2023-08-29
backgroundMetadataUpdater can log ErrNoTopicsToUpdateMetadata repetitively if client.conf.Metadata.Full is false


dnwe:
##### Description

If the user has set `client.conf.Metadata.Full` to `false` then the refresh of metadata will log an error until the consumer has been assigned one or more topic partitions to refresh. It might also periodically log this if (e.g.,) metadata responses are throttled and don't contain any topics in the result

##### Versions

| Sarama | Kafka | Go  |
| ------ | ----- | --- |
| v1.41.0 | 3.5.1 | 1.21.0 |

##### Configuration

```go
client.conf.Metadata.Full = false
```

##### Logs

```
Client background metadata update: kafka: no specific topics to update metadata
```

##### Additional Context

N/A

dnwe:
Looks like this had also previously been rightly reported as an issue with the admin client https://github.com/IBM/sarama/issues/2069

famarting:
do you think this change would be a good way to address this issue? https://github.com/IBM/sarama/pull/2830

Issue #2069, opened 2021-11-19
Bug: cluster admin refresh medatadata report error when set `Metadata.Full` to false.


3AceShowHand:
##### Versions

When there is no topic in the Kafka cluster, metadata cannot be reported, if we set `Metadata.Full` to false.

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*

| Sarama | Kafka | Go |
|--------|-------|----|
| v1.12.7 |3.0.0  |1.17.2  |

##### Configuration

```
Config.Metadata.Full = false
```

What configuration values are you using for Sarama and Kafka?
``` go
```

![Screen Shot 2021-11-19 at 11 20 33 PM](https://user-images.githubusercontent.com/7138436/142647376-a83e12a4-afe1-4eef-b9c4-d880dcd41b55.png)

##### Logs

When filing an issue please provide logs from Sarama and Kafka if at all
possible. You can set `sarama.Logger` to a `log.Logger` to capture Sarama debug
output.

<details><summary>logs: CLICK ME</summary>
<p>

```
```

</p>
</details>

##### Problem Description

nkostoulas:
The docs do state:

> ErrNoTopicsToUpdateMetadata is returned when Meta.Full is set to false but no specific topics were found to update the metadata.

I too don't get the intuition of this though.

PR #2830, opened 2024-03-11
fix metadata disable full scan

famarting:
The changes in this PR address, but don't fix directly, this issues:
- https://github.com/IBM/sarama/issues/2069
- https://github.com/IBM/sarama/issues/2630

Also this will be of great help in Dapr: https://github.com/dapr/components-contrib/issues/3247

With this changes now it will be possible to set `config.Metadata.Full = false` but it would require the user to set a topic via
`config.Metadata.InitialTopics = []string{"dummy-topic"}` , with this the first request to fetch the metadata will either create that topic if it doesn't exist or it will just fetch its metadata


famarting:
ping, is there any interest in this contribution? I'm happy to debate it further or change anything that is needed

PR #2848, opened 2024-04-01
fix: provide IsMessageSizeTooLarge() function

### LINKED ISSUE #2655: regression: unable to handle detect errors related to messages being too large
(creating issue from comments I made on #2628 a number weeks ago)

The recent change to error handling in #2628 broke some of our running code. We had code that would:

```golang
if err == sarama.ErrMessageSizeTooLarge {
    // log a message, bump a metric, but don't retry
}
```

ie for our use-case, we want to ignore messages related to a message being too large. It's OK for us to throw these out. We used to check a size before sending into Sarama, but then we had some cases where our payload was under the limit, but once headers etc were added, it was over, so we switched to checking this error code. 

Noting that https://github.com/IBM/sarama/issues/543 reports that previous error code has been returned for at least 8 years, this seems a reasonably significant change... (and a good example of [Hyrum's Law](https://nordicapis.com/what-does-hyrums-law-mean-for-api-design/) )

It's now not clear to me how I should check for size related issues.

Interestingly the last PR which touched that line of code was 4 years ago in 2019 where https://github.com/IBM/sarama/pull/1262 reverts a change a few days earlier (https://github.com/IBM/sarama/pull/1218) that renamed a bunch of error variables because:

> renaming ErrXxx constants is an API changing change https://github.com/IBM/sarama/issues/1261

(that change is arguably less impactful than this one because it could be detected/corrected at compile time whereas we only caught this at runtime)

ae-govau:
This provides a method for producers to workaround the regression introduced by #2628. Fixes #2655.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -167,6 +167,22 @@ func (err ConfigurationError) Error() string {
 	return "kafka: invalid configuration (" + string(err) + ")"
 }
 
+const (
+	// messageSizeTooLargePrefix used by newMessageSizeTooLargeConfigurationError and IsMessageSizeTooLarge
+	messageSizeTooLargePrefix = "Attempt to produce message larger than configured Producer.MaxMessageBytes"
+)
+
+// newMessageSizeTooLargeConfigurationError creates an error returned when a message payload is
+// attempted to be published that exceeds the configured maximum message size. We return a specific
+// string that is checked for by IsMessageSizeTooLarge() as several consumers of this library
+// need to differentiate this error from other ConfigurationErrors, and the actual type returned
+// has changed multiple times during the evolution of this library. While the string check is a hack,
+// it's better that it's done once within this library, than multiple users reimplementing in their
+// own clients.
+func newMessageSizeTooLargeConfigurationError(msgSize, configuredSize int) ConfigurationError {
+	return ConfigurationError(fmt.Sprintf("%s: %d > %d", messageSizeTooLargePrefix, msgSize, configuredSize))
+}
+
 // KError is the type of error that can be returned directly by the Kafka broker.
 // See https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ErrorCodes
 type KError int16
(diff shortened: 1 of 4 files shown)


errors.go
```diff
@@ -455,3 +471,18 @@ func (err KError) Error() string {
...
+func IsMessageSizeTooLarge(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if errors.Is(err, ErrMessageSizeTooLarge) {
+		return true
+	}
+
+	return strings.Contains(err.Error(), messageSizeTooLargePrefix)
```
puellanivis, line 487: This _really_ shouldn’t be something that we bake in. Checking the string value of the error is considered The Wrong Way™ to check error conditions. And once we provide this, the API is kind of stuck with it no matter what.

I think returning this as a `ConfigurationError` is itself faulty. We should be returning a type that allows `errors.Is(err, ErrMessageSizeTooLarge)` to cover all the functionality necessary.

ae-govau: I'd argue encapsulating the string check as a private implementation detail of the library does the opposite of baking it in. e.g. currently users need to bake-in this logic (we have, as have others). Moving it to a function means that users won't need to update their logic each time this library chooses the change the error type that is returned (or changes the text in it).

The precise error returned for this condition has changed at least 3 times over the years so far:

- #2628 
- #1262
- #1218

I agree it's not great, but I'm not sure that there is an alterative to this change that doesn't change the error type returned for a 4th time. (Although I'd be supportive of a simple revert of #2628 which would return `ErrMessageSizeTooLarge` again).

puellanivis: I mean, it’s better than depending on raw strings copied into all our codebases, for sure. But I worry that it’s too much like putting wallpaper over a hole in the wall.

There will be a non-zero number of codebases that aren’t going to change their code to use this function, and just keep using the raw string, because “hey, if it ain’t broke, don’t fix it.” (Also, finding priority for maintenance work is almost always a struggle, since there’s a fairly ephemeral value to product.) And these people are still going to find themselves stuck out if/when the error text changes, and the breaking behavior change for these people will again be relatively silent until it causes the same critical/issue/alert/whatever that brought us all here.

🤷‍♀️ I’m just not sure… like I said, I don’t know if this improves the situation in any practical way. And in a year or two, someone will be looking at this code and wondering why we would ever double down on something much of Go considers an anti-pattern.

ae-govau: Have opened alternate PR #2851 and CCed you there. I prefer the approach in #2851 but the downside is that it changes the error type (again).

puellanivis: I mean, I kind of hope no one removed the `ErrMessageSizeTooLarge` handling right? 😰 Because this was still a value that the server could be returning, and one would likely want to handle the same as if the client refused to even send the message.

ae-govau:
Closing in favour of #2851

PR #2851, opened 2024-04-03
fix: return ErrMessageSizeTooLarge when message is too large

### LINKED ISSUE #2137: ErrMessageSizeTooLarge indicates a kafka server response when the error is in the Producer
##### Versions

| Sarama | Kafka | Go |
|--------|-------|----|
|  1.29.0 |  any | 1.17  |

##### Configuration

``` go
defaults
```

##### Logs

Logs aren’t especially important here (not a bug), except to say that the relevant error is `kafka server: Message was too large, server rejected it to avoid allocation error.`  

##### Problem Description

When the message size [exceeds](https://github.com/Shopify/sarama/blob/54a84a712aa0640cc71b27d6ad1d68a293e64e62/async_producer.go#L369-L372) `Producer.MaxMessageBytes`, the [error message](https://github.com/Shopify/sarama/blob/a37f45dcee6b1e95240c17a4c60c4a960d50eaeb/errors.go#L248-L249) indicates that the error is coming from the Kafka server, when in fact it’s a Producer error.

In our application, we recently spent a fair amount of time working with a customer, chasing down this issue -- we unproductively focused on the server while the error was in our software. https://github.com/cockroachdb/cockroach/issues/76258. The downstream server had a very large max message size (hundreds of MB), but the errors persisted, and it took a while to recognize that `Producer.MaxMessageBytes` (defaulted to 1MB) was the culprit.

We’d like to see a more accurate and descriptive error message.

Thanks for the consideration, I can mock up a PR to illustrate a preferred solution if it would be helpful.

ae-govau:
For most of this library's existence it has returned ErrMessageSizeTooLarge when the message exceeded the configured size.

For a short period in 2019 this error was renamed (#1218) but shortly revered back (#1262). Later in 2023 this error was changed to a ConfigurationError (#2628) to fix #2137, however this has caused issues with clients who rely on the previous error code being distinct from other ConfigurationError conditions (#2655).

This commit reverts to previous behaviour, and adds a test to pickup if this changes again in the future.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -450,9 +450,8 @@ func (p *asyncProducer) dispatcher() {
 			continue
 		}
 
-		size := msg.ByteSize(version)
-		if size > p.conf.Producer.MaxMessageBytes {
-			p.returnError(msg, ConfigurationError(fmt.Sprintf("Attempt to produce message larger than configured Producer.MaxMessageBytes: %d > %d", size, p.conf.Producer.MaxMessageBytes)))
+		if msg.ByteSize(version) > p.conf.Producer.MaxMessageBytes {
+			p.returnError(msg, ErrMessageSizeTooLarge)
 			continue
 		}
 
--- a/sync_producer_test.go
+++ b/sync_producer_test.go
@@ -58,8 +58,8 @@ func TestSyncProducer(t *testing.T) {
 		Value:    ByteEncoder(make([]byte, config.Producer.MaxMessageBytes+1)), // will exceed default max size, e.g. configuration side
 		Metadata: "test",
 	})
-	if err != ErrMessageSizeTooLarge {
-		t.Error("expected err to be ErrMessageSizeTooLarge - many people rely on this, please do not change with searching for previous discussions")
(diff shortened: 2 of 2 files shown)


ae-govau:
@puellanivis - this might be a better approach?  I know you like the `errors.Is()` but I added the equality check for the err in the tests as I suspect that's what most code-bases currently do.

@shermanCRL , @dnwe and @hindessm who reported #2137 and made #2628 which this effectively reverts who may be interested in this discussion and have comments.

An altnernate PR which doesn't change the current types returns is proposed in #2848.  Personally I prefer this approach in this PR (ie restore previous behaviour), but that unresolves #2137 which may or may not be OK.

puellanivis:
I mean, one of the first tests `errors.Is` does is check to make sure the errors are comparable and then `err == target`. So, testing with equality here isn’t so bad (and is a stronger assertion for the package API tests, that we’re returning directly the `ErrMessageSizeTooLarge` and not any wrapped version.

puellanivis:
To be more clear, I’m more behind this revert, rather than adding a new function to test an error string “safely”.

ae-govau:
Pushed additional commit to fix the lint. Force-push because I forgot the DCO the first time.

ae-govau:
It's not clear to me why that one test failed in one env. Any chance that's a flaky test?

PR #2858, opened 2024-04-10
Fixing error in using chan `consumer.ready` for waiting consumer has been set up

g0ne150:
Fixing error in using chan `consumer.ready` for waiting consumer has been set up.

### DIFF
--- a/examples/consumergroup/main.go
+++ b/examples/consumergroup/main.go
@@ -116,7 +116,7 @@ func main() {
 			if ctx.Err() != nil {
 				return
 			}
-			consumer.ready = make(chan bool)
+			consumer.ready <- true
 		}
 	}()
 


examples/consumergroup/main.go
```diff
@@ -116,7 +116,7 @@ func main() {
 			if ctx.Err() != nil {
 				return
 			}
-			consumer.ready = make(chan bool)
+			consumer.ready <- true
```
puellanivis, line 119: First, when signaling a broadcast-type semaphore semantic with channels, the proper action is to use `close()` as is performed in `consumer.Setup()` below, which is already being called by `client.Consume(…, …, &consumer)`.

In fact, this code should in all likelihood panic from trying to send on a closed channel.

The only way to exit out of this consumer goroutine is either via `log.Panicf()` where we don’t need to worry about the `<-consumer.ready` blocking forever, but also when `ctx.Err() != nil`, where we don’t do anything to signal to the main goroutine that it should also end.

A more appropriate bugfix here would be to have `<-consumer.ready` also listen for the `<-ctx.Done()`, which should in turn also be canceled by a `defer cancel()` at the top of the consumer goroutine. In such a case the main goroutine can also notify something like, “consumer died before it was ready, bailing out without starting up.”

PR #2864, opened 2024-04-18
refactor: isValidVersion readibility

CodingCookieRookie:


fetch_request.go
```diff
@@ -281,7 +281,7 @@ func (r *FetchRequest) headerVersion() int16 {
 }
 
 func (r *FetchRequest) isValidVersion() bool {
```
hindessm, line 283: This is clearly not equivalent. I suspect you automated the fix and match incorrectly on `<= 1` against `<= 11`? Personally I prefer the consistency of a range check in all the cases even when it is a simple range as it is easier to update (and easier to maintain when we switch to generating code from protocol json).

CodingCookieRookie: Yes thanks for noticing, missed those during the automated fix for all, edited it. I see, sure that makes sense as well :)

puellanivis:
I’m unsure of the value of this change. There is no functional behavior change, and we lose the consistency to the version range checks elsewhere else.

Issue #2797, opened 2024-02-11
Possible Message Batching Bug?


road-cycling:
##### Description

I'm noticing Kafka Broker batched message submissions are getting rejected due to the message being too large. Initially I thought this was due to the post compressed message size being too large however, logs added to the Kafka.Errors feedback channel (as shown below) displayed messages as small as 900B being rejected.

Logs
```
2024/02/11 05:32:57 Message Too Large(497922): kafka: Failed to produce message to topic metrics-debug: kafka server: Message was too large, server rejected it to avoid allocation error
2024/02/11 05:32:57 Message Too Large(2394856): kafka: Failed to produce message to topic metrics-debug: kafka server: Message was too large, server rejected it to avoid allocation error
2024/02/11 05:32:57 Message Too Large(919): kafka: Failed to produce message to topic metrics-debug: kafka server: Message was too large, server rejected it to avoid allocation error
2024/02/11 05:32:57 Message Too Large(188553): kafka: Failed to produce message to topic metrics-debug: kafka server: Message was too large, server rejected it to avoid allocation error
2024/02/11 05:32:57 Message Too Large(3805): kafka: Failed to produce message to topic metrics-debug: kafka server: Message was too large, server rejected it to avoid allocation error
2024/02/11 05:32:57 Message Too Large(1742883): kafka: Failed to produce message to topic metrics-debug: kafka server: Message was too large, server rejected it to avoid allocation error
2024/02/11 05:32:57 Message Too Large(6544): kafka: Failed to produce message to topic metrics-debug: kafka server: Message was too large, server rejected it to avoid allocation error
2024/02/11 05:32:57 Message Too Large(1216): kafka: Failed to produce message to topic metrics-debug: kafka server: Message was too large, server rejected it to avoid allocation error
2024/02/11 05:32:57 Message Too Large(1164809): kafka: Failed to produce message to topic metrics-debug: kafka server: Message was too large, server rejected it to avoid allocation error
2024/02/11 05:32:57 Message Too Large(6546): kafka: Failed to produce message to topic metrics-debug: kafka server: Message was too large, server rejected it to avoid allocation error
2024/02/11 05:32:57 Message Too Large(903): kafka: Failed to produce message to topic metrics-debug: kafka server: Message was too large, server rejected it to avoid allocation error
2024/02/11 05:32:57 Message Too Large(4911): kafka: Failed to produce message to topic metrics-debug: kafka server: Message was too large, server rejected it to avoid allocation error
[...]

PR #2892, opened 2024-05-08
fix: allow skipping per message size check

### LINKED ISSUE #2142: Compressed size is not used for validating message in producer when compression is enabled
##### Versions

Sarama Version : v1.29.1
Go version: 1.13

##### Configuration

``` 
Producer.Compression = sarama.CompressionGZIP
Producer.MaxMessageBytes = 2097164
```

##### Logs

`kafka: Failed to produce message to topic: kafka server: Message was too large, server rejected it to avoid allocation error.`

##### Problem Description
Even after setting producer compression by default to `sarama.CompressionGZIP`. We are noticing in asyn_producer.go message byte size validation checks against uncompressed message even though Kafka broker allows compressed message based on max size settings . This is cause issues sending message that has very low compressed size compared to uncompressed message. Can we get some help on it?
https://github.com/Shopify/sarama/blob/6693712f54b76066ea239255c30585832983947d/async_producer.go#L367

ae-govau:
This is useful when producer compression is enabled, as it allows for much larger messages to be sent to the broker (providing the entire batch they are in stays small enough).

The previous workaround suggested in #2142 causes other ill-effects related to message batching (such as in #2797).

Fixes #2142, probably also #2797 and may help mitigate not having merged #2851.

It's not clear how easy it is to write a test for this but happy to try if we're happy enough with this approach.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -450,10 +450,12 @@ func (p *asyncProducer) dispatcher() {
 			continue
 		}
 
-		size := msg.ByteSize(version)
-		if size > p.conf.Producer.MaxMessageBytes {
-			p.returnError(msg, ConfigurationError(fmt.Sprintf("Attempt to produce message larger than configured Producer.MaxMessageBytes: %d > %d", size, p.conf.Producer.MaxMessageBytes)))
-			continue
+		if !p.conf.Producer.SkipPerMessageSizeTest {
+			size := msg.ByteSize(version)
+			if size > p.conf.Producer.MaxMessageBytes {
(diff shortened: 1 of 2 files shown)


ae-govau:
This still doesn't solve everything - https://github.com/IBM/sarama/blob/0ab2bb77aeca321f41a0953a8c6f52472607a59e/produce_set.go#L253-L255 still is checking a before-compression amount for calculating the batch, but I think the above is probably better than erroneously returning errors for small messages when `MaxMessageBytes` is set too high.

PR #2893, opened 2024-05-08
fix(doc): correct JVM's config name corresponding to MaxWaitTime

abhipranay:
Fixed JVM config name for `MaxWaitTime` config in docs.

Incorrect JVM config: `fetch.wait.max.ms` <-- docs mention this.
[Correct JVM config](https://kafka.apache.org/36/documentation.html#consumerconfigs_fetch.max.wait.ms): `fetch.max.wait.ms`

### DIFF
--- a/config.go
+++ b/config.go
@@ -387,7 +387,7 @@ type Config struct {
 		// default is 250ms, since 0 causes the consumer to spin when no events are
 		// available. 100-500ms is a reasonable range for most cases. Kafka only
 		// supports precision up to milliseconds; nanoseconds will be truncated.
-		// Equivalent to the JVM's `fetch.wait.max.ms`.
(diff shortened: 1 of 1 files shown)


dnwe:
@abhipranay thanks, this is a good spot and worth correcting.

One small clarification is that `fetch.wait.max.ms` was previously correct and was used in the original Consumer implementation in Apache Kafka (which was eventually [deleted in 2.1.0](https://github.com/apache/kafka/commit/cc4dce94af)) that Sarama implemented in its first incarnation in the early days. When Kafka implemented [Consumer Groups](https://github.com/apache/kafka/commit/c24740c7b0) in 0.8.2 they renamed the property to fetch.max.wait.ms

dnwe (approved):

PR #2901, opened 2024-05-18
chore: broker

datager:


client.go
```diff
@@ -742,6 +742,18 @@ func (client *client) updateBroker(brokers []*Broker) {
 			Logger.Printf("client/broker remove invalid broker #%d with %s", broker.ID(), broker.Addr())
 		}
 	}
+
+	// if already created client.brokers connections, need close client.seedBrokers connections to avoid connection leak
+	if len(client.brokers) > 0 {
+		for _, b := range client.seedBrokers {
+			Logger.Printf("[close seed broker] start %v %v\n", b.ID(), b.Addr())
+			if err := b.Close(); err != nil {
+				Logger.Printf("[close seed broker] err %v %v\n", b.ID(), b.Addr())
```
puellanivis, line 751: We probably want to log what the error was, right?

datager: thanks a lot for review, i commited again to log err, could you please review again~
```go
		for _, b := range client.seedBrokers {
			Logger.Printf("[close seed broker] id: %v, addr: %v, start\n", b.ID(), b.Addr())
			if err := b.Close(); err != nil {
				Logger.Printf("[close seed broker] id: %v, addr: %v, err: %v\n", b.ID(), b.Addr(), err)
```

puellanivis: Yes, this looks better.

I can’t speak to the correctness of the code though.

PR #2909, opened 2024-05-23
fix: declare assignor variable for examples & clean up log format

kumakichi:


puellanivis (commented):
Yep, looks like these changes are appropriate, as they fix true errors.

dnwe (approved):
LGTM too — thanks!

PR #2914, opened 2024-05-25
feat: add metadata field to consumer message

fufuceng:
Hello everyone!

I've been using sarama for a while both in production and in my personal projects. One of the issues I'm lacking is that we don't have the chance to transfer information in the message we consume. I added the Metadata field in the produced messages into the consumed message structure. I think it will be very useful especially for tracing. 

I am waiting for your comments, thanks in advance.

### DIFF
--- a/async_producer_test.go
+++ b/async_producer_test.go
@@ -1633,6 +1633,26 @@ func (b *appendInterceptor) OnConsume(msg *ConsumerMessage) {
 	b.i++
 }
 
+type metadataInterceptor struct {
+	text string
+}
+
+func (m *metadataInterceptor) OnSend(msg *ProducerMessage) {
+	if m.text == "fail" {
+		panic("hey, the interceptor has failed")
+	}
+
+	msg.Metadata = m.text
+}
+
+func (m *metadataInterceptor) OnConsume(msg *ConsumerMessage) {
+	if m.text == "fail" {
+		panic("hey, the interceptor has failed")
+	}
+
+	msg.Metadata = m.text
+}
+
 func testProducerInterceptor(
 	t *testing.T,
 	interceptors []ProducerInterceptor,
@@ -1695,12 +1715,16 @@ func TestAsyncProducerInterceptors(t *testing.T) {
 		},
 		{
 			name:         "interceptor chain",
-			interceptors: []ProducerInterceptor{&appendInterceptor{i: 0}, &appendInterceptor{i: 1000}},
+			interceptors: []ProducerInterceptor{&appendInterceptor{i: 0}, &appendInterceptor{i: 1000}, &metadataInterceptor{text: "sarama"}},
 			expectationFn: func(t *testing.T, i int, msg *ProducerMessage) {
 				v, _ := msg.Value.Encode()
-				expected := TestMessage + strconv.Itoa(i) + strconv.Itoa(i+1000)
-				if string(v) != expected {
-					t.Errorf("Interceptor should have incremented the value, got %s, expected %s", v, expected)
+				expectedValue := TestMessage + strconv.Itoa(i) + strconv.Itoa(i+1000)
+				if string(v) != expectedValue {
+					t.Errorf("Interceptor should have incremented the value, got %s, expected %s", v, expectedValue)
+				}
+
+				if str, ok := msg.Metadata.(string); ok != true || str != "sarama" {
+					t.Errorf("Interceptor should have set the metadata, got %s, expected sarama", msg.Metadata)
 				}
 			},
 		},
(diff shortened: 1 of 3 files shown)


async_producer_test.go
```diff
@@ -1695,12 +1715,16 @@ func TestAsyncProducerInterceptors(t *testing.T) {
...
 				v, _ := msg.Value.Encode()
-				expected := TestMessage + strconv.Itoa(i) + strconv.Itoa(i+1000)
-				if string(v) != expected {
-					t.Errorf("Interceptor should have incremented the value, got %s, expected %s", v, expected)
+				expectedValue := TestMessage + strconv.Itoa(i) + strconv.Itoa(i+1000)
+				if string(v) != expectedValue {
+					t.Errorf("Interceptor should have incremented the value, got %s, expected %s", v, expectedValue)
+				}
+
+				if str, ok := msg.Metadata.(string); ok != true || str != "sarama" {
```
puellanivis, line 1726: Do not compare `bool` values to true or false, just use `!ok`

async_producer_test.go
```diff
@@ -1695,12 +1715,16 @@ func TestAsyncProducerInterceptors(t *testing.T) {
...
 		{
 			name:         "interceptor chain",
-			interceptors: []ProducerInterceptor{&appendInterceptor{i: 0}, &appendInterceptor{i: 1000}},
+			interceptors: []ProducerInterceptor{&appendInterceptor{i: 0}, &appendInterceptor{i: 1000}, &metadataInterceptor{text: "sarama"}},
 			expectationFn: func(t *testing.T, i int, msg *ProducerMessage) {
 				v, _ := msg.Value.Encode()
-				expected := TestMessage + strconv.Itoa(i) + strconv.Itoa(i+1000)
-				if string(v) != expected {
-					t.Errorf("Interceptor should have incremented the value, got %s, expected %s", v, expected)
+				expectedValue := TestMessage + strconv.Itoa(i) + strconv.Itoa(i+1000)
```
puellanivis, line 1721: There’s not really a reason we need to disambiguate `expected` here. There’s only one expected value here, so `expected` is a reasonable name for it.

consumer.go
```diff
@@ -21,6 +21,11 @@ type ConsumerMessage struct {
 	Topic      string
 	Partition  int32
 	Offset     int64
+
+	// This field is used to hold arbitrary data you wish to include
+	// Sarama completely ignores this field and is only to be used for
+	// pass-through data.
+	Metadata interface{}
```
puellanivis, line 28: We already have `Headers` … why wouldn’t we use that instead?

puellanivis (commented):
I’m failing to understand why this change should be at all necessary. We can already pass metadata in the `Headers` field, and the provided tests don’t indicate how this new Metadata field would be at all more useful than the existing Headers.

async_producer_test.go
fufuceng: this actually controls the value field, maybe it would be better if I change the `str` variable below it to `expectedMetadata`. What do you think?

consumer.go
fufuceng: Together with the interceptor, I think it will be useful for trace. Context or span can be moved down with this field. For example, in the otelsarama library, an extra channel is opened, messages are read and thrown back into this channel. It can cause performance problems under heavy load.  
This is the usage scenario for me, but those who need it in the OnConsume interceptor can move the information they want from here. Or those who want can also use this fielder in the handlers they write.

fufuceng: The header section consists of fields of type []byte and is highly related to kafka messages. I thought it would be nice to have a field that is not controlled by Sarama, but only by the user.

fufuceng: I updated the code according to the message above, please let me know if you want to go with `expected`. I will update to it again.

async_producer_test.go
fufuceng: I updated the code according to the message above, please let me know if you want to go with `expected`. I will update to it again.

async_producer_test.go
fufuceng: fixed, thanks

consumer.go
puellanivis: I suppose that’s where my concern comes from: the name isn’t immediately clear that this is a user-controlled side-channel… it sounds like it’s something related to Kafka functionality. As well, what if two interceptors try to use this field for cross-purposes? They would end up clobbering each other’s use, right?

As for putting contexts into the field, my concern is double from the note on the usage from the ̀ context` package:
> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.

(Yes, `http.Request` has a context embedded in it, this is because `http` is under go1 compat rules, and thus cannot change the signature of `http.ServeHTTP`, but they still had to put the context somewhere.)

fufuceng: https://github.com/IBM/sarama/commit/1568fa18fd6b00d379a98dc8acc6e9d14d2a940e

I followed the same approach with ProducerMessage struct. When we did this, I honestly wasn't too worried as the responsibility lies with the user. Nothing in Sarama should depend on/use this field.

puellanivis: Huh… 🤷‍♀️ yeah, ok then.

Issue #2907, opened 2024-05-22
High CPU usage in `newBrokerProducer.func2` 


dzy176:
##### Description

We found our prog running with high CPU usage.
I found a same closed issue -> https://github.com/IBM/sarama/issues/2251, but no reply to tell us how to resolve it.
In that case, sarama version is 1.33.0, but we already use a very new version 1.43.0 :(   , so I create a new issue here for help. 

##### Versions

| Sarama | Kafka | Go  |
| ------ | ----- | --- |
|  1.43.0      |3.6.1 | 1.22    |

##### Pprof
```shell
# go tool pprof http://127.0.0.1:9102/debug/pprof/profile
Fetching profile over HTTP from http://127.0.0.1:9102/debug/pprof/profile
Saved profile in /root/pprof/pprof.exporter.samples.cpu.001.pb.gz
exporter
Type: cpu
Time: May 22, 2024 at 1:47pm (CST)
Duration: 30.17s, Total samples = 30.05s (99.59%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top -cum
Showing nodes accounting for 27.04s, 89.98% of 30.05s total
Dropped 56 nodes (cum <= 0.15s)
Showing top 10 nodes out of 17
      flat  flat%   sum%        cum   cum%
     0.42s  1.40%  1.40%     29.94s 99.63%  github.com/IBM/sarama.(*asyncProducer).newBrokerProducer.func2
         0     0%  1.40%     29.94s 99.63%  github.com/IBM/sarama.withRecover
     8.28s 27.55% 28.95%     29.31s 97.54%  runtime.selectgo
     0.87s  2.90% 31.85%      9.43s 31.38%  runtime.selunlock
     1.33s  4.43% 36.27%      9.39s 31.25%  runtime.sellock
         0     0% 36.27%      8.56s 28.49%  runtime.unlock (inline)
     0.01s 0.033% 36.31%      8.56s 28.49%  runtime.unlockWithRank (inline)
     8.07s 26.86% 63.16%      8.55s 28.45%  runtime.unlock2
         0     0% 63.16%      8.06s 26.82%  runtime.lock (inline)
     8.06s 26.82% 89.98%      8.06s 26.82%  runtime.lock2
(pprof) list github.com/IBM/sarama
Total: 30.05s
[...]

PR #2922, opened 2024-06-05
Fix cpu spin loop when pending is closed while buf is not empty

AleksandrShibanov:
it can be situation when 
on line 854 pending is ok
on line 860 we add res to buf
on line 861 pending closes in other goroutine
and then there will be cpu spin loop. 

I will be glad to see your suggestions and comments


AleksandrShibanov:
https://github.com/IBM/sarama/issues/2907

puellanivis (commented):
While this resolves the immediate symptom, I’m unsure if this is a proper solution.

Issue #2923, opened 2024-06-06
Sincerely ask, is this redundant?


Trinoooo:
Version：Latest, v1.43.2
Description：
```
func safeAsyncClose(b *Broker) {
	tmp := b // local var prevents clobbering in goroutine
	go withRecover(func() {
		if connected, _ := tmp.Connected(); connected {
			if err := tmp.Close(); err != nil {
				Logger.Println("Error closing broker", tmp.ID(), ":", err)
			}
		}
	})
}
```
annotation says "local var prevents clobbering in goroutine", how it works? I mean when  `safeAsyncClose` is called, the local var `b` is just a copy pointing to the `Broker`, even if goroutine overwrites b, other goroutines will not perceive it.

thanks for help :)

puellanivis:
Looking at this, the code originates from https://github.com/IBM/sarama/commit/237cfb5380c1d4322b8f2b90f132a90f98e1778e

In this change, we can see that originally, the code had to duplicate the variable loop-local to avoid the issue solved by `loopvar`:
```
	for _, broker := range client.brokers {
		myBroker := broker // NB: block-local prevents clobbering
		go withRecover(func() { myBroker.Close() })
	}
```

But when this was extracted into a synchronous function call, that copying indeed became unnecessary.

PR #2925, opened 2024-06-11
fix(client): refactor duplicated replica+partition logic

Trinoooo:
1. extract the same logic in Replicas, InSyncReplicas, OfflineReplicas. Improve reusability, maintainability, and consistency
2. remove unnecessary local variable and misleading annotion. see https://github.com/IBM/sarama/issues/2923(#2923)


puellanivis (commented):
Looks all reasonable to me.

dnwe (approved):

PR #2933, opened 2024-07-04
feat: update go directive to 1.20

mauri870:
The github.com/klauspost/compress recently [deprecated go 1.19](https://github.com/klauspost/compress/pull/926) and moved to a go 1.20 directive in go.mod

### DIFF
--- a/examples/consumergroup/go.mod
+++ b/examples/consumergroup/go.mod
@@ -1,6 +1,6 @@
 module github.com/IBM/sarama/examples/consumer
 
-go 1.19
+go 1.20
 
 require github.com/IBM/sarama v1.43.1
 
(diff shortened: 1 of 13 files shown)


mauri870:
Wondering why we don't move to a supported go version, such as 1.21 and 1.22?

dnwe:
@mauri870 thanks for the PR — a bump to 1.20 is fine to allow us to keep pulling in compress updates. We don't tend to bump the go directive in `go.mod` further than that until we (or one of our direct dependencies) is using some newer part of the Go language that requires a newer toolchain in order to compile. Whilst Go themselves officially only maintain support for the latest N and N-1 versions, there will be various people using Sarama from Debian stable / Ubuntu LTS / RedHat Enterprise etc. and potentially using the Go toolchain supplied and maintained by the vendor with backported security patches etc.

dnwe (approved):

PR #2935, opened 2024-07-13
Add handling a new error code - 91

dimonleksin:
Go version 1.20
Kafka version 3.7.1(latest)
IF call DescribeUserScramCredentials with non-existent user - kafka returned error code 91 and description "Attempt to describe a user credential that does not exist: <user-name>"
In a Kafka Protocol Specification this not written, but in my practice this true.


errors.go
```diff
@@ -173,98 +173,99 @@ type KError int16
...
+	ErrFencedInstancedId                   KError = 82 // Errors.FENCED_INSTANCE_ID
+	ErrEligibleLeadersNotAvailable         KError = 83 // Errors.ELIGIBLE_LEADERS_NOT_AVAILABLE
+	ErrElectionNotNeeded                   KError = 84 // Errors.ELECTION_NOT_NEEDED
+	ErrNoReassignmentInProgress            KError = 85 // Errors.NO_REASSIGNMENT_IN_PROGRESS
+	ErrGroupSubscribedToTopic              KError = 86 // Errors.GROUP_SUBSCRIBED_TO_TOPIC
+	ErrInvalidRecord                       KError = 87 // Errors.INVALID_RECORD
+	ErrUnstableOffsetCommit                KError = 88 // Errors.UNSTABLE_OFFSET_COMMIT
+	ErrThrottlingQuotaExceeded             KError = 89 // Errors.THROTTLING_QUOTA_EXCEEDED
+	ErrProducerFenced                      KError = 90 // Errors.PRODUCER_FENCED
+	ErrDescribeUserScramCredentialNotExist KError = 91 // User not founded with describe users
```
puellanivis, line 268: https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/server/ZkAdminManager.scala#L805

The error does not seem to be attached to “Scram” specifically. 

The code is itself just “resource not found” https://fossies.org/dox/kafka-3.7.1-src/enumorg_1_1apache_1_1kafka_1_1common_1_1protocol_1_1Errors.html#a0232a0a06a6b03b6868b74862c6393e3

dimonleksin: Yes, code named by "resource not found", but used this error only with users. So, i can rename this in next commit?

puellanivis: Whatever this name is when merged becomes the name that it has to be to keep from now until we can make a breaking change.

So, ideally, we want this error to be descriptive but also generic enough to cover all of its usage. And breaking with the common format and concurrence with `Errors.RESOURCE_NOT_FOUND` would violate the principle least astonishment.

PR #2939, opened 2024-07-18
fix(producer): treat ErrKafkaStorageError as retriable

richardartoul:
This is retriable according to the spec: https://kafka.apache.org/protocol.html


puellanivis:
Indeed. 👍

richardartoul:
Could we merge this?

dnwe (approved):

Issue #2940, opened 2024-07-19
Re-ordering of offset commit requests can cause committed offset to move "backwards".


prestona:
##### Description

Offsets committed by multiple consumer group sessions can be re-ordered by the time they arrive at Kafka, potentially resulting in duplicate message delivery, should the offset be used to resume consumption.

If a single sarama.ConsumerGroup is consuming from more than one partition, then reordering of offset commit requests can occur.
This is because of a race condition between the goroutines used to run the ConsumerClaim method for different topic partitions.

When ConsumerGroupSession.Commit() is called, this ends up calling into offsetManager.flushToBroker().
This in turn:
1. Builds a offset commit request, using any so-far uncommitted offsets across all of the partitions belonging to the group
2. Finds the group's coordinator (if not already known)
3. Sends the offset commit request to the broker that is the group coordinator.

However, there is no locking to prevent the interleaving of two (or more) go routines between the building an offset commit request, and sending the request to Kafka.

For example, consider the following interactions between two goroutines P0 and P1 - each belonging to the same consumer group, but consuming from different partitions of a topic:
1. P0 marks offset 10 and calls commit (the uncommitted offsets tracked by the session are: {P0 -> 11})
2. P0 builds the offset commit request (containing {P0 -> 11}).
3. P0 goroutine yields, and P1 goroutine is scheduled
4. P1 marks offset 20 and calls commit (the uncommitted offsets tracked by the session are: {P0 -> 11, P1 -> 21})
5. P1 goroutine yields, and P0 goroutine is scheduled
6. P0 sends the offset commit ({P0 -> 11}) request, gets the response, and resumes running code in the ConsumerClaim method
7. P0 marks offset 11 and calls commit (the uncommitted offsets tracked by the session are {0->12, 1->21})
8. P0 builds the offset commit request (containing {0->12, 1->21}).
9. P0 sends its offset commit request (containing {0->12, 1->21})
8. P0 goroutine yields, and P1 goroutine is scheduled
10. P1 sends its offset commit request (containing {P0 -> 11, P1 -> 21}).

From the Kafka broker's perspective, it receives:
1. Offset commit request (step 6): {P0 -> 11}
2. Offset commit request (step 9) {P0 -> 12, P2 -> 21}
3. Offset commit request (step 10): {P0 -> 11, P2 -> 21}

This causes the offset committed for P0 to move "backwards" from 12 to 11.

Due to the nature of this race condition, it is more likely to occur if:
[...]

PR #2409, opened 2022-12-30
fix(metrics): fix race when accessing metric registry

vincentbernat:
A race condition was introduced in
5b04c98528ddcbb2e23cfaea937428e31ba232b2 (feat(metrics): track consumer-fetch-response-size) when passing the metric registry around to get additional metrics. Notably, `handleResponsePromise()` could access the registry after the broker has been closed and is tentatively being reopened. This triggers a data race because `b.metricRegistry` is being set during `Open()` (as it is part of the configuration).

We fix this by reverting the addition of `handleResponsePromise()` as a method to `Broker`. Instead, we provide it with the metric registry as an argument. An alternative would have been to get the metric registry before the `select` call. However, removing it as a method make it clearer than this function is not allowed to access the broker internals as they are not protected by the lock and the broker may not be alive any more.

All the following calls to `b.metricRegistry` are done while the lock is held:

- inside `Open()`, the lock is held, including inside the goroutine
- inside `Close()`, the lock is held
- `AsyncProduce()` has a contract that it must be called while the broker is open, we keep a copy of the metric registry to use inside the callback
- `sendInternal()`, has a contract that the lock should be held
- `authenticateViaSASLv1()` is called from `Open()` and `sendWithPromise()`, both of them holding the lock
- `sendAndReceiveSASLHandshake()` is called from
- `authenticateViaSASLv0/v1()`, which are called from `Open()` and `sendWithPromise()`

I am unsure about `responseReceiver()`, however, it is also calling `b.readFull()` which accesses `b.conn`, so I suppose it is safe.

This leaves `sendAndReceive()` which is calling `send()`, which is calling `sendWithPromise()` which puts a lock. We move the lock to `sendAndReceive()` instead. `send()` is only called from `sendAndReceiver()` and we put a lock for `sendWithPromise()` other caller.

The test has been stolen from #2393 from @samuelhewitt. #2393 is an alternative proposal using a RW lock to protect `b.metricRegistry`.

Fix #2320

PR #2941, opened 2024-07-19
Ensure ordering of offset commit requests

prestona:
Add locking to prevent concurrent calls to commit from potentially being re-ordered between determining which offsets to include in the offset commit request, and sending the request to Kafka.

Note: we think this can be improved upon (by not adding more locks). A potential improvement would be to use the [broker lock](https://github.com/IBM/sarama/blob/d2246cc9b9df113f3ffed7c442fd81e8b9cf736e/broker.go#L31), and only hold this until the request is sent (avoiding holding a lock for the entire network round-trip time). However, it looks like #2409 has made [sendAndReceive](https://github.com/IBM/sarama/blob/d2246cc9b9df113f3ffed7c442fd81e8b9cf736e/broker.go#L1038) hold the broker lock until a response is received from the broker, so we need to understand that a little better first.

Fixes: #2940

### DIFF
--- a/offset_manager_test.go
+++ b/offset_manager_test.go
@@ -3,6 +3,8 @@ package sarama
 import (
 	"errors"
 	"fmt"
+	"runtime"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


dnwe (approved):
Fix LGTM, though it would be helpful if we could possibly get a reproduce of the out-of-order commits into the FV suite in a followup PR?

prestona:
> Fix LGTM, though it would be helpful if we could possibly get a reproduce of the out-of-order commits into the FV suite in a followup PR?

Yep. Good point. The manual reproduce needed quite a few repetitions before we hit the window, but we can take a look and see what an FV suite test would look like, and how often it can be made to detect the condition.

offset_manager_test.go
```diff
@@ -105,6 +107,111 @@ func TestNewOffsetManager(t *testing.T) {
...
+			func() {
+				mu.Lock()
+				defer mu.Unlock()
+				if outOfOrder == "" {
+					for partition, offset := range req.blocks["topic"] {
+						last := lastOffset[partition]
+						if last > offset.offset {
+							outOfOrder =
+								fmt.Sprintf("out of order commit to partition %d, current committed offset: %d, offset in request: %d",
+									partition, last, offset.offset)
```
puellanivis, line 157: Consider using a `sync/atomic.Value` with a pointer to a `struct{ partition, last, offset IntType }` shape struct, to avoid the inherent serialization caused by mutex locking.

offset_manager_test.go
```diff
@@ -105,6 +107,111 @@ func TestNewOffsetManager(t *testing.T) {
...
+	om, err := NewOffsetManagerFromClient("group", testClient)
+	if err != nil {
+		t.Error(err)
+	}
+	defer safeClose(t, om)
+
+	const numPartitions = 10
+	const commitsPerPartition = 1000
+
+	wg := &sync.WaitGroup{}
```
puellanivis, line 190: Prefer simply declaring a variable `var wg sync.WaitGroup` rather than assigning a zero value / pointer to zero value.

offset_manager_test.go
```diff
@@ -105,6 +107,111 @@ func TestNewOffsetManager(t *testing.T) {
...
+		},
+	})
+	testClient, err := NewClient([]string{seedBroker.Addr()}, NewTestConfig())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer safeClose(t, testClient)
+	om, err := NewOffsetManagerFromClient("group", testClient)
+	if err != nil {
+		t.Error(err)
```
puellanivis, line 183: Can the test keep going after this?  `om` is almost certainly a nil pointer, so the first use is going to `nil` pointer exception, right?

offset_manager_test.go
```diff
@@ -105,6 +107,111 @@ func TestNewOffsetManager(t *testing.T) {
...
+				pom.MarkOffset(int64(c+1), "")
+				om.Commit()
+			}
+			wg.Done()
+		}()
+	}
+
+	wg.Wait()
+	mu.Lock()
+	defer mu.Unlock()
```
puellanivis, line 209: This lock should be unnecessary. The `wg.Wait()` ensures all writes should be complete, and performs all the synchronization necessary.

offset_manager_test.go
```diff
@@ -105,6 +107,111 @@ func TestNewOffsetManager(t *testing.T) {
 	}
 }
 
+// Test that the correct sequence of offset commit messages is sent to a broker when
+// multiple goroutines for a group are committing offsets at the same time
+func TestOffsetManagerCommitSequence(t *testing.T) {
+	mu := &sync.Mutex{}
```
puellanivis, line 113: Same as below, just go ahead and declare the variable instead: `var mu sync.Mutex`

offset_manager_test.go
```diff
@@ -105,6 +107,111 @@ func TestNewOffsetManager(t *testing.T) {
...
+				}
+			}
+			return resp
+		},
+		"OffsetCommitRequest": func(r *request) (rest encoderWithHeader) {
+			req := r.body.(*OffsetCommitRequest)
+			func() {
+				mu.Lock()
+				defer mu.Unlock()
+				if outOfOrder == "" {
```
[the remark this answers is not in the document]
prestona: Superseded by switch to sync/atomic.Pointer

offset_manager_test.go
prestona: Good idea. I ended up using an `atomic.Pointer, as it avoids the need to make a type assertion when calling `Load()`.

offset_manager_test.go
prestona: Fixed.

offset_manager_test.go
prestona: Well spotted. Substituted with `t.Fatal(...)`.

offset_manager_test.go
prestona: Superseded by switch to sync/atomic.Pointer. The intent was to ensure that the error message assigned to `outOfOrder` (by the goroutine running the handler in the mock broker) was guaranteed to be observed by the goroutine running the test function.

offset_manager_test.go
prestona: Superseded by switch to sync/atomic.Pointer - which I've declared as per your suggestion.

prestona:
@puellanivis, thanks for taking the time to review. I hope I've addressed your concerns in 21e6e58. I've also cherry-picked this change onto #2947 to keep that consistent.

offset_manager_test.go
puellanivis: Yeah, I figured it was added out of an abundance of caution. The information here (helpful apart from even this code review) is that the `wg.Wait()` itself is synchronizing, as all calls to `wg.Done()` _happen before_ the return of `wg.Wait()`, the same as if you had used a mutex.

puellanivis (commented):
Looks acceptable.

dnwe:
Superseded by 2947

PR #2942, opened 2024-07-19
Fix bugs related to Idempotency and Strict ordering when MaxOpenRequests=1

richardartoul:
I found a lot of bugs by just running a synthetic workload and randomly restarting brokers and changing partition assignments.


richardartoul:
Sorry opened this P.R too early, was trying to open it against my own fork

Issue #2860, opened 2024-04-13
Does sarama still guarantee message ordering?


wenyihu6:
##### Description

Based on kafka documentation, "allowing retries while setting enable.idempotence to false and max.in.flight.requests.per.connection to greater than 1 will potentially change the ordering of records".

Sarama has a default configuration with a max of 3 [retries] (https://github.com/IBM/sarama/blob/4ad35041300e1c15ba58b630745ac8eb05f30c10/config.go#L530), false idempotence, and a max of 5 [in flight requests per connection](https://github.com/IBM/sarama/blob/4ad35041300e1c15ba58b630745ac8eb05f30c10/config.go#L513).

Does this make the [ordering guarantee](https://github.com/IBM/sarama/wiki/Frequently-Asked-Questions#does-saramas-producer-guarantee-message-ordering) in sarama no longer true?

awoods187:
This is related to this bug https://github.com/IBM/sarama/issues/2619. We are observing this having a challenging impact on our downstream customers of Cockroach Labs

richardartoul:
Should be fixed by #2943 if someone can review

dnwe:
Believed to be fixed by https://github.com/IBM/sarama/pull/3422

3AceShowHand:
@dnwe Does #3422 is designed to work ONLY with `Producer.Idempotent` set to true ?


I still meet the out of order by the following configuration:
```
config.Producer.Retry.Max = 5
config.Net.MaxOpenRequests = 1
```

3AceShowHand:
I think #3422 did not fully close the retry ordering issue for non-idempotent producers.

The intent of #3422 looks broader than idempotent-only behavior: the PR description says partition muting prevents a subsequent batch for the same partition from being sent before  the first one is acknowledged, and `TestAsyncProducerRetryOrdering` explicitly covers the default non-idempotent producer with both `MaxOpenRequests=1` and `MaxOpenRequests=5`.

However, we observed a remaining out-of-order case in a our usage scenario where the`Producer.Idempotent` was not enabled, `Net.MaxOpenRequests=1` was set, and the data was dispatched to the same Kafka partition in the monotonic increasing order. 

During Kafka broker/controller failures, Kafka partition data contained a newer row
  before an older retry for the same topic-partition:

  - newer: commitTs `467053165727973505`
  - older: commitTs `467053165675545243`

[...]

Issue #2803, opened 2024-02-15
Sarama Async Producer Encounters 'Out of Order' Error: what are the reasons?


ghost:
##### Description

We are encountering an error (once every few weeks) while using the async producer in our Kafka setup. The error message encountered is as follows:

> assertion failed: message out of sequence added to a batch

This error seems to originate from the following line in the Sarama library:
[produce_set.go#L89](https://github.com/IBM/sarama/blob/33633d891de73c74ba05138b83d37233aa83850f/produce_set.go#L89)

The occurrence of this error is sporadic, and we are struggling to understand the underlying cause or identify any corrective measures. It appears that, occasionally, messages are being added to the batch in an incorrect order.

We are seeking insights or suggestions on what might be triggering this error. Our investigations have considered network issues as a potential cause; however, we have not found any corresponding logs or indicators to substantiate this theory when the error occurs.

##### Versions

| Sarama | Kafka | Go  |
| ------ | ----- | --- |
|  v1.42.1      |  2.6.2     |   1.20.6  |

##### Configuration

```go
	config := sarama.NewConfig()
	config.Version = version
	config.Consumer.Group.Rebalance.Strategy = sarama.NewBalanceStrategySticky()
	config.Producer.RequiredAcks = sarama.WaitForAll
	config.Producer.Idempotent = true
	config.Net.MaxOpenRequests = 1
	config.Producer.Retry.Max = 100000
	config.Producer.Retry.Backoff = 100 * time.Millisecond
	config.Producer.Return.Successes = true
	config.Producer.Return.Errors = true
	config.Producer.Partitioner = sarama.NewHashPartitioner
```

##### Logs

We are facing the error detailed at the following location:
[produce_set.go#L89](https://github.com/IBM/sarama/blob/33633d891de73c74ba05138b83d37233aa83850f/produce_set.go#L89)

##### Additional Context

All messages are dispatched using an asynchronous producer, configured with a high retry count to ensure message delivery even in the event of transient Kafka broker failures. Despite this, we observe that occasionally a message fails to be added to the batch, rendering it ineligible for any retry mechanism in Sarama.

nevillus:
With the setup described above, we encountered some instances of "The broker received an out of order sequence number" errors on the Kafka server. These occurrences are very rare too, but we are wondering if this could indicate an issue with how the messages are being pushed, leading to them being ordered incorrectly.

dnwe:
[...]

Issue #2619, opened 2023-08-22
AsyncProducer produces messages in out-of-order when retries happen


dethi:
##### Versions

| Sarama | Kafka | Go |
|--------|-------|----|
|  >= v1.31.0 | v3.5.0 |  1.20.7 |

##### Configuration

What configuration values are you using for Sarama and Kafka?
``` go
config.Version = sarama.V2_7_0_0
// Read success and errors
config.Producer.Return.Successes = true
config.Producer.Return.Errors = true
// Wait for Ack from all
config.Producer.RequiredAcks = sarama.WaitForAll
// Reduce internal buffer size to simplify the reproduction
config.ChannelBufferSize = 10
// Allow only one open requests at a time
config.Net.MaxOpenRequests = 1
// Retry almost indefinitely
config.Producer.Retry.Max = 1_000_000_000
// Reduce the max size, to flush more often and reproduce easily
config.Producer.MaxMessageBytes = 100
// We are only gonna produce to one partition to simplify the reproduction
config.Producer.Partitioner = sarama.NewManualPartitioner
```

##### Logs & Reproduction Code

I have created a Gist with all the information required to reproduce (code, configuration, etc), including some example logs:
https://gist.github.com/dethi/406fb5562030d8d0fa76db18d95bbbbe

###### Expectation

When producing message to the same partition, the [AsyncProducer guarantee the ordering of the produced messages](https://github.com/IBM/sarama/wiki/Frequently-Asked-Questions#does-saramas-producer-guarantee-message-ordering) even when retries are required, as long as `config.Net.MaxOpenRequests = 1`.

Idempotency shouldn't matter here.

###### Reality

Up until 1.31.0, the expectation hols. But the behaviour changed when [request pipelining was introduced to the AsyncProducer](https://github.com/IBM/sarama/pull/2094). Now, retries cause message to be published out of order.

An easy way to see this is by enabling `config.Producer.Idempotent`. It will result in the AsyncProducer returning an error to the caller when retries happen (like the partition leader disappear, i.e. broker disconnect):
```
kafka: Failed to produce message to topic sarama-outoforder-bug: kafka server: The broker received an out of order sequence number
```
When idempotency is not enabled, Sarama will publish successfully the messages, but in out-of-order.

###### Code Analysis / Flow of produced messages

(follow the links as you read each paragraph)

[...]

PR #2094, opened 2021-12-24
Support request pipelining in AsyncProducer

slaunay:
# Enhancement

Although there is a `Config.Net.MaxOpenRequests` parameter, the `AsyncProducer` uses at most one in-flight produce request when producing to a single Kafka broker.

As shown in #1539 a while back, using request pipelining increases throughput when writing to a Kafka cluster over a high latency network link and something that is recommended when implementing the Kafka protocol as described in
[their network section](https://kafka.apache.org/protocol#protocol_network):

> The broker's request processing allows only a single in-flight request per
> connection in order to guarantee this ordering. Note that clients can (and
> ideally should) use non-blocking IO to implement request pipelining and
> achieve higher throughput. i.e., clients can send requests even while
> awaiting responses for preceding requests since the outstanding requests
> will be buffered in the underlying OS socket buffer.

`MaxOpenRequests` defaults to 5 but if you were to execute the following code:
```go
for {
  // Blocks till we get a response from the broker or an error is returned
  res, err := broker.Produce(req)
  // Check for error or else use response
}
```
then you would only see a single in-flight `Produce` request.
That is, to send the next request you must first wait for the `responseReceiver` Go routine to read the current response bytes and return it to calling Go routine.

Obviously, if it takes a while to read such response because of network latency or slow replication, the calling Go routine will seat idle instead of sending more records therefore reducing the throughput (more on that in the `Testing done`
section).

Unfortunately this is how the `AsyncProducer` currently works, at least how the `brokerProducer` "bridge" Go routine processes batches (i.e. `*produceSet`):
https://github.com/Shopify/sarama/blob/635bcf350a7b8a92b4da0aacd288ee09311e673d/async_producer.go#L693-L707

By adding a new `Broker.AyncProduce` receiver and using it in the `AsyncProducer`, we can achieve proper pipelining of `Produce` requests.

The new receiver relies on a callback to eventually receive the response:
```go
for {
  // Blocks once we reach MaxOpenRequests in-flight requests
  err := broker.AsyncProduce(req, func(res *ProduceResponse, err error) {
    // Check for response error or else use response
  })
  // Check for request error
}
```

The same could technically be done with `Broker.Produce` and N go routines:
```go
for {
  go func() {
    res, err := broker.Produce(req)
    // Check for error or else use response
  }
}
```
but it gets tricky to get the back-pressure behavior from the internal `Brokers.responses` channel and avoid creating too many Go routines.

[...]

PR #2943, opened 2024-07-19
Fix broken error handling around Idempotent producer + Ensure strict ordering when Net.MaxOpenRequests = 1

richardartoul:
There have been numerous issues filed lately that:

1. Strict ordering no longer works after [request pipelining](https://github.com/IBM/sarama/pull/2094) was introduced even when `Net.MaxOpenRequests` is set to 1.
2. Requests can still end up failing / mis-sequenced when Idempotent producer is enabled.

I was able to reproduce these issues both locally and in a staging environment and have fixed both of them. P.R has many comments explaining the changes.

Relevant issues:

1. #2619
2. #2803
3. #2860

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -1255,6 +1255,12 @@ func (p *asyncProducer) retryBatch(topic string, partition int32, pSet *partitio
 		msg.retries++
 	}
 
+	// Previous versions of this library did not retry the call to p.client.Leader(topic, partition)
+	// if it failed. That was problematic because it meant during periods of time where we couldn't
+	// figure out who the leader was (because it had just died, or there was a network problem, or
+	// whatever) idempotent Produce requests would fail immediately instead of retrying for awhile
+	// as expected. This retry loop is very important since prematurely (and unnecessarily) failing
+	// an idempotent batch is ~equivalent to data loss.
 	succeeded := false
 	for i := 0; i < p.conf.Producer.Retry.Max; i++ {
 		// it's expected that a metadata refresh has been requested prior to calling retryBatch
--- a/async_producer.go
+++ b/async_producer.go
@@ -741,7 +741,7 @@ func (pp *partitionProducer) flushRetryBuffers() {
 		}
 
 		for _, msg := range pp.retryState[pp.highWatermark].buf {
-			if pp.parent.conf.Producer.Idempotent && msg.retries == 0 && msg.flags == 0 && !msg.hasSequence {
+			if pp.parent.conf.Producer.Idempotent && msg.retries == 0 && msg.flags == 0 {
 				if msg.hasSequence {
 					panic("assertion failure: reassigning producer epoch and sequence number to message that already has them")
 				}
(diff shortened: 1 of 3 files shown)


richardartoul:
Hmm... I think it may still be possible for concurrent requests to be issued sendResponse writes to a channel, and once the response is picked up off that channel there could be an inflight request from the next request + an inflight retry actually

async_producer.go
```diff
@@ -1168,18 +1268,39 @@ func (p *asyncProducer) retryBatch(topic string, partition int32, pSet *partitio
...
+
+		bp := p.getBrokerProducer(leader)
+		bp.output <- produceSet
+		p.unrefBrokerProducer(leader, bp)
+		succeeded = true
+		break
+	}
+
+	if !succeeded {
+		Logger.Printf("Failed retrying batch for %v-%d because of %v while looking up for new leader, no more retries\n", topic, partition)
```
puellanivis, line 1299: [nitpick] Newlines at the end should be unnecessary for loggers? (I mean, this is generally the case, but I don’t know if that is specifically true here.)

Three `%` verbs are specified but only two arguments are given.

async_producer.go
```diff
@@ -1168,18 +1268,39 @@ func (p *asyncProducer) retryBatch(topic string, partition int32, pSet *partitio
...
-	if err != nil {
-		Logger.Printf("Failed retrying batch for %v-%d because of %v while looking up for new leader\n", topic, partition, err)
+	// Previous versions of this library did not retry the call to p.client.Leader(topic, partition)
+	// if it failed. That was problematic because it meant during periods of time where we couldn't
+	// figure out who the leader was (because it had just died, or there was a network problem, or
+	// whatever) idempotent Produce requests would fail immediately instead of retrying for awhile
+	// as expected. This retry loop is very important since prematurely (and unnecessarily) failing
+	// an idempotent batch is ~equivalent to data loss.
+	succeeded := false
+	for i := 0; i < p.conf.Producer.Retry.Max; i++ {
```
puellanivis, line 1278: Suggest using a different variable name if we’re trying retries/tries rather than indices.

[off-by-one smell] Are we counting retries, or tries? That is, if I’ve asked for 5 retries max, then that’s 6 total tries.

puellanivis (commented):
I’m somewhat concerned by the number of `panic()`s in the code. Are these really so serious that they need to potentially crash the whole program? and/or kill off a processing goroutine, without any information up-the-chain that processing has terminated?

lifepuzzlefun:
any progress on this ？I think this bug is very urgent to fix

jgersdorf:
As we are seeing also a lot of `kafka server: The broker received an out of order sequence number` errors, we would be interessted in this fix.

jgersdorf:
I guess @richardartoul has not answered to @puellanivis review comments yet?

@richardartoul : Do you think you can continue with this PR, os is this PR abandoned?

ulpcan:
We are waiting for this MR to be continued

richardartoul:
Hey guys, I'm really sorry but I wasn't able to convince myself this P.R was correct or materially improved the situation. There is definitely a bug in the library, but I'm suspicious there is more than one issue and I timeboxed myself to two days on this and don't have more time to continue trying to fix it

PR #2947, opened 2024-07-23
fix(consumer): maintain ordering of offset commit requests

prestona:
Use the Broker.lock to prevent concurrent calls to commit from potentially being re-ordered between determining which offsets to include in the offset commit request, and sending the request to Kafka. This has the benefit that the lock is only held while the OffsetCommitRequest is sent, not for the period of time that Sarama is waiting for the response from the broker, improving the rate at which commit requests can be made.

The down side is that (without further re-factoring to the Broker) this change duplicates some of the code from the Broker into the offset manager. Potentially, making the code more fragile.

### DIFF
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -251,18 +251,23 @@ func (om *offsetManager) Commit() {
 }
 
 func (om *offsetManager) flushToBroker() {
-	req := om.constructRequest()
-	if req == nil {
-		return
-	}
-
 	broker, err := om.coordinator()
 	if err != nil {
 		om.handleError(err)
 		return
 	}
 
-	resp, err := broker.CommitOffset(req)
+	// Care needs to be taken to unlock this. Don't want to defer the unlock as this would
+	// cause the lock to be held while waiting for the broker to reply.
+	broker.lock.Lock()
+	req := om.constructRequest()
+	if req == nil {
+		broker.lock.Unlock()
+		return
+	}
+	resp, rp, err := sendOffsetCommit(broker, req)
+	broker.lock.Unlock()
+
 	if err != nil {
 		om.handleError(err)
 		om.releaseCoordinator(broker)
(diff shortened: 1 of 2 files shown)


offset_manager.go
```diff
@@ -251,28 +251,52 @@ func (om *offsetManager) Commit() {
...
 }
 
+func sendOffsetCommit(coordinator *Broker, req *OffsetCommitRequest) (*OffsetCommitResponse, *responsePromise, error) {
+	resp := new(OffsetCommitResponse)
+	responseHeaderVersion := resp.headerVersion()
+	promise, err := coordinator.send(req, true, responseHeaderVersion)
+	if err != nil {
+		return nil, nil, err
+	}
+	return resp, promise, err
```
puellanivis, line 297: `err` here must be `nil`. It’s usually better to use the stronger statement. (That is, the value must be `nil` so making it `nil` allows readers to not have to consider what value `err` might have here.)

The alternative is to just replace the whole `if err != nil { return nil, nil, err } return resp, promise, err` with `return resp, promise, err`.

prestona: Thanks for the careful review @puellanivis. I've added a further commit to based on your first suggestion.

I prefer keeping the if statement, as the alternative (`return resp, promise, err`) would return a non-nil `resp` value in the error case - which didn't seem quite right.

puellanivis: 👍 Yeah, I agree on keeping the `if` for the same reasons. Returning a non-nil `resp` in that case would be unexpected.

offset_manager_test.go
```diff
@@ -102,6 +107,105 @@ func TestNewOffsetManager(t *testing.T) {
...
+						Offset: 0,
+						Err:    ErrNoError,
+					}
+				}
+			}
+			return resp
+		},
+		"OffsetCommitRequest": func(r *request) encoderWithHeader {
+			req := r.body.(*OffsetCommitRequest)
+			if outOfOrder.Load() == nil {
```
puellanivis, line 147: Oh, I guess one point still: switch this to a guard statement.

```
if outOfOrder.Load() != nil {
	return
}
```

This allows for far less indention of the primary code path.

puellanivis: Although… I guess, you’re doing more work outside of just checking offsets… 🤦‍♀️ ignore.

dnwe (approved):
LGTM. After some further review and discussion lets go ahead and merge this version of the fix

PR #2954, opened 2024-08-07
fix(ci): correct docker-compose install

dnwe:
The base actions image no longer appears to include an old version of `docker-compose` so there's no thing to replace. Switch to installing `docker-compose` in a directory alongside the `docker` binary.

Also bump docker-compose to the latest release whilst making changes here.


prestona (approved):
LGTM

PR #2962, opened 2024-08-14
draft: added support incremental rebalance in one session

Sergey-Belyakov:
The idea is to get rid of stop the world during rebalancing without changing the main consumer interfaces. During the response from the heartbeat ErrRebalanceInProgress, the execution of fetch requests continues. In parallel, the reading from the departed partitions is closed.

tests is coming

### DIFF
--- a/config.go
+++ b/config.go
@@ -304,6 +304,8 @@ type Config struct {
 				Interval time.Duration
 			}
 			Rebalance struct {
+				// IsIncremental to allow don't stopping the rebalance process when a new member joins the group.
+				IsIncremental bool
 				// Strategy for allocating topic partitions to members.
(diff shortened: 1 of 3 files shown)


prestona:
Hi @Sergey-Belyakov - thanks for opening this PR. I wanted to check if you are aware of #2608, which also proposes changes to the re-balance strategy?

Sergey-Belyakov:
> Hi @Sergey-Belyakov - thanks for opening this PR. I wanted to check if you are aware of #2608, which also proposes changes to the re-balance strategy?

Hi! I have seen this MP, but I wanted to offer an incremental rebalancing option without changing the main interfaces of the consumer

PR #2979, opened 2024-09-07
Zstd encoder limit and raise encoder cache size

rtreffer:
This PR addresses #2965 with a different approach.

This PR addresses 2 issues with the current implementation
1. The number of in-use zstd encoders can exceed GOMAXPROCS if a large number of goroutines are used
2. The number of cached encoders is too low for highly parallel sarama use, leading to repeated encoder creation and thus low throughput

The PR preserves the following property
- Encoders are lazy created

The memory behavior of applications can change slightly.
Before applying the patch:

- The maximum memory usage from encoders was tied to the concurrency (goroutines) but would shrink to 1 on idle

After applying the patch:
- The maximum memory usage from encoders is tied to the peak parallelism per compression level

This should not change the worst case for the great majority of users, but it might be relevant in cases where applications were alternating between high sarama use and other uses.

There are 2 new benchmarks and a testing flag (zstdTestingDisableConcurrencyLimit) to verify the concurrency limiting.
I've also added some more information to the tests (like setting the bytes so throughput can be measured).
Here is a sample output from my machine (AMD framework 13):
```
# go test -benchmem -run=^$ -test.v -bench ^BenchmarkZstdMemory github.com/IBM/sarama
goos: linux
goarch: amd64
pkg: github.com/IBM/sarama
cpu: AMD Ryzen 7 7840U w/ Radeon  780M Graphics     
BenchmarkZstdMemoryConsumption
BenchmarkZstdMemoryConsumption-16                             16          68034969 ns/op        2959.16 MB/s            96.00 (gomaxprocs)               1.000 (goroutines)     21974595 B/op           815 allocs/op
BenchmarkZstdMemoryConsumptionConcurrency
BenchmarkZstdMemoryConsumptionConcurrency-16                  39          30498097 ns/op        8801.71 MB/s             4.000 (gomaxprocs)            256.0 (goroutines)       86327669 B/op          1479 allocs/op
BenchmarkZstdMemoryNoConcurrencyLimit
BenchmarkZstdMemoryNoConcurrencyLimit-16                      21          52053651 ns/op        5156.90 MB/s             4.000 (gomaxprocs)            256.0 (goroutines)       437548737 B/op         2196 allocs/op
PASS
ok      github.com/IBM/sarama   3.566s
```

### DIFF
--- a/zstd.go
+++ b/zstd.go
@@ -38,6 +38,9 @@ func newZstdEncoder(params ZstdEncoderParams) *zstd.Encoder {
 	}
 	zstdEnc, _ := zstd.NewWriter(nil, zstd.WithZeroFrames(true),
 		zstd.WithEncoderLevel(encoderLevel),
+		// 1MB is the default max message size in Kafka. We will usually not see messages larger than this.
+		// This results in a ~20% performance boost on high compression levels.
+		zstd.WithWindowSize(1024*1024),
 		zstd.WithEncoderConcurrency(1))
 	return zstdEnc
 }
--- a/zstd.go
+++ b/zstd.go
@@ -1,14 +1,11 @@
 package sarama
 
 import (
+	"runtime"
 	"sync"
 
 	"github.com/klauspost/compress/zstd"
 )
-
-// zstdMaxBufferedEncoders maximum number of not-in-use zstd encoders
-// If the pool of encoders is exhausted then new encoders will be created on the fly
-const zstdMaxBufferedEncoders = 1
 
 type ZstdEncoderParams struct {
 	Level int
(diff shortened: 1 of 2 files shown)


zstd.go
```diff
@@ -20,35 +17,64 @@ var zstdDecMap sync.Map
...
+		zstd.WithEncoderConcurrency(1))
+	return zstdEnc
+}
+
 func getZstdEncoder(params ZstdEncoderParams) *zstd.Encoder {
+	zstdMutex.Lock()
+	defer zstdMutex.Unlock()
+
+	limit := runtime.GOMAXPROCS(0)
+	for zstdCheckedOutEncoders >= limit && !zstdTestingDisableConcurrencyLimit {
```
HenryCaiHaiying, line 50: 1. Is `zstdCheckedOutEncoders` and the capacity of of zstdEncoderChannel are doing the same kind of control?
2. Is seems hard to follow if the `limit` value from line 49 are different from line 52 or line 29, might need some comment or explanation.

puellanivis: Some process could have changed the `GOMAXPROCS` in between calls. The proper behavior after calling a `sync.Semaphore.Wait()` is to refresh all values before performing any tests.

PR #2992, opened 2024-09-30
fix(lint): resolve IDENTICAL_BRANCHES issue in broker

frzifus:
The same code is executed when the condition "err != nil" is true or false, because the code in the if-then branch and after the if statement is identical.


puellanivis (commented):
10/10 no notes.

frzifus:
> 10/10 no notes.

mh?

puellanivis:
> mh?

I cannot Approve PRs here, but the PR is solid, and clean.

dnwe (approved):

Issue #1358, opened 2019-04-17
AsyncProducer retries causing OOM


hagen1778:
##### Versions

| Sarama | Kafka | Go |
|--------|-------|----|
| 1.20.1 |1.0.0  |1.11.3  |

##### Configuration

``` go
        conf := sarama.NewConfig()
	conf.Version = sarama.V0_11_0_0
	conf.Net.MaxOpenRequests = 5
	conf.Net.ReadTimeout = 5s
	conf.Net.DialTimeout = 5s
	conf.Net.WriteTimeout = 5s
	conf.Metadata.Retry.Max = 0
	conf.Producer.RequiredAcks = -1
	conf.Producer.Timeout = 5s
	conf.Producer.Flush.Bytes = 16MB
	conf.Producer.Flush.Frequency = 5s
	conf.Producer.Return.Errors = true
	conf.Producer.Return.Successes = true
```

##### Logs

Sarama logs recorded right before memory usage goes up:
<details><summary>logs: CLICK ME</summary>
<p>

```
producer/leader/<topic>/58 abandoning broker 10137
producer/leader/<topic>/14 selected broker 10128
producer/broker/10137 state change to [closed] on <topic>/80
producer/broker/10128 state change to [open] on <topic>/12
client/metadata fetching metadata for [<topic>] from broker <addr>:9092
producer/broker/10247 state change to [retrying] on <topic>/46 because kafka server: Request exceeded the user-specified time limit in the request.
producer/broker/10137 state change to [closed] on <topic>/18
producer/leader/<topic>/30 selected broker 10247
producer/broker/10247 state change to [retrying] on <topic>/30 because kafka server: Request exceeded the user-specified time limit in the request.
producer/broker/10137 state change to [open] on <topic>/58
producer/leader/<topic>/3 abandoning broker 10247
producer/broker/10247 state change to [closed] on <topic>/30
producer/broker/10247 state change to [closed] on <topic>/3
producer/leader/<topic>/55 state change to [normal]
producer/broker/10139 state change to [retrying] on <topic>/7 because kafka server: Request exceeded the user-specified time limit in the request.
```

</p>
</details>

##### Problem Description
The go-application writes messages into `AsyncProducer`. It also reads from `Success` and `Error` channels. The average memory usage of application is about 8GB. 
The problem is that during peak load application memory consumption could raise up to 4x which causes OOM.

[...]

PR #3026, opened 2024-12-05
feat: Prevent retry buffer in AsyncProducer to go OOM

wanwenli:
This commit adds an optional configuration to Sarama's retry mechanism to limit the size of the retry buffer. The change addresses issues #1358 and #1372 by preventing unbounded memory growth when retries are backlogged or brokers are unresponsive.

Key updates:
- Added `Producer.Retry.MaxBufferLength` configuration to control the maximum number of messages stored in the retry buffer.
- Implemented logic to handle overflow scenarios, ensuring non-flagged messages are either retried or sent to the errors channel, while flagged messages are re-queued.

This enhancement provides a safeguard against OOM errors in high-throughput or unstable environments while maintaining backward compatibility (unlimited buffer by default).

buf length 10k

Additionally, as demonstrated in the graph above, the retry buffer’s length is effectively constrained, confirming that the new feature functions as intended.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -13,6 +13,13 @@ import (
 	"github.com/rcrowley/go-metrics"
 )
 
+// ErrProducerRetryBufferOverflow is returned when the bridging retry buffer is full and OOM prevention needs to be applied.
+var ErrProducerRetryBufferOverflow = errors.New("retry buffer full: message discarded to prevent buffer overflow")
+
+// minFunctionalRetryBufferLength is the lower limit of Producer.Retry.MaxBufferLength for it to function.
+// Any non-zero maxBufferLength but less than this lower limit is pushed to the lower limit.
+const minFunctionalRetryBufferLength = 4 * 1024
+
 // AsyncProducer publishes Kafka messages using a non-blocking API. It routes messages
 // to the correct broker for the provided topic-partition, refreshing metadata as appropriate,
 // and parses responses for errors. You must read from the Errors() channel or the
@@ -1207,6 +1214,11 @@ func (bp *brokerProducer) handleError(sent *produceSet, err error) {
 // effectively a "bridge" between the flushers and the dispatcher in order to avoid deadlock
 // based on https://godoc.org/github.com/eapache/channels#InfiniteChannel
 func (p *asyncProducer) retryHandler() {
+	maxBufferSize := p.conf.Producer.Retry.MaxBufferLength
+	if 0 < maxBufferSize && maxBufferSize < minFunctionalRetryBufferLength {
+		maxBufferSize = minFunctionalRetryBufferLength
+	}
+
 	var msg *ProducerMessage
 	buf := queue.New()
 
@@ -1227,6 +1239,19 @@ func (p *asyncProducer) retryHandler() {
 		}
 
 		buf.Add(msg)
+
+		if maxBufferSize > 0 && buf.Length() >= maxBufferSize {
+			msgToHandle := buf.Peek().(*ProducerMessage)
+			if msgToHandle.flags == 0 {
+				select {
+				case p.input <- msgToHandle:
+					buf.Remove()
+				default:
+					buf.Remove()
+					p.returnError(msgToHandle, ErrProducerRetryBufferOverflow)
+				}
+			}
+		}
 	}
 }
 
--- a/async_producer_test.go
+++ b/async_producer_test.go
@@ -2244,6 +2244,82 @@ func TestTxnCanAbort(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestPreventRetryBufferOverflow(t *testing.T) {
+	broker := NewMockBroker(t, 1)
+	defer broker.Close()
+	topic := "test-topic"
+
+	metadataRequestHandlerFunc := func(req *request) (res encoderWithHeader) {
+		r := new(MetadataResponse)
+		r.AddBroker(broker.Addr(), broker.BrokerID())
+		r.AddTopicPartition(topic, 0, broker.BrokerID(), nil, nil, nil, ErrNoError)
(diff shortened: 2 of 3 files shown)


wanwenli:
@dnwe @puellanivis @rtreffer @klauspost I’ve submitted a PR addressing the retry buffer OOM issue (#1358, #1372) by introducing an optional configuration to limit the retry buffer’s size. The changes aim to prevent unbounded memory growth while maintaining backward compatibility (unlimited buffer by default). Your review and feedback would be greatly appreciated!

async_producer.go
```diff
@@ -13,6 +13,9 @@ import (
 	"github.com/rcrowley/go-metrics"
 )
 
+// ErrProducerRetryBufferOverflow is returned when the bridging retry buffer is full and OOM prevention needs to be applied.
+var ErrProducerRetryBufferOverflow = errors.New("discard a message and surface it as an error to prevent retry buffer to overflow")
```
puellanivis, line 17: This error message does not very well explain what the error here is. I think it’s suggesting a resolution, rather than describing what’s wrong?

I would probably recommend `what went wrong: how to resolve it` as a format if we want to suggest a resolution.

async_producer.go
```diff
@@ -1227,6 +1231,19 @@ func (p *asyncProducer) retryHandler() {
 		}
 
 		buf.Add(msg)
+
+		if maxBufferSize > 0 && buf.Length() >= maxBufferSize {
```
puellanivis, line 1235: It may make sense to put a higher lower limit on the maxBuffer, since setting this to `1` would basically block all sending forever? This would see any non-zero maxBuffer but less than this lower limit to be push to the lower limit.

async_producer.go
```diff
@@ -1227,6 +1231,19 @@ func (p *asyncProducer) retryHandler() {
...
+			msgToHandle := buf.Peek().(*ProducerMessage)
```
puellanivis, line 1236: Is there not a race condition here? In that, we merely `Peek()` and then operate based on what we peeked, but the `buf.Remove` is not guaranteed to be removing the same `*ProducerMessage` that we just peeked?

wanwenli: Thank you for raising this concern! In this implementation, the buf queue is exclusively accessed within the retryHandler function, which runs in a single goroutine. Since there is no concurrent access to buf, operations like Peek() and Remove() are safe and sequential, ensuring the correct message is handled.

Additionally, the same pattern of using Peek() followed by Remove() is already used in lines 1223 and 1224 of this file. https://github.com/IBM/sarama/blob/c1c3fa4c761722df6914c953cdc1bd3f05b74b23/async_producer.go#L1223 This consistency reinforces that the operations are expected to be thread-safe in the current single-threaded context.

Please correct me if i'm wrong.

puellanivis: 👍 Sounds good. My first thoughts are always concern about concurrency. But if `buf` is scope limited to this one goroutine, then you’re right. There’s no concerns about concurrency, as it’s exclusively held.

async_producer.go
wanwenli: @puellanivis your recommendation totally makes sense. I've added a const named `minFunctionalRetryBufferLength` at the top of this go file and used it in this `retryHandler` method.

In addition, I've modified the error message of `ErrProducerRetryBufferOverflow`, based on your comment. please take a look again. Thank you very much.

async_producer.go
```diff
@@ -13,6 +13,14 @@ import (
 	"github.com/rcrowley/go-metrics"
 )
 
+// ErrProducerRetryBufferOverflow is returned when the bridging retry buffer is full and OOM prevention needs to be applied.
+var ErrProducerRetryBufferOverflow = errors.New("retry buffer is full: message discarded to " +
+	"prevent buffer overflow. Check other errors and logs to determine the cause of excessive retries")
```
puellanivis, line 18: I don’t know what this particular project’s position is on line length, but usually the recommendation is to not unnaturally break up lines just to fit an arbitrary line length. https://go.dev/wiki/CodeReviewComments#line-length

Here, I think splitting up the string doesn’t improve understanding the code, and it can potentially break up “grepability” of log lines. In the logs, I will see `message discarded to prevent buffer overflow` but if I were to try and grep for this line, I wouldn’t find any matches.

Also, https://go.dev/wiki/CodeReviewComments#error-strings error strings should not end sentences, or start sentences, as they might be concatted with `fmt.Errorf()` to expand the log context.

async_producer.go
```diff
@@ -13,6 +13,14 @@ import (
...
+
+// minFunctionalRetryBufferLength is the lower limit of Producer.Retry.MaxBufferLength for it to function.
+// Any non-zero maxBufferLength but less than this lower limit is pushed to the lower limit.
+const minFunctionalRetryBufferLength = 5000
```
puellanivis, line 22: A limit of `4 * 1024` would be a better choice than a decimal round number. `io.Copy` as well defaults to 32 KiB.

async_producer.go
wanwenli: a good callout. I've shortened the error message to one line. thanks

async_producer.go
wanwenli: updated. please take another look

puellanivis (commented):
I don’t see any concerns. Though, I don’t know how appropriate a minimum functional retry buffer of 4 KiB would be. (I just knew it should be a power of 2 for being more allocator friendly.)

FYI: I cannot actually “approve” or even “resolve conversation”.

dnwe (approved):
@wanwenli thanks for submitting your PR and thanks @puellanivis for the review and giving useful early feedback. Looking over the changes I'm happy with the implementation and it provides a useful tuning mechanism for backpressure. Thanks again!

PR #3045, opened 2024-12-23
chore(test): include kafka 4.0.0 in FV testing

dnwe:
Whilst 4.0 isn't yet released, in order to allow us to do some up-front testing, compile from src and include in FV runs

Note: we have to include a server.properties for ZooKeeper mode as the 4.0 release no longer includes it. They only ship a Kraft-mode config file and we still want to run in ZooKeeper mode for now as its more convenient for the FV setup to be consistent


hindessm (approved):
LGTM

PR #3053, opened 2025-01-05
fix(admin): add retries for GroupCoordinator errors

dnwe:
- retry admin operations that rely on talking to the group coordinator for a given ID if the coordinator has changed from the cached value, or is not available
- also fixup controller retry in ElectLeaders to use the err from response rather than sendAndReceive
- also rename isErrNotController to isRetriableControllerError for consistency

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -3,6 +3,7 @@ package sarama
 import (
 	"errors"
 	"fmt"
+	"io"
 	"math/rand"
 	"strconv"
 	"sync"
@@ -144,6 +145,10 @@ type ClusterAdmin interface {
 	// locally cached value if it's available.
 	Controller() (*Broker, error)
 
+	// Coordinator returns the coordinating broker for a consumer group. It will
+	// return a locally cached value if it's available.
+	Coordinator(group string) (*Broker, error)
+
 	// Remove members from the consumer group by given member identities.
 	// This operation is supported by brokers with version 2.3 or higher
 	// This is for static membership feature. KIP-345
@@ -195,14 +200,25 @@ func (ca *clusterAdmin) Controller() (*Broker, error) {
 	return ca.client.Controller()
 }
 
+func (ca *clusterAdmin) Coordinator(group string) (*Broker, error) {
+	return ca.client.Coordinator(group)
+}
+
 func (ca *clusterAdmin) refreshController() (*Broker, error) {
 	return ca.client.RefreshController()
 }
 
-// isErrNotController returns `true` if the given error type unwraps to an
-// `ErrNotController` response from Kafka
-func isErrNotController(err error) bool {
-	return errors.Is(err, ErrNotController)
+// isRetriableControllerError returns `true` if the given error type unwraps to
+// an `ErrNotController` or `EOF` response from Kafka
+func isRetriableControllerError(err error) bool {
+	return errors.Is(err, ErrNotController) || errors.Is(err, io.EOF)
+}
+
+// isRetriableGroupCoordinatorError returns `true` if the given error type
+// unwraps to an `ErrNotCoordinatorForConsumer`,
+// `ErrConsumerCoordinatorNotAvailable` or `EOF` response from Kafka
+func isRetriableGroupCoordinatorError(err error) bool {
+	return errors.Is(err, ErrNotCoordinatorForConsumer) || errors.Is(err, ErrConsumerCoordinatorNotAvailable) || errors.Is(err, io.EOF)
 }
 
 // retryOnError will repeatedly call the given (error-returning) func in the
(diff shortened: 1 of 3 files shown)


admin.go
```diff
@@ -281,14 +288,17 @@ func (ca *clusterAdmin) CreateTopic(topic string, detail *TopicDetail, validateO
...
 		if err != nil {
 			return err
 		}
 		request := NewMetadataRequest(ca.conf.Version, topics)
 		response, err = controller.GetMetadata(request)
-		if isErrNotController(err) {
+		if err != nil {
+			return err
+		}
+		if isRetriableControllerError(err) {
```
puellanivis, line 301: Logically, this can only ever be called with `err == nil` due to lines 298 through 301.

admin.go
```diff
@@ -1043,72 +1064,111 @@ func (ca *clusterAdmin) ListConsumerGroups() (allGroups map[string]string, err e
...
+		if err != nil {
+			return err
+		}
 
-	return coordinator.FetchOffset(request)
+		defer func() {
+			if err != nil && isRetriableGroupCoordinatorError(err) {
+				_ = ca.client.RefreshCoordinator(group)
+			}
+		}()
```
puellanivis, line 1079: If you’re going to be accessing the `err` in a `defer` like this, it’s recommended to then name the return up on line 1069: `func() (err error)` this means that the properly scoped `err` will be set by the `return ErrWhatever`

This then prevents accidentally missing an `err = ErrWhatever` and even more so, guards against an (accidental) shadowing of `err` from preventing the ability to be able to do that assignment at all.

admin.go
```diff
@@ -1043,72 +1064,111 @@ func (ca *clusterAdmin) ListConsumerGroups() (allGroups map[string]string, err e
...
+
+		response, err = coordinator.FetchOffset(request)
+		if err != nil {
+			return err
+		}
+		if !errors.Is(response.Err, ErrNoError) {
+			err = response.Err
+			return err
```
puellanivis, line 1087: With the named return, this can just be `return response.Err` and the assignment to the function-return scoped `err` is automatic.

admin.go
```diff
@@ -1043,72 +1064,111 @@ func (ca *clusterAdmin) ListConsumerGroups() (allGroups map[string]string, err e
...
 		partitions: map[string][]int32{
 			topic: {partition},
 		},
 	}
 
-	resp, err := coordinator.DeleteOffsets(request)
-	if err != nil {
-		return err
-	}
+	return ca.retryOnError(isRetriableGroupCoordinatorError, func() error {
```
puellanivis, line 1105: Same as above, we should name the return value here, so the `defer` can guarantee usage of the correct returning value.

admin.go
```diff
@@ -1043,72 +1064,111 @@ func (ca *clusterAdmin) ListConsumerGroups() (allGroups map[string]string, err e
...
 	}
 	if ca.conf.Version.IsAtLeast(V2_0_0_0) {
 		request.Version = 1
 	}
 
-	resp, err := coordinator.DeleteGroups(request)
-	if err != nil {
-		return err
-	}
+	return ca.retryOnError(isRetriableGroupCoordinatorError, func() error {
```
puellanivis, line 1143: [repeat] name return

admin.go
```diff
@@ -1043,72 +1064,111 @@ func (ca *clusterAdmin) ListConsumerGroups() (allGroups map[string]string, err e
...
-	}
+		response, err = coordinator.DeleteGroups(request)
+		if err != nil {
+			return err
+		}
 
-	return nil
+		groupErr, ok := response.GroupErrorCodes[group]
+		if !ok {
+			return ErrIncompleteResponse
```
puellanivis, line 1162: [inconsistency]🤔 It’s unclear if leaving out the `err = ErrIncompleteResponse` is intentional or a bug?

`isRetriableGroupCoordinatorError(ErrIncompleteResponse)` should be false, so the lack of assignment to `err` should be correct either way. But having everything consistent and parallel helps cognitive load. (That is, we don’t have to wonder if skipping the `err` assignment is a bug or not.)

admin.go
```diff
@@ -1284,24 +1344,43 @@ func (ca *clusterAdmin) AlterClientQuotas(entity []QuotaEntityComponent, op Clie
...
+		GroupId: group,
 	}
 	for _, instanceId := range groupInstanceIds {
 		groupInstanceId := instanceId
 		request.Members = append(request.Members, MemberIdentity{
 			GroupInstanceId: &groupInstanceId,
 		})
 	}
-	return controller.LeaveGroup(request)
+	err := ca.retryOnError(isRetriableGroupCoordinatorError, func() error {
```
puellanivis, line 1362: [repeat] name return

admin.go
dnwe: Yeah I had planned to make that change, was still just getting something together that works in practice and passes the retry tests 😅

hindessm (approved):
LGTM

admin.go
puellanivis: 👍 Work first, polish second. :smart:

PR #3055, opened 2025-01-07
fix(ci): restore the Kafka 4.0.0 FV

dnwe:
Now that [KAFKA-17616](https://issues.apache.org/jira/browse/KAFKA-17616) has landed on the kafka [4.0 branch](https://github.com/apache/kafka/tree/4.0) it can no longer be run in ZooKeeper mode.

We'll need to work on a strategy for templating the docker-compose.yml in order to run KRaft quorum services instead of ZooKeeper if KAFKA_VERSION >= 4.0.0, but for now just pin to a commit before the removal started.


hindessm (approved):
LGTM

PR #3076, opened 2025-01-10
fix(sasl): add nilguard around token to prevent panic

hoo47:
### description
- Currently, there is no `Token` nil inspection.
- as-is: `if token.Extensions != nil && len(token.Extensions) > 0 {`
- to-be: `if token != nil && token.Extensions != nil && len(token.Extensions) > 0 {`

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -1570,7 +1570,11 @@ func (b *Broker) createSaslAuthenticateRequest(msg []byte) *SaslAuthenticateRequ
 func buildClientFirstMessage(token *AccessToken) ([]byte, error) {
 	var ext string
 
-	if token != nil && token.Extensions != nil && len(token.Extensions) > 0 {
+	if token == nil {
+		return []byte{}, fmt.Errorf("failed to build client first message: token is nil")
+	}
+
+	if len(token.Extensions) > 0 {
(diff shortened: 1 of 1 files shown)


broker.go
```diff
@@ -1570,7 +1570,7 @@ func (b *Broker) createSaslAuthenticateRequest(msg []byte) *SaslAuthenticateRequ
 func buildClientFirstMessage(token *AccessToken) ([]byte, error) {
 	var ext string
 
-	if token.Extensions != nil && len(token.Extensions) > 0 {
+	if token != nil && token.Extensions != nil && len(token.Extensions) > 0 {
```
puellanivis, line 1573: The `token.Extensions != nil` test is unnecessary as well. Since `token.Extensions == nil` implies `len(token.Extensions) == 0`.

puellanivis, line 1573: Actually, line 1580 dereferences `token` as well.

We should probably instead add a `if token == nil { return a short-circuit literal }` to avoid all possible derefs.

puellanivis:
👍

dnwe (approved):

dnwe:
Thanks! LGTM

PR #3081, opened 2025-01-16
fix(test): consumer group fetch request messages

stsmurf:
There is a bug in the tests with

```go
NewMockSequence(
    NewMockFetchResponse(t, 1).
        SetMessage("my-topic", 0, 0, StringEncoder("foo")).
        SetMessage("my-topic", 0, 1, StringEncoder("bar")),
    NewMockFetchResponse(t, 1),
)
```

This sets the message on the first `NewMockFetchResponse` twice and does not set the message on the second response. This was covered up in the test because of the `break` in the for loop over `claim.Messages()`. The update now passes the messages through a channel back to the test for assertion.  

In addition, the `context.WithCancel` was not having any affect on the test. Closing the group causes the context to be Done. This is verified by utilizing the same channel.

### DIFF
--- a/consumer_group_test.go
+++ b/consumer_group_test.go
@@ -13,20 +13,22 @@ import (
 )
 
 type handler struct {
-	*testing.T
-	cancel context.CancelFunc
+	messageCh chan *ConsumerMessage
 }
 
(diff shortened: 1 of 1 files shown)


dnwe (approved):
Thanks! Good spot

PR #3088, opened 2025-01-28
feat(producer): add MaxBufferBytes to limit retry buffer size

wanwenli:
# Summary

This PR introduces the MaxBufferBytes configuration to Sarama’s producer, allowing users to limit the retry buffer size based on the total byte size of messages. This complements the existing MaxBufferLength configuration, which limits the number of messages in the retry buffer. The implementation ensures more robust control over memory usage and prevents potential OOM issues when dealing with large messages.

The changes are built upon the foundation laid in https://github.com/IBM/sarama/pull/3026, further enhancing the retry buffer’s configurability and usability.

# Key Changes

- Added a new Producer.Retry.MaxBufferBytes configuration.
- The type of MaxBufferBytes is int64, as the maximum value of int32 (approximately 2 GB) is insufficient for buffers holding many messages of large average size, such as 1 MB. Using int64 ensures the buffer can handle higher limits if needed.
- Ensured that any value below the functional lower limit of 32 MB is pushed to this minimum.
- Updated the retry handler to enforce both MaxBufferBytes and MaxBufferLength constraints.
- Added relevant unit tests to validate functionality for both configurations, including edge cases and scenarios with unlimited settings.

This enhancement allows users to fine-tune producer behavior, particularly when working with large message payloads or memory-constrained environments.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -16,9 +16,16 @@ import (
 // ErrProducerRetryBufferOverflow is returned when the bridging retry buffer is full and OOM prevention needs to be applied.
 var ErrProducerRetryBufferOverflow = errors.New("retry buffer full: message discarded to prevent buffer overflow")
 
-// minFunctionalRetryBufferLength is the lower limit of Producer.Retry.MaxBufferLength for it to function.
-// Any non-zero maxBufferLength but less than this lower limit is pushed to the lower limit.
-const minFunctionalRetryBufferLength = 4 * 1024
+const (
+	// minFunctionalRetryBufferLength defines the minimum number of messages the retry buffer must support.
+	// If Producer.Retry.MaxBufferLength is set to a non-zero value below this limit, it will be adjusted to this value.
+	// This ensures the retry buffer remains functional under typical workloads.
+	minFunctionalRetryBufferLength = 4 * 1024
+	// minFunctionalRetryBufferBytes defines the minimum total byte size the retry buffer must support.
+	// If Producer.Retry.MaxBufferBytes is set to a non-zero value below this limit, it will be adjusted to this value.
+	// A 32 MB lower limit ensures sufficient capacity for retrying larger messages without exhausting resources.
+	minFunctionalRetryBufferBytes = 32 * 1024 * 1024
+)
 
 // AsyncProducer publishes Kafka messages using a non-blocking API. It routes messages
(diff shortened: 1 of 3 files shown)


async_producer_test.go
```diff
@@ -2199,57 +2200,86 @@ func TestPreventRetryBufferOverflow(t *testing.T) {
...
+					Value: val,
+				}
+				producer.Input() <- msg
+			}
+
+			producer.AsyncClose()
+			wg.Wait()
+
+			require.Equal(t, successes+producerErrors, tt.numMessages, "Expected all messages to be processed")
+			require.True(t, errorFound, "Expected at least one error matching ErrProducerRetryBufferOverflow")
```
puellanivis, line 2280: There’s not really a reason why these should be `require` instead of `assert`. https://go.dev/wiki/TestComments#keep-going

wanwenli: Thank you for your feedback! It helped me learn new things, and I’ve updated my code accordingly.

wanwenli:
@dnwe can you please review this PR? Thanks!

async_producer.go
```diff
@@ -1239,15 +1258,18 @@ func (p *asyncProducer) retryHandler() {
 		}
 
 		buf.Add(msg)
+		currentByteSize += int64(msg.ByteSize(version))
 
-		if maxBufferSize > 0 && buf.Length() >= maxBufferSize {
+		if (maxBufferLength > 0 && buf.Length() >= maxBufferLength) || (maxBufferBytes > 0 && currentByteSize >= maxBufferBytes) {
```
puellanivis, line 1263: [improvement?] Hm… 🤔  an idea would be to maybe invert the `if`, so the body could be an early `continue`. This follows https://go.dev/wiki/CodeReviewComments#indent-error-flow

puellanivis (commented):
The only thing left to comment about is a possible alternative, that may or may not be an improvement.

Otherwise, this looks like it does what it intends.

async_producer.go
wanwenli: thanks. I've taken your advice and inverted the `if` condition via De Morgan's laws. This improvement also reduces the indentation level.

dnwe (approved):
Thank you for this contribution, and for refining it based on the review feedback. The changes LGTM!

PR #3089, opened 2025-01-31
fix: remove redundant nil check

knbr13:
[mocks/async_producer.go](https://github.com/IBM/sarama/blob/e2fa7bf18c064719197abba64a2ce14c8227a92d/mocks/async_producer.go#L77) remove unnecessay nil check.

```
var slice []byte
if slice == nil || len(slice) == 0 {}
// is the same as
if len(slice) == 0 {}
```
because `len(slice)` where slice == nil is always equal to 0,

`staticcheck` will highlight this optimization with the following message:
```
mocks/async_producer.go:77:7: should omit nil check; len() for []*github.com/IBM/sarama/mocks.producerExpectation is defined as zero (S1009)
```

for reference: [staticcheck S1009](https://staticcheck.dev/docs/checks#S1009)

### DIFF
--- a/mocks/async_producer.go
+++ b/mocks/async_producer.go
@@ -74,7 +74,7 @@ func NewAsyncProducer(t ErrorReporter, config *sarama.Config) *AsyncProducer {
 				partitioners[msg.Topic] = partitioner
 			}
 			mp.l.Lock()
-			if mp.expectations == nil || len(mp.expectations) == 0 {
+			if len(mp.expectations) == 0 {
 				mp.expectations = nil
 				mp.t.Errorf("No more expectation set on this mock producer to handle the input message.")
 			} else {


puellanivis (commented):
Yep. 👍

dnwe:
Interesting that we’re not getting this flagged by staticcheck in our golangci-lint runs as we definitely have sc enabled [here](https://github.com/IBM/sarama/blob/e2fa7bf18c064719197abba64a2ce14c8227a92d/.golangci.yml#L75)

puellanivis:
> Interesting that we’re not getting this flagged by staticcheck in our golangci-lint runs as we definitely have sc enabled [here](https://github.com/IBM/sarama/blob/e2fa7bf18c064719197abba64a2ce14c8227a92d/.golangci.yml#L75)

https://github.com/IBM/sarama/blob/e2fa7bf18c064719197abba64a2ce14c8227a92d/.golangci.yml#L57
```
  disable-all: true
```

This might be related?

dnwe:
> ```
>   disable-all: true
> ```
> 
> This might be related?

That’s just the config equivalent of the cli flag to disable all the default linters and only enable explicitly defined ones

dnwe:
Very strange, run via golangci-lint we get no results for staticcheck even if we disable all config and default excludes:
```
% golangci-lint-1.62.2 run --no-config --disable-all --enable=staticcheck --exclude-use-default=false .
```

But standalone staticcheck flags a few different things including S1009:
```
% staticcheck ./...
balance_strategy.go:993:30: func (*partitionMovements).isLinked is unused (U1000)
balance_strategy.go:1028:30: func (*partitionMovements).in is unused (U1000)
balance_strategy.go:1043:30: func (*partitionMovements).hasCycles is unused (U1000)
balance_strategy.go:1075:30: func (*partitionMovements).isSticky is unused (U1000)
balance_strategy.go:1093:6: func indexOfSubList is unused (U1000)
metrics_helpers_test.go:120:6: func minMaxHistogramValidator is unused (U1000)
mocks/async_producer.go:77:7: should omit nil check; len() for []*github.com/IBM/sarama/mocks.producerExpectation is defined as zero (S1009)
mocks/mocks.go:55:41: error strings should not be capitalized (ST1005)
mocks/mocks.go:56:41: error strings should not be capitalized (ST1005)
```

knbr13:
the `staticcheck` linter used by `golangci-lint` isn't the same as the standalone staticcheck tool. 
according to `golangci-lint`'s documentation, the author of `staticcheck` does not approve its use as a library inside `golangci-lint`. 
this could explain why some checks, like [S1009](https://staticcheck.dev/docs/checks/#S1009), are flagged when running `staticcheck` directly but not via `golangci-lint`.

golangci-lint-docs-staticcheck

for reference: https://golangci-lint.run/usage/linters/#staticcheck

dnwe (approved):
Lgtm

PR #3101, opened 2025-02-17
fix: Fix the problem when some partitions have no leader 

liutao365:
When some topic partitions have no leader due to Kafka broker failures, the Sarama consumer group should be able to continue consuming partitions that have leaders and resume consuming the partitions that previously had no leader once they return to normal. This pull request addresses this issue.


client.go
```diff
@@ -113,6 +113,9 @@ type Client interface {
 	// LeastLoadedBroker retrieves broker that has the least responses pending
 	LeastLoadedBroker() *Broker
 
+	// check if partition is readable
+	ParttionNotReadable(topic string, partition int32) bool
```
puellanivis, line 117: [typo]
```suggestion
	PartitionNotReadable(topic string, partition int32) bool
```

client.go
```diff
@@ -1283,3 +1286,14 @@ type nopCloserClient struct {
...
+
+func (client *client) ParttionNotReadable(topic string, partition int32) bool {
+	pm := client.metadata[topic][partition]
+	if pm == nil {
+		return true
+	}
+	if pm.Leader == -1 {
+		return true
+	}
+	return false
```
puellanivis, line 1298: [simplify] We can simply return the bool conditional itself, rather than using an `if cond { return true } return false`:
```suggestion
	return pm.Leader == -1
```

consumer_group.go
```diff
@@ -864,6 +864,23 @@ func newConsumerGroupSession(ctx context.Context, parent *consumerGroup, claims
...
+			if parent.client.ParttionNotReadable(topic, partition) {
+				// partition not readable, wait for it to become readable
+				go func(topic string, partition int32) {
+					for {
+						if parent.client.ParttionNotReadable(topic, partition) {
+							time.Sleep(5 * time.Second)
+						} else {
+							break
+						}
+					}
```
puellanivis, line 876: [simplify] We can loop directly on conditional:
```suggestion
					for parent.client.ParttionNotReadable(topic, partition) {
						time.Sleep(5 * time.Second)
					}
```

Additionally, there are concerns about spinning up a goroutine without a clear lifetime. As well, that this seems to be an indefinite retry, potentially leaving the consumer in a degraded state without any clarity that it is retrying.

liutao365:
Add test case:
1. create a more than 6 partitions and 1 replication-factor (simplify the test case) topic and produce some millions messages .
2. shut down one of the kafka brokers, which will cause some of the topic partitions lose leader.
3. run the expamples/consumergroup/main.go and add a configuration 
image
4. restore the broker previously shut down.
5. describe the consumergroup and check all the partittions have no consumer lag

consumer_group.go
```diff
@@ -864,6 +864,26 @@ func newConsumerGroupSession(ctx context.Context, parent *consumerGroup, claims
...
+				// partition not readable, wait for it to become readable
+				go func(topic string, partition int32) {
+					for parent.client.PartitionNotReadable(topic, partition) {
+						select {
+						case <-ctx.Done():
+							return
+						case <-parent.closed:
+							return
+						default:
+							time.Sleep(5 * time.Second)
```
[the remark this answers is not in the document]
liutao365: good advice👍🏻

dnwe:
@liutao365 thanks for proposing this change, the approach looks good to me – I added a commit to fixup the client locking (the FV was failing for the race condition) and another commit to slightly refactor the consume partition code into a single named func to cover both paths. Can you take a look and confirm you're happy? Also @puellanivis if you wouldn't mind re-reviewing that would be great too.

We should probably add a unittest to cover this scenario, but I'm happy for us to do that under a follow-up PR as it is probably worth landing this fix sooner rather than later

liutao365:
> @liutao365 thanks for proposing this change, the approach looks good to me – I added a commit to fixup the client locking (the FV was failing for the race condition) and another commit to slightly refactor the consume partition code into a single named func to cover both paths. Can you take a look and confirm you're happy? Also @puellanivis if you wouldn't mind re-reviewing that would be great too.
> 
> We should probably add a unittest to cover this scenario, but I'm happy for us to do that under a follow-up PR as it is probably worth landing this fix sooner rather than later

Hi, @dnwe , I've red the changes and its better than my original commit, I'm happy with this improvements.

prestona (approved):
LGTM!

consumer_group.go
```diff
@@ -861,21 +861,37 @@ func newConsumerGroupSession(ctx context.Context, parent *consumerGroup, claims
...
+		sess.consume(topic, partition)
+	}
 
-				// consume a single topic/partition, blocking
-				sess.consume(topic, partition)
-			}(topic, partition)
+	// start consuming each topic partition in a goroutine
+	for topic, partitions := range claims {
+		for _, partition := range partitions {
+			go consumeTopicPartition(sess, topic, partition)
```
puellanivis, line 894: I mean, this function is now used exclusively here, and only here. So, assigning it to a local variable doesn’t seem all that necessary. Instead, we could just use:
```
go func(topic string, partition int32) {
  // function body here.
}(topic, partition)
```

(Also, there’s no reason to pass `sess` through, since it’s already available by closure. `topic` and `partition` are loopvars, so we still want to pass them through as parameters to get a local copy for go 1.21, which would otherwise misbehave if the variables were accessed through closure.)

Also, this would let us pull the `sess.waitGroup.Add(1)` out of the function body, but still keep it spatially close to the paired `Done()`. We usually want to prefer adding to the waitgroup outside of the `go func()` because we cannot be sure at what time the new goroutine will run, and thus performing the `Add` in the goroutine means there is a race condition within which a `wg.Wait()` could fail to block properly.

dnwe: Pulling the anonymous go routine func out into a named one was mostly aesthetic / personal preference, but in this scenario where it was nested within a double for loop, I felt it made it easier to at a glance understand what the current code was doing (spinning off a blocking goroutine for each topic-partition) and decoupled _what_ that goroutine was doing from the loop

In general I agree about incrementing waitGroups outside of the go routine. However, in this case we're only ever using the waitGroup.Wait() in cleanup to give any active consumers time to complete. If we haven't started the goroutine then the consumer won't be active and if it starts post-cleanup the sess context will already be done so consume won't proceed.

puellanivis: I can see esthetics of removing indentions. I personally like to avoid assigning things that are only used once to temporary values because one of the two hard problems in computers is naming things.

For me, this kind of breaks up the code flow (one of the ideas for why golang uses goroutines rather than promises to get async behavior). While, sure, it is nice to remove the indention levels of the anonymous go func, it still only executes in one place, and that’s the for loop. Like we would recommend spitting the body of a for loop out if we weren’t starting a new goroutine with it right? So, it just feels weird. It just kind of feels a bit too clever.

But that said, it’s not wrong, so… 🤷‍♀️

As for the `WaitGroup` handling, I can see an argument for why in this particular instance it isn’t necessarily racey, I like to practice something I call “code hygiene”. Like, even if I know my hands aren’t dirty in this specific case before I start handling food, I still make myself wash them anyways, because it helps build and reinforce good habits. (Though, of course, always beware of good habits becoming unnecessary as the language makes them unnecessary. c.f. the loopvar guarding.)

dnwe: No worries, I pondered this some more and agree — have moved back to an in-line approach by replacing last commit

PR #3108, opened 2025-02-27
produce: set MaxTimestamp

rockwotj:
Similar to other clients, set the MaxTimestamp computed from the records
in the batch.


puellanivis (commented):
I see nothing to comment on.

dnwe (approved):
Thanks for the contribution!

Issue #3107, opened 2025-02-26
sync.Pool for expectation channels in SyncProducer


kasimtj:
#### Description

We encountered excessive channel creations in SyncProducer.SendMessage/SendMessages methods, application profiling showed that 10% of cpu time spent on channel creation for callback purpose. Lifetime of this channels are short and seems that they can be reused instead of being closed 

#### Additional context

Profiling from production application

<img width="456" alt="Image" src="https://github.com/user-attachments/assets/ee3d9716-c7f5-4b70-b6cc-2df0160b5b23" />

I ran benchmarks locally, results are here https://github.com/IBM/sarama/pull/3106

dnwe:
Thanks, closing as fixed by #3109

PR #3109, opened 2025-02-27
feat(producer): add sync pool for channel reuse

kasimtj:
Fix for this issue https://github.com/IBM/sarama/issues/3107


sync_producer.go
```diff
@@ -110,32 +116,38 @@ func verifyProducerConfig(config *Config) error {
...
+	if pErr != nil {
 		return -1, -1, pErr.Err
 	}
 
 	return msg.Partition, msg.Offset, nil
 }
 
 func (sp *syncProducer) SendMessages(msgs []*ProducerMessage) error {
-	expectations := make(chan chan *ProducerError, len(msgs))
+	doneIDx := make(chan int, len(msgs))
```
puellanivis, line 133: Hm… is this `idx` to mean `index`? Then it should be `doneIdx`. The rule on capitalizing `ID` is when it’s short for “identification”. Maybe even just `doneidx` or `indices` also work, though?

puellanivis (commented):
That the only thing I can find to comment about is a variable name says good things for the PR. 😉

sync_producer.go
kasimtj: changes to indices, looks cleaner

kasimtj:
> I can see nothing worth commenting about.

should i tag maintainers for a review or just wait for someone?

dnwe (approved):
Thanks for the PR and issue/benchmark comparison. Changed look good to me, approved

PR #3113, opened 2025-03-02
chore: bump minimum Go version to 1.23.0

dnwe:
Dependencies are all starting to move to go 1.23.0 going forwards, so we'll sadly have to start bumping Sarama's go directive more aggressively as well in order to keep up. The Go maintainers have unfortunately decided that:

> A module’s go line must declare a version greater than or equal to the
> go version declared by each of the modules listed in require
> statements.

and

> Before Go 1.21, Go toolchains treated the go line as an advisory
> requirement: if builds succeeded the toolchain assumed everything
> worked, and if not it printed a note about the potential version
> mismatch. Go 1.21 changed the go line to be a mandatory requirement
> instead.

So we really have no option here, especially with the golang.org/x modules bumping to 1.23.0 as soon as 1.24.0 was released, we can no longer continue to support older Go toolchains.

Also bump CI testing to use the "oldstable" and "stable" aliases for N-1 and N going forwards as we're unlikely to want to test with anything else.

### DIFF
--- a/.github/workflows/apidiff.yml
+++ b/.github/workflows/apidiff.yml
@@ -15,6 +15,10 @@ on:
 permissions:
   contents: read  # for actions/checkout to fetch code
 
+env:
+  # Use the Go toolchain installed by setup-go
+  GOTOOLCHAIN: local
+
 jobs:
   apidiff:
     runs-on: ubuntu-latest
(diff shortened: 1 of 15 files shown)


hindessm (approved):
LGTM

Issue #3119, opened 2025-03-05
Issue hitting max records in a batch


EvanJGunn:
### The Issue
My team is getting this invalid array length error in our consumer:
`{"level":"info","msg":"consumer/broker/1395569342 disconnecting due to error processing FetchRequest: kafka: error decoding packet: invalid array length","time":"2025-02-27 20:17:36"}`

I tracked the error down and it looks like its happening here:
https://github.com/IBM/sarama/blob/9bf344fc6f900fc6d3bc2b0746d63e4f21017771/record_batch.go#L160
https://github.com/IBM/sarama/blob/9bf344fc6f900fc6d3bc2b0746d63e4f21017771/real_decoder.go#L116

Unless I am mistaken, it appears that the maximum number of records in a batch is set at `131070`.

### The Question
Is there a reason the max records in a batch is set at `131070`?

### More Background
My team is consuming a lot of small records from confluents warpstream product. For whatever reason, it looks like Fetch Max gets ignored by the warpstream product, so that unfortunately is not a solution for us.

puellanivis:
Following commits back to the [first commit with this limit](https://github.com/IBM/sarama/commit/c9508e324af8d5e9877a7de159ead30a5bb81363#diff-972d571100b3acb73abcdcec3be40d47a7b8672f90163456fde0be2cdf82d2bdR42-R44):

```
	if tmp > pd.avail() || tmp > 2*math.MaxUint16 {
		return -1, errors.New("kafka getArrayCount: unreasonably long array")
	}
```

It seems it was just an arbitrary choice to prevent trying to allocate “unreasonable” array lengths. We probably ought to limit it against how many objects could remain if the whole rest of the packet were the array. (Or the number of bytes remaining.) This should prevent most malicious encodings?

richardartoul:
@puellanivis yeah I think what you're suggesting makes sense. Usually an invalid encoding will result in a stupendously large number anyways, there's a lot more wiggle room, but kafka batches can be really large in terms of record count, especially if the messages are small

rmb938:
Hey All, I made a PR to fix this issue https://github.com/IBM/sarama/pull/3120

EvanJGunn:
I tested using this branch and it appears to work perfectly. Thank you.

dnwe:
> My team is consuming a lot of small records from confluents warpstream product. For whatever reason, it looks like Fetch Max gets ignored by the warpstream product, so that unfortunately is not a solution for us.

[...]

PR #3120, opened 2025-03-05
fix(decoder): use configurable limit for max number of records in a record batch

rmb938:
Replace the `tmp > 2*math.MaxUint16` arbitrary number safety blanket with `tmp > MaxResponseSize` as we already have that constant (that a user can change via sarama.MaxResponseSize = ...) which is what we use for MaxBytes in a FetchRequest.

Also added a test for large record counts.

Fixes: #3119

### DIFF
--- a/real_decoder.go
+++ b/real_decoder.go
@@ -121,6 +121,20 @@ func (rd *realDecoder) getArrayLength() (int, error) {
 	return tmp, nil
 }
 
+func (rd *realDecoder) getArrayLengthNoLimit() (int, error) {
+	if rd.remaining() < 4 {
+		rd.off = len(rd.raw)
+		return -1, ErrInsufficientData
+	}
+	tmp := int(int32(binary.BigEndian.Uint32(rd.raw[rd.off:])))
+	rd.off += 4
+	if tmp > rd.remaining() {
+		rd.off = len(rd.raw)
+		return -1, ErrInsufficientData
+	}
+	return tmp, nil
+}
+
 func (rd *realDecoder) getCompactArrayLength() (int, error) {
 	n, err := rd.getUVarint()
 	if err != nil {
--- a/real_decoder.go
+++ b/real_decoder.go
@@ -115,22 +115,8 @@ func (rd *realDecoder) getArrayLength() (int, error) {
 	if tmp > rd.remaining() {
 		rd.off = len(rd.raw)
 		return -1, ErrInsufficientData
-	} else if tmp > 2*math.MaxUint16 {
+	} else if tmp > int(MaxResponseSize) {
 		return -1, errInvalidArrayLength
-	}
-	return tmp, nil
-}
-
-func (rd *realDecoder) getArrayLengthNoLimit() (int, error) {
-	if rd.remaining() < 4 {
-		rd.off = len(rd.raw)
-		return -1, ErrInsufficientData
-	}
-	tmp := int(int32(binary.BigEndian.Uint32(rd.raw[rd.off:])))
-	rd.off += 4
-	if tmp > rd.remaining() {
-		rd.off = len(rd.raw)
-		return -1, ErrInsufficientData
 	}
 	return tmp, nil
 }
--- a/record_batch.go
+++ b/record_batch.go
@@ -157,7 +157,11 @@ func (b *RecordBatch) decode(pd packetDecoder) (err error) {
 		return err
 	}
 
-	numRecs, err := pd.getArrayLength()
+	// Using NoLimit because a single record batch could contain
+	// more then 2*math.MaxUint16 records. The packet decoder will
+	// check to make sure the array is not greater than the
+	// remaining bytes.
+	numRecs, err := pd.getArrayLengthNoLimit()
 	if err != nil {
 		return err
 	}
--- a/record_batch.go
+++ b/record_batch.go
@@ -157,11 +157,7 @@ func (b *RecordBatch) decode(pd packetDecoder) (err error) {
 		return err
 	}
 
-	// Using NoLimit because a single record batch could contain
-	// more then 2*math.MaxUint16 records. The packet decoder will
-	// check to make sure the array is not greater than the
-	// remaining bytes.
-	numRecs, err := pd.getArrayLengthNoLimit()
+	numRecs, err := pd.getArrayLength()
 	if err != nil {
 		return err
 	}
--- a/record_test.go
+++ b/record_test.go
@@ -3,7 +3,8 @@
 package sarama
 
 import (
-	"fmt"
+	"encoding/binary"
+	"math"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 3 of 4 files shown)


record_test.go
```diff
@@ -254,3 +255,40 @@ func TestRecordBatchDecoding(t *testing.T) {
...
+		6,        // Header Key Length
+		8, 9, 10, // Header Key
+		4,      // Header Value Length
+		11, 12, // Header Value
+	}
+
+	batch := RecordBatch{}
+	err := decode(encodedBatch, &batch, nil)
+	if err != ErrInsufficientData {
+		t.Fatal(fmt.Errorf("was suppose to get ErrInsufficientData, instead got: %w", err))
```
puellanivis, line 292:
```suggestion
		t.Fatal("was suppose to get ErrInsufficientData, instead got:", err)
```

record_test.go
```diff
@@ -254,3 +255,40 @@ func TestRecordBatchDecoding(t *testing.T) {
...
+		6, // Value Length
+		5, 6, 7,
+		2,        // Number of Headers
+		6,        // Header Key Length
+		8, 9, 10, // Header Key
+		4,      // Header Value Length
+		11, 12, // Header Value
+	}
+
+	batch := RecordBatch{}
```
puellanivis, line 289:
```suggestion
	var batch RecordBatch
```

record_test.go
```diff
@@ -254,3 +255,40 @@ func TestRecordBatchDecoding(t *testing.T) {
...
+		2,               // Version
+		91, 48, 202, 99, // CRC
+		0, 0, // Attributes
+		0, 0, 0, 0, // Last Offset Delta
+		0, 0, 1, 88, 141, 205, 89, 56, // First Timestamp
+		0, 0, 0, 0, 0, 0, 0, 0, // Max Timestamp
+		0, 0, 0, 0, 0, 0, 0, 0, // Producer ID
+		0, 0, // Producer Epoch
+		0, 0, 0, 0, // First Sequence
+		0, 1, 255, 255, // Number of Records - 1 + 2*math.MaxUint16
```
puellanivis, line 273: I’m pretty sure this evaluates to `2 * math.MaxUint16 - 1`, and thus would not trip up the old `getArrayLength`. Indeed:
```
[fix3119|∆1]
localhost sarama$ git diff
diff --git a/record_batch.go b/record_batch.go
index 0100e74..564650a 100644
--- a/record_batch.go
+++ b/record_batch.go
@@ -161,7 +161,7 @@ func (b *RecordBatch) decode(pd packetDecoder) (err error) {
        // more then 2*math.MaxUint16 records. The packet decoder will
        // check to make sure the array is not greater than the
        // remaining bytes.
-       numRecs, err := pd.getArrayLengthNoLimit()
+       numRecs, err := pd.getArrayLength()
        if err != nil {
                return err
        }
[fix3119|∆1]
localhost sarama$ go test
PASS
ok      github.com/IBM/sarama   29.316s
```

real_decoder.go
```diff
@@ -121,6 +121,20 @@ func (rd *realDecoder) getArrayLength() (int, error) {
 	return tmp, nil
 }
 
+func (rd *realDecoder) getArrayLengthNoLimit() (int, error) {
```
puellanivis, line 124: I’m not entirely sure why we would want to keep the two implementations around?

rmb938: I didn't want to change the behavior of everything else using the `getArrayLength` function. But if that's a worry I can just update  `getArrayLength` instead.

record_test.go
rmb938: Yep, left that in during testing, will remove.

record_test.go
rmb938: It's `+1`. The reason `getArrayLength` works is because they both check `if tmp > rd.remaining() {` first which returns ErrInsufficientData. They return this error because the test encodedBatch has a small length, I didn't want to make a giant array just for the test, but I can to show `getArrayLength` doesn't work.

puellanivis: Ah… that’s right. `MaxUint16` is `0xFFFF` so double is `0x1_FFFE`, then plus one `0x1_FFFF`.

I understand you don’t want to make a long array constant, because it would be bad coding… but having a test that doesn’t fail when the change is reverted is also against the whole design of testing.

Consider:
```diff
localhost sarama$ git diff record_test.go
diff --git a/record_test.go b/record_test.go
index 2afc265..f435f86 100644
--- a/record_test.go
+++ b/record_test.go
@@ -257,12 +257,12 @@ func TestRecordBatchDecoding(t *testing.T) {
 }

 func TestRecordBatchInvalidNumRecords(t *testing.T) {
-       encodedBatch := []byte{
+       encodedBatch := append([]byte{
                0, 0, 0, 0, 0, 0, 0, 0, // First Offset
-               0, 0, 0, 70, // Length
+               0, 2, 0, 70, // Length
                0, 0, 0, 0, // Partition Leader Epoch
                2,               // Version
-               91, 48, 202, 99, // CRC
+               0xe0, 0xa6, 0xe9, 0x57, // CRC
                0, 0, // Attributes
                0, 0, 0, 0, // Last Offset Delta
                0, 0, 1, 88, 141, 205, 89, 56, // First Timestamp
@@ -284,7 +284,7 @@ func TestRecordBatchInvalidNumRecords(t *testing.T) {
                8, 9, 10, // Header Key
                4,      // Header Value Length
                11, 12, // Header Value
-       }
+       }, make([]byte, 0x2_0000)...)

        batch := RecordBatch{}
        err := decode(encodedBatch, &batch, nil)
localhost sarama$ go test -run TestRecordBatchInvalidNumRecords
--- FAIL: TestRecordBatchInvalidNumRecords (0.00s)
    record_test.go:292: was suppose to get ErrInsufficientData, instead got: kafka: error decoding packet: invalid array length
FAIL
exit status 1
FAIL    github.com/IBM/sarama   0.002s
```

real_decoder.go
puellanivis: Makes sense. Let’s see what the maintainers say. 👍

record_test.go
rmb938: Yes that's true, I started working on a proper test right after I sent that :)

record_test.go
```diff
@@ -254,3 +256,50 @@ func TestRecordBatchDecoding(t *testing.T) {
 		}
 	}
 }
+
+func TestRecordBatchLargeNumRecords(t *testing.T) {
+	numOfRecords := 10 + (2 * math.MaxUint16)
```
puellanivis, line 261: Not sure how useful making this parameterizable is, when the CRC will change if the value changes, but then I like that it’s apparent and inarguable what length we’re encoding.

puellanivis (commented):
👍 from me

record_test.go
puellanivis: Who knows… 🤷‍♀️ I was working real fast and dirty on the code, since it wasn’t being committed anywhere. 😂 Just had to demonstrate the idea.

record_test.go
rmb938: Yea I only set it as a var to make it obvious what it is. I kept forgetting what I had it set to during testing, so it's probably useful just to have it easily readable like you said haha.

record_batch.go
```diff
@@ -157,7 +157,11 @@ func (b *RecordBatch) decode(pd packetDecoder) (err error) {
 		return err
 	}
 
-	numRecs, err := pd.getArrayLength()
+	// Using NoLimit because a single record batch could contain
+	// more then 2*math.MaxUint16 records. The packet decoder will
```
gunli, line 161: "then" should be "than"

dnwe:
@rmb938 commented over on the issue [here](https://github.com/IBM/sarama/issues/3119#issuecomment-2710216498), but I think we probably want to just keep the existing getArrayLength func rather than adding a new one, and just change the magic number to use the existing constant we have for FetchBytes perhaps?

rmb938:
@dnwe Yep makes sense to me, done.

nkawli:
Hit the same issue in my setup. When will we get the new release with this fix ?

PR #3128, opened 2025-03-18
fix: send null instead of empty string when describing default client quotas

petedannemann:
DescribeClientQuota currently fails with the following error message logged by the Kafka broker when trying to describe default client quotas:
```
org.apache.kafka.common.errors.InvalidRequestException: Request specified MATCH_TYPE_DEFAULT, but also specified a match string
```
This is because we are currently sending an empty string (`""`) instead of a null as the match string

The Java client simply ignores the match string here to avoid this issue, so I propose we do the same https://github.com/apache/kafka/blob/3ed590288fd773902f7791959e8f081ff937c144/clients/src/main/java/org/apache/kafka/common/requests/DescribeClientQuotasRequest.java#L54-L55

### DIFF
--- a/describe_client_quotas_request.go
+++ b/describe_client_quotas_request.go
@@ -87,7 +87,7 @@ func (d *QuotaFilterComponent) encode(pe packetEncoder) error {
 			return err
 		}
 	} else if d.MatchType == QuotaMatchDefault {
-		if err := pe.putString(""); err != nil {
+		if err := pe.putNullableString(nil); err != nil {
 			return err
 		}
 	} else {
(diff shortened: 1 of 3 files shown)


dnwe (approved):
Thanks! This does indeed look like a bug in the original implementation, thanks for the fix

petedannemann:
Thanks for the review, my mistake for missing the unit test there, I think I may have only ran the functional tests on my machine

petedannemann:
@dnwe curious if you have a date in mind for the next Sarama release? I was hoping to leverage this feature here https://github.com/Mongey/terraform-provider-kafka/pull/527

PR #3144, opened 2025-04-15
fix(decoder): handle null arrays correctly

dnwe:
Array lengths in the Kafka protocol are encoded as a 32-bit signed integer in big-endian format. For a null value this corresponds to the byte sequence FF FF FF FF which is intended to represent null as a length of -1. To correctly read that we need to cast the value to int32 before widening to int.

### DIFF
--- a/real_decoder.go
+++ b/real_decoder.go
@@ -306,7 +306,7 @@ func (rd *realDecoder) getInt32Array() ([]int32, error) {
 		rd.off = len(rd.raw)
 		return nil, ErrInsufficientData
 	}
-	n := int(binary.BigEndian.Uint32(rd.raw[rd.off:]))
+	n := int(int32(binary.BigEndian.Uint32(rd.raw[rd.off:])))
 	rd.off += 4
 
 	if rd.remaining() < 4*n {
@@ -335,7 +335,7 @@ func (rd *realDecoder) getInt64Array() ([]int64, error) {
 		rd.off = len(rd.raw)
 		return nil, ErrInsufficientData
 	}
-	n := int(binary.BigEndian.Uint32(rd.raw[rd.off:]))
+	n := int(int32(binary.BigEndian.Uint32(rd.raw[rd.off:])))
 	rd.off += 4
 
 	if rd.remaining() < 8*n {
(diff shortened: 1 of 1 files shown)


puellanivis (commented):
> ARRAY: Represents a sequence of objects of a given type T. Type T can be either a primitive type (e.g. STRING) or a structure. First, the length N is given as an INT32. Then N instances of type T follow. A null array is represented with a length of -1. In protocol documentation an array of T instances is referred to as [T].
—https://kafka.apache.org/protocol.html

Checks out. The cast to `int32` first is necessary to get signedness before then casting to a maybe-larger `int`.

I’m not sure that we’re still handling -1 length NULL arrays properly though. Should we be returning `errInvalidArrayLength` or should we be returning a `[]T(nil)`?

dnwe:
Yeah the latter I think. @vladoatanasov spotted the original problem whilst using the decoder code elsewhere. I was mildly surprised we'd never hit any problems due to this, but I can only guess it's an uncommon response for the client to decode and we obviously don't have any unittest coverage of this case

vladoatanasov:
Hey @puellanivis, I stumbled upon this issue while implementing the DescribeConfigs API [here](https://github.com/IBM/opentalaria/blob/main/protocol/describe_configs_request.go#L13). The spec states `The configuration keys to list, or null to list all configuration keys.` for the configuration keys array. And that's how clients behave, they send -1, aka nil when no config keys are specified. Tried this behavior with Sarama and Kafka cli. So I believe the decoder should return `nil` when the array size is -1

puellanivis:
🤔 This probably shouldn’t still be in draft mode? The change is sound. (Even though it’s a little weird that it has a double cast, it’s correct.)

dnwe:
@puellanivis yeah I think I intended to come back and add a unittest before getting a review and merging, but obviously forgot about it

vladoatanasov:
@dnwe it's probably worth it adding a code comment as to why a double cast is needed. I had already forgotten about this use-case. Someone looking at this code in the future might be wondering what's going on.

puellanivis:
🤔 I think a unit test might be a bit hard, because it would never fail in a 32-bit environment. But then, I think we get plenty of 64-bit environment testing done.

I have +1’ed already the idea of adding a comment about why a double cast is necessary. I remembered pretty fast why it’s there, but it is kind of an astonishing thing to need to do.

PR #3158, opened 2025-05-14
feat(broker): negotiate MetadataRequest version

trapped:
The latest `github.com/IBM/sarama` version fails to connect to Redpanda brokers (latest version too):

- Redpanda rejects the request with `Unsupported version 10 for metadata API`
- sarama fails with `kafka: client has run out of available brokers to talk to: EOF`

<details>
<summary>Sarama logs</summary>
```
2025/05/13 18:26:29 Initializing new client
2025/05/13 18:26:29 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
2025/05/13 18:26:29 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
2025/05/13 18:26:29 client/metadata fetching metadata for all topics from broker localhost:9092
2025/05/13 18:26:29 Connected to broker at localhost:9092 (unregistered)
2025/05/13 18:26:29 client/metadata got error from broker -1 while fetching metadata: EOF
2025/05/13 18:26:29 Closed connection to broker localhost:9092
2025/05/13 18:26:29 client/metadata no available broker to send metadata request to
2025/05/13 18:26:29 client/brokers resurrecting 1 dead seed brokers
2025/05/13 18:26:29 Error while sending ApiVersionsRequest to broker localhost:9092: kafka: broker not connected
2025/05/13 18:26:30 client/metadata retrying after 250ms... (2 attempts remaining)
2025/05/13 18:26:30 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
2025/05/13 18:26:30 client/metadata fetching metadata for all topics from broker localhost:9092
2025/05/13 18:26:30 Connected to broker at localhost:9092 (unregistered)
2025/05/13 18:26:30 client/metadata got error from broker -1 while fetching metadata: EOF
2025/05/13 18:26:30 Closed connection to broker localhost:9092
2025/05/13 18:26:30 client/metadata no available broker to send metadata request to
2025/05/13 18:26:30 client/brokers resurrecting 1 dead seed brokers
2025/05/13 18:26:30 client/metadata retrying after 250ms... (1 attempts remaining)
2025/05/13 18:26:30 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
2025/05/13 18:26:30 client/metadata fetching metadata for all topics from broker localhost:9092
2025/05/13 18:26:30 Connected to broker at localhost:9092 (unregistered)
2025/05/13 18:26:30 client/metadata got error from broker -1 while fetching metadata: EOF
2025/05/13 18:26:30 Closed connection to broker localhost:9092
2025/05/13 18:26:30 client/metadata no available broker to send metadata request to
2025/05/13 18:26:30 client/brokers resurrecting 1 dead seed brokers
2025/05/13 18:26:30 client/metadata retrying after 250ms... (0 attempts remaining)
2025/05/13 18:26:30 ClientID is the default of 'sarama', you should consider setting it to something application-specific.
2025/05/13 18:26:30 client/metadata fetching metadata for all topics from broker localhost:9092
2025/05/13 18:26:30 Connected to broker at localhost:9092 (unregistered)
2025/05/13 18:26:30 client/metadata got error from broker -1 while fetching metadata: EOF
2025/05/13 18:26:30 Closed connection to broker localhost:9092
2025/05/13 18:26:30 client/metadata no available broker to send metadata request to
2025/05/13 18:26:30 client/brokers resurrecting 1 dead seed brokers
2025/05/13 18:26:30 Closing Client
```
</details>

The culprit being simply that:
1. Sarama chooses request versions solely according to the provided `config.Version`, [even though it sends an ApiVersionsRequest](https://github.com/IBM/sarama/blob/43f8559f9dc49d7bc46f81adec99aed43d423bf8/broker.go#L189-L201)
2. [MetadataRequest uses versions 8-9-10 for Kafka versions >= v2.4.0.0](https://github.com/IBM/sarama/blob/43f8559f9dc49d7bc46f81adec99aed43d423bf8/metadata_request.go#L26-L29)
3. [Redpanda only supports versions 0 to 8](https://github.com/redpanda-data/redpanda/blob/bee4479bf54ae2d9af54b07b1203179791b0f166/src/v/kafka/server/handlers/metadata.h#L28), and it correctly advertises it

-------

This PR adds a `brokerAPIVersions` field to the `Broker` type, storing the `ApiVersionsResponse` values _if `conf.Version.IsAtLeast(V2_4_0_0) && conf.ApiVersionsRequest`_; request versions are then validated before sending requests.

Additionally, adds a `NewNegotiatedMetadataRequest` constructor that behaves like `NewMetadataRequest`, but considers the broker's advertised min/max API versions; this is then used in the client when refreshing metadata.

These changes allow connecting to Redpanda while maintaining `config.Version = sarama.MaxVersion`; they are not a complete implementation of version selection. Hopefully though this can be a starting point.

### DIFF
--- a/.whitesource
+++ b/.whitesource
@@ -1,0 +1,3 @@
+{
+  "settingsInheritedFrom": "ibm-mend-config/mend-config@main"
+}


broker.go
```diff
@@ -1013,6 +1027,16 @@ func (b *Broker) sendInternal(rb protocolBody, promise *responsePromise) error {
 		return ErrUnsupportedVersion
 	}
 
+	// validate the request is using a supported API version
+	if b.brokerAPIVersions != nil && b.brokerAPIVersions[rb.key()] != [2]int16{0, 0} {
```
puellanivis, line 1031: [unnecessary guards] The test for `b.brokerAPIVersions != nil` is not necessary. It is not an error to access a `nil` map, it simply functions the same as an empty map. It is only assigning into a map that does not work.

Thus, when `b.brokerAPIVersions == nil` then `b.brokerAPIVersions[rb.key()] == [2]int16{ 0, 0 }`.

broker.go
```diff
@@ -33,6 +34,10 @@ type Broker struct {
 	responses     chan *responsePromise
 	done          chan bool
 
+	// brokerAPIVersions stores the broker-advertised min/max API versions for each
+	// request key, if ApiVersionsRequest is enabled.
+	brokerAPIVersions map[int16][2]int16
```
puellanivis, line 39: [premature optimization]🤔 This `[2]int16` does not save any space over a `struct{ min, max uin16 }` and the array is far less transparent about which is the minimum value, and which is the maximum value.

In fact, the `ApiVersionsResponseKey` itself is only 4 `int16` rather than 2. And there is not really going to be _that_ many entries anyways, right? So, without profiling indicating a need to optimize anything here, I would recommend that using a `map[int16]*ApiVersionsResponseKey` would be far more easy for humans to understand and keep straight.

broker.go
```diff
@@ -1013,6 +1027,16 @@ func (b *Broker) sendInternal(rb protocolBody, promise *responsePromise) error {
 		return ErrUnsupportedVersion
 	}
 
+	// validate the request is using a supported API version
+	if b.brokerAPIVersions != nil && b.brokerAPIVersions[rb.key()] != [2]int16{0, 0} {
+		minMax := b.brokerAPIVersions[rb.key()]
+		if rb.version() < minMax[0] || rb.version() > minMax[1] {
+			return fmt.Errorf("%w: unsupported API version %d for %s, supported versions are [%d-%d]",
+				// protocolBody is a *sarama.XXXRequest
+				ErrUnsupportedVersion, rb.version(), reflect.TypeOf(rb).Elem().Name(), minMax[0], minMax[1])
```
puellanivis, line 1036: There is no need to import `reflect` just to get a type name, as `fmt.Printf` provides a specific verb for this: `%T`.
```suggestion
			return fmt.Errorf("%w: unsupported API version %d for %T, supported versions are [%d-%d]",
				// protocolBody is a *sarama.XXXRequest
				ErrUnsupportedVersion, rb.version(), rb, minMax.MinVersion, minMax.MaxVersion)
```

metadata_request.go
```diff
@@ -45,6 +45,32 @@ func NewMetadataRequest(version KafkaVersion, topics []string) *MetadataRequest
 	return m
 }
 
+func NewNegotiatedMetadataRequest(version KafkaVersion, brokerMinMaxVersions [2]int16, topics []string) (*MetadataRequest, error) {
```
puellanivis, line 48: I would recommend using a `minVersion, maxVersion int16` for ease of understanding which of the two `int16`s are which.

broker.go
trapped: TIL! Thanks 🙂

metadata_request.go
trapped: Updated

broker.go
trapped: Updated

broker.go
trapped: Updated

broker.go
```diff
@@ -187,16 +191,25 @@ func (b *Broker) Open(conf *Config) error {
...
 					Logger.Printf("Error while sending ApiVersionsRequest to broker %s: %s\n", b.addr, err)
+					return
+				}
+
+				b.lock.Lock()
+				defer b.lock.Unlock()
+
+				b.brokerAPIVersions = make(map[int16]*ApiVersionsResponseKey, len(apiVersionsResponse.ApiKeys))
+				for i, key := range apiVersionsResponse.ApiKeys {
+					b.brokerAPIVersions[key.ApiKey] = &apiVersionsResponse.ApiKeys[i]
```
puellanivis, line 212:
```suggestion
				for _, key := range apiVersionsResponse.ApiKeys {
					key := key // loopvar guard only necessary before go1.22
					b.brokerAPIVersions[key.ApiKey] = &key
```

broker.go
```diff
@@ -1013,6 +1026,14 @@ func (b *Broker) sendInternal(rb protocolBody, promise *responsePromise) error {
 		return ErrUnsupportedVersion
 	}
 
+	// validate the request is using a supported API version
+	if apiVersions, ok := b.brokerAPIVersions[rb.key()]; ok {
+		if rb.version() < apiVersions.MinVersion || rb.version() > apiVersions.MaxVersion {
+			return fmt.Errorf("%w: unsupported API version %d for %T, supported versions are [%d-%d]",
```
puellanivis, line 1032: Proper interval notation would be `[%d,%d]`, though `%d-%d` is just as reasonable as a non–math-geek version.

puellanivis (commented):
Really, only coming up with lint, which is a good sign. 🙂

broker.go
trapped: Good call!

client.go
```diff
@@ -985,7 +985,20 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int,
 		}
 
 		req := NewMetadataRequest(client.conf.Version, topics)
+
+		// negotiate request version if enabled
+		if apiVersion, ok := broker.brokerAPIVersions[req.key()]; ok {
+			negotiatedReq, err := NewNegotiatedMetadataRequest(client.conf.Version, apiVersion.MinVersion, apiVersion.MaxVersion, topics)
```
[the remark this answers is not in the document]
trapped: I had initially written it like that - here's how it'd look like:

```go
var (
	req = NewMetadataRequest(client.conf.Version, topics)
	err error
)

// negotiate request version if enabled
if apiVersion := broker.brokerAPIVersions[req.key()]; apiVersion != nil {
	req, err = NewNegotiatedMetadataRequest(client.conf.Version, apiVersion.MinVersion, apiVersion.MaxVersion, topics)
}

if err != nil {
	Logger.Printf("client/metadata failed to negotiate metadata request version: %v", err)
	return retry(err)
}
```

My gripe with it is that the error handling here is really only meant for `NewNegotiatedMetadataRequest`'s error, and is not really useful unless we are in that conditional block. It could make sense to extract it from there if we had more/other errors to handle, though.

broker.go
```diff
@@ -1013,6 +1026,14 @@ func (b *Broker) sendInternal(rb protocolBody, promise *responsePromise) error {
 		return ErrUnsupportedVersion
 	}
 
+	// validate the request is using a supported API version
+	if apiVersions, ok := b.brokerAPIVersions[rb.key()]; ok {
```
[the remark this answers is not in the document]
trapped: Updated

broker.go
trapped: Updated

client.go
```diff
@@ -985,7 +985,20 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int,
 		}
 
 		req := NewMetadataRequest(client.conf.Version, topics)
+
+		// negotiate request version if enabled
+		if apiVersion, ok := broker.brokerAPIVersions[req.key()]; ok {
```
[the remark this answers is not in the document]
trapped: Updated

broker.go
trapped: I'll keep it non-math-geeky then, for simplicity 🙂

puellanivis (commented):
I have run out of anything to comment on. :)

trapped:
@puellanivis thanks for the review! I'd love to run the CI on it to see if there's anything I've missed.

dnwe (commented):
@trapped thanks for spotting the incompatibility with redpanda and coming up with this PR to start to adhere to the apiversions responses

I know you added it to avoid changing the existing NewMetadataRequest signature, but I was wondering if we could avoid the need for `NewNegotiatedMetadataRequest` and the changes in client.go (which take the broker lock) if we instead introduced a new func to the protocolBody interface something like `restrictApiVersion(minVersion, maxVersion int16) error` which gives an error if the existing version is less than minVersion and silently rolls down the version if it exceeds maxVersion. Then we replace your "validate the request is using a supported API version" in sendInternal to call `rb.restrictProtocolVersions(apiVersion.MinVersion, apiVersions.MaxVersion)` if there is a value in b.brokerAPIVersions?

The implementation in metadata_request.go would be something like:
```go
func (m *MetadataRequest) restrictApiVersion(minVersion, maxVersion int16) error {
	if m.Version < minVersion {
		return fmt.Errorf("%w: unsupported API version %d for %T, supported versions are %d-%d",
			ErrUnsupportedVersion, m.Version, m, minVersion, maxVersion)
	}
	m.Version = max(m.Version, maxVersion)
	return nil
}
```

Which I think would still give the behaviour that Version in sarama.Config can be used by a user to pin to "no more than was supported by this KafkaVersion" whilst also adhering to "no more than is supported by the ApiVersionsResponse maxVersion values either"

trapped:
Hey @dnwe! Of course, I'm happy to make wider changes 🙂 

Yours seems like a really nice idea. It should be easy to then add the method to every other request, at which point we'd essentially have this "version negotiation" for everything I guess.

I'll come up with the changes ASAP!

dnwe:
@trapped 😅 well it was just an initial idea, so don't discard what you have here, but if you have time and motivation then feel free to give it a go on a separate branch and see how it might look

The useful property is that it can all be done as "internal" api changes so we don't have to worry about anyone that might be using the protocol types externally, which gives us more freedom to experiment and redesign as we go along

trapped:
@dnwe yeah, makes sense 🙂 I'll share a link to another draft PR here soon.

dnwe:
@trapped thanks! One thing I forgot to take account of in my original snippet was that maxVersion supported by the remote broker can and will exceed the max version of the protocol that we've actually implemented the encoding/decoding for within Sarama. I'd also incorrectly used `max` where we probably wanted `min` semantics, as we probably want the old Version field from sarama.Config to pin the _maximum_ versions that are used, and people could set Version to sarama.MaxVersion to opt-in to relying on api versions negotiated usage instead.

e.g., for FetchRequest we've only currently implemented up to V11 but Kafka 2.7 and newer will advertise V12+ – so we need to include that in the sum.

So for MetadataRequest probably something like this:
```go
func (m *MetadataRequest) restrictApiVersion(minVersion, maxVersion int16) error {
	maxEncodedVersion := min(10, maxVersion)
	if m.Version < minVersion {
		return fmt.Errorf("%w: unsupported API version %d for %T, supported versions are %d-%d",
			ErrUnsupportedVersion, m.Version, m, minVersion, maxEncodedVersion)
	}
	m.Version = min(m.Version, maxEncodedVersion)
	return nil
}
```

PR #3160, opened 2025-05-19
fix(ci): bump golangci-lint to v2

dnwe:
The golangci-lint action is refusing to use the older version of golangci-lint now, so bump to latest v2 and run `golangci-lint migrate` to convert the config file to the latest format


hindessm (approved):
LGTM (I appreciate that you took the time to put the comments back after migration!)

Issue #3191, opened 2025-06-11
Sarama Consumer DeadLock


dominicqi:
### 🐞 Description

A deadlock can occur when manually committing offsets using Sarama.  
This issue is caused by inconsistent lock acquisition order between `pomsLock` and `brokerLock`.

### 🔢 Versions

| Sarama  | Kafka | Go    |
|---------|-------|-------|
| 1.43.3  |       |       |

### ⚙️ Configuration

```go
saramaConfig.Consumer.Offsets.AutoCommit.Enable = false
```
Goroutine Stack
Goroutine A holds pomsLock, then tries to acquire brokerLock:

```
goroutine 829602 [sync.Mutex.Lock, 665 minutes]:
sync.runtime_SemacquireMutex
sync.(*Mutex).lockSlow
sync.(*RWMutex).Lock
github.com/IBM/sarama.(*offsetManager).releaseCoordinator
github.com/IBM/sarama.(*offsetManager).handleResponse
github.com/IBM/sarama.(*offsetManager).flushToBroker
github.com/IBM/sarama.(*offsetManager).Commit
github.com/IBM/sarama.(*consumerGroupSession).Commit
github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kafkareceiver.(*logsConsumerGroupHandler).ConsumeClaim

```
Goroutine B holds brokerLock, then tries to acquire pomsLock:
```
goroutine 829612 [sync.RWMutex.RLock, 665 minutes]:
sync.runtime_SemacquireRWMutexR
sync.(*RWMutex).RLock
github.com/IBM/sarama.(*offsetManager).constructRequest
github.com/IBM/sarama.(*offsetManager).flushToBroker
github.com/IBM/sarama.(*offsetManager).Commit
github.com/IBM/sarama.(*consumerGroupSession).Commit
github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kafkareceiver.(*logsConsumerGroupHandler).ConsumeClaim

```
[constructRequest](https://github.com/IBM/sarama/blob/c610319d027db2af1dcdd1151331b644cbaee9ef/offset_manager.go#L351) holds brokerLock, then tries to acquire pomsLock.

[handleResponse](https://github.com/IBM/sarama/blob/c610319d027db2af1dcdd1151331b644cbaee9ef/offset_manager.go#L371) holds pomsLock, and later calls releaseCoordinator, which requires brokerLock.

[...]

PR #3193, opened 2025-06-13
feat: fix deadlock in Commit #3191

dominicqi:
Fix Issue   #3191

### DIFF
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -285,8 +285,10 @@ func (om *offsetManager) flushToBroker() {
 func sendOffsetCommit(coordinator *Broker, req *OffsetCommitRequest) (*OffsetCommitResponse, *responsePromise, error) {
 	resp := new(OffsetCommitResponse)
 	responseHeaderVersion := resp.headerVersion()
+
 	coordinator.lock.Lock()
 	defer coordinator.lock.Unlock()
+
 	promise, err := coordinator.send(req, true, responseHeaderVersion)
 	if err != nil {
 		return nil, nil, err
--- a/offset_manager.go
+++ b/offset_manager.go
@@ -257,16 +257,11 @@ func (om *offsetManager) flushToBroker() {
 		return
 	}
 
-	// Care needs to be taken to unlock this. Don't want to defer the unlock as this would
-	// cause the lock to be held while waiting for the broker to reply.
-	broker.lock.Lock()
 	req := om.constructRequest()
 	if req == nil {
-		broker.lock.Unlock()
 		return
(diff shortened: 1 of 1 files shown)


offset_manager.go
```diff
@@ -290,6 +285,8 @@ func (om *offsetManager) flushToBroker() {
 func sendOffsetCommit(coordinator *Broker, req *OffsetCommitRequest) (*OffsetCommitResponse, *responsePromise, error) {
 	resp := new(OffsetCommitResponse)
 	responseHeaderVersion := resp.headerVersion()
+	coordinator.lock.Lock()
+	defer coordinator.lock.Unlock()
```
puellanivis, line 290: I’m not sure how much of the code is following “paragraphing”, but I think it would be best if these two lines were set apart with an empty line above and below. It will help them standout better, and as a sort of semantic “horizontal bar” it’s usually helpful to set them out like that as well.

dominicqi: Yes, I agree.

puellanivis (commented):
I don’t see anything that should break from this… I’ve not proven that it would solve the deadlock. But it’s a good step.

dominicqi:
> I don’t see anything that should break from this… I’ve not proven that it would solve the deadlock. But it’s a good step.

I thought this change might help break the deadlock by decoupling the two locks — `broker.lock` and `om.pomslock.RLock`.

puellanivis:
> > I don’t see anything that should break from this… I’ve not proven that it would solve the deadlock. But it’s a good step.
> 
> I thought this change might help break the deadlock by decoupling the two locks — `broker.lock` and `om.pomslock.RLock`.

Yes, I think it should work as well. But I’m not _certain_ that it would. The phrasing was a bit of a play on:

> `Beware of bugs in the above code; I have only proved it correct, not tried it.'' —Knuth

dominicqi:
> > > I don’t see anything that should break from this… I’ve not proven that it would solve the deadlock. But it’s a good step.
> > 
> > 
> > I thought this change might help break the deadlock by decoupling the two locks — `broker.lock` and `om.pomslock.RLock`.
> 
> Yes, I think it should work as well. But I’m not _certain_ that it would. The phrasing was a bit of a play on:
> 
> > `Beware of bugs in the above code; I have only proved it correct, not tried it.'' —Knuth

Haha, yeah — it’s not been properly verified yet.

puellanivis:
I still continue to approve this PR.

dnwe:
I think this was resolved via #3477 but feel free to reopen if not

PR #3221, opened 2025-07-12
fix(consumer): stuck on the batch with zero records length

sterligov:
This pull request addresses the known issue described in [KAFKA-5443](https://issues.apache.org/jira/browse/KAFKA-5443), which was initially analyzed in [#2053](https://github.com/IBM/sarama/pull/2053). A prior attempt to fix it was made in [#2057](https://github.com/IBM/sarama/pull/2057), but the fix was applied incorrectly and did not fully resolve the root cause.

I’ve provided a more detailed explanation of the underlying issue and the reasoning behind this fix in [this comment](https://github.com/IBM/sarama/pull/3150#issuecomment-3064757832).

### DIFF
--- a/fetch_response.go
+++ b/fetch_response.go
@@ -49,9 +49,6 @@ type FetchResponseBlock struct {
 	// transactional records prior to this offset have been decided (ABORTED or
 	// COMMITTED)
 	LastStableOffset int64
-	// RecordsNextOffset contains the next consecutive offset following this response block.
-	// This field is computed locally and is not part of the server's binary response.
-	RecordsNextOffset *int64
 	// LogStartOffset contains the current log start offset.
 	LogStartOffset int64
 	// AbortedTransactions contains the aborted transactions.
@@ -64,6 +61,10 @@ type FetchResponseBlock struct {
 
 	Partial bool
 	Records *Records // deprecated: use FetchResponseBlock.RecordsSet
+
+	// recordsNextOffset contains the next consecutive offset following this response block.
+	// This field is computed locally and is not part of the server's binary response.
+	recordsNextOffset *int64
 }
 
 func (b *FetchResponseBlock) decode(pd packetDecoder, version int16) (err error) {
@@ -152,7 +153,7 @@ func (b *FetchResponseBlock) decode(pd packetDecoder, version int16) (err error)
 			return err
 		}
 
-		b.RecordsNextOffset, err = records.nextOffset()
+		b.recordsNextOffset, err = records.nextOffset()
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


fetch_response.go
```diff
@@ -150,7 +150,7 @@ func (b *FetchResponseBlock) decode(pd packetDecoder, version int16) (err error)
 			return err
 		}
 
-		b.LastRecordsBatchOffset, err = records.recordsOffset()
+		b.RecordsNextOffset, err = records.nextOffset()
```
puellanivis: I guess, since we’re more caching a value here rather than this `RecordsNextOffset` existing properly in the serialized communication, that there’s no reason not to precalculate it… also, since it gets stored as a pointer, incrementing it would require testing for possible `nil` anyways…

puellanivis (commented):
I suppose, the only feedback I could give is that it would be nice to have something explaining that this field doesn’t actually appear “on the wire”? I mean, that was my first thought, “wait, are we renaming an encoded field in the fetch response block?” But it’s not, so that is why renaming it shouldn’t be a big deal, right?

But also, this is technically changing an exported field on an exported type… 😬 Although, a quick search of github suggests the only people referencing the field are from forks/clones.…

sterligov:
> I suppose, the only feedback I could give is that it would be nice to have something explaining that this field doesn’t actually appear “on the wire”?

I've added a comment explaining this field.

> But also, this is technically changing an exported field on an exported type… 😬 Although, a quick search of github suggests the only people referencing the field are from forks/clones.…

In my opinion, changing the name is not a problem, this field is only used in a few places and it was added in a previous PR that solved this problem in a way that was not quite correct.

fetch_response.go
sterligov: Unfortunately I don't have time to figure out how to do it more succinctly, I just based it on a previous pull request that tried to fix this bug.

puellanivis: I’ve lost context of what I precisely I was meaning, but I think my comment turned into “but yeah, this is probably the best was to do it now, regardless of my first concern.”

puellanivis:
We could consider unexporting the name as well? This would make it perfectly clear that it’s not a part of the exported API?

fetch_response.go
```diff
@@ -64,6 +61,10 @@ type FetchResponseBlock struct {
 
 	Partial bool
 	Records *Records // deprecated: use FetchResponseBlock.RecordsSet
+
+	// recordsNextOffset contains the next consecutive offset following this response block.
+	// This field is computed locally and is not part of the server's binary response.
```
puellanivis, line 66: 👍

PR #3223, opened 2025-07-16
fix: sync response header version to clamped request header

trapped:
Smallish refactor changing the internal structure of `responsePromise`: instead of storing a (potentially stale) response header version, it now stores a pointer to the response's `protocolBody` (if any).

This allows `sendInternal` to easily update the response's `Version` after version negotiation. (request and response versions must match; _header versions_ instead don't map 1-1, meaning we can't rely on the request's header version for decoding responses).

Fixes the issue from [this comment](https://github.com/IBM/sarama/pull/3209#issuecomment-3078126116).

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -130,7 +130,7 @@ type SCRAMClient interface {
 type responsePromise struct {
 	requestTime   time.Time
 	correlationID int32
-	headerVersion int16
+	response      protocolBody
 	handler       func([]byte, error)
 	packets       chan []byte
 	errors        chan error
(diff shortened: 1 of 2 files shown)


dnwe:
@trapped historically in the functional tests we pin the Sarama config version to match the remote Kafka version we are testing against, but it might be worth updating that [here](https://github.com/IBM/sarama/blob/57be9bdd0a5cd133dc27edf35560b057074b9508/functional_test.go#L96-L110) to just always set MaxVersion and UseApiVersions and that should shake out this sort of issue too (where Sarama supports a newer version than the remote)

trapped:
@dnwe thanks for the suggestion, that makes a lot of sense. I've just pushed a new commit with that change. Let's see if CI breaks.

dnwe:
Cool, that testing eventually worked with apiversions by default. I've taken those commits off again so we can merge this PR with the single focus and then I'll raise a separate PR to move to "enable apiversions by default" as a followup

dnwe (approved):

Issue #3224, opened 2025-07-16
Metadata request burst


cupcicm:
##### Description

I work at Confluent, in the Warpstream team. Both confluent internally and its customers are using sarama to drive workloads that write Gigabytes per second of traffic to Kafka and Warpstream.
I believe that sarama can overwhelm the brokers with metadata requests when the number of brokers is high and the number of partitions they each hold is high. 

Let’s imagine the setup. We have 3 brokers, and each of them is the leader for 1000 partitions. There are also 100 producers producing to these partitions.
Let us now imagine that one broker shuts down.
Inside each producer this will happen:
1. Something will try to make the producer produce to one of the partitions that broker was the leader for, and will notice that the broker is not there anymore.
2. Sarama will detect that it is `abandoned` and the code [here](https://github.com/IBM/sarama/blob/57be9bdd0a5cd133dc27edf35560b057074b9508/async_producer.go#L662) will trigger. It will set the `brokerProducer` of the partition to `nil` which will trigger a metadata refresh.
3. Sarama will do this in turn for the 1000 partitions, triggering 1000 metadata refreshes, one for each partition.

Overall, the 100 producers will, in a very short amount of time, issue 100,000 metadata requests to the broker, so that they can know where the new leader is now.

We have customers who are facing this issue on a Warpstream cluster, where the issue is exacerbated by the fact an agent can declare itself the leader for a lot of partitions. 

We would like to fix this and we are willing to contribute a patch that
1. Allows to run a single metadata request per sarama client in parallel. If other parts of the code call `RefreshMetadata()` they will be queued after the current one finishes
2. Adds a configuration boolean that allows (as an opt-in) to enable this behavior.

Would you be open to integrating this patch ?

##### Versions

| Sarama | Kafka | Go  |
| ------ | ----- | --- |
|   1.45.2     |    4.0   |  1.24   |

##### Configuration

```go

	config := sarama.NewConfig()
	config.Version = sarama.MaxVersion
	config.Metadata.RefreshFrequency = 60 * time.Second
	config.Producer.Flush.MaxMessages = 1_000_000
	config.Producer.Flush.Frequency = 25 * time.Millisecond
	config.Producer.MaxMessageBytes = 64_000_000
	config.Producer.Flush.Bytes = 16_000_000

	config.Net.ReadTimeout = 90 * time.Second
	config.Consumer.Group.Rebalance.Timeout = 60 * time.Second
	config.Consumer.Group.Session.Timeout = 45 * time.Second
[...]

PR #3225, opened 2025-07-16
Adds a component that knows how to run a single concurrent metadata refresh

cupcicm:
If other goroutines ask for another metadata refresh, we are going to accumulate all requested topics into a slice and we will queue another metadata request immediately after the current one is done. 
It is possible to enable this behavior by setting `config.Metadata.SingleFlight = true` in your configuration.

I confirmed that this fixes https://github.com/IBM/sarama/issues/3224 in our setup.


metadata.go
```diff
@@ -0,0 +1,221 @@
...
+	allTopics bool
+	chans     []chan error
+
+	// This is the function that gets called when to refresh the metadata.
+	// It is called with the list of all topics that need to be refreshed
+	// or with nil if all topics need to be refreshed.
+	refresh func(topics []string) error
+}
+
+func newMetadataRefresh(f func(topics []string) error) *currentRefresh {
```
ryanworl: you could re-use your type MetadataRefresh here no?

cupcicm: ah sorry it's misleading, I have changed the names. Let me update.

cupcicm: OK, updated.

client_test.go
```diff
@@ -814,7 +816,7 @@ func TestClientMetadataTimeout(t *testing.T) {
 	}
 
 	for _, tc := range tests {
-		t.Run(tc.name, func(t *testing.T) {
+		var test = func(t *testing.T, singleFlight bool) {
```
puellanivis:
```suggestion
		test := func(t *testing.T, singleFlight bool) {
```

puellanivis: I don’t think I see anywhere that `singleFlight` is used in this function body?

client_test.go
```diff
@@ -871,23 +873,45 @@ func TestClientMetadataTimeout(t *testing.T) {
 			}
 
 			safeClose(t, c)
+		}
+		t.Run("singleflight_"+tc.name, func(t *testing.T) {
+			test(t, true)
+		})
+		t.Run("concurrent_"+tc.name, func(t *testing.T) {
+			test(t, false)
 		})
```
puellanivis: Consider instead:
```go
for singleFlight := range []bool{ true, false } {
	t.Run(fmt.Sprintf("%s_singleflight_is_%t", tc.name, singleFlight), func(t *testing.T) {
		// no need to change the function code, we get `singleFlight` now through closure rather than as a function argument.
	})
}
```

With this, we avoid the need to name the anonymous function.

client_test.go
```diff
@@ -871,23 +873,45 @@ func TestClientMetadataTimeout(t *testing.T) {
...
 func TestClientUpdateMetadataErrorAndRetry(t *testing.T) {
 	seedBroker := NewMockBroker(t, 1)
-
-	metadataResponse1 := new(MetadataResponse)
-	metadataResponse1.AddBroker(seedBroker.Addr(), 1)
-	seedBroker.Returns(metadataResponse1)
+	var called atomic.Int32
+
+	seedBroker.setHandler(func(req *request) (res encoderWithHeader) {
+		require.EqualValues(t, 3, req.body.key(), "this test sends only Metadata requests")
```
puellanivis: Do not sure `require` here. The danger here is that we could be running this on a separate goroutine from the one running the test.

Note the condition on https://pkg.go.dev/testing@go1.24.5#T.Fatal

Also consider that using a regular `if` block is more than adequate:
```go
if req.body.key() != 3 /* can we get a name in here instead? */ {
	t.Error("this test should only send Metadata requests")
	return
}
```

client_test.go
```diff
@@ -871,23 +873,45 @@ func TestClientMetadataTimeout(t *testing.T) {
...
+		var topics = req.body.(*MetadataRequest).Topics
+		var resp = new(MetadataResponse)
```
puellanivis: Prefer shorthand declaration over `var foo = typedValue` expressions.
```suggestion
		topics := req.body.(*MetadataRequest).Topics
		resp := new(MetadataResponse)
```

client_test.go
```diff
@@ -871,23 +873,45 @@ func TestClientMetadataTimeout(t *testing.T) {
...
+		for _, topic := range topics {
```
puellanivis: Looks like `topics` exists only on two lines, where it is declared, and where it is used. And assigning to a local variable doesn’t provide any extra clarity over just using the value directly:
```suggestion
		for _, topic := range req.body.(*MetadataRequest).Topics {
```

client_test.go
```diff
@@ -911,6 +931,68 @@ func TestClientUpdateMetadataErrorAndRetry(t *testing.T) {
 	waitGroup.Wait()
 	safeClose(t, client)
 	seedBroker.Close()
+	// The refresh metadata is always for the same topic,
+	// it should have been batched.
+	require.Less(t, called.Load(), int32(7))
```
puellanivis: The test failure from this message is going to end up looking like:
```
--- Fail: TestClientUpdateMetadataErrorAndRetry
    client_test.go: 936
        	Error Trace:	/path/to/package/client_test.go:936
        	Error:      	"7" is not less than "7"
        	Test:      	TestLess
```

This provides no useful information.

```suggestion
	if count := called.Load(); count >= 7 {
		t.Errorf("Refresh metadata was called %d times, this should be less than 7.", count)
	}
```

client_test.go
```diff
@@ -911,6 +931,68 @@ func TestClientUpdateMetadataErrorAndRetry(t *testing.T) {
...
+		return resp
+	})
+
+	config := NewTestConfig()
+	config.Metadata.SingleFlight = true
+	client, err := NewClient([]string{seedBroker.Addr()}, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+	waitGroup := sync.WaitGroup{}
```
puellanivis: There’s no need to assign zero values:
```suggestion
	var waitGroup sync.WaitGroup
```

client_test.go
```diff
@@ -911,6 +931,68 @@ func TestClientUpdateMetadataErrorAndRetry(t *testing.T) {
...
+	waitGroup.Add(1000)
+	for i := 0; i < 1000; i++ {
+		go func() {
+			defer waitGroup.Done()
+			require.NoError(t, client.RefreshMetadata("topic1"))
+			require.NoError(t, client.RefreshMetadata("topic2"))
+			require.Error(t, client.RefreshMetadata("topic3"))
+			topics, err := client.Topics()
+			require.NoError(t, err)
+			require.Len(t, topics, 2)
```
puellanivis: These are now _definitely_ running in a separate goroutine from the one running the test.

This goroutine needs to not use `require`, `t.Fatal`, or `t.FailNow`.

config.go
```diff
@@ -167,6 +167,14 @@ type Config struct {
...
 		// if it is configured to do so (`auto.create.topics.enable` is true). Defaults to true.
 		AllowAutoTopicCreation bool
+
+		// SingleFlight controls whether to send a single metadata refresh request at a given time
+		// or whether to allow anyone to refresh the metadata concurrently.
+		// If this is set to true and the client needs to refresh the metadata from different goroutines,
+		// the requests will be batched together so that a single refresh is sent at a time.
+		// See https://github.com/IBM/sarama/issues/3224 for more details.
+		// SingleFlight defaults to false.
+		SingleFlight bool
```
puellanivis, line 177: I feel like there should be a more specific name than just “SingleFlight” which could broadly maybe mean all messages are done single flight.

metadata.go
```diff
@@ -0,0 +1,221 @@
+package sarama
+
+import (
+	"sync"
+)
+
+type MetadataRefresh func(topics []string) error
```
puellanivis: Is it necessary that this be exported?

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+
+// currentRefresh makes sure sarama does not issue metadata requests
+// in parallel. If we need to refresh the metadata for a list of topics,
+// this struct will check if a refresh is already ongoing, and if so, it will
+// accumulate the list of topics to refresh in the next refresh.
+// When the current refresh is over, it will queue a new metadata refresh call
+// with the accumulated list of topics.
+type currentRefresh struct {
+	sync.Mutex
```
puellanivis: [style] Do not just embed a mutex. This will expose everyone who has an instance of the struct to call the `Lock` and `Unlock` functions of the mutex, even though that is definitively an implementation detail.
```suggestion
	mu sync.Mutex
```

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+	chans     []chan error
+
+	// This is the function that gets called when to refresh the metadata.
+	// It is called with the list of all topics that need to be refreshed
+	// or with nil if all topics need to be refreshed.
+	refresh func(topics []string) error
+}
+
+func newCurrentRefresh(f func(topics []string) error) *currentRefresh {
+	mr := &currentRefresh{
```
puellanivis: [style] There is need to assign something to a local variable just to return it:
```suggestion
	return &currentRefresh{
```

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+		topicsMap: make(map[string]struct{}),
+		topics:    make([]string, 0),
+		refresh:   f,
+		chans:     make([]chan error, 0),
+	}
```
puellanivis: [style] There is no need to `make` empty slices: https://go.dev/wiki/CodeReviewComments#declaring-empty-slices
```suggestion
		refresh:   f,
	}
```

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+		r.topics = append(r.topics, topic)
+	}
+}
+
+func (r *nextRefresh) AddTopics(topics []string) {
+	if len(topics) == 0 {
+		r.allTopics = true
+		return
+	}
+	r.topics = append(r.topics, topics...)
```
puellanivis: This type has a mutex in it. There should be a mutex being held while modifying at least one of the fields.

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+		topics:    make([]string, 0),
+		refresh:   f,
+		chans:     make([]chan error, 0),
+	}
+	return mr
+}
+
+// AddTopicsFrom adds topics from the next refresh to the current refresh.
+// You need to hold the lock to call this method.
+func (r *currentRefresh) AddTopicsFrom(next *nextRefresh) {
```
puellanivis: Do we need to export this receiver method?

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+}
+
+// nextRefresh holds the list of topics we will need
+// to refresh in the next refresh.
+// When a refresh is ongoing, calls to RefreshMetadata() are
+// accumulated in this struct, so that we can immediately issue another
+// refresh when the current refresh is over.
+type nextRefresh struct {
+	sync.Mutex
```
puellanivis:
```suggestion
	mu sync.Mutex
```

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+// AddTopicsFrom adds topics from the next refresh to the current refresh.
+// You need to hold the lock to call this method.
+func (r *currentRefresh) AddTopicsFrom(next *nextRefresh) {
+	if next.allTopics {
+		r.allTopics = true
+		return
+	}
+	if len(next.topics) > 0 {
+		r.AddTopics(next.topics)
+		next.topics = next.topics[:0]
```
puellanivis: Here we truncate `next.topics` but not the `nesxt.allTopics` above?

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+func (r *nextRefresh) AddTopics(topics []string) {
+	if len(topics) == 0 {
+		r.allTopics = true
+		return
+	}
+	r.topics = append(r.topics, topics...)
+}
+
+func (r *currentRefresh) HasTopics(topics []string) bool {
+	if len(topics) == 0 {
```
puellanivis, line 85: Not sure I feel about the different semantics of `len(topics) == 0` being conflated with `allTopics`

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+	}
+	return true
+}
+
+// Start starts a new refresh.
+// The refresh is started in a new goroutine, and this function
+// returns a channel on which the caller can wait for the refresh
+// to complete.
+// You need to hold the lock to call this method.
```
puellanivis, line 105: If we need to be holding a mutex lock to call this method, then it should not also be exported. If it is exported, then we logically have no way to force compliance of holding the lock.

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+type singleFlightMetadataRefresher struct {
+	current *currentRefresh
+	next    *nextRefresh
+}
+
+func newSingleFlightRefresher(f func(topics []string) error) MetadataRefresh {
+	var refresher = &singleFlightMetadataRefresher{
+		current: newCurrentRefresh(f),
+		next: &nextRefresh{
+			topics: make([]string, 0),
```
puellanivis:
```suggestion
```

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+func (m *singleFlightMetadataRefresher) Refresh(topics []string) error {
+	for {
+		var current = m.current
+		current.Lock()
+		if !current.Ongoing() {
+			// If no refresh is ongoing, we can start a new one, in which
+			// we add the topics that have been accumulated in the next refresh
+			// and the topics that have been provided by the caller.
+			m.next.Lock()
+			m.current.AddTopicsFrom(m.next)
```
puellanivis: Here we refer to `m.current` but on line 188 , we copy that value into `current` I guess to save the `m.` on each use.

The more consistent way to handle this would be to just use `m.current` everywhere. (Unless for some reason, we need to replace the `m.current` value, in which we would obviously need a local variable pointer to hold the old current.

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+// It makes sure a sarama client never issues more than one metadata refresh
+// in parallel.
+type singleFlightMetadataRefresher struct {
+	current *currentRefresh
+	next    *nextRefresh
+}
+
+func newSingleFlightRefresher(f func(topics []string) error) MetadataRefresh {
+	var refresher = &singleFlightMetadataRefresher{
+		current: newCurrentRefresh(f),
```
puellanivis: `newCurrentRefresh` is only ever called here. We can just do it ourselves:
```suggestion
		current: &currentRefresh{
			f: f,
		},
```

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+		return
+	}
+	for _, topic := range topics {
+		if _, ok := r.topicsMap[topic]; ok {
+			continue
+		}
+		r.topicsMap[topic] = struct{}{}
+		r.topics = append(r.topics, topic)
+	}
+}
```
puellanivis: I’m not sure I see what value there is in keeping two copies of the `topics`. One as a slice, and the other as a map. We don’t need to order them in the outgoing request, right?

And since this is the only possible place we could be assigning into the map, and we had to grab a mutex to even be, we can lazy initialize the map:
```suggestion
func (r *currentRefresh) addTopics(topics []string) {
	if len(topics) == 0 {
		r.allTopics = true
		return
	}
	if r.topics == nil {
		// We know we're going to be trying to assign at least one value into this map.
		// So we can just test ahead of each addTopics.
		r.topics = make(map[string]struct{})
	}
	for _, topic := range topics {
		r.topics[topic] = struct{}{} // faster to just assign directly rather than test first
	}
}
```

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+		panic("waiting for a refresh that is not ongoing")
+	}
+	var ch = make(chan error, 1)
+	r.chans = append(r.chans, ch)
+	return ch
+}
+
+// Ongoing returns true if the refresh is ongoing.
+// You need to hold the lock to call this method.
+func (r *currentRefresh) Ongoing() bool {
```
puellanivis: This convenience function provides no convenience.

The `currentRefresh` lock must already be held, so we need to restrict this function from being called outside of this scope, so exporting it is not useful.

And since there is otherwise no locking, we could just export `r.ongoing` directly instead. Otherwise, we already can access `r.ongoing` ourselves.

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+// If a refresh was already ongoing for this list of topics, the function
+// waits on that refresh to complete, and returns its result.
+// If a refresh was already ongoing for a different list of topics, the function
+// accumulates the list of topics to refresh in the next refresh, and queues that refresh.
+// If no refresh is ongoing, it will start a new refresh, and return its result.
+func (m *singleFlightMetadataRefresher) Refresh(topics []string) error {
+	for {
+		var current = m.current
+		current.Lock()
+		if !current.Ongoing() {
```
puellanivis: To point of the inconvenience of this function, it’s far more convenient here to just access the field directly.
```suggestion
		if !current.ongoing {
```

metadata.go
```diff
@@ -0,0 +1,221 @@
...
+			// If no refresh is ongoing, we can start a new one, in which
+			// we add the topics that have been accumulated in the next refresh
+			// and the topics that have been provided by the caller.
+			m.next.Lock()
+			m.current.AddTopicsFrom(m.next)
+			m.next.Unlock()
+			current.AddTopics(topics)
+			var ch = current.Start()
+			current.Unlock()
```
puellanivis: Requiring each code path to properly unlock this mutex explicitly is error prone.

Perhaps, we should compose all these functions into a `ch, needToRetry := m.current.refreshTopics(topics)`

metadata_test.go
```diff
@@ -0,0 +1,84 @@
...
+		// There should never be two refreshes that run concurrently,
+		// so it's safe to change the map with no lock.
+		for _, topic := range topics {
+			called[topic]++
+		}
+		return nil
+	})
+	go func() {
+		require.NoError(t, refresh([]string{"topic1"}))
+	}()
```
puellanivis: From the other test file, you cannot call `t.FailNow` on a goroutine that is not the goroutine running the test.

metadata_test.go
```diff
@@ -0,0 +1,84 @@
...
+	go func() {
+		// This one and the next one will be batched together because the first
+		// call is still ongoing when they run.
+		// So we will issue a single refresh call, for topic2 and topic3.
+		time.Sleep(10 * time.Millisecond)
```
puellanivis: I feel like we could use channels to gate execution guaranteeing that the sequence of steps taken are those that are expected, without needing to use an arbitrary wait time “just to be sure”.

```go
first := make(chan struct{})
refresh := newSingleFlightRefresher(func(topics []string) error {
	select {
	case <-first: // already closed
	default:
		close(first)
	}
	// There should never be two refreshes that run concurrently,
	// so it's safe to change the map with no lock.
	for _, topic := range topics {
		called[topic]++
	}
	time.Sleep(10 * time.Millisecond) // unsure how to guarantee the second and third refresh have looped.
	return nil
})
var wg sync.WaitGroup
wg.Add(1)
go func() {
	defer wg.Done()
	assert.NoError(t, refresh([]string{"topic1"}))
}()
wg.Add(1)
go func() {
	defer wg.Done()
	<-first
	assert.NoError(t, refresh([]string{"topic3", "topic4"}))
}()
wg.Add(1)
go func() {
	defer wg.Done()
	<-first
	assert.NoError(t, refresh([]string{"topic3", "topic4"}))
}()
wg.Wait()
```

client_test.go
cupcicm: 👍 , fixing

client_test.go
cupcicm: 👍

client_test.go
cupcicm: Would you prefer that I used 
```
if xxx {
  t.Error()
}
```

or `assert.EqualValues` ? (rather than `require.` )

client_test.go
cupcicm: sorry, fixing 👍

client_test.go
cupcicm: fixing 👍

client_test.go
cupcicm: OK, fixing

client_test.go
cupcicm: Done.

client_test.go
cupcicm: Switching to assert. Tell me if you prefer the `if ... t.Error` pattern

metadata.go
cupcicm: not really, no, let me unexport it.

metadata.go
cupcicm: 👍 , done.

metadata.go
cupcicm: 👍 , done.

metadata.go
cupcicm: 👍, done

metadata.go
cupcicm: The caller holds the mutex while calling it, for consistency with `AddTopicsFrom` (although I could definitely lock inside the function. Let me address the other comments and come back to this)

metadata.go
cupcicm: No, fixed.

metadata.go
cupcicm: Fixed.

metadata.go
cupcicm: Let me clear the slice when the topics are added instead (in `nextRefresh.addTopics`)

metadata.go
cupcicm: True, done.

metadata.go
cupcicm: Fixed.

metadata.go
cupcicm: Ah ! Sorry! No, it's because in a previous iteration of my design I was replacing m.current by the "next refresh" but it's not necessary anymore. Let me just use `m.current` everywhere.

metadata.go
cupcicm: Done.

metadata.go
cupcicm: The only reason why I keep the map is to make the `hasTopics` function use it rather than linearly check in the slice for each of the topics.
If you think it's OK to loop through the slice all the time I can remove the map. WDYT ? (I think that some clusters have a lot of topics 😅 )

metadata.go
cupcicm: You're right, sorry about this.

metadata.go
cupcicm: Fixed.

metadata_test.go
cupcicm: replaced with `assert`. Tell me if you want me to do `if ... { t.Error() }` instead.

client_test.go
cupcicm: Fixing

metadata.go
cupcicm: I agree with you but I feel that it's not "my fault" so to speak ?
if the caller calls "RefreshMetadata(nil)" then it means "I want to refresh the data for all topics". This means that the `Refresh()` function needs to understand that `Refresh(nil)` means that too. This means that either in `Refresh` I make a "if topics == nil check that m.current.allTopics = true" or that I make this function understand that "nil" means "I want to know if all topics are in current. WDYT ?

metadata.go
cupcicm: ❗ very nice !
I hadn't thought about this, let me do this it's going to be clearer 👍

cupcicm:
Thank you for your very thoughtful and very precise review @puellanivis
I have (I think) addressed all your comments except the naming one. 

One possibly controversial change I made is the last commit, where I used the lower-level function you made me write (`refreshOrQueue`) in the test, which allows to resolve the race condition of "this request has been successfully enqueued but I haven't waited for the result to become available yet. If we allow ourselves to test this primitive (rather than the full `Refresh()`) we can test with no `time.Sleep`. If we don't, ... it's complicated (I think it's impossible because there is no point where we can assert we have enqueued the request without having to wait for the result).

About the naming of the configuration parameter, I wanted to discuss it with you. WDYT about
1. config.Metadata.BatchRequests
2. config.Metadata.DeduplicateRequests

? I personally liked "single flight" since I think it contains my idea of piggy backing on the previous refresh although it's a bit more complicated than usual singleflighting since we need to remember the topics.

client_test.go
puellanivis: Usually the former is more direct and clear. In this case, the `if cond { t.Error() }` pattern can provide a more useful test failure message than just “3 != 3”.

metadata.go
puellanivis: I mean, also that we should probably set `next.allTopic = false` in the other code path as well.

client_test.go
puellanivis: It is not as big of a deal `assert` vs `if … t.Error` We just need to be sure not to use `require` in separate goroutines.

metadata.go
puellanivis: It does seem like the `RefreshMetdata` request packet does use a zero-length topics list to mean “all topics”. 👍 I suppose we’re fine, and we can discard this comment.

metadata.go
puellanivis: I mean, there’s no reason why we cannot just use the map, and iterate over the map to make the slice to send?

puellanivis:
I think it’s ok to use the more internal `refreshOrQueue` to remove race conditions, as long as the exported behavior relying upon that internal function is reasonably transparent. (Simple enough that we can just look at the code and see if it is valid or not.)

Unsure which new configuration parameter would be appropriate. `config.Metadata.SingleFlight` might not even be so bad, since my concern was mostly on just knowing that it applied to metadata requests only. 🤔 I probably would prefer less “deduplicate” as we are avoiding duplicate metadata refreshes, but we’re also merging in all outstanding topics together, not just removing duplicate ones. So maybe something to mean “batching” is not such a bad idea.

client_test.go
```diff
@@ -4,13 +4,17 @@ package sarama
...
 	"sync"
 	"sync/atomic"
 	"syscall"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
 	"github.com/rcrowley/go-metrics"
```
puellanivis: All 3rd party imports should be group together:
```suggestion
	"github.com/rcrowley/go-metrics"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
```

client_test.go
```diff
@@ -911,6 +929,70 @@ func TestClientUpdateMetadataErrorAndRetry(t *testing.T) {
...
+func TestClientRefreshesMetadataConcurrently(t *testing.T) {
+	seedBroker := NewMockBroker(t, 1)
+
+	seedBroker.setHandler(func(req *request) (res encoderWithHeader) {
+		time.Sleep(10 * time.Millisecond)
+		require.EqualValues(t, 3, req.body.key(), "this test sends only Metadata requests")
+		topics := req.body.(*MetadataRequest).Topics
+		resp := new(MetadataResponse)
+		for _, topic := range topics {
+			if topic == "topic1" {
```
puellanivis: We could use a `switch topic` here instead?

metadata.go
```diff
@@ -0,0 +1,234 @@
...
+// If a refresh was already ongoing for a different list of topics, the function
+// accumulates the list of topics to refresh in the next refresh, and queues that refresh.
+// If no refresh is ongoing, it will start a new refresh, and return its result.
+func (m *singleFlightMetadataRefresher) Refresh(topics []string) error {
+	for {
+		ch, queued := m.refreshOrQueue(topics)
+		if !queued {
+			return <-ch
+		}
+		<-ch
```
puellanivis, line 187: Good… 👍 we throw away the error from the previous request if we’re needing to refresh again after blocking on one.

metadata.go
```diff
@@ -0,0 +1,234 @@
...
+	go func() {
+		err := r.refresh(topics)
+		r.mu.Lock()
+		r.ongoing = false
+		for _, ch := range r.chans {
+			ch <- err
+			close(ch)
+		}
+		r.clear()
+		r.mu.Unlock()
```
puellanivis: We could use a defer here.
```suggestion
		r.mu.Lock()
		defer r.mu.Unlock()

		r.ongoing = false
		for _, ch := range r.chans {
			ch <- err
			close(ch)
		}
		r.clear()
```

metadata_test.go
```diff
@@ -0,0 +1,93 @@
...
+func TestMetadataRefresh(t *testing.T) {
+	releaseRefresh := make(chan struct{})
+	refresh := newMetadataRefresh(func(topics []string) error {
+		<-releaseRefresh
+		return nil
+	})
+
+	ch, queued := refresh.refreshOrQueue([]string{"topic1"})
+	// It's the first call, it's not queued.
+	require.False(t, queued)
```
puellanivis: As explained before, these tests produce more or less useless test messages. Meanwhile, with the `if … Error` pattern, we can put the comment that you’ve added for each line as the failure message:
```go
if queued {
  t.Error("The first call should not queue.")
}
```

†: technically you can also use `require.False(t, queued, "the first call should not queue")`, but I still recommend the `if … Error` pattern in this case. Basically, no boolean needs to run through a `require` or `assert` since the only time it could fail is when its value is the other value. So there is never a point in printing out `False != True` or `True != False`. But not all tests have to be `if … Error`, assert is fine in some contexts.

client_test.go
cupcicm: Done.

client_test.go
cupcicm: Done.

metadata.go
cupcicm: OK, sorry. I am adding a `clear()` function to make it clearer what's happening.

metadata.go
cupcicm: 👍

metadata.go
cupcicm: Done.

metadata_test.go
cupcicm: OK. Changing them

cupcicm (commented):
@puellanivis Would you prefer if the feature is called "config.Metadata.DeduplicateRequests" ?

dnwe:
> @puellanivis Would you prefer if the feature is called "config.Metadata.DeduplicateRequests" ?

@cupcicm I think SingleFlight is fine, as mentioned it is already grouped under Metadata and conveys “Use a single merged metadata request per flight (true) or parallel requests (false)” and is understood as evocative of https://pkg.go.dev/golang.org/x/sync/singleflight

After some testing I will probably flip the default to true before cutting the next release so the config would just remain as an opt-out anyway – if we even choose to offer it at all

cupcicm:
OK. @puellanivis / @dnwe I have pushed an update that's exactly the same but with the signed commit, sorry. 

I think we are good then? I believe I have addressed all your comments? Please tell me if that's not the case.

epot:
@puellanivis / @dnwe what are we missing to be able to merge this? Thanks!

dnwe:
@epot nothing blocking, just been on holiday for most of July. I intend to merge this week and then do some post-merge testing with the main branch at scale

epot:
Ok great, thanks for the update!

client_test.go
```diff
@@ -813,81 +816,99 @@ func TestClientMetadataTimeout(t *testing.T) {
...
 func TestClientUpdateMetadataErrorAndRetry(t *testing.T) {
 	seedBroker := NewMockBroker(t, 1)
-
-	metadataResponse1 := new(MetadataResponse)
-	metadataResponse1.AddBroker(seedBroker.Addr(), 1)
-	seedBroker.Returns(metadataResponse1)
+	var called atomic.Int32
+
+	seedBroker.setHandler(func(req *request) (res encoderWithHeader) {
+		assert.EqualValues(t, 3, req.body.key(), "this test sends only Metadata requests")
```
puellanivis: 🤔 Now that I think about it, we want to ensure the handler exits out here, otherwise the `req.body.(*MetadataRequest)` would likely panic, so we probably want either:
```go
if !assert.EqualValues(…) {
	return
}
```
or
```go
if req.body.key() != 3 {
	…
	return
}
```

client_test.go
```diff
@@ -911,6 +928,71 @@ func TestClientUpdateMetadataErrorAndRetry(t *testing.T) {
...
+		t.Errorf("Refresh metadata was called %d times, this should be less than 7.", count)
+	}
+}
+
+func TestClientRefreshesMetadataConcurrently(t *testing.T) {
+	seedBroker := NewMockBroker(t, 1)
+
+	seedBroker.setHandler(func(req *request) (res encoderWithHeader) {
+		time.Sleep(10 * time.Millisecond)
+		require.EqualValues(t, 3, req.body.key(), "this test sends only Metadata requests")
```
puellanivis: Same as above, this shouldn’t be a `require` but should also test to make sure either the `assert` passed, or use a simple `if` test.

metadata.go
```diff
@@ -0,0 +1,240 @@
...
+	ongoing   bool
+	topicsMap map[string]struct{}
+	topics    []string
+	allTopics bool
+	chans     []chan error
+
+	// This is the function that gets called when to refresh the metadata.
+	// It is called with the list of all topics that need to be refreshed
+	// or with nil if all topics need to be refreshed.
+	refresh func(topics []string) error
```
puellanivis: [lint] A common style of ordering parameters is to have anything that can be accessed without holding the mutex above the mutex, and anything that needs the mutex below.

Since this is is a write-once value, it’s thread-safe to access without holding the lock.

metadata.go
```diff
@@ -0,0 +1,240 @@
...
+
+// start starts a new refresh.
+// The refresh is started in a new goroutine, and this function
+// returns a channel on which the caller can wait for the refresh
+// to complete.
+// You need to hold the lock to call this method.
+func (r *currentRefresh) start() chan error {
+	r.ongoing = true
+	ch := make(chan error, 1)
+	r.chans = append(r.chans, ch)
```
puellanivis: [polish] We should be able to use `ch := wait()` here?

metadata_test.go
```diff
@@ -0,0 +1,103 @@
...
+	require.NoError(t, <-ch3)
+	require.NoError(t, <-ch4)
+}
+
+func TestMetadataRefreshConcurrency(t *testing.T) {
+	var firstRefreshChans []chan error
+	var lock sync.Mutex
+	releaseRefresh := make(chan struct{})
+	refresh := newMetadataRefresh(func(topics []string) error {
+		<-releaseRefresh
```
puellanivis: [naming] 🤔 I think maybe “release” might not be the right word. I might expect that channel with such a name to be one that we close, and then everything forever is released. We could maybe use the more narrow semantic `step`?

puellanivis (commented):
Except for two minor concerns in the testing, everything else is just polishing.

PS: This means, once the test concerns are addressed, there’s no need to wait on any rereview.

metadata.go
epot: done

client_test.go
epot: done

client_test.go
epot: done

metadata.go
epot: sure

metadata_test.go
epot: renamed it to `step`

epot:
@puellanivis, I tried to address all your comments. Manu asked me to take over his PR while he was on holidays, but I am not necessarily super familiar with everything he did so I hope I didn't mess anything up. Could you take another look please?

client_test.go
```diff
@@ -886,7 +885,10 @@ func TestClientUpdateMetadataErrorAndRetry(t *testing.T) {
 	var called atomic.Int32
 
 	seedBroker.setHandler(func(req *request) (res encoderWithHeader) {
-		assert.EqualValues(t, 3, req.body.key(), "this test sends only Metadata requests")
+		if req.body.key() != 3 {
+			t.Fatal("this test sends only Metadata requests")
```
puellanivis: You were not here for the earlier comment. `t.FailNow` specifically explains:
> FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test.

Please replace this with `Error`.

client_test.go
```diff
@@ -940,7 +942,10 @@ func TestClientRefreshesMetadataConcurrently(t *testing.T) {
 
 	seedBroker.setHandler(func(req *request) (res encoderWithHeader) {
 		time.Sleep(10 * time.Millisecond)
-		require.EqualValues(t, 3, req.body.key(), "this test sends only Metadata requests")
+		if req.body.key() != 3 {
+			t.Fatal("this test sends only Metadata requests")
```
puellanivis: Same as above, this should be `t.Error`

metadata_test.go
puellanivis: You have unfortunately renamed the wrong thing. The `refresh` itself was fine, and appropriate. It’s the `releaseRefresh` that should be `stepRefresh`.

puellanivis (commented):
Just a two things.

client_test.go
epot: ah sorry !

client_test.go
epot: yes!

metadata_test.go
epot: ahah sorry I wasn't sure but I didn't want to try to be smart. Fixed now!

epot:
@puellanivis I pushed again, let me know!

puellanivis:
OK, I cannot see anything to comment about. 👍

epot:
great! so what happens now?

dnwe:
Please can you fix the DCO signoff?

epot:
Just did it @dnwe !

epot:
I noticed you triggered more tests and one is failing. Should I look into it or is it known as flaky?

dnwe:
Triggered a re-run to see

epot:
Looks good now :)

epot:
Great! Thanks :) Do you know when you plan to release it? Is there some place I can subscribe to know when it happens?

epot:
@dnwe sorry for the ping, did you see my message above? I am curious when this is going to be released. Thanks!

dnwe:
@epot no worries, yes it will be released soon, should be this week – with this and the ApiVersions adherence, there's a reasonable amount of change included so I have a few different teams doing some pre-release testing before I cut the tag

dnwe:
Note also that you can `go get github.com/IBM/sarama@main` at any time to use the latest main branch version in your app's go.mod/go.sum so you can start using this level of code right away too

epot:
Yes we are already doing that, but a bunch of customers where asking us when it would be released. Thanks a lot for the update!

PR #3227, opened 2025-07-21
chore: fix inconsistent function name in comment

stellrust:
fix inconsistent function name in comment


dnwe (approved):
Thanks for the catch!

PR #3234, opened 2025-08-03
feat: always send ApiVersionsRequest and fallback to v0

dnwe:
We previously pinned to only sending if Version was set to v2.4.0.0, but we should always send the request regardless, but fallback to sending a v0 request (without client info) if the v3 request does not succeed

In functional tests we always use MaxVersion in config an ApiVersionsRequest enabled to negotiate versions.

In unittests we disable ApiVersionsRequest by default and pin Version. A number of these tests use step-by-step ordered protocol expectations and it isn't convenient to update them all to also include an ApiVersionsRequest roundtrip unnecessarily, so we disable it in plain unittests by default, keeping it as opt-in.

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -175,8 +175,6 @@ func (b *Broker) Open(conf *Config) error {
 		return err
 	}
 
-	usingApiVersionsRequests := conf.Version.IsAtLeast(V2_4_0_0) && conf.ApiVersionsRequest
-
 	b.lock.Lock()
 
 	if b.metricRegistry == nil {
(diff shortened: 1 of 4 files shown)


broker.go
```diff
@@ -222,12 +220,14 @@ func (b *Broker) Open(conf *Config) error {
...
-		if usingApiVersionsRequests {
-			apiVersionsResponse, err := b.sendAndReceiveApiVersions()
+		if conf.ApiVersionsRequest {
+			apiVersionsResponse, err := b.sendAndReceiveApiVersions(3)
 			if err != nil {
 				Logger.Printf("Error while sending ApiVersionsRequest to broker %s: %s\n", b.addr, err)
-				// Don't return here - continue without API version discovery
-			} else {
+				// send a v0 request in case remote cluster is < 2.4.0.0
+				apiVersionsResponse, _ = b.sendAndReceiveApiVersions(0)
```
trapped, line 228: Nice 👌 I missed this detail in my original PR.

PR #3236, opened 2025-08-04
Hopefully temporary fix for problem noticed on Confluent 4.0 Brokers

rwilsonncsa:
Problem: CreateTopic request with version 3 protocol fails with EOF error. This change adds ConfluentRequestVersionBump config option that overrides protocol version sent in CreateTopic request (setting version to 4) when true.


rwilsonncsa:
Problem was discovered as Confluent not handling ReplicationFactor of -1 correctly. This change is not needed.

Issue #3233, opened 2025-08-01
CreateTopic fails with Confluent Cloud


blazerw:
##### Description

CreateTopic no longer works with Confluent cloud. The call returns with an EOF error. (We've found that setting the request.Version to 4 fixes the issue.

##### Versions

1.42.5

| Sarama | Kafka | Go  |
| ------ | ----- | --- |
| 1.42.5 | 4.0.x | 1.23.1 |

##### Additional Context

This code change fixed it for us in ./admin.go line 260 (https://github.com/IBM/sarama/blob/main/admin.go#L260)

```
    if ca.conf.Version.IsAtLeast(V3_0_0_0) {
        // Version 4 is the same as version 3 (except for unknown reason)
        request.Version = 4
    } else if ca.conf.Version.IsAtLeast(V2_0_0_0) {
        // Version 3 is the same as version 2 (brokers response before throttling)
        request.Version = 3
    } else if ca.conf.Version.IsAtLeast(V0_11_0_0) {
        // Version 2 is the same as version 1 (response has ThrottleTime)
        request.Version = 2
    } else if ca.conf.Version.IsAtLeast(V0_10_2_0) {
        // Version 1 adds validateOnly.
        request.Version = 1
    }
```

dnwe:
@blazerw hmm that seems odd. The validVersions for CreateTopic for Kafka 4.0.x are 2-7 (minVersion being bumped to 2) so as long as the `Version` field in sarama.Config is at least V0_11_0_0 then a v2 or newer request will be sent and should be supported

Ref: https://github.com/apache/kafka/blob/4.0.0/clients/src/main/resources/common/message/CreateTopicsRequest.json#L21-L34

blazerw:
I thought I'd add the log output (doesn't seem to be more than INFO level, though). This is without my version hack.
```
   time=2025-08-04T13:15:18.937-05:00 level=INFO msg="Initializing new client"
   time=2025-08-04T13:15:18.937-05:00 level=INFO msg="client/metadata fetching metadata for all topics from broker blibbety.us-east-1.aws.confluent.cloud:9092"
   time=2025-08-04T13:15:19.319-05:00 level=INFO msg="Connected to broker at blibbety.us-east-1.aws.confluent.cloud:9092 (unregistered)"
   time=2025-08-04T13:15:19.433-05:00 level=INFO msg="client/brokers registered new broker #0 at b0-blibbety.us-east-1.aws.confluent.cloud:9092"
[...]

PR #3238, opened 2025-08-05
feat(protocol): support CreateTopicRequest V4

dnwe:
Add support for using CreateTopicRequest V4 against Kafka 2.4 and newer clusters. Protocol-wise this is identical to the existing V3, but the version indicates that the server should accept `-1` as the partition and replicate-factor values to indicate "use the server side defaults" (KIP-464).

Contributes-to: IBM/sarama#3233

### DIFF
--- a/create_topics_request.go
+++ b/create_topics_request.go
@@ -20,16 +20,29 @@ func (c *CreateTopicsRequest) setVersion(v int16) {
 	c.Version = v
 }
 
-func NewCreateTopicsRequest(version KafkaVersion, topicDetails map[string]*TopicDetail, timeout time.Duration) *CreateTopicsRequest {
+func NewCreateTopicsRequest(
+	version KafkaVersion,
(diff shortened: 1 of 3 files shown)


create_topics_request.go
```diff
@@ -20,16 +20,29 @@ func (c *CreateTopicsRequest) setVersion(v int16) {
 	c.Version = v
 }
 
-func NewCreateTopicsRequest(version KafkaVersion, topicDetails map[string]*TopicDetail, timeout time.Duration) *CreateTopicsRequest {
+func NewCreateTopicsRequest(
+	version KafkaVersion,
+	topicDetails map[string]*TopicDetail,
+	timeout time.Duration,
+	validateOnly bool,
```
puellanivis, line 27: 😬 this is a breaking change.

Fortunately, it seems no one is actually using this exported function though.

PR #3243, opened 2025-08-07
chore: replace unnecessary go-multierror dependency

bestbug456:
Hello,

This proposal of changes remove the hashicorp dependencies, after scratching my head for a while I was not able to seen a good reason for keeping this dependencies and I hope that this commit was able to achieve it without losing any functionalities, at least from test execution everything looks like as before. The reason why I'm proposing this change is for two reason:
* I don't think is worth as dependence for the little contribution that is doing now days (hoping to not have misunderstood it)
* I personally find confusing when a project with MIT license use a third party library with GPL library, especially if the perimeter is very little like this one

Thanks for your time in reading this pull request!

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -89,8 +89,9 @@ var ErrTxnUnableToParseResponse = errors.New("transaction manager: unable to par
 // ErrUnknownMessage when the protocol message key is not recognized
 var ErrUnknownMessage = errors.New("kafka: unknown protocol message key")
 
-// Deprecated: MultiErrorFormat specifies the formatter applied to format multierrors.
-// Please use errors.Join() instead
+// MultiErrorFormat specifies the formatter applied to format multierrors.
+//
+// Deprecated: Please use [errors.Join] instead.
 func MultiErrorFormat(es []error) string {
 	if len(es) == 1 {
 		return es[0].Error()


errors.go
```diff
@@ -135,11 +115,51 @@ func Wrap(sentinel error, wrapped ...error) sentinelError {
 }
 
 func multiError(wrapped ...error) error {
```
puellanivis: Since Go 1.20, we have [errors.Join](https://pkg.go.dev/errors@go1.24.5#Join)

errors.go
```diff
@@ -91,24 +89,6 @@ var ErrTxnUnableToParseResponse = errors.New("transaction manager: unable to par
 // ErrUnknownMessage when the protocol message key is not recognized
 var ErrUnknownMessage = errors.New("kafka: unknown protocol message key")
 
-// MultiErrorFormat specifies the formatter applied to format multierrors. The
-// default implementation is a condensed version of the hashicorp/go-multierror
-// default one
-var MultiErrorFormat multierror.ErrorFormatFunc = func(es []error) string {
```
puellanivis, line 97: 🤔 There seems to be at least 6 references to this exported variable: https://github.com/search?q=sarama.MultiErrorFormat&type=code

We should instead mark it as deprecated, and maybe replace it with:
```
func MultiErrorFormat(es []error) string {
	/* we can go ahead and keep the code the same, to avert breaks of Hyrum’s Law */
}
```

Since the code seems to be of the following form, there shouldn’t be any need for this to be: 1) a variable. b) a specific type, let alone a [hapax](https://en.wiktionary.org/wiki/hapax) use of another module:
```
return errors.New(sarama.MultiErrorFormat(errs))
```

bestbug456: Thanks for pointing out, honestly surprising. That been say we if we want to keep the variable `MultiErrorFormat` and mark it as deprecated we can not have the function with the same name, Therefore I've add as proposal a function called `MultiErrorComposer`. Let me know if this was what you had in mind :)

errors.go
bestbug456: done

errors.go
puellanivis: We shouldn’t need to keep the variable at all. It being a variable to a function, or a function on its own only matters in what type it is. Right now, it’s a named type, but as simple function it is now an untyped function. Since its shape matches that of `multierror.ErrorFormatFunc` even if I have the most restrictive code:
```go
	var fn multierror.ErrorFormatFunc
	fn = sarama.MultiErrorFormat // this works just fine still
```

Even this still works fine:
```go
func foo(fn fn multierror.ErrorFormatFunc) {
}
func bar() {
	foo(sarama.MultiErrorFormat) // this also works fine
}
```

errors.go
```diff
@@ -91,10 +89,25 @@ var ErrTxnUnableToParseResponse = errors.New("transaction manager: unable to par
...
+// Deprecated: MultiErrorFormat specifies the formatter applied to format multierrors.
+// Please move to MultiErrorComposer function.
```
puellanivis: The deprecation notice should point people to just use `errors.Join()` instead, and that they should expect a differently formatted error as a result.

errors.go
```diff
@@ -91,10 +89,25 @@ var ErrTxnUnableToParseResponse = errors.New("transaction manager: unable to par
...
+var MultiErrorFormat = func(es []error) string {
```
puellanivis: Even if we go with keeping this as a variable, It’s even simpler than this; there is no need to repeat the function at all, you can just use a function as a value:
```suggestion
var MultiErrorFormat = MultiErrorComposer
```

But like I said, just convert this to a function. It’s still usable in all the same ways as a variable (except that it cannot have its value reassigned, but no one is doing that, and no one should be doing it, under the principle of don’t mutate global variables of another package.)

bestbug456: Hello!! :) 
On this I think is more a discussion on what we want to do, for me if someone is using this variable we should try to break the usage less as possible, pointed that is deprecated but keep the core functionalities I think is a good balance for a patch/minor release, in the meantime we provide the new function `MultiErrorComposer`, this will allow us when the "v2" of sarama will be out to remove the deprecated function. 
Personally, I'm not a big fan in this case to hide under the hood a reference to a new function since I would like to permit evolution of the function without bind it with the previous work. Therefore, I think we should have more a "branch out" from the previous work rather than "merge" the 2 together, grant some times for people to adapt and then remove definitely the variable. I would like to ear your opinion about of this approach

puellanivis: This is what I’m trying to explain to you. The only “breaking” change of changing this from a variable to a function, is if someone assigns a new function to this variable, which [no one seems to be doing at all](https://github.com/search?q=%22MultiErrorFormat+%3D%22&type=code), and they should never be doing anyways. The only use of that code being aliasing this function variable into their own local function variable.

All right-hand use assignments, uses in expressions, and use as a function parameter will see an untyped function with the same shape, and will coerce the type into whatever concrete type it is being assigned into. This is why you can have a `func(int16)` and call `func(42)`, and why you can use a plain `func foo(w http.ResponseWriter, r *http.Request)` as the `http.HandlerFunc` in `http.ServeMux.HandleFunc`, because the untyped function will get coerced to the named type in the function parameter.

Changing this from a `var` to a `func` is a safe change, because any code it would cause to break [should fail at compile time](https://go.dev/play/p/70KMkF4GFnI) and such code was never thread-safe already. Not to mention, [all public uses of the function](https://github.com/search?q=%22sarama.MultiErrorFormat%22&type=code) outside of this don’t even depend upon any typing.

PS: Actually, the only operators that can even operate on function types are: … `==`. And even then it is restricted exclusively to `== nil` (or vice-versa).

puellanivis: Also, deprecated means “we’re going to keep this working exactly the way it was†, and we’re not going to change it”. We _do not need_ this function, there is `errors.Join` now.

PS: even our own use of the `MultiErrorFormat` would be removed with this PR.

bestbug456: Thanks for your time, let me know if now the code is what you did expected or if there is something else.

errors.go
```diff
@@ -91,10 +89,9 @@ var ErrTxnUnableToParseResponse = errors.New("transaction manager: unable to par
 // ErrUnknownMessage when the protocol message key is not recognized
 var ErrUnknownMessage = errors.New("kafka: unknown protocol message key")
 
-// MultiErrorFormat specifies the formatter applied to format multierrors. The
-// default implementation is a condensed version of the hashicorp/go-multierror
-// default one
-var MultiErrorFormat multierror.ErrorFormatFunc = func(es []error) string {
+// Deprecated: MultiErrorFormat specifies the formatter applied to format multierrors.
+// Please use errors.Join() instead
```
puellanivis: Godoc should still start with the exported symbol name. The deprecation should follow the first paragraph as a separate paragraph.
```suggestion
// MultiErrorFormat specifies the formatter applied to format multierrors.
//
// Deprecated: Please use [errors.Join] instead.
```

It could potentially also still include the information about how it’s copying hashicorp’s multierror despite us no longer including the import. Since its style still(?) copies their format.

bestbug456: I've updated the deprecation notice, running `go fmt errors.go` it does not produce any code changes.

dnwe (approved):

PR #3260, opened 2025-08-12
chore(style): refactor compress.go for readability

dnwe:
Split out the gzip and lz4 case blocks similarly to the zstdCompress func


puellanivis:
👍

prestona (approved):

PR #3276, opened 2025-09-02
refactor to use reflect.TypeFor

Sahil-4555:
**What type of PR is this?**
> Other

**What does this PR do? Why is it needed?**
This PR replaces `reflect.TypeOf` with the newer `reflect.TypeFor`. The change removes the need for dummy values, avoids unnecessary allocations, and improves type safety with cleaner code.

**Which issues(s) does this PR fix?**
No specific issue; this is a code quality improvement using modern Go reflection.

### DIFF
--- a/mockbroker.go
+++ b/mockbroker.go
@@ -90,7 +90,7 @@ func (b *MockBroker) SetHandlerByMap(handlerMap map[string]MockResponse) {
 		fnMap[k] = v
 	}
 	b.setHandler(func(req *request) (res encoderWithHeader) {
-		reqTypeName := reflect.TypeOf(req.body).Elem().Name()
+		reqTypeName := reflect.TypeFor[protocolBody]().Name()
 		mockResponse := fnMap[reqTypeName]
 		if mockResponse == nil {
 			return nil
@@ -108,7 +108,7 @@ func (b *MockBroker) SetHandlerFuncByMap(handlerMap map[string]requestHandlerFun
 		fnMap[k] = v
 	}
 	b.setHandler(func(req *request) (res encoderWithHeader) {
-		reqTypeName := reflect.TypeOf(req.body).Elem().Name()
+		reqTypeName := reflect.TypeFor[protocolBody]().Name()
 		return fnMap[reqTypeName](req)
 	})
 }


Sahil-4555:
### Why You Might Use `reflect.TypeFor` Instead of `reflect.TypeOf` in Go

### Background

In Go, the `reflect` package provides two similar-looking functions:

-  `reflect.TypeOf(x)`
Takes a value `x` and returns its `reflect.Type`.

  ```go
  t := reflect.TypeOf(123) // type: int
  ```

-  `reflect.TypeFor[T]()` _(introduced in Go 1.22_)
A generic function that returns the `reflect.Type` for a type parameter `T` without needing a value.

  ```go
  t := reflect.TypeFor[int]() // type: int
  ```

###  Why Use `reflect.TypeFor` Instead of `reflect.TypeOf`?

 1. **No value needed**
   -  `reflect.TypeOf` requires an actual value at runtime.
  -  If you only know the type (not a value), you have to create a dummy value:
    
 ```go
    t := reflect.TypeOf((*MyStruct)(nil)).Elem()
```
    

   -  `reflect.TypeFor` works directly with the type parameter:
```go
   t := reflect.TypeFor[MyStruct]()
```    
   This is cleaner and avoids allocating or creating dummy values.
 
 2. **Compile-time type safety**
- With `reflect.TypeOf`, the type is determined from a runtime value, so mistakes may only show up at runtime.
-  With `reflect.TypeFor`, the type is determined at compile time from the generic type parameter, so it's safer and clearer.

 3. **Better for generic code**
  - In generic functions, you often have a type parameter `T` but no value of type `T`.
   `reflect.TypeOf`  can’t be used without creating a zero value: 
 
  ```go
    func PrintType[T any]() {
        var zero T
        fmt.Println(reflect.TypeOf(zero))
    }
  ```

  With `reflect.TypeFor`, you can simply do:
  
  ```go
    func PrintType[T any]() {
        fmt.Println(reflect.TypeFor[T]())
    }
  ```

 4. **Avoids unnecessary allocations**
  -  Creating a dummy value for `reflect.TypeOf` may allocate memory (especially for composite types).
  -  `reflect.TypeFor` avoids this overhead entirely.

### Summary Table

| Feature | `reflect.TypeOf` | `reflect.TypeFor` (Go 1.22+) |
|---------|------------------|-------------------------------|
| Requires a value | ✅ Yes | ❌ No |
| Works without allocation | ❌ Sometimes allocates | ✅ Yes |
| Compile-time type safety | ❌ Runtime only | ✅ Compile-time |
| Good for generics | ❌ Needs zero value | ✅ Directly works |

###  ✅ Recommendation

**Use `reflect.TypeFor` when:**
* You know the type at compile time (especially in generic code).
* You don't have or don't want to create a value.
* You want cleaner, safer, and allocation-free code.

**Use `reflect.TypeOf` when:**
* You already have a value and want its type at runtime.

mockbroker.go
```diff
@@ -90,7 +90,7 @@ func (b *MockBroker) SetHandlerByMap(handlerMap map[string]MockResponse) {
 		fnMap[k] = v
 	}
 	b.setHandler(func(req *request) (res encoderWithHeader) {
-		reqTypeName := reflect.TypeOf(req.body).Elem().Name()
+		reqTypeName := reflect.TypeFor[protocolBody]().Name()
```
puellanivis, line 93: This is not the same behavior. Because `protocolBody` is an interface, it gets passed into `reflect.TypeOf` without any conversion. Thus we’re getting the underlying type of what is _in_ `protocolBody` not that type itself. Here this would be a pointer to some concrete protocol body type, _not_ the `protocolBody` type itself.

The mistake here is that it is common that people will use `reflect.TypeOf(&interfacePrototype).Elem()` to get the type of the interface itself, since the dereference results in passing in a pointer-to-interface type into the `any` type parameter to `TypeOf`.

But that’s not the pattern we’re using here.

https://go.dev/play/p/ZwNsnoHSovl

puellanivis (commented):
This code is erroneous in principle and cannot actually be changed to accomplish a switch to using `reflect.TypeFor`.

puellanivis:
> **Use `reflect.TypeOf` when:**
> 
> * You already have a value and want its type at runtime.

To be absolutely clear. This is the very specific situation the existing code is using `TypeOf` for.

mockbroker.go
Sahil-4555: Thanks for pointing this out. My intention with switching to `reflect.TypeFor[protocolBody]()` was to simplify the reflection call, but I now realize it doesn’t capture the same behavior — it only returns the interface type itself, while the original code correctly gave us the underlying concrete type inside `req.body`. I’ll revert this change to preserve the intended behavior.

PR #3277, opened 2025-09-02
chore: refactor to use modern atomic types

Sahil-4555:
**What type of PR is this?**
> Other

**What does this PR do? Why is it needed?**
This PR updates our atomic operations to use `Go 1.19's` newer typed atomic approach instead of the old function-based method. We're changing fields like `int64` to `atomic.Int64` so we can write cleaner code like `child.retries.Add(1)` instead of `atomic.AddInt32(&child.retries, 1)`. This removes the need for pointer handling and gives us better compile-time safety when multiple threads access the same variables in  operations.

**Which issues(s) does this PR fix?**
The function-based atomics allowed accidental non-atomic access to shared variables, causing race conditions in concurrent  operations. The verbose `atomic.AddInt32(&child.retries, 1)` syntax was harder to maintain. Typed atomics enforce atomic-only access at compile time and use cleaner method calls, preventing threading bugs.

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -29,7 +29,7 @@ type Broker struct {
 	conn          net.Conn
 	connErr       error
 	lock          sync.Mutex
-	opened        int32
+	opened        atomic.Int32
 	responses     chan *responsePromise
 	done          chan bool
 
@@ -162,7 +162,7 @@ func NewBroker(addr string) *Broker {
 // follow it by a call to Connected(). The only errors Open will return directly are ConfigurationError or
 // AlreadyConnected. If conf is nil, the result of NewConfig() is used.
 func (b *Broker) Open(conf *Config) error {
-	if !atomic.CompareAndSwapInt32(&b.opened, 0, 1) {
+	if !b.opened.CompareAndSwap(0, 1) {
 		return ErrAlreadyConnected
 	}
 
@@ -189,7 +189,7 @@ func (b *Broker) Open(conf *Config) error {
 		if b.connErr != nil {
 			Logger.Printf("Failed to connect to broker %s: %s\n", b.addr, b.connErr)
 			b.conn = nil
-			atomic.StoreInt32(&b.opened, 0)
+			b.opened.Store(0)
 			return
 		}
 		if conf.Net.TLS.Enable {
@@ -254,7 +254,7 @@ func (b *Broker) Open(conf *Config) error {
 					Logger.Printf("Error while closing connection to broker %s (due to SASL v0 auth error: %s): %s\n", b.addr, b.connErr, err)
 				}
 				b.conn = nil
-				atomic.StoreInt32(&b.opened, 0)
+				b.opened.Store(0)
 				return
 			}
 		}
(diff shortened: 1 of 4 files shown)


broker.go
```diff
@@ -29,7 +29,7 @@ type Broker struct {
 	conn          net.Conn
 	connErr       error
 	lock          sync.Mutex
-	opened        int32
+	opened        atomic.Int32
```
puellanivis: This looks like we want ̀ atomic.Bool` instead.

consumer.go
```diff
@@ -411,9 +411,9 @@ type partitionConsumer struct {
 	responseResult error
 	fetchSize      int32
 	offset         int64
-	retries        int32
+	retries        atomic.Int32
 
-	paused int32
+	paused atomic.Int32 // accessed atomically, 0 = not paused, 1 = paused
```
puellanivis: Again we want `atomic.Bool`

mocks/consumer.go
```diff
@@ -223,15 +223,15 @@ func (c *Consumer) ExpectConsumePartition(topic string, partition int32, offset
...
-			errors:              make(chan *sarama.ConsumerError, c.config.ChannelBufferSize),
+			t:                  c.t,
+			topic:              topic,
+			partition:          partition,
+			offset:             offset,
+			messages:           make(chan *sarama.ConsumerMessage, c.config.ChannelBufferSize),
+			suppressedMessages: make(chan *sarama.ConsumerMessage, c.config.ChannelBufferSize),
+			errors:             make(chan *sarama.ConsumerError, c.config.ChannelBufferSize),
 		}
+		c.partitionConsumers[topic][partition].highWaterMarkOffset.Store(highWatermarkOffset)
```
puellanivis: Alternatively:
```
consumer := &PartitionConsumer{
	…
}
consumer.highWaterMarkOffset.Store(highWatermarkOffset)
c.partitionConsumers[topic][partition] = consumer
```

mocks/consumer.go
```diff
@@ -345,15 +345,15 @@ func (pc *PartitionConsumer) Messages() <-chan *sarama.ConsumerMessage {
...
+	return pc.highWaterMarkOffset.Load()
 }
 
 // Pause implements the Pause method from the sarama.PartitionConsumer interface.
 func (pc *PartitionConsumer) Pause() {
 	pc.l.Lock()
 	defer pc.l.Unlock()
 
-	pc.suppressedHighWaterMarkOffset = atomic.LoadInt64(&pc.highWaterMarkOffset)
+	pc.suppressedHighWaterMarkOffset.Store(pc.highWaterMarkOffset.Load())
```
puellanivis: `x.Store(y.Load())` cannot be be composed into one atomic operation. The Load will still be atomic, but then the Store will happen in a separate atomic operation.

Fortunately, we don’t have to worry about doing any sort of coördination here, because `suppressedHighWaterMarkOffset` is already only accessed while holding the `pc.l` mutex, and so it does not have to be atomic.

puellanivis (commented):
I don’t see anything to comment on. 👍

dnwe (approved):
Thanks! A good piece of housekeeping

PR #3279, opened 2025-09-05
feat: add SASLv1 support for Kerberos

dnwe:
For some reason we were still pinning the old v0 SASL when using GSSAPI and this doesn't work if an ApiVersionsRequest is sent before the auth flow.
Add support for sending the krb5 bytes in the v1 SASL SaslAuthenticate protocol wrapping.

Note: this is just a draft for now as although it is tested and working against a GSSAPI enabled cluster, the underlying code is very much a copy-and-paste of the existing v0 auth flow, just wrappering the bytes in the SaslAuthenticate rather than sending them to the broker directly. This needs more work to tidyup the implementation

### DIFF
--- a/gssapi_kerberos.go
+++ b/gssapi_kerberos.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"net"
 	"strings"
 	"time"
 
@@ -202,7 +203,7 @@ func (krbAuth *GSSAPIKerberosAuth) initSecContext(bytes []byte, kerberosClient K
 }
 
 func (krbAuth *GSSAPIKerberosAuth) spn(broker *Broker) string {
-	host := strings.SplitN(broker.addr, ":", 2)[0]
+	host, _, _ := net.SplitHostPort(broker.addr)
 	var spn string
 	if krbAuth.Config.BuildSpn != nil {
 		spn = krbAuth.Config.BuildSpn(broker.conf.Net.SASL.GSSAPI.ServiceName, host)
--- a/gssapi_kerberos.go
+++ b/gssapi_kerberos.go
@@ -117,7 +117,8 @@ func (krbAuth *GSSAPIKerberosAuth) newAuthenticatorChecksum() []byte {
 func (krbAuth *GSSAPIKerberosAuth) createKrb5Token(
 	domain string, cname types.PrincipalName,
 	ticket messages.Ticket,
-	sessionKey types.EncryptionKey) ([]byte, error) {
+	sessionKey types.EncryptionKey,
+) ([]byte, error) {
 	auth, err := types.NewAuthenticator(domain, cname)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


gssapi_kerberos.go
```diff
@@ -200,63 +201,122 @@ func (krbAuth *GSSAPIKerberosAuth) initSecContext(bytes []byte, kerberosClient K
...
-	if err != nil {
-		Logger.Printf("Kerberos client error: %s", err)
-		return err
-	}
-	// Construct SPN using serviceName and host
-	// default SPN format: <SERVICE>/<FQDN>
-
-	host := strings.SplitN(broker.addr, ":", 2)[0] // Strip port part
+func (krbAuth *GSSAPIKerberosAuth) spn(broker *Broker) string {
+	host := strings.SplitN(broker.addr, ":", 2)[0]
```
puellanivis: I would recommend we switch to `strings.Cut`:
```suggestion
	host, _, _ := strings.Cut(broker.addr, ":")
```

gssapi_kerberos.go
```diff
@@ -117,7 +117,8 @@ func (krbAuth *GSSAPIKerberosAuth) newAuthenticatorChecksum() []byte {
 func (krbAuth *GSSAPIKerberosAuth) createKrb5Token(
 	domain string, cname types.PrincipalName,
 	ticket messages.Ticket,
-	sessionKey types.EncryptionKey) ([]byte, error) {
+	sessionKey types.EncryptionKey,
+) ([]byte, error) {
```
puellanivis: I approve matching the `Struct{}` multiline format, but I wonder, do we even need to break up this line at all?

gssapi_kerberos.go
```diff
@@ -200,63 +201,122 @@ func (krbAuth *GSSAPIKerberosAuth) initSecContext(bytes []byte, kerberosClient K
...
 	if krbAuth.Config.BuildSpn != nil {
 		spn = krbAuth.Config.BuildSpn(broker.conf.Net.SASL.GSSAPI.ServiceName, host)
 	} else {
 		spn = fmt.Sprintf("%s/%s", broker.conf.Net.SASL.GSSAPI.ServiceName, host)
 	}
+	return spn
+}
+
+// Login will use the given KerberosClient to login and get a ticket for the given spn.
+func (krbAuth *GSSAPIKerberosAuth) Login(kerberosClient KerberosClient, spn string) (*messages.Ticket, error) {
```
puellanivis: I know it’s existing code, but `kerberosClient` seems overly specific. Surely, `client` would be clear enough?

gssapi_kerberos.go
```diff
@@ -200,63 +201,122 @@ func (krbAuth *GSSAPIKerberosAuth) initSecContext(bytes []byte, kerberosClient K
...
+		authResponse, err := authSendReceiver(token)
+		if err != nil {
+			Logger.Printf("SASL Kerberos authenticate error as %s: %s", principal, err)
+			return err
+		}
+
+		if authResponse.Err != ErrNoError {
+			Logger.Printf("SASL Kerberos authentication failed as %s: %s", principal, authResponse.Err)
+			return authResponse.Err
+		}
```
puellanivis: :squint: I think `authSendReceiver` takes care of converting an `authResponse.Err` into the `error` return value?

gssapi_kerberos.go
```diff
@@ -200,63 +201,122 @@ func (krbAuth *GSSAPIKerberosAuth) initSecContext(bytes []byte, kerberosClient K
...
 			}
-		} else if krbAuth.step == GSS_API_FINISH {
+		case GSS_API_FINISH:
+			return nil
+		}
+	}
+}
+
+// AuthorizeV2 performs the SASL v2 GSSAPI authentication with the Kafka broker.
+func (krbAuth *GSSAPIKerberosAuth) AuthorizeV2(broker *Broker, authSendReceiver func(authBytes []byte) (*SaslAuthenticateResponse, error)) error {
```
puellanivis: 😬 I’m not sure how I feel about passing in a fairly arbitrary function here.

puellanivis (commented):
Understanding this is a draft. I mean, it all looks pretty good so far. I think all I commented on was style things. 😂

gssapi_kerberos.go
dnwe: Yes good spot, I added this one to my [grab-bag of semgrep rules](https://github.com/dgryski/semgrep-go/pull/70) :)

hindessm: https://pkg.go.dev/net#SplitHostPort ?

gssapi_kerberos.go
dnwe: Yeah as you've probably seen, this is mimicking what the other `b.sendAndReceiveSASLFoo` funcs currently accept and use and was just the simplest way to get a V2 flow working

gssapi_kerberos.go
dnwe: Yes it's a bit of a stylistic choice. I hadn't really looked at the gssapi code before so I found the long func declarations hard to read whilst first becoming accustomed to the code. Added a 'style' commit to more consistently apply this across the file when the func declarations exceed a threshold length

gssapi_kerberos.go
dnwe: Good spot, removed

gssapi_kerberos.go
dnwe: Fair, have added a commit to fix this consistently across the file

hindessm (approved):

Issue #1722, opened 2020-06-11
Write i/o timeout when Closing TLS connection


KenjiTakahashi:
##### Versions

| Sarama | Kafka | Go |
|--------|-------|----|
| v1.26.4/0189d59 | 2.5.0 | 1.14.4 |

##### Configuration

Really the important part is TLS Enabled, i.e.
``` go
config.Net.TLS.Enable = true
config.Net.TLS.Config = tlsConfig
```

##### Logs

Again, the relevant part is this log.

```
[SARAMA] 2020/06/11 10:24:59 Error while closing connection to broker localhost:9092: write tcp [::1]:37390->[::1]:9092: i/o timeout
```

##### Problem Description

After enabling TLS, the above error happens on client `Close()` call. Without TLS, everything works fine.
I went down to it and figured that the culprit is https://github.com/Shopify/sarama/blob/0189d59e5253ee7aa2a9eb50bffd18a69460630d/broker.go#L309
but that's probably because it's the first write executed on the connection?

Going further, I found out that removing deadline from https://github.com/Shopify/sarama/blob/0189d59e5253ee7aa2a9eb50bffd18a69460630d/broker.go#L708
(or setting it to `time.Time{}`) removes the problem.
Then I found this report https://github.com/golang/go/issues/13828 stating that `SetWriteDeadline` does not seem to work properly on TLS Clients. Which appears to be (still) true.

Honestly, not sure what can realistically be done about it in `sarama`, removing deadlines does not sound like viable solution. Still thought I will share my findings.

d1egoaz:
maybe @dnwe has more information about this.
He added something similar here https://github.com/Shopify/sarama/pull/1529

PR #3283, opened 2025-09-08
fix: add read deadline to tls write

bvalente:
Related to:
- https://github.com/golang/go/issues/13828
- https://github.com/IBM/sarama/issues/1722

We're using https://github.com/Mongey/terraform-provider-kafka to manage Kafka Topics with Terraform. Recently we've changed from Plaintext communications to AWS IAM Authentication. When doing so, our provider sometimes would hang indefinitely on some plans. We pinned this to the `kafka.t3.small` cluster tiers, as these have several limitations, including a maximum of 4 TCP connections per second.

While debugging the provider, we understood that the Call Stack was stuck on writing to the cluster, more specifically right on the first communication that it was trying to do with the clusters. Reading through the code, we found a very interesting comment for the Write function of the TLS package.

https://github.com/golang/go/blob/go1.23.0/src/crypto/tls/conn.go#L1192-L1195
```
// As Write calls [Conn.Handshake], in order to prevent indefinite blocking a deadline
// must be set for both [Conn.Read] and Write before Write is called when the handshake
// has not yet completed. See [Conn.SetDeadline], [Conn.SetReadDeadline], and
// [Conn.SetWriteDeadline].
```
Based on this, TLS requires both Write and Read Deadlines to be set because the Write function may do a handshake on the fist communication, and the handshake both Writes and Reads.

I believe that in our case, since we are working with brokers that don't have a very reliable network, sometimes the handshake would not progress on the server side, and we would indefinitely wait for a Read that would never come.

After implementing this change in our local workstation, instead of experiencing indefinite hanging, the program would finally report some time of error:
```
Error: kafka: client has run out of available brokers to talk to: read tcp 10.xxx.xxx.xxx:59582->10.xxx.xxx.xxx:9098: i/o timeout
```

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -958,11 +958,20 @@ func (b *Broker) readFull(buf []byte) (n int, err error) {
 	return io.ReadFull(b.conn, buf)
 }
 
-// write  ensures the conn WriteDeadline has been setup before making a
+// write ensures the conn Deadline has been setup before making a
 // call to conn.Write
 func (b *Broker) write(buf []byte) (n int, err error) {
-	if err := b.conn.SetWriteDeadline(time.Now().Add(b.conf.Net.WriteTimeout)); err != nil {
+	now := time.Now()
+	if err := b.conn.SetWriteDeadline(now.Add(b.conf.Net.WriteTimeout)); err != nil {
 		return 0, err
+	}
+	// TLS connections require both read and write deadlines to be set
+	// to avoid handshake indefinite blocking
+	// see https://github.com/golang/go/blob/go1.23.0/src/crypto/tls/conn.go#L1192-L1195
+	if b.conf.Net.TLS.Enable {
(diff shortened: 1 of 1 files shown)


puellanivis (commented):
Only thing I could think is to join the `time.Now()` calls into a common local variable, so they’re both based on the same “now”.

But yeah, good change all over otherwise. 👍

bvalente:
@puellanivis thank you for the review

I addressed your comment, and force pushed after rebasing with master

puellanivis (commented):
Looks great. :)

bvalente:
Hello @puellanivis, what would be the process to get this merged and tagged? Is there a timeline, or anything I can do from our side? 🙂

puellanivis:
Sometimes reviews from IBM can take a while. I don’t actually have any ability to even approve in my code review, let alone merge anything. I’m just a third-party F/OSS contributor helping out with code reviews.

PR #3292, opened 2025-09-15
feat: support KIP-848 ListConsumerGroups protocol

hindessm:
Add TypesFilter to request and GroupType to group data.


list_groups_response.go
```diff
@@ -115,13 +123,20 @@ func (r *ListGroupsResponse) decode(pd packetDecoder, version int16) error {
 		r.Groups[groupId] = protocolType
 
 		if r.Version >= 4 {
+			var groupData = GroupData{}
```
puellanivis: [style] Avoid assigning zero values. This is done automatically.
```suggestion
			var groupData GroupData
```

PR #3293, opened 2025-09-15
feat: support more describe log dirs versions

hindessm:


admin.go
```diff
@@ -1176,14 +1176,24 @@ func (ca *clusterAdmin) DescribeLogDirs(brokerIds []int32) (allLogDirs map[int32
...
+			} else if ca.conf.Version.IsAtLeast(V2_0_0_0) {
 				request.Version = 1
 			}
 			response, err := b.DescribeLogDirs(request)
 			if err != nil {
 				errChan <- err
 				return
 			}
+			if !errors.Is(response.ErrorCode, ErrNoError) {
+				errChan <- err
```
puellanivis: This value is known to be `nil` because if it were non-nil it would have been sent to the channel on line 1190.

describe_log_dirs_request.go
```diff
@@ -21,34 +21,68 @@ type DescribeLogDirsRequestTopic struct {
...
+	isFlexible := r.Version >= 2
 
-	for _, d := range r.DescribeTopics {
-		if err := pe.putString(d.Topic); err != nil {
+	if isFlexible {
+		if len(r.DescribeTopics) == 0 {
+			pe.putUVarint(0)
+		} else {
+			pe.putCompactArrayLength(len(r.DescribeTopics))
+		}
```
puellanivis: This code works out to be the same as below, except with `putCompactArrayLength` instead of `putArrayLength`:
```suggestion
		length := len(r.DescribeTopics)
		if length == 0 {
			length = -1
		}
		pe.putCompactArrayLength(length)
```

In fact, the two code paths are now mostly the same and we can just use:
```go
	length := len(r.DescribeTopics)
	if length == 0 {
		// In order to query all topics we must send null
		length = -1
	}
	if isFlexible {
		pe.putCompactArrayLength(length)
	} else {
		if err := pe.putArrayLength(length); err != nil {
			return err
		}
	}
```

describe_log_dirs_request.go
```diff
@@ -21,34 +21,68 @@ type DescribeLogDirsRequestTopic struct {
...
+	for _, d := range r.DescribeTopics {
+		if isFlexible {
+			if err := pe.putCompactString(d.Topic); err != nil {
+				return err
+			}
+
+			if err := pe.putCompactInt32Array(d.PartitionIDs); err != nil {
+				return err
+			}
+			if isFlexible {
```
puellanivis: This test is already known to be true, because we’re in an `if isFlexible` conditional.

dnwe:
@hindessm thanks! I added a simple FV to PR just to exercise it across the kafka versions within GHA

admin.go
hindessm: Oops. It should be `errChan <- response.ErrorCode`. I'll fix it.

describe_log_dirs_request.go
hindessm: I was trying to be consistent with the slightly similar code in `offset_fetch_request.go` handling of `partitions` but I agree the suggested code is much nicer (and functionally the same). (I prefer the strict `nil` check `offset_fetch_request.go` uses for `partitions` to the `len` check used here.)

dnwe (approved):
Thanks, LGTM now

PR #3297, opened 2025-09-18
chore: apply modernize fixes from gopls

dnwe:
Apply a selection of the gopls [modernize](https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize) fixes category-by-category:
```
go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -category=$CAT ./...
```

PR #3302, opened 2025-09-22
fix: update map rather than create a new map

hindessm:
This PR was originally just minor refactoring to improve map allocations. But a bug was found while responding to the excellent PR feedback from Cassondra.

Cassondra rightly points out that the explicit capacities are hints so don't always reduce (re)allocations. In fact, particularly when the capacities are small, they make no difference, but it is still good practice to be explicit when we know the size.

### DIFF
--- a/balance_strategy.go
+++ b/balance_strategy.go
@@ -321,7 +321,8 @@ func (s *stickyBalanceStrategy) balance(currentAssignment map[string][]topicPart
 	// make sure we are getting a more balanced assignment; otherwise, revert to previous assignment
 	if !initializing && reassignmentPerformed && getBalanceScore(currentAssignment) >= getBalanceScore(preBalanceAssignment) {
 		currentAssignment = deepCopyAssignment(preBalanceAssignment)
-		currentPartitionConsumer = maps.Clone(preBalancePartitionConsumers)
+		clear(currentPartitionConsumer)
+		maps.Copy(currentPartitionConsumer, preBalancePartitionConsumers)
 	}
 
 	// add the fixed assignments (those that could not change) back


admin.go
```diff
@@ -249,7 +249,7 @@ func (ca *clusterAdmin) CreateTopic(topic string, detail *TopicDetail, validateO
 		return errors.New("you must specify topic details")
 	}
 
-	topicDetails := make(map[string]*TopicDetail)
+	topicDetails := make(map[string]*TopicDetail, 1)
 	topicDetails[topic] = detail
```
puellanivis: I recommend using complex literals when all the fields are known:
```suggestion
	topicDetails := map[string]*TopicDetail{
		topic: detail,
	}
```

admin.go
```diff
@@ -479,7 +479,7 @@ func (ca *clusterAdmin) CreatePartitions(topic string, count int32, assignment [
 		return ErrInvalidTopic
 	}
 
-	topicPartitions := make(map[string]*TopicPartition)
+	topicPartitions := make(map[string]*TopicPartition, 1)
 	topicPartitions[topic] = &TopicPartition{Count: count, Assignment: assignment}
```
puellanivis:
```suggestion
	topicPartitions := map[string]*TopicPartition{
		topic: &TopicPartition{
			Count:      count,
			Assignment: assignment,
		},
	}
```

admin.go
```diff
@@ -615,8 +615,8 @@ func (ca *clusterAdmin) DeleteRecords(topic string, partitionOffsets map[int32]i
 		partitionPerBroker[broker] = append(partitionPerBroker[broker], partition)
 	}
 	for broker, partitions := range partitionPerBroker {
-		topics := make(map[string]*DeleteRecordsRequestTopic)
-		recordsToDelete := make(map[int32]int64)
+		topics := make(map[string]*DeleteRecordsRequestTopic, 1)
```
puellanivis: Move this down to line 623, and convert it and lines 623–625 to a complex literal.

admin.go
```diff
@@ -1032,7 +1032,7 @@ func (ca *clusterAdmin) ListConsumerGroups() (allGroups map[string]string, err e
 				return
 			}
 
-			groups := make(map[string]string)
+			groups := make(map[string]string, len(response.Groups))
 			maps.Copy(groups, response.Groups)
```
puellanivis:
```suggestion
			groups := maps.Clone(response.Groups)
```

puellanivis (commented):
N.B. that adding a capacity hint to `make(map[K]V)` is merely a hint, and could be ignored by the compiler/runtime…

Also, it’s still better that we provide it when we have it.

admin.go
hindessm: Rather than just do this one, I decided to make an additional commit with all of the uncontentious changes from your earlier comments on #3297.

One of the changes, which looks logically correct to me, causes a linter error so I may live to regret making the wider changes.

admin.go
hindessm: Done.

balance_strategy.go
```diff
@@ -313,17 +313,15 @@ func (s *stickyBalanceStrategy) balance(currentAssignment map[string][]topicPart
...
 
 	reassignmentPerformed := s.performReassignments(sortedPartitions, currentAssignment, prevAssignment, sortedCurrentSubscriptions, consumer2AllPotentialPartitions, partition2AllPotentialConsumers, currentPartitionConsumer)
 
 	// if we are not preserving existing assignments and we have made changes to the current assignment
 	// make sure we are getting a more balanced assignment; otherwise, revert to previous assignment
 	if !initializing && reassignmentPerformed && getBalanceScore(currentAssignment) >= getBalanceScore(preBalanceAssignment) {
 		currentAssignment = deepCopyAssignment(preBalanceAssignment)
-		currentPartitionConsumer = make(map[topicPartitionAssignment]string, len(preBalancePartitionConsumers))
-		maps.Copy(currentPartitionConsumer, preBalancePartitionConsumers)
+		currentPartitionConsumer = maps.Clone(preBalancePartitionConsumers)
```
puellanivis: Huh. This was even called out when it was implemented: https://github.com/IBM/sarama/pull/1416/files#r314581692

Turns out, the code it’s copying used `currentPartitionConsumer.clear()` which deletes all the entries from the underlying map referenced, rather than assigning a whole brand new map into it. In this way, the previous references to the map would reflect these updates, even when they’re not directly used later in this function. However, this code just constructs a whole new map, rather than clearing out the existing map, so that other references to the same map do not reflect the changes made… different maps.

The whole `StickyAssignor.java` file is _completely_ different now… but updating to copy that code is obviously out of scope.

I would instead recommend here:
```
clear(currentPartitionConsumer)
maps.Copy(currentPartitionConsumer, preBalancePartitionConsumers)
```

This should be the intended original code, rather than this subtly bugged(?) code.

puellanivis:
I can find nothing further to comment about.

hindessm:
FYI: Somewhat related to this, I added a comment about the use of maps in the API to https://github.com/IBM/sarama/wiki/Ideas-that-will-break-backwards-compatibility.

puellanivis:
> FYI: Somewhat related to this, I added a comment about the use of maps in the API to https://github.com/IBM/sarama/wiki/Ideas-that-will-break-backwards-compatibility.

I don’t seem to see any way to add any discussion of things, though? Like:
* logrus is now in maintenance mode. It is advised that new code not use it; besides, we have `log/slog` now.


(Saving further comments for a better place to discuss.)

PR #3303, opened 2025-09-22
wip: Flexible experiment

hindessm:
I plan to update all the protocol support to v2.4 level. While investigating what was involved, I started experimenting with slightly different approach to dealing with flexible versions (as several of the required protocol updates are just flexible version updates).

Before I go to far down this path, I'd be interested in any feedback so I thought I'd open a wip PR. (I've included a couple of protocol updates just to exercise the new mechanism and get a feel for writing with it but I'll probably break up the eventual non-draft PR.)

### DIFF
--- a/create_topics_request.go
+++ b/create_topics_request.go
@@ -33,8 +33,9 @@ func NewCreateTopicsRequest(
 	}
 	switch {
 	case version.IsAtLeast(V2_4_0_0):
+		// Version 5 is the first flexible version
 		// Version 4 makes partitions/replicationFactor optional even when assignments are not present (KIP-464)
-		r.Version = 4
+		r.Version = 5
 	case version.IsAtLeast(V2_0_0_0):
 		// Version 3 is the same as version 2 (brokers response before throttling)
 		r.Version = 3
@@ -49,6 +50,7 @@ func NewCreateTopicsRequest(
 }
 
 func (c *CreateTopicsRequest) encode(pe packetEncoder) error {
+	pe.setFlexible(c.Version >= 5)
 	if err := pe.putArrayLength(len(c.TopicDetails)); err != nil {
 		return err
 	}
@@ -67,10 +69,12 @@ func (c *CreateTopicsRequest) encode(pe packetEncoder) error {
 		pe.putBool(c.ValidateOnly)
 	}
 
+	pe.maybePutEmptyTaggedFieldArray()
 	return nil
 }
 
 func (c *CreateTopicsRequest) decode(pd packetDecoder, version int16) (err error) {
+	pd.setFlexible(version >= 5)
 	n, err := pd.getArrayLength()
 	if err != nil {
 		return err
@@ -103,7 +107,9 @@ func (c *CreateTopicsRequest) decode(pd packetDecoder, version int16) (err error
 
 		c.Version = version
 	}
-
+	if _, err := pd.maybeGetEmptyTaggedFieldArray(); err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


real_encoder.go
```diff
@@ -217,3 +237,7 @@ func (re *realEncoder) pop() error {
 func (re *realEncoder) metricRegistry() metrics.Registry {
 	return re.registry
 }
+
+func (re *realEncoder) setFlexible(v bool) {
```
puellanivis, line 241: I’m not sure how much we need to be using setters, when we are our own package. We can reach into the struct and just set the value directly, no need to have a setter?

Oh… because it’s all accessed through an interface. 😐

puellanivis:
🤔 Now that I think about it… we’re using an interface for encoders already right?

… Why not provide two implementations one that uses varints, and one that doesn’t, and then just get a flexible or not direct implementation, and save execution from a bunch of if blocks. (Though, they’ll likely end up getting correctly predicted a lot in general.)

hindessm:
I wondered about this too. If we did that, the caller would be responsible for creating the correct decoder implementation? And, If so, we'd need an `IsFlexibleVersion() bool` function on the protocol messages to enable that decision to be made. (Not that this is necessarily problem but I'm just trying to consider the implications.)

In the encoder case, we'd need four encoders to handle the prep and real cases for flexible and non flexible. And again the caller would be responsible for picking the correct ones.

puellanivis:
We could just as easily combine the process into a `func getTheCorrectKindOfEncoder(isFlexible bool) encoder` sort of function as well?

And indeed. We trade the if-complexity for space-complexity. This is what the brainstorming draft is about, right? Figure out where we might want to put this.

I think we could also however clean up a lot of the encoders and decoders, so long as we never need to use _both_ the fixed, and compact encodings at the same time?

hindessm:
> This is what the brainstorming draft is about, right? Figure out where we might want to put this.

Absolutely. (And I really appreciate your feedback!) There is inherent complexity in this area; we just need to figure out how to hide it so it isn't a distraction most of the time. The initial refactoring removed 330+ lines and made the message code more readable which in itself validates that trying to do something to hide the complexity is worth doing.

I'm distracted by something else right now - CreateTopics V5 needs tagged fields so I'm prototyping handling them - but I'll try the distinct flexible/non-flexible encoder/decoder approach out to see how it is to use (possibly on another branch like this one).

hindessm:
I've created a [branch](https://github.com/hindessm/sarama/tree/flexible-experiment-split-implementations) with a very rough first attempt at splitting the flexible and non-flexible decoding. (The encoding will be trickier as we need to split the prep and real encoders.)
I had hoped it would be nicer but so far I not convinced. I might still take a look at the encoding side and consider if there is a way to make this approach cleaner.

hindessm:
I've updated my [branch](https://github.com/hindessm/sarama/tree/flexible-experiment-split-implementations) with the encoding side.
There are a few bits that would definitely need some tidying up.
The split up encoder and decoder variants do make the code a bit nicer but the casting to find the correct flexible wrapper and figuring out when it is needed are pretty horrible.
I'm not sure which I prefer so I'd appreciate comments on either approach.

hindessm:
@puellanivis I've done a bit of clean up of my other [branch](https://github.com/hindessm/sarama/tree/flexible-experiment-split-implementations) that took the approach you suggested (assuming I understood you correctly). I'm actually now pretty happy with that approach.
I've added a couple of message updates to the branch and the work needed for those updates seems pretty reasonable.
I am tempted to force push the first two commits from that branch to this PR and take it out of draft. But I'd love to know what you think.

dnwe:
Yes between these two I definitely think I prefer the other [branch](https://github.com/hindessm/sarama/tree/flexible-experiment-split-implementations) – it has turned out quite well

dnwe:
Closing as superseded by (merged) https://github.com/IBM/sarama/pull/3317

puellanivis:
After looking (late) through many of the new code, it looks good whatever decision you guys made. 😂

hindessm:
@puellanivis Assuming I understood your comment correctly, then we ended up adopting the approach you proposed [here](https://github.com/IBM/sarama/pull/3303#issuecomment-3327996238). I've added quite a few more flexible message versions since and haven't regretted the approach we've taken as it makes the code much easier to read and the updates required are as small as I'd hoped. Thanks very much for your input on this topic.

PR #3308, opened 2025-09-25
chore: support V5 ListOffsets

dnwe:
V5 adds a new error code, OFFSET_NOT_AVAILABLE, but is otherwise identical to V4

### DIFF
--- a/offset_request.go
+++ b/offset_request.go
@@ -52,6 +52,28 @@ type OffsetRequest struct {
 	blocks         map[string]map[int32]*offsetRequestBlock
 }
 
+func NewOffsetRequest(version KafkaVersion) *OffsetRequest {
+	request := &OffsetRequest{}
+	if version.IsAtLeast(V2_2_0_0) {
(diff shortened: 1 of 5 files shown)


offset_request.go
```diff
@@ -52,6 +52,28 @@ type OffsetRequest struct {
 	blocks         map[string]map[int32]*offsetRequestBlock
 }
 
+func NewOffsetRequest(version KafkaVersion) *OffsetRequest {
```
puellanivis, line 55: Did we want to export this? (I suspect yes, because we export a lot of request constructors).

dnwe: Yeah at the moment we export all the protocol, that may change when we move to a v2 api, but for now we keep them the same

puellanivis: Well, I think mentioned elsewhere, for a v2 we probably want to split up the protocol from the API. That’s what I also did with `sftp` v2.

Building a transparent shim of the protocol, and then the complex API hiding implementation details really simplified a lot of the exposed API, while keeping the protocol open and standardized as well.

PR #3311, opened 2025-09-26
chore(ci): adopt gotestsum and re-run flakes

dnwe:
Switch to using gotestsum as a runner in the Makefile and move tparse into the .github/workflows as it is GitHub Actions specific and so that we can run it with `always()` rather than only when tests pass.

Additionally add missing coverage dump for unittest runs (we already had it for FV runs) and also output teststat information in Makefile which highlights the slowest tests and any re-run flakes.


hindessm (approved):
Nice.

PR #3312, opened 2025-09-26
fix: non-flexible ElectLeadersRequest V0/V1 encode/decode

hindessm:
And clean up the misleading layout/comment of the []bytes in the test:

Test layout for V2 `[]byte` was:
```
	0, 0, // empty tagged fields
	0, 39, 16, 0, // timeout 10000
```
but, while the bytes were correct the comments were misleading as the encoding is really:
```
	0,            // empty tagged fields
	0, 0, 39, 16, // timeout 10000
	0, // empty tagged fields
```


dnwe:
Aside: we probably should have a functional test for elect leaders as I feel like this has likely never properly worked?

PR #3317, opened 2025-09-30
refactor: Flexible encoding/decoding refactoring

hindessm:
Modify the flexible/compact handling to make the protocol code simpler.

Many thanks to @puellanivis for the feedback on #3303 which lead to this approach being used.


dnwe (approved):
Thanks! This will certainly make it much simpler and cleaner to extend our protocol support to the latest versions

hindessm:
This change will have been worth the effort to me just to avoid having to decide in which manner to fix the inconsistent `putNullableCompactString` and `getCompactNullableString` method names. :-)

PR #3324, opened 2025-10-02
fix(fvt): simplify retry using testify's EventuallyWithT

dnwe:
Experiment with using testify to make this assertion block cleaner and more reliable. Also allow up to 10 seconds wait time.

### DIFF
--- a/functional_admin_test.go
+++ b/functional_admin_test.go
@@ -13,12 +13,13 @@ import (
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestFuncAdminQuotas(t *testing.T) {
 	const (
-		maxRetries = 10
-		retryDelay = 100 * time.Millisecond
+		waitFor = 10 * time.Second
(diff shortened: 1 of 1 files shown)


functional_admin_test.go
```diff
@@ -60,29 +61,20 @@ func TestFuncAdminQuotas(t *testing.T) {
...
-		t.Fatalf("Expected one quota entry for default user, found: %v", quotas)
-	}
+	require.EventuallyWithT(t, func(t *assert.CollectT) {
+		quotas, err = adminClient.DescribeClientQuotas(
+			[]QuotaFilterComponent{defaultUserFilter},
+			false,
+		)
+		require.NoError(t, err)
+		require.NotEmpty(t, quotas, "Expected not empty quotas for default user")
+		require.Len(t, quotas, 1, "Expected one quota entry for default user")
```
puellanivis, line 76: 😬 if these are beign called from within a goroutine, that technically violates a precondition of `testing.T.FailNow`:
> FailNow must be called from the goroutine running the test or benchmark function, not from other goroutines created during the test. Calling FailNow does not stop those other goroutines.
https://pkg.go.dev/testing@go1.25.1#T.FailNow

PR #3327, opened 2025-10-02
fix: decoder issues

hindessm:
Couple of issues spotted while debugging protocol message updates.

### DIFF
--- a/real_decoder.go
+++ b/real_decoder.go
@@ -470,7 +470,7 @@ func (rd *realFlexibleDecoder) getString() (string, error) {
 	}
 
 	if length < 0 {
-		return "", errInvalidByteSliceLength
+		return "", errInvalidStringLength
 	}
 	tmpStr := string(rd.raw[rd.off : rd.off+length])
 	rd.off += length
--- a/real_decoder.go
+++ b/real_decoder.go
@@ -444,13 +444,31 @@ func (rd *realFlexibleDecoder) getBytes() ([]byte, error) {
 	return rd.getRawBytes(length)
 }
 
+func (rd *realFlexibleDecoder) getStringLength() (int, error) {
+	length, err := rd.getUVarint()
+	if err != nil {
+		return 0, err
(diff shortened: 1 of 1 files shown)


dnwe (approved):

real_decoder.go
```diff
@@ -444,29 +444,45 @@ func (rd *realFlexibleDecoder) getBytes() ([]byte, error) {
...
+	length, err := rd.getUVarint()
+	if err != nil {
+		return 0, err
+	}
+
+	n := int(length - 1)
+
+	switch {
+	case n < -1:
+		return 0, errInvalidStringLength
```
puellanivis, line 457: 🤔 Duplicated testing, though here we’re only testing `-1`, below we test for the stronger `< 0`.

hindessm: I don't recall any more but I probably wanted to keep it consistent with the non-flexible version. There are a few issues like this already that I'd like to tackle. For instance, when we get array lengths there are checks like:
```
	if tmp > rd.remaining() {
		rd.off = len(rd.raw)
		return -1, ErrInsufficientData
	} else if tmp > int(MaxResponseSize) {
		return -1, errInvalidArrayLength
	}
```
The second check is redundant since we know the total buffer size is less than MaxResponseSize as it was check after we read the length in the header the data arrived in so the remaining check covers it already.
The remaining check would be better made by the caller since we may need to check for, for example, `4*tmp` (for int32 array) or `8*tmp` (for int64 array). Checking `1*tmp`only makes sense in the compact string case and even then it is potentially a significant underestimate.

puellanivis: Hm… those checks ought to probably be inverted anyways. Returning an invalid array length has value above simply returning insufficient data. 🤔 

And yeah, checking the remaining for an array… 🤔 maybe we could pass in a byte size, allowing us to properly check for insufficient data. But as you say… maybe better to just do the check downstream.

PR #3328, opened 2025-10-02
fix: improve KIP-511 behaviour on older Kafka clusters

dnwe:
Rather than failing with "not enough bytes to read" when parsing the ApiVersionsResponse (and always sending a V0 as fallback), instead follow the KIP and expect a V0 ApiVersionsResponse to have been sent back with the supported ApiVersions apikey version included in it.


hindessm (approved):
LGTM

Issue #3310, opened 2025-09-25
Recent flakey FV tests


dnwe:
For a while we had a good run where the FV was stable and rarely failed, but recently there seems to have been an increase in flakiness with a number of tests randomly timing out or failing assertions. Need to investigate what can be done to fixup wherever the holes are

TestFuncTxnProduceWithBrokerFailure occasionally times out (e.g., [here](https://github.com/IBM/sarama/actions/runs/18004740295/job/51222208971))

TestConsumerFetchFollowerFailover occasionally times out (e.g., [here](https://github.com/IBM/sarama/actions/runs/17883905146/job/50855066523))

TestFuncConsumerGroupPartitioningStateful fails assertion "functional_consumer_group_test.go:93: ttl exceeded, expected handlers 5 but got 0" (e.g., [here](https://github.com/IBM/sarama/actions/runs/17824302806/job/50673742549))

PR #3329, opened 2025-10-02
chore(fvt): improve testFuncConsumerGroupMember

dnwe:
- ensure ConsumeClaim returns when it should
- use require.EventuallyWithT
- add a bunch of t.Logf so it's clearer what state each consumer reaches

Contributes-to: #3310


hindessm (approved):
LGTM

PR #3337, opened 2025-10-03
fix: correct ApiVersionsResponse handling of ErrUnsupportedVersion

dnwe:
Improve the api_versions_response_test.go to be more representative of the different good and bad responses it might receive.

Subsequently fix the ApiVersionsResponse decoding so that it correctly downgrades the decoder from flexible to non-flexible after reading the ErrorCode of UnsupportedVersion

### DIFF
--- a/api_versions_response.go
+++ b/api_versions_response.go
@@ -1,6 +1,8 @@
 package sarama
 
-import "time"
+import (
+	"time"
+)
 
 // ApiVersionsResponseKey contains the APIs supported by the broker.
 type ApiVersionsResponseKey struct {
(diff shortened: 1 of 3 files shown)


dnwe:
@hindessm any thoughts on how we might handle this FIXME? See that and the testcase with t.Skip(...)

dnwe:
> @hindessm any thoughts on how we might handle this FIXME? See that and the testcase with t.Skip(...)

(from @hindessm):
```go
func downgradeFlexibleDecoder(pd packetDecoder) packetDecoder {
  if f, ok := pd.(realFlexibleDecoder); ok {
    return f.realDecoder
  }
  return pd
}
```

hindessm (approved):
LGTM

PR #3339, opened 2025-10-04
feat: support JoinGroup V6 protocol

hindessm:
Version 6 is the first flexible version

The other commits are a fix for a bug that broken the FVTs for this protocol bump and a very minor refactor that I spotted while debugging.

### DIFF
--- a/real_decoder.go
+++ b/real_decoder.go
@@ -479,7 +479,7 @@ func (rd *realFlexibleDecoder) getStringLength() (int, error) {
 
 func (rd *realFlexibleDecoder) getString() (string, error) {
 	length, err := rd.getStringLength()
-	if err != nil {
+	if err != nil || length == -1 {
 		return "", err
 	}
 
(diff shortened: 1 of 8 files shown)


real_decoder.go
```diff
@@ -479,7 +479,7 @@ func (rd *realFlexibleDecoder) getStringLength() (int, error) {
 
 func (rd *realFlexibleDecoder) getString() (string, error) {
 	length, err := rd.getStringLength()
-	if err != nil {
+	if err != nil || length == -1 {
```
dnwe, line 482: The commit msg is fix: string length -1 should be allowed’ which is a little confusing when the behaviour is ‘exit early if length is -1’

Also, whilst we do end up with effectively ‘return "", nil’ it’s a little weird to conjoin it with the using ‘if err != nil’ pattern. Shouldn’t we just have a separate if for this case?

hindessm: Allowed as in "return here with no error rather than return with `errInvalidStringLength` further down" doesn't seem confusing to me?

While I don't necessarily disagree, I was matching the existing code in the non-flexible decoder.

dnwe (approved):

PR #3384, opened 2025-11-20
fix: close broken tcp connections

Asphaltt:
Prevent TCP connection leaks that error occurred to kernel space's socket. Previously, when errors occurred on TCP sockets, connections would remain in TCP_CLOSE state indefinitely without being properly freed.

The leaked connections caused a cascade of kernel-level issues:
- TCP sockets remain in TCP_CLOSE state indefinitely
- Their sk_receive_queues retain FINACK skb packets
- These skbs hold pages allocated from page_pool
- page_pool_release_retry() stalls because pages cannot be freed while still referenced by the TCP stack

The fix ensures that TCP connections are properly closed when socket errors occur, preventing resource leaks at both the application and kernel levels.

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -157,6 +157,27 @@ func NewBroker(addr string) *Broker {
 	return &Broker{id: -1, addr: addr}
 }
 
+func (b *Broker) getSockError() error {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	if b.conn == nil {
+		return nil
+	}
+
+	conn := b.conn
+	if c, ok := conn.(*bufConn); ok {
+		conn = c.Conn
+	}
+	if c, ok := conn.(*tls.Conn); ok {
+		conn = c.NetConn()
+	}
+	if c, ok := conn.(*net.TCPConn); ok {
+		return getTCPConnSockError(c)
+	}
+	return nil
+}
+
 // Open tries to connect to the Broker if it is not already connected or connecting, but does not block
 // waiting for the connection to complete. This means that any subsequent operations on the broker will
 // block waiting for the connection to succeed or fail. To get the effect of a fully synchronous Open call,
--- a/broker.go
+++ b/broker.go
@@ -158,7 +158,10 @@ func NewBroker(addr string) *Broker {
 }
 
 func (b *Broker) getSockError() error {
-	b.lock.Lock()
+	// skip socket health checks while another operation owns broker state
+	if !b.lock.TryLock() {
+		return nil
+	}
 	defer b.lock.Unlock()
 
 	if b.conn == nil {
@@ -308,18 +311,12 @@ func (b *Broker) Open(conf *Config) error {
 		if conf.Net.SASL.Enable && !useSaslV0 {
 			b.connErr = b.authenticateViaSASLv1()
 			if b.connErr != nil {
-				close(b.responses)
-				<-b.done
-				b.responses = nil
-				b.done = nil
-				err = b.conn.Close()
+				err = b.closeLocked()
 				if err == nil {
 					DebugLogger.Printf("Closed connection to broker %s due to SASL v1 auth error: %s\n", b.addr, b.connErr)
 				} else {
 					Logger.Printf("Error while closing connection to broker %s (due to SASL v1 auth error: %s): %s\n", b.addr, b.connErr, err)
 				}
-				b.conn = nil
-				b.opened.Store(false)
 				return
 			}
 		}
@@ -398,11 +395,11 @@ func (b *Broker) closeLocked() error {
 	if b.responses != nil {
 		close(b.responses)
 	}
+	// close the socket before waiting so in-flight reads can exit
+	err := b.conn.Close()
 	if b.done != nil {
 		<-b.done
 	}
-
-	err := b.conn.Close()
 
 	b.conn = nil
 	b.responses = nil
--- a/client.go
+++ b/client.go
@@ -684,6 +684,38 @@ func (client *client) randomizeSeedBrokers(addrs []string) {
 	}
 }
 
+func (client *client) checkSeedBrokersHealth(brokers []*Broker) []*Broker {
+	if len(brokers) == 0 {
+		return nil
+	}
+
+	healthyBrokers := make([]*Broker, 0, len(brokers))
+	for _, broker := range brokers {
+		if err := broker.getSockError(); err != nil {
+			Logger.Printf("client/seedbrokers close seed broker #%d at %s due to socket error: %v", broker.ID(), broker.Addr(), err)
+			safeAsyncClose(broker)
+			continue
+		}
+
+		healthyBrokers = append(healthyBrokers, broker)
+	}
+
+	return healthyBrokers
+}
+
+func (client *client) checkBrokersHealth() {
+	for id, broker := range client.brokers {
+		if err := broker.getSockError(); err != nil {
+			Logger.Printf("client/brokers close broker #%d at %s due to socket error: %v", broker.ID(), broker.Addr(), err)
+			safeAsyncClose(broker)
+			delete(client.brokers, id)
+		}
+	}
+
+	client.seedBrokers = client.checkSeedBrokersHealth(client.seedBrokers)
+	client.deadSeeds = client.checkSeedBrokersHealth(client.deadSeeds)
+}
+
 func (client *client) updateBroker(brokers []*Broker) {
 	if client.brokers == nil {
 		return
@@ -1052,6 +1084,14 @@ func (client *client) updateMetadata(data *MetadataResponse, allKnownMetaData bo
 
 	client.lock.Lock()
 	defer client.lock.Unlock()
+
+	// Check health of existing brokers, including seed brokers, dead
+	// seed brokers, and registered brokers.
+	// - if error occurred on broker's tcp socket, close the tcp
+	//   connection.
+	// - if it's seed broker or dead seed broker, remove it from
+	//   the list.
+	client.checkBrokersHealth()
 
 	// For all the brokers we received:
 	// - if it is a new ID, save it
--- a/client.go
+++ b/client.go
@@ -684,9 +684,9 @@ func (client *client) randomizeSeedBrokers(addrs []string) {
 	}
 }
 
-func (client *client) checkSeedBrokersHealth(brokers []*Broker) []*Broker {
+func (client *client) checkSeedBrokersHealth(brokers []*Broker) {
 	if len(brokers) == 0 {
-		return nil
+		return
 	}
 
 	for _, broker := range brokers {
@@ -695,8 +695,6 @@ func (client *client) checkSeedBrokersHealth(brokers []*Broker) []*Broker {
 			safeAsyncClose(broker)
 		}
 	}
-
-	return brokers
 }
 
 func (client *client) checkBrokersHealth() {
@@ -708,8 +706,8 @@ func (client *client) checkBrokersHealth() {
 		}
 	}
 
-	client.seedBrokers = client.checkSeedBrokersHealth(client.seedBrokers)
-	client.deadSeeds = client.checkSeedBrokersHealth(client.deadSeeds)
+	client.checkSeedBrokersHealth(client.seedBrokers)
+	client.checkSeedBrokersHealth(client.deadSeeds)
 }
 
 func (client *client) updateBroker(brokers []*Broker) {
--- a/client.go
+++ b/client.go
@@ -689,18 +689,14 @@ func (client *client) checkSeedBrokersHealth(brokers []*Broker) []*Broker {
 		return nil
 	}
 
-	healthyBrokers := make([]*Broker, 0, len(brokers))
 	for _, broker := range brokers {
 		if err := broker.getSockError(); err != nil {
 			Logger.Printf("client/seedbrokers close seed broker #%d at %s due to socket error: %v", broker.ID(), broker.Addr(), err)
 			safeAsyncClose(broker)
-			continue
 		}
-
-		healthyBrokers = append(healthyBrokers, broker)
 	}
 
-	return healthyBrokers
+	return brokers
 }
 
 func (client *client) checkBrokersHealth() {
--- a/broker_test.go
+++ b/broker_test.go
@@ -197,6 +197,52 @@ func TestBrokerFailedRequest(t *testing.T) {
 	}
 }
 
+func TestBrokerClose(t *testing.T) {
+	t.Run("interrupts active async response reads", func(t *testing.T) {
+		mockBroker := NewMockBroker(t, 0)
+		defer mockBroker.Close()
+
+		broker := NewBroker(mockBroker.Addr())
+		conf := NewTestConfig()
+		conf.ApiVersionsRequest = false
+		conf.Net.ReadTimeout = 30 * time.Second
+
+		require.NoError(t, broker.Open(conf))
+
+		request := ProduceRequest{}
+		request.RequiredAcks = WaitForLocal
+
+		responseErrs := make(chan error, 1)
+		err := broker.AsyncProduce(&request, func(_ *ProduceResponse, err error) {
+			responseErrs <- err
+		})
+		require.NoError(t, err)
+
+		closeErrs := make(chan error, 1)
+		go func() {
+			closeErrs <- broker.Close()
+		}()
+
+		select {
+		case err := <-closeErrs:
+			require.NoError(t, err)
+		case <-time.After(250 * time.Millisecond):
+			require.FailNow(t, "Close blocked with an active async response read")
+		}
+
+		select {
+		case err := <-responseErrs:
+			require.Error(t, err)
+		case <-time.After(time.Second):
+			require.FailNow(t, "timed out waiting for the async produce callback")
+		}
+
+		connected, err := broker.Connected()
+		require.NoError(t, err)
+		require.False(t, connected)
+	})
+}
+
 // closeImmediatelyDialer is a test dialer that returns a net.Conn whose peer is
 // already closed. This reliably triggers a transport-level failure (e.g. EOF)
 // during ApiVersions negotiation in Broker.Open.
--- a/client_test.go
+++ b/client_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"sync"
 	"sync/atomic"
 	"syscall"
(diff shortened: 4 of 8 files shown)


broker.go
```diff
@@ -156,6 +158,39 @@ func NewBroker(addr string) *Broker {
...
 }
 
+func (b *Broker) getSockOpt(conn *net.TCPConn, opt int) (int, error) {
+	file, err := conn.File()
+	if err != nil {
+		return 0, fmt.Errorf("failed to get file from connection: %w", err)
+	}
+	defer file.Close() // Close the duplicated file descriptor
+
+	sockErr, err := syscall.GetsockoptInt(int(file.Fd()), syscall.SOL_SOCKET, opt)
```
puellanivis:
```
$ GOOS=windows go test
# github.com/IBM/sarama [github.com/IBM/sarama.test]
./broker.go:168:40: cannot use int(file.Fd()) (value of type int) as syscall.Handle value in argument to syscall.GetsockoptInt
./broker.go:184:50: undefined: syscall.SO_ERROR
FAIL    github.com/IBM/sarama [build failed]
```

broker.go
```diff
@@ -192,6 +227,7 @@ func (b *Broker) Open(conf *Config) error {
 			b.opened.Store(false)
 			return
 		}
+		b.connTCP = b.conn.(*net.TCPConn) // for getSockError
```
puellanivis: Why can’t we `b.conn.(*net.TCPConn)` where necessary, rather than alias `b.conn` only as a concrete type, instead of an `interface`?

broker.go
```diff
@@ -156,6 +158,39 @@ func NewBroker(addr string) *Broker {
 	return &Broker{id: -1, addr: addr}
 }
 
+func (b *Broker) getSockOpt(conn *net.TCPConn, opt int) (int, error) {
+	file, err := conn.File()
+	if err != nil {
+		return 0, fmt.Errorf("failed to get file from connection: %w", err)
+	}
+	defer file.Close() // Close the duplicated file descriptor
```
puellanivis: Maybe `conn.SyscallConn`, so we can then use `sysconn.Control(f func(fd uintptr))`?

Although, regardless, we would need to optionally cast to `syscall.Handle` for Windows. Though, `GOOS=windows` also doesn’t seem to have `SO_ERROR` even though microsoft’s references say it actually does?

broker.go
Asphaltt: No.

If tls is enabled, `b.conn` will be a `*tls.Conn`.

```go
		if conf.Net.TLS.Enable {
			b.conn = tls.Client(b.conn, validServerNameTLS(b.addr, conf.Net.TLS.Config))
		}
```

broker.go
Asphaltt: After checking both the `syscall` package and `golang.org/x/sys`, it appears that `SO_ERROR` is not defined on Windows. As a result, retrieving the socket error is unnecessary for Windows.

However, the current code triggers a syntax error on Windows because `syscall.SO_ERROR` does not exist there. To address this, I would like to factor out the `syscall.GetsockoptInt()` call on Windows, so that the build and IDE/editor no longer complain.

broker.go
puellanivis: Right, and a `if conn, ok := b.conn.(*net.TCPConn); ok {` will take care of the case when `b.conn` is not a `*net.TCPConn`.

PS: in fact, testing for validity will test for successful type conversion anyways, so `if conn, _ := b.conn.(*net.TCPConn); conn != nil {` is equally sufficient.

dnwe: Yes we need to do a type assertion check because also the user can put whatever Dialer they want via `Config.Net.Proxy.Dialer`  Please guard the type assertion here, or make the type on the struct generic and only enable the SO_ERROR probing later when the returned connection really is a *net.TCPConn

client.go
```diff
@@ -684,6 +684,38 @@ func (client *client) randomizeSeedBrokers(addrs []string) {
...
+			Logger.Printf("client/brokers skipped broker #%d at %s due to socket error: %v", broker.ID(), broker.Addr(), err)
+			safeAsyncClose(broker)
+			delete(client.brokers, id)
+		}
+	}
+
+	client.seedBrokers = client.checkSeedBrokersHealth(client.seedBrokers)
+	client.deadSeeds = client.checkSeedBrokersHealth(client.deadSeeds)
+}
+
```
dnwe: If we delete from seedBrokers (bootstraps) what re-populates them?

I don't really understand why this new health check is bolted on as a separate pass (client.checkBrokersHealth + checkSeedBrokersHealth) instead of piggy‑backing on existing metadata flows - was there a strong motivation for doing so?

During LeastLoadedBroker/Leader/RefreshMetadata, when a broker returns an error, we already close and deregister it so we could put a check for socket errors there I think? Similarly In updateBroker, while reconciling broker lists, we could probe each existing broker and skip any with SO_ERROR.

internal/sockopt/sockopt_other.go
```diff
@@ -0,0 +1,11 @@
...
+
+package sockopt
+
+import "syscall"
+
+const SockoptError = syscall.SO_ERROR
+
+func GetSockoptInt(fd uintptr, opt int) (int, error) {
+	return syscall.GetsockoptInt(int(fd), syscall.SOL_SOCKET, opt)
+}
```
dnwe: "other" is a bit vague, if we're going to have this then lets match Go's convention of calling it sockopt_unix.go or sockopt_posix.go

I also wonder if we should just have these in the single top-level sarama package but use unexported names for the funcs rather than adding an ./internal/ package?

dnwe (commented):
Thanks for reporting and investigating this issue, as well as raising a PR!

A few thoughts:

broker.go
Asphaltt: >  because also the user can put whatever Dialer they want via `Config.Net.Proxy.Dialer`

Make sense. 

I'll update it to `b.connTCP, _ = b.conn.(*net.TCPConn)` as it won't get socket error if `b.connTCP == nil`.

client.go
Asphaltt: Thanks for your explanation.

> If we delete from seedBrokers (bootstraps) what re-populates them?

In that case, should the seedBrokers be re-opened instead of being removed when we detect `SO_ERROR`?

> why this new health check is bolted on as a separate pass

I'm not entirely sure where the most appropriate place is to perform this health check, as I'm not deeply familiar with all parts of this library. I initially attempted to implement it inside `updateBrokers()`, but the issue still occurred.

With this PR, broken TCP connections would at least be detected and closed every 10 minutes.

> During LeastLoadedBroker/Leader/RefreshMetadata, when a broker returns an error, we already close and deregister it so we could put a check for socket errors there I think?

I'm not certain that covers all cases. For brokers that rarely get used and don't fetch data for a long time, their TCP connections may remain broken for an extended period without triggering any of those paths.

internal/sockopt/sockopt_other.go
puellanivis: The weird thing is that Windows does implement this part of the POSIX standard, Golang just doesn’t have the `SO_ERROR` definition. So `_posix.go` would be misleading.

Convention is to simply add `no` to the build tag suffix: `sockopt_nowindows.go`

internal/sockopt/sockopt_windows.go
```diff
@@ -0,0 +1,9 @@
+//go:build windows
```
puellanivis: `plan9` also doesn’t have `SO_ERROR`, or `GetSockoptInt`. (Though, in this case, I don’t think we support it, as it already doesn’t build because `go-metrics` tries to use a non-existent `syslog.Writer`.)

Also, neither `js/wasm` nor `wasip1/wasm` work. These ones at least seem to compile prior to this change.

You may want:
```suggestion
//go:build windows || wasm
```

Though, this will require renaming the file, as the `_windows.go` suffix means this build tag for `windows` is currently redundant, and unnecessary.

broker.go
```diff
@@ -156,6 +159,39 @@ func NewBroker(addr string) *Broker {
...
+
+	sockErr, err := b.getSockOpt(b.connTCP, sockopt.SockoptError)
+	if err != nil {
+		return fmt.Errorf("failed to get socket error: %w", err)
+	}
+	if sockErr != 0 {
+		return syscall.Errno(sockErr)
+	}
+	return nil
+}
```
puellanivis, line 182: As mentioned, using `SyscallConn` and `RawConn.Control` remove the need to `dup` the underlying file descriptor: 
```golang
func (b *Broker) getSockError() error {
	b.lock.Lock()
	defer b.lock.Unlock()

	if conn, _ := b.conn.(*net.TCPConn); conn != nil {
		rawConn, err := conn.SyscallConn()
		if err != nil {
			return err
		}

		var sockErr int
		var opErr error

		err = rawConn.Control(func(fd uintptr) {
				sockErr, opErr = sockopt.GetSockoptInt(fd, sockopt.SockoptError)
		})
		if err != nil {
			return fmt.Errorf("failed to get socket error: %w", err)
		}

		if opErr != nil {
			return fmt.Errorf("failed to get socket error: %w", err)
		}

		if sockErr != 0 {
			return syscall.Errno(sockErr)
		}
	}

	return nil
}
```

Asphaltt: LGTM.

I'll update it like this.

Thanks!

internal/sockopt/sockopt_windows.go
Asphaltt: It would be better to rely on `//go:build unix`, since the issue has only been observed on Linux so far. From there, we can exclude the OSes that do not define `SO_ERROR`:

```go
//go:build unix && !android && !illumos && !ios && !hurd
```

With this condition, the socket error check will apply to the following Unix-like systems:

`aix`, `darwin`, `dragonfly`, `freebsd`, `linux`, `netbsd`, `openbsd`, and `solaris`.

---

`//go:build unix` is short for the following OSes:

```go
// https://github.com/golang/go/blob/e88be8a128d167c5ab91eabc998ab364370c702e/src/cmd/dist/build.go#L1063C4-L1063C10

// unixOS is the set of GOOS values matched by the "unix" build tag.
// This is the same list as in internal/syslist/syslist.go.
var unixOS = map[string]bool{
	"aix":       true,
	"android":   true,
	"darwin":    true,
	"dragonfly": true,
	"freebsd":   true,
	"hurd":      true,
	"illumos":   true,
	"ios":       true,
	"linux":     true,
	"netbsd":    true,
	"openbsd":   true,
	"solaris":   true,
}
```

broker.go
```diff
@@ -156,6 +157,17 @@ func NewBroker(addr string) *Broker {
...
 
+func (b *Broker) getSockError() error {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	if b.connTCP == nil {
+		return nil
+	}
+
+	return getTCPConnSockError(b.connTCP)
```
puellanivis: I continue to have concerns about using a different variable to store the same value that would already be available via `b.conn.(*net.TCPConn)`.

There is no need to coördinate handling both the `conn` and `connTCP` variables to ensure that they remain consistent with each other. In context of the [Swiss-Cheese Model](https://en.wikipedia.org/wiki/Swiss_cheese_model) of accident causation, forcing developers to manually ensure the consistency of two variables when it can be driven by just the one variable would open up a significant hole in one of the layers of swiss cheese, making accidents during refactoring more likely.

```suggestion
	connTCP, _ := b.conn.(*net.TCPConn)
	if connTCP == nil {
		return nil
	}

	return getTCPConnSockError(connTCP)
```

Asphaltt: Thanks for your explanation.

I understand your concerns now.

broker.go
```diff
@@ -156,6 +156,24 @@ func NewBroker(addr string) *Broker {
...
+		return nil
+	}
+
+	conn := b.conn
+	if c, ok := conn.(*tls.Conn); ok {
+		conn = c.NetConn()
+	}
+	if c, ok := conn.(*net.TCPConn); ok {
+		return getTCPConnSockError(c)
+	}
```
puellanivis, line 180: [improvement] We probably want to test to ensure that the values we’re getting out of the type assertion are not typed nil pointers.

This also means that testing for validity (non-nil) removes the need to explicitly test the success of the type assertion itself; as the result of a failed `, ok` type assertion is the zero value, which is here the typed nil pointer:
```suggestion
	conn := b.conn
	if c, _ := conn.(*tls.Conn); c != nil {
		conn = c.NetConn()
	}
	if c, _ := conn.(*net.TCPConn); c != nil {
		return getTCPConnSockError(c)
	}
```

(Note, we can remove the test for `b.conn == nil` because if it be true, then all type assertions will fail, because as an interface, `v == nil` means that it is the untyped nil, and thus cannot be assigned to any type. Example: where a typed `nil` will not compare equal to `nil` if in an interface: [`any((*net.TCPConn)(nil)) != nil`](https://go.dev/play/p/wd_hEIalvUy)

puellanivis (commented):
Looks good just one small possible improvement. (The code, as is, is fine though.)

broker.go
Asphaltt: Thanks for the suggestion! I appreciate the insight about typed nil pointers.

However, I prefer to keep the current implementation for the following reasons:

1. **Explicit intent**: The `if b.conn == nil` check at the beginning makes it immediately clear that we're guarding against nil connections. While it's true that type assertions will fail on untyped nil, this explicit check documents the precondition and makes the code easier to understand at first glance.

2. **Readability**: Using `ok` in type assertions is a common Go idiom that signals "we're checking if this conversion is safe." Testing `c != nil` after discarding `ok` is less conventional and may confuse readers who expect to see the standard `if x, ok := ...` pattern.

3. **Future-proofing**: If `NetConn()` or the connection wrapping logic changes in the future, the explicit checks make it easier to understand the control flow and add debugging.

While your suggestion is technically correct and slightly more concise, I believe the current version optimizes for clarity and maintainability, which is especially important for connection management code where bugs can be subtle and hard to debug.

What do you think? If there's a strong preference in the project for the more compact style, I'm happy to adjust.

puellanivis: Yes, I am already well aware of the `ok` pattern, and the nuances behind it, which is specifically why I have been bringing up this point.

Testing that the type-assertion succeeded is not always sufficient—as specifically in this case, a type-nil pointer being in the interface will result in a panic on the receiver method calls. So, we should also be testing for validity of the typed pointer `c`: `c != nil`. So, at the very minimum, we should have `ok && c != nil`, which is a test for success and validity.

I don’t see how my code would be any less future proof than yours. First, the go1 compatibility guarantee is in effect for `crypto/tls`, so the `NetConn` behavior shall not change radically enough to force either into breaking.

Asphaltt: I don't believe it's our responsibility to check whether `b.conn` is a typed-nil pointer here.

If `b.conn` were a typed-nil pointer—which could only originate from `Config.Net.Proxy.Dialer`—calling any method on it would panic when attempting to serve Kafka traffic. This is clearly not expected behavior and would indicate a bug in the user's custom `Dialer` implementation, not in sarama itself.

In other words, `b.conn == nil` (untyped nil) means the connection hasn't been established yet, which is a valid and expected state that we should handle.

broker.go
```diff
@@ -156,6 +156,24 @@ func NewBroker(addr string) *Broker {
...
+	return nil
```
dnwe, line 181: Have you tried this to see if it works in practice?

Looking at the code b.conn is stored as `*bufConn` so I feel like these are both just going to always get `!ok` and always return nil, so none of the other code will have run

elsewhere we unwrap from bufConn and grab the underlying Conn like this:
```go
	conn := b.conn
	if bconn, ok := b.conn.(*bufConn); ok {
		conn = bconn.Conn
	}
	if tc, ok := conn.(*tls.Conn); ok {
		// etc.
	}
```

Asphaltt: Thank you for pointing that out.

After adding `c, ok := b.conn.(*bufConn); ok`, it now works as expected:

```log
[Sarama] 2025/12/01 11:59:32 client/seedbrokers close seed broker #-1 at xxx.io:9093 due to socket error: broken pipe
[Sarama] 2025/12/01 12:09:32 client/brokers close broker #1081 at 10.xx.xx.xx:9093 due to socket error: broken pipe
```

---

However, the current approach has become quite complicated:

```go
	conn := b.conn
	if c, ok := conn.(*bufConn); ok {
		conn = c.Conn
	}
	if c, ok := conn.(*tls.Conn); ok {
		conn = c.NetConn()
	}
	if c, ok := conn.(*net.TCPConn); ok {
		return getTCPConnSockError(c)
	}
	return nil
```

I ended up paying the complexity cost, and it feels a bit painful—especially since this approach wasn't originally mine.

With a small change, such as adding a comment to the dedicated field `b.connTCP`, e.g. `// used for getSockError`, the solution would be much cleaner:

```go
return getTCPConnSockError(b.connTCP)
```

puellanivis: I don’t know. I don’t think it’s particularly all that complicated?

It’s basically just three `c  = maybeDeref(c)` operations after all. It’s quite simple to track each operation individually, and in a simple series.

Asphaltt:
PING

---

Over time, `go.mod` and `go.sum` may conflict with the files on the main branch. When that happens, I will need to resolve those conflicts.

puellanivis:
I have no lingering concerns that need resolution.

Asphaltt:
Hi @dnwe, do you have any remaining concerns that need further discussion?

If not, could you please approve this PR?

dnwe:
@puellanivis can you take a look at the reworked version to see if you agree with my updates? They’re along the lines of what I wrote in https://github.com/IBM/sarama/pull/3384#discussion_r2547361998

client.go
```diff
@@ -684,6 +684,34 @@ func (client *client) randomizeSeedBrokers(addrs []string) {
...
+	}
+
+	for _, broker := range brokers {
+		if err := broker.getSockError(); err != nil {
+			Logger.Printf("client/seedbrokers close seed broker #%d at %s due to socket error: %v", broker.ID(), broker.Addr(), err)
+			safeAsyncClose(broker)
+		}
+	}
+
+	return brokers
```
puellanivis, line 699: I’m not sure what the value is in returning the same slice that was passed in as an argument? I mean, it doesn’t look like there’s any chance that the slice has been modified at all?

One usually only wants to return a value that was passed in as an argument if there’s a chance that it could have reallocated. (like `append` and `slices.Grow`), while otherwise, we just perform the operation on the argument, and don’t return anything at all `slices.Sort`)

puellanivis (commented):
Seems mostly fine? Just the one thing.

dnwe (approved):

Issue #3391, opened 2025-11-27
kafka producer panic on updateBroker


3AceShowHand:
##### Versions

| Sarama | Kafka | Go  |
| ------ | ----- | --- |
|  v1.41.2      |       |     |

##### Configuration

```go

```

##### Logs

```
panic: assignment to entry in nil map

goroutine 17419 [running]:
github.com/IBM/sarama.(*client).updateBroker(0x405b0fdc20, {0x4053092f40, 0x7, 0x2766c?})
	github.com/IBM/sarama@v1.41.2/client.go:729 +0x2c0
github.com/IBM/sarama.(*client).updateMetadata(0x405b0fdc20, 0x40738555e0, 0x0)
	github.com/IBM/sarama@v1.41.2/client.go:1098 +0xa4
github.com/IBM/sarama.(*client).tryRefreshMetadata(0x405b0fdc20, {0x405facc300, 0x1, 0x1}, 0xa, {0x48e11e0?, 0x405b0fdc01?, 0xbb70180?})
	github.com/IBM/sarama@v1.41.2/client.go:1040 +0x7fc
github.com/IBM/sarama.(*client).RefreshMetadata(0x405b0fdc20, {0x405facc300, 0x1, 0x1})
	github.com/IBM/sarama@v1.41.2/client.go:560 +0x110
github.com/IBM/sarama.(*client).LeaderAndEpoch(0x405b0fdc20, {0x406a928f90, 0x12}, 0x14)
	github.com/IBM/sarama@v1.41.2/client.go:503 +0x94
github.com/IBM/sarama.(*client).Leader(0x403ff4a6e8?, {0x406a928f90?, 0x403e1a9200?}, 0x4efe700?)
	github.com/IBM/sarama@v1.41.2/client.go:492 +0x20
github.com/IBM/sarama.(*partitionProducer).dispatch(0x4051978720)
	github.com/IBM/sarama@v1.41.2/async_producer.go:632 +0x4c
github.com/IBM/sarama.withRecover(0x403ff2ea00?)
	github.com/IBM/sarama@v1.41.2/utils.go:43 +0x40
created by github.com/IBM/sarama.(*asyncProducer).newPartitionProducer in goroutine 12219
	github.com/IBM/sarama@v1.41.2/async_producer.go:600 +0x1a8

```

<details><summary>logs: CLICK ME</summary>
<p>

```

```

</p>
</details>

dnwe:
Presumed fixed

PR #3393, opened 2025-11-27
fix(client): add nilguards to updateBroker

dnwe:
This _should_ be a rare edge case as the updateBroker func is only called from updateMetadata which does already do an up-front `client.Close()` check under read-lock before then acquiring the write lock. However, there's potentially a small window of opportunity that if client.Close() was called whilst metadata refresh was in-flight and for whatever reason the updateMetadata goroutine gets pre-empted in-between the readlock release and the write lock acquire then the client could have been closed and so `client.brokers` will be nil.

I wouldn't have expected this to ever happen, but it was reported by a user in an older Sarama version in #3391 and there's no harm in adding the nilguard just in case.

### DIFF
--- a/client.go
+++ b/client.go
@@ -685,9 +685,16 @@ func (client *client) randomizeSeedBrokers(addrs []string) {
 }
 
 func (client *client) updateBroker(brokers []*Broker) {
+	if client.brokers == nil {
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


hindessm (approved):
LGTM

PR #3399, opened 2025-12-02
chore: finish up the move to atomic types

puellanivis:
This completes the change in #3277 and rolls out all atomics usage to using the types, rather than the bare functions.

This already found a few usages that had inconsistently used atomic functions on some values.

### DIFF
--- a/functional_consumer_group_test.go
+++ b/functional_consumer_group_test.go
@@ -378,18 +378,19 @@ func (s *testFuncConsumerGroupSink) Close() map[string][]string {
 
 type testFuncConsumerGroupMember struct {
 	ConsumerGroup
-	clientID     string
-	claims       map[string]int
+	t        *testing.T
+	clientID string
+	isCapped bool
+	sink     *testFuncConsumerGroupSink
+
 	generationId atomic.Int32
 	state        atomic.Int32
 	handlers     atomic.Int32
-	errs         []error
 	maxMessages  atomic.Int32
-	isCapped     bool
-	sink         *testFuncConsumerGroupSink
 
-	t  *testing.T
-	mu sync.RWMutex
+	mu     sync.RWMutex
+	claims map[string]int
+	errs   []error
 }
 
 func defaultConfig(clientID string) *Config {


mocks/consumer.go
```diff
@@ -398,7 +398,8 @@ func (pc *PartitionConsumer) YieldMessage(msg *sarama.ConsumerMessage) *Partitio
 	msg.Partition = pc.partition
 
 	if pc.paused {
-		msg.Offset = atomic.AddInt64(&pc.suppressedHighWaterMarkOffset, 1) - 1
+		msg.Offset = pc.suppressedHighWaterMarkOffset
+		pc.suppressedHighWaterMarkOffset++
```
puellanivis, line 402: Rechecked this, and yes, the usage is indeed always made under lock.

functional_consumer_group_test.go
```diff
@@ -380,11 +380,11 @@ type testFuncConsumerGroupMember struct {
 	ConsumerGroup
 	clientID     string
 	claims       map[string]int
-	generationId int32
-	state        int32
-	handlers     int32
+	generationId atomic.Int32
+	state        atomic.Int32
+	handlers     atomic.Int32
 	errs         []error
```
puellanivis: :squint: I’m a bit suspicious of this… other values are atomics… taking a dive to check stuff out.

functional_consumer_group_test.go
```diff
@@ -378,18 +378,19 @@ func (s *testFuncConsumerGroupSink) Close() map[string][]string {
...
-	state        int32
-	handlers     int32
-	errs         []error
-	maxMessages  int32
-	isCapped     bool
-	sink         *testFuncConsumerGroupSink
+	t        *testing.T
+	clientID string
+	isCapped bool
+	sink     *testFuncConsumerGroupSink
```
puellanivis, line 384: These are set at initialisation and never changed after, and thus can be accessed freely without mutex.

functional_consumer_group_test.go
```diff
@@ -378,18 +378,19 @@ func (s *testFuncConsumerGroupSink) Close() map[string][]string {
...
-	t  *testing.T
-	mu sync.RWMutex
+	generationId atomic.Int32
+	state        atomic.Int32
+	handlers     atomic.Int32
+	maxMessages  atomic.Int32
+
+	mu     sync.RWMutex
+	claims map[string]int
+	errs   []error
```
puellanivis, line 393: These are the two fields that can update after initialization, and are already correctly protected by mutexes. Their movement to here is just to match convention of putting mutex protected values below the mutex that protects them.

dnwe (approved):
Thanks! LGTM

Issue #1162, opened 2018-09-03
ClusterAdmin should reconnect broker after disconnected


fatkun:
##### Versions

*Please specify real version numbers or git SHAs, not just "Latest" since that changes fairly regularly.*
Sarama Version: 647feef69a1a7e0e23cccceeb786176bc4fcf5ea
Kafka Version: 1.1.0
Go Version:1.10.3

##### Configuration
```
	conf := sarama.NewConfig()
	conf.Version = sarama.V1_1_0_0
	conf.ClientID = busRef.Name + "-provisioner"
	clusterAdmin, err := sarama.NewClusterAdmin(brokers, conf)
```

##### Logs

> write tcp 172.17.0.34:43046->10.99.115.9:9092: write: broken pipe

##### Problem Description

code: https://github.com/knative/eventing/blob/master/pkg/buses/kafka/bus.go

reporduce step
1. new clusterAdmin
2. sleep some time
3. create topic
4. the err raised forever after the first time

It seems just like this issue : https://github.com/Shopify/sarama/issues/866

we should close broker connection after error just like client does?

chandradeepak:
@fatkun , are you expecting the functionality where we close broker connection when an error happens as described in #866 to be there in cluster Admin also so that it can be used for further create topic ?

neosab:
> 4\. the err raised forever after the first time

For me, the error gets resolved during the next cluster metadata refresh but ideally I would expect this to happen when an error occurs while creating the topic.

```
{"level":"info","ts":"2018-08-31T18:20:41.525Z","logger":"bus","caller":"kafka/bus.go:205","msg":"Provisioning topic default.k8s-event-flow on bus backed by Kafka","channels.knative.dev/bus":"kafka","channels.knative.dev/busType":"kafka","channels.knative.dev/busComponent":"provisioner"}
[...]

PR #3406, opened 2025-12-15
fix(admin): retry on network errors by resetting controller connection

DCjanus:
- **Context**  
  - Admin calls can get stuck after a network failure (broken pipe / connection reset) because the cached controller connection isn’t discarded. See #1162.  
  - The current admin retry only considers `ErrNotController`/`EOF` as retriable (definition at [`admin.go#L214-L225`](https://github.com/IBM/sarama/blob/e7ee950c053b9c06cc2df2be1ec33b4f94f3767d/admin.go#L214-L225)), so network errors bubble out without healing the connection.

- **What’s changed**  
  - Treat network failures (`net.Error`, `io.ErrUnexpectedEOF`, `EPIPE`) as retriable in admin/controller paths.  
  - When such an error occurs inside `retryOnError`, close the cached controller connection before the next attempt; the subsequent call to `Controller()` in each admin operation (e.g. `CreateTopic` at [`admin.go#L265-L288`](https://github.com/IBM/sarama/blob/e7ee950c053b9c06cc2df2be1ec33b4f94f3767d/admin.go#L265-L288)) reopens a fresh TCP socket automatically.

- **Why it’s safe**  
  - The close happens only after a request has already failed with a network error, inside the admin retry loop (`admin.go#L230-L242`). Healthy calls are untouched.  
  - Reconnect reuses the existing controller metadata; if the broker has changed, the next retry still hits the existing `ErrNotController` path, which already triggers `RefreshController()`—same behavior as today, just without reusing a broken socket.  
  - Scope is limited to ClusterAdmin; producer/consumer code paths are unchanged.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -6,9 +6,10 @@ import (
 	"io"
 	"maps"
 	"math/rand"
+	"net"
 	"strconv"
-	"syscall"
 	"sync"
+	"syscall"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


DCjanus:
After further consideration, I realized that my fix was incorrect, and I need to think more about a more appropriate implementation.

PR #3412, opened 2025-12-19
fix(broker): auto-close broken connections

DCjanus:
## Background
- Reproduces and fixes the "cannot recover after TCP reset" issue tracked in [IBM/sarama#1162](https://github.com/IBM/sarama/issues/1162).
- An earlier attempt in [PR #3406](https://github.com/IBM/sarama/pull/3406) required scattered fixes in many call sites and was hard to maintain. This revision centralizes the logic inside Broker.
- Note: The design idea is mine, but the implementation and wording were assisted by AI (I am not a native speaker). If the tone feels a bit mechanical, please kindly excuse it, and thank you for reviewing.

## Approach
- Add `shouldCloseBrokerConn` to detect fatal transport errors inside Broker.
- When triggered, immediately call `closeInnerLocked` to close/reset the connection and preserve the original error in `connErr`, so callers can re-establish via explicit `Open()`.

## Changes
- `broker.go`: auto-close and reset state on transport errors during ApiVersions negotiation and `sendAndReceive`; extract lock-held helper `closeInnerLocked`.
- Unit test: `TestBrokerOpenApiVersionsTransportError` uses a custom dialer to reliably hit EOF, ensuring the connection is marked unusable and `Open` is retryable.
- Functional test: `TestFuncAdminNetworkErrorClosesControllerConnection` injects a one-shot `reset_peer` via toxiproxy to prove `Open` triggers reconnection and metadata requests recover; adds proxy helper utilities.
- Test env: `docker compose down/rm` always includes `--profile zookeeper` to avoid leftovers with Kafka < 4.
- Dependency bumps: `github.com/klauspost/compress` → 1.18.2, `golang.org/x/sync` → 0.18.0 (no behavior change).

## Verification
- Unit: `go test -run '^TestBrokerOpenApiVersionsTransportError$' -count=1 ./...`
- Functional: `go test -tags functional -run '^TestFuncAdminNetworkErrorClosesControllerConnection$' -count=1 ./...`

## Compatibility
No breaking API changes; only closes broken connections earlier on transport errors and relies on explicit `Open` to reconnect.

### DIFF
--- a/functional_client_test.go
+++ b/functional_client_test.go
@@ -27,6 +27,11 @@ func TestFuncConnectionFailure(t *testing.T) {
 }
 
 func TestFuncAdminNetworkErrorClosesControllerConnection(t *testing.T) {
+	// Goal (IBM/sarama#1162): verify controller reconnection semantics after a TCP reset.
+	// Expected flow:
+	// 1) First metadata request succeeds.
+	// 2) Injected TCP reset makes the next metadata request fail.
+	// 3) Explicit Open triggers automatic reconnection and the subsequent metadata request succeeds.
 	checkKafkaVersion(t, "0.11.0.0")
 	setupFunctionalTest(t)
 	defer teardownFunctionalTest(t)
@@ -44,25 +49,12 @@ func TestFuncAdminNetworkErrorClosesControllerConnection(t *testing.T) {
 	}
 	defer safeClose(t, adminClient)
 
-	ca, ok := adminClient.(*clusterAdmin)
-	if !ok {
-		t.Fatalf("expected *clusterAdmin, got %T", adminClient)
-	}
-
-	controller, err := ca.Controller()
+	controller, err := adminClient.Controller()
 	if err != nil {
 		t.Fatal(err)
 	}
 	if controller.ID() < 0 {
-		_, controllerID, err := ca.DescribeCluster()
-		if err != nil {
-			t.Fatal(err)
-		}
-		controller, err = ca.findBroker(controllerID)
-		if err != nil {
-			t.Fatal(err)
-		}
-		_ = controller.Open(config)
+		t.Fatalf("expected controller broker ID to be resolved, got %d", controller.ID())
 	}
 
 	// Warm up the connection so the proxy toxic applies to an established TCP session.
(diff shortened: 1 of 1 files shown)


broker.go
```diff
@@ -340,19 +362,28 @@ func (b *Broker) Close() error {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
+	b.connErr = nil
+	return b.closeInnerLocked()
+}
+
+// closeInnerLocked is like closeInner but requires b.lock to be held by caller.
```
puellanivis: ¿ 😕 ? I don’t see any `closeInner` anywhere.

DCjanus: > ¿ 😕 ? I don’t see any `closeInner` anywhere.

Ah, yeah — good catch.

`closeInner` was a function I tried out earlier, but it got removed as the implementation evolved.
  
I forgot to remove the reference from the comment. I’ll fix that.

DCjanus: > ¿ 😕 ? I don’t see any `closeInner` anywhere.

Fixed in [f69de2e](https://github.com/IBM/sarama/pull/3412/changes/f69de2ec01993001c56e124ea62f8b391b9af777).

Thanks @puellanivis,  for your time and careful review.

puellanivis:
> I also wasn't totally sure on the naming of `closeInnerLocked`, although I couldn't immediately think of a better name for it

Yeah, the common pattern in Go is usually to provide a `close` receiver method, and a `Close` receiver method that calls the unexported receiver method. Since the lowercase name cannot be called by anyone outside of this package, it’s inherently “inner”. Even when a lock must be held when calling it. However, I can agree that a `Locked` suffix is useful, and prevents a collision with the builtin `close` function name, though since it’s a receiver method, scope and form of usage is usually sufficient disambiguation.

DCjanus:
> I also wasn't totally sure on the naming of closeInnerLocked, although I couldn't immediately think of a better name for it

I’m considering renaming `closeInnerLocked` to `closeLocked` since it’s a broker method and the lock requirement is implied.

Does that feel clear enough, or would you prefer something more explicit like `closeConnAndResetLocked`?

DCjanus:
> I wonder if we should make this a dedicated func on broker to avoid the repetition?

Just introduced a `maybeCloseLocked` helper in [this commit](https://github.com/IBM/sarama/pull/3412/commits/506ed17a3e5bd38a68fe9b3eb861380e4c73eefd) to consolidate the repeated close-on-error logic.

DCjanus:
I went through the comments pass and made the notes shorter. The comment-trimming commit is [a138d2c](https://github.com/IBM/sarama/pull/3412/commits/a138d2cfacee34c8214eb25cc1410fc150218154). 🤝

About the extra comments: I do use an agent, but I’m still getting familiar with this codebase. I asked the agent to explain parts of the design so I could check its output. That made it add more comments than needed for reviewers. I removed most of them and kept only the useful ones. If I removed too much, please feel free to adjust the comments directly.

There was earlier feedback that `closeInnerLocked` felt off. I renamed it to `closeLocked` (commit: [6f8546b](https://github.com/IBM/sarama/pull/3412/changes/6f8546bfa4f628ba39c78f84b9cb07dce711915b)); my thinking is that a lowercase method name already implies “inner,” so `Inner` felt redundant.

If you prefer a different name or any other changes, please feel free to push directly to this branch — maintainer edits are enabled.

dnwe:
LGTM. I've rebase the commits into a clean set and will approve and merge once the FV checks have all passed 👍🏻

PR #3419, opened 2026-01-05
feat(admin): add KIP-396 list/alter offsets APIs

### LINKED ISSUE #1669: KIP-396: Add Reset/List Offsets Operations to AdminClient
Add support for [KIP-396](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=97551484) which was added to the Java AdminClient in the Kafka 2.5.0

> Resetting the offsets of a consumer group is a relatively common operation. It's usually done to allow an application to reprocess data or otherwise to skip/drop records. Currently (thanks to KIP-122) it can be done using the kafka-consumer-group.sh tool however it would be nice to be able to to it directly with the AdminClient. Applications can more easily depend on the AdminClient API rather than running the ConsumerGroupsCommand class or worse reimplementing the logic. The ability to retrieve committed offsets is already available in the AdminClient and has proved very useful. Additionally KIP-496 is adding the ability to delete committed offsets. This justifies integrating the consumer group reset/alter functionality directly in the AdminClient.

DCjanus:
## Summary

- Add Admin APIs for KIP-396 `ListOffsets` and `AlterConsumerGroupOffsets` to support bulk offset queries and group offset commits.
- Provide Go types aligned with KIP-396 semantics (`OffsetAndMetadata`).

## Key changes

- Add `ListOffsets` and `AlterConsumerGroupOffsets` to `ClusterAdmin` and wire protocol requests.
- Add result structs with leader epoch support and commit metadata for admin offset commits.
- Implement broker fan-out for list offsets and coordinator commit path for alter offsets.
- Add functional tests covering timestamp list offsets and admin offset commits.

## Constraints / tradeoffs

- `ListOffsets` currently accepts `int64` queries (earliest/latest/timestamp) to stay consistent with existing offset conventions; we can introduce an explicit spec type in a follow-up if needed.

## Notes

- Spec reference: [KIP-396](https://cwiki.apache.org/confluence/display/KAFKA/KIP-396%3A+Add+Offset+Management+APIs+to+AdminClient)
- Closes [#1669](https://github.com/IBM/sarama/issues/1669)

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -115,6 +115,20 @@ type ClusterAdmin interface {
 	// List the consumer group offsets available in the cluster.
 	ListConsumerGroupOffsets(group string, topicPartitions map[string][]int32) (*OffsetFetchResponse, error)
 
+	// ListOffsets lists offsets for the specified topic partitions.
+	// Each value is OffsetNewest, OffsetOldest, or a timestamp in milliseconds.
+	// Results are keyed by topic/partition and include per-partition errors.
+	//
+	// For oldest/newest requests, Kafka may return a valid offset while timestamp is -1.
+	// To get the exact message timestamp, fetch the record at that offset.
+	// This operation is supported by brokers with version 0.10.1.0 or higher.
+	ListOffsets(partitions map[string]map[int32]int64, options *ListOffsetsOptions) (map[string]map[int32]*OffsetResult, error)
+
+	// AlterConsumerGroupOffsets alters offsets for the specified group by committing the provided offsets and metadata.
+	// The request targets the group's coordinator and returns per-partition results in the response.
+	// This operation is not transactional so it may succeed for some partitions while fail for others.
+	AlterConsumerGroupOffsets(group string, offsets map[string]map[int32]OffsetAndMetadata, options *AlterConsumerGroupOffsetsOptions) (*OffsetCommitResponse, error)
+
 	// Deletes a consumer group offset
 	DeleteConsumerGroupOffset(group string, topic string, partition int32) error
 
@@ -235,6 +249,17 @@ func isRetriableListTopicsError(err error) bool {
 }
 
 func isRetriableListTopicsTimeoutError(err error) bool {
+	var netErr net.Error
+	return errors.As(err, &netErr) && netErr.Timeout()
+}
+
+// isRetriableBrokerError returns `true` if the given error is a retryable
+// transport error or a timeout.
+func isRetriableBrokerError(err error) bool {
+	return errors.Is(err, ErrNotConnected) || shouldCloseBrokerConn(err) || isTimeoutError(err)
+}
+
+func isTimeoutError(err error) bool {
 	var netErr net.Error
 	return errors.As(err, &netErr) && netErr.Timeout()
 }
--- a/admin.go
+++ b/admin.go
@@ -245,12 +245,7 @@ func isRetriableListTopicsError(err error) bool {
 		return true
 	}
 
-	return isRetriableListTopicsTimeoutError(err)
-}
-
-func isRetriableListTopicsTimeoutError(err error) bool {
-	var netErr net.Error
-	return errors.As(err, &netErr) && netErr.Timeout()
+	return isTimeoutError(err)
 }
 
 // isRetriableBrokerError returns `true` if the given error is a retryable
@@ -473,7 +468,7 @@ func (ca *clusterAdmin) ListTopics() (map[string]TopicDetail, error) {
 		metadataReq := NewMetadataRequest(ca.conf.Version, nil)
 		metadataResp, err := b.GetMetadata(metadataReq)
 		if err != nil {
-			if isRetriableListTopicsTimeoutError(err) {
+			if isTimeoutError(err) {
 				_ = b.Close()
 			}
 			return err
@@ -517,7 +512,7 @@ func (ca *clusterAdmin) ListTopics() (map[string]TopicDetail, error) {
 
 		describeConfigsResp, err := b.DescribeConfigs(describeConfigsReq)
 		if err != nil {
-			if isRetriableListTopicsTimeoutError(err) {
+			if isTimeoutError(err) {
 				_ = b.Close()
 			}
 			return err
--- a/admin_offsets.go
+++ b/admin_offsets.go
@@ -1,0 +1,182 @@
+package sarama
+
+import (
+	"errors"
+	"sync"
+)
+
+// ListOffsetsOptions configures how offsets are fetched.
+type ListOffsetsOptions struct {
+	IsolationLevel IsolationLevel
+}
+
+// OffsetResult contains the response for a single topic partition.
+type OffsetResult struct {
+	Offset      int64
+	Timestamp   int64
+	LeaderEpoch int32
+	Err         error
+}
+
+// OffsetAndMetadata describes the offset commit for a single partition.
+type OffsetAndMetadata struct {
+	Offset   int64
+	Metadata string
+	// LeaderEpoch contains the leader epoch of the last consumed record.
+	// It is used by the broker to fence stale commits after leader changes.
+	// Use the epoch returned by offset fetch/list APIs, or -1 to omit.
+	LeaderEpoch int32
+}
+
+// AlterConsumerGroupOffsetsOptions configures how offsets are committed.
+// It is currently empty and reserved for future Kafka protocol options.
+type AlterConsumerGroupOffsetsOptions struct{}
+
+func (ca *clusterAdmin) ListOffsets(partitions map[string]map[int32]int64, options *ListOffsetsOptions) (map[string]map[int32]*OffsetResult, error) {
+	type topicPartition struct {
+		topic     string
+		partition int32
+	}
+
+	type brokerOffsetRequest struct {
+		request    *OffsetRequest
+		partitions []topicPartition
+	}
+
+	type brokerOffsetResult struct {
+		result map[topicPartition]*OffsetResult
+		err    error
+	}
+
+	if len(partitions) == 0 {
+		return nil, ConfigurationError("no partitions provided")
+	}
+
+	if options == nil {
+		options = &ListOffsetsOptions{}
+	}
+
+	requests := make(map[*Broker]*brokerOffsetRequest)
+	for topic, topicOffsets := range partitions {
+		for partition, offsetQuery := range topicOffsets {
+			broker, _, err := ca.client.LeaderAndEpoch(topic, partition)
+			if err != nil {
+				return nil, err
+			}
+
+			req := requests[broker]
+			if req == nil {
+				req = &brokerOffsetRequest{
+					request: NewOffsetRequest(ca.conf.Version),
+				}
+				req.request.IsolationLevel = options.IsolationLevel
+				requests[broker] = req
+			}
+			req.request.AddBlock(topic, partition, offsetQuery, 1)
+			req.partitions = append(req.partitions, topicPartition{topic: topic, partition: partition})
+		}
+	}
+
+	results := make(chan brokerOffsetResult)
+	var wg sync.WaitGroup
+
+	for broker, req := range requests {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			var resp *OffsetResponse
+			err := ca.retryOnError(isRetriableBrokerError, func() error {
+				var err error
+				_ = broker.Open(ca.client.Config())
+				resp, err = broker.GetAvailableOffsets(req.request)
+				return err
+			})
+			if err != nil {
+				results <- brokerOffsetResult{err: err}
+				return
+			}
+			broker.handleThrottledResponse(resp)
+
+			partitionResults := make(map[topicPartition]*OffsetResult, len(req.partitions))
+			for _, tp := range req.partitions {
+				block := resp.GetBlock(tp.topic, tp.partition)
+				if block == nil {
+					partitionResults[tp] = &OffsetResult{Err: ErrIncompleteResponse}
+					continue
+				}
+				partitionResults[tp] = &OffsetResult{
+					Offset:      block.Offset,
+					Timestamp:   block.Timestamp,
+					LeaderEpoch: block.LeaderEpoch,
+					Err:         block.Err,
+				}
+			}
+
+			results <- brokerOffsetResult{result: partitionResults}
+		}()
+	}
+
+	go func() {
+		wg.Wait()
+		close(results)
+	}()
+
+	allResults := make(map[string]map[int32]*OffsetResult, len(partitions))
+	var errs []error
+	for res := range results {
+		if res.err != nil {
+			errs = append(errs, res.err)
+		}
+		for tp, info := range res.result {
+			if allResults[tp.topic] == nil {
+				allResults[tp.topic] = make(map[int32]*OffsetResult)
+			}
+			allResults[tp.topic][tp.partition] = info
+		}
+	}
+
+	return allResults, errors.Join(errs...)
+}
+
+func (ca *clusterAdmin) AlterConsumerGroupOffsets(group string, offsets map[string]map[int32]OffsetAndMetadata, _ *AlterConsumerGroupOffsetsOptions) (*OffsetCommitResponse, error) {
+	if len(offsets) == 0 {
+		return nil, ConfigurationError("no offsets provided")
+	}
+
+	var response *OffsetCommitResponse
+	request := NewOffsetCommitRequest(ca.conf, group)
+
+	var commitTimestamp int64
+	if request.Version == 1 {
+		commitTimestamp = ReceiveTime
+	}
+
+	for topic, topicOffsets := range offsets {
+		for partition, offset := range topicOffsets {
+			request.AddBlockWithLeaderEpoch(topic, partition, offset.Offset, offset.LeaderEpoch, commitTimestamp, offset.Metadata)
+		}
+	}
+
+	err := ca.retryOnError(isRetriableGroupCoordinatorError, func() (err error) {
+		defer func() {
+			if err != nil && isRetriableGroupCoordinatorError(err) {
+				_ = ca.client.RefreshCoordinator(group)
+			}
+		}()
+
+		coordinator, err := ca.client.Coordinator(group)
+		if err != nil {
+			return err
+		}
+
+		response, err = coordinator.CommitOffset(request)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+
+	return response, err
+}
--- a/admin_offsets.go
+++ b/admin_offsets.go
@@ -7,6 +7,8 @@ import (
 
 // ListOffsetsOptions configures how offsets are fetched.
 type ListOffsetsOptions struct {
+	// IsolationLevel selects between ReadUncommitted (default) and ReadCommitted
+	// when fetching the latest offset. Only honored by brokers running v0.11+.
 	IsolationLevel IsolationLevel
 }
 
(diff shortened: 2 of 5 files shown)


admin_offsets.go
```diff
@@ -0,0 +1,229 @@
...
+
+// TopicPartitionID identifies a specific partition in a topic.
+type TopicPartitionID struct {
+	Topic     string
+	Partition int32
+}
+
+// OffsetSpec specifies which offset to look up for a partition.
+type OffsetSpec struct {
+	timestamp int64
```
puellanivis: I’m not sure why we need a ton of constructors, when we can simply export this value.

After all, given the functionality provided, we can already arbitrarily mutate any given `OffsetSpec` and access the field at will:
```go
offspec := OffsetSpecLatest()
offspec = OffsetSpecForTimestamp(arbitraryTimestamp)
go func() {
	offspec = OffsetSpecForTimestamp(OffsetNewest)
}()
ts := offspec.Timestamp() // write-read race condition
```

admin_offsets.go
```diff
@@ -0,0 +1,229 @@
...
+			}
+			req.request.IsolationLevel = options.IsolationLevel
+			requests[broker] = req
+		}
+		req.request.AddBlock(tp.Topic, tp.Partition, spec.timestamp, 1)
+		req.partitions = append(req.partitions, tp)
+	}
+
+	results := make(chan brokerOffsetResult, len(requests))
+	wg := sync.WaitGroup{}
```
puellanivis: [style] There is no need to assign zero values to variables, just simply declare them:
```suggestion
	var wg sync.WaitGroup
```

admin_offsets.go
```diff
@@ -0,0 +1,229 @@
...
+					Err:         block.Err,
+				}
+			}
+
+			results <- brokerOffsetResult{result: partitionResults}
+		}(req)
+	}
+
+	wg.Wait()
+	close(results)
```
puellanivis: Perform this in a goroutine, so that we can process results as they come in, rather than waiting for all of them to return results before processing any of them:
```suggestion
	go func() {
		wg.Wait()
		close(results)
	}()
```

This also means we don’t need to buffer the channel either.

admin_offsets.go
```diff
@@ -0,0 +1,229 @@
...
+
+	allResults := make(map[TopicPartitionID]*ListOffsetsResult, len(partitions))
+	var firstErr error
```
puellanivis:
```go
var errs []error
for res := range results {
	if res.err != nil {
		errs = append(errs, res.err)
	}
	...
}

return allResults errors.Join(errs...)
```

admin_offsets.go
```diff
@@ -0,0 +1,229 @@
...
+		req.request.AddBlock(tp.Topic, tp.Partition, spec.timestamp, 1)
+		req.partitions = append(req.partitions, tp)
+	}
+
+	results := make(chan brokerOffsetResult, len(requests))
+	wg := sync.WaitGroup{}
+
+	for _, req := range requests {
+		wg.Add(1)
+		go func(req *brokerOffsetRequest) {
```
puellanivis: How many requests are we likely to be spinning off here?

It can happen that spinning off too many goroutines will actually be performance-degrading rather than performance-enhancing.

admin_offsets.go
```diff
@@ -0,0 +1,229 @@
...
+		request.Version = 6
+	}
+	if conf.Version.IsAtLeast(V2_3_0_0) {
+		request.Version = 7
+	}
+
+	if request.Version >= 2 && request.Version < 5 {
+		request.RetentionTime = -1
+		if conf.Consumer.Offsets.Retention > 0 {
+			request.RetentionTime = int64(conf.Consumer.Offsets.Retention / time.Millisecond)
```
puellanivis:
```suggestion
			request.RetentionTime = conf.Consumer.Offsets.Retention.Milliseconds()
```

admin.go
```diff
@@ -115,6 +115,20 @@ type ClusterAdmin interface {
...
+	//
+	// For oldest/newest requests, Kafka may return a valid offset while timestamp is -1.
+	// To get the exact message timestamp, fetch the record at that offset.
+	// This operation is supported by brokers with version 0.10.1.0 or higher.
+	ListOffsets(partitions map[TopicPartitionID]OffsetSpec, options *ListOffsetsOptions) (map[TopicPartitionID]*ListOffsetsResult, error)
+
+	// AlterConsumerGroupOffsets alters offsets for the specified group by committing the provided offsets and metadata.
+	// The request targets the group's coordinator and returns per-partition results in the response.
+	// This operation is not transactional so it may succeed for some partitions while fail for others.
+	AlterConsumerGroupOffsets(group string, offsets map[TopicPartitionID]OffsetAndMetadata, options *AlterConsumerGroupOffsetsOptions) (*OffsetCommitResponse, error)
```
puellanivis: Our current design uses `map[string]map[int32]V` adding this `TopicPartitionID` might have been nice if we had started with it, but we haven’t really. https://pkg.go.dev/github.com/IBM/sarama#AlterPartitionReassignmentsResponse

Providing two different ways to do something is likely to increase confusion over any benefits of flattening the maps.

It also reduces the ability to iterate over topics individually, without needing to then iterate over all `Topic` × `Partition` combinations and select for `Topic`.

admin_offsets.go
```diff
@@ -0,0 +1,229 @@
...
+	return s.timestamp
+}
+
+// ListOffsetsOptions configures how offsets are fetched.
+type ListOffsetsOptions struct {
+	IsolationLevel IsolationLevel
+}
+
+// ListOffsetsResult contains the response for a single topic partition.
+type ListOffsetsResult struct {
```
puellanivis: I’m not sure this is an `Offsets` result? As it seems to be single offset result?

admin_offsets.go
```diff
@@ -0,0 +1,229 @@
...
+type brokerOffsetRequest struct {
+	broker     *Broker
+	request    *OffsetRequest
+	partitions []TopicPartitionID
+}
+
+type brokerOffsetResult struct {
+	result map[TopicPartitionID]*ListOffsetsResult
+	err    error
+}
```
puellanivis: We could also isolate/scope these types into the `ListOffsets` right?

admin_offsets.go
```diff
@@ -0,0 +1,229 @@
...
+	for tp, spec := range partitions {
+		broker, _, err := ca.client.LeaderAndEpoch(tp.Topic, tp.Partition)
+		if err != nil {
+			return nil, err
+		}
+
+		req := requests[broker]
+		if req == nil {
+			req = &brokerOffsetRequest{
+				broker:  broker,
```
puellanivis: I’m not sure why we double track the `broker`? We’re using it as the key, and as a field in the value?

We could just `for broker, req := range requests { … }` below, right?

admin_offsets.go
DCjanus: > This also means we don’t need to buffer the channel either.

make results channel unbuffered in https://github.com/IBM/sarama/pull/3419/commits/51b33dc48ebe98299937b4b6b00673fa69ab85ce

admin_offsets.go
DCjanus: > ```go
> var errs []error
> for res := range results {
> 	if res.err != nil {
> 		errs = append(errs, res.err)
> 	}
> 	...
> }
> 
> return allResults errors.Join(errs...)
> ```

Addressed in commit bce2339 by aggregating all errors with errors.Join.

admin_offsets.go
DCjanus: > We could also isolate/scope these types into the `ListOffsets` right?

Updated in commit 243445e (scoped the helper types inside ListOffsets).

admin_offsets.go
DCjanus: > I’m not sure why we double track the `broker`? We’re using it as the key, and as a field in the value?
> 
> We could just `for broker, req := range requests { … }` below, right?

Addressed in commit ea32b6e by removing the redundant broker field and passing the broker via the map key in the loop.

admin_offsets.go
DCjanus: Renamed `ListOffsetsResult` to `OffsetResult` to reflect the single-partition result; done in commit f9150fe.

admin_offsets.go
DCjanus: > How many requests are we likely to be spinning off here?
> 
> It can happen that spinning off too many goroutines will actually be performance-degrading rather than performance-enhancing.

Thanks for the note. The Java AdminClient doesn’t appear to impose an explicit concurrency cap either — it groups by broker and issues one in‑flight request per broker via the admin I/O loop.

Key entry point (no concurrency limiting logic):
- [`KafkaAdminClient#listOffsets`](https://github.com/apache/kafka/blob/6887bb791e42862b456016aff47d86bf1988a3bb/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L4260-L4269)

In practice, the Java I/O loop is equivalent to spawning goroutines in Go: both drive concurrent in‑flight requests without a per‑broker cap.

Given typical Kafka clusters are well below `1,000` brokers, I believe even `10,000` concurrent in‑flight requests should be relatively easy for modern hardware to handle.

admin_offsets.go
DCjanus: > I’m not sure why we need a ton of constructors, when we can simply export this value.
> 
> After all, given the functionality provided, we can already arbitrarily mutate any given `OffsetSpec` and access the field at will:
> 
> ```go
> offspec := OffsetSpecLatest()
> offspec = OffsetSpecForTimestamp(arbitraryTimestamp)
> go func() {
> 	offspec = OffsetSpecForTimestamp(OffsetNewest)
> }()
> ts := offspec.Timestamp() // write-read race condition
> ```

Agree that we should not mix two different input styles for the same concept (as noted in https://github.com/IBM/sarama/pull/3419#discussion_r2668626040).

To stay consistent with existing Sarama APIs (e.g. `Client.GetOffset(topic, partition, time int64)` which uses `OffsetOldest`/`OffsetNewest`), I removed `OffsetSpec` and made `ListOffsets` take `int64` directly (pass `OffsetOldest`/`OffsetNewest` or a millisecond timestamp).

Done in commit f3e40c7.

admin.go
DCjanus: > Our current design uses `map[string]map[int32]V` adding this `TopicPartitionID` might have been nice if we had started with it, but we haven’t really. https://pkg.go.dev/github.com/IBM/sarama#AlterPartitionReassignmentsResponse
> 
> Providing two different ways to do something is likely to increase confusion over any benefits of flattening the maps.
> 
> It also reduces the ability to iterate over topics individually, without needing to then iterate over all `Topic` × `Partition` combinations and select for `Topic`.

Thanks for the point about keeping the API shape consistent. I reverted to the existing `map[string]map[int32]V` style so callers can iterate by topic without scanning all partitions.

This removes `TopicPartitionID` from `ListOffsets` and aligns the input/return maps with the rest of Sarama’s admin APIs.

Done in commit be65797.

admin_offsets.go
```diff
@@ -0,0 +1,215 @@
...
+			req.partitions = append(req.partitions, topicPartition{topic: topic, partition: partition})
+		}
+	}
+
+	results := make(chan brokerOffsetResult)
+	var wg sync.WaitGroup
+
+	for broker, req := range requests {
+		wg.Add(1)
+		go func(broker *Broker, req *brokerOffsetRequest) {
```
puellanivis: [nitpick] We’re compiling this under `go.mod` `go version v1.24.0`, which should be well enough into the change of semantics for loopvars that removed the need to passing these variables through arguments, rather than accessing them by simple closure. 🤔 Though, I’m not sure how we would want to approach changing this common pattern, even if it is now unnecessary.

admin_offsets.go
```diff
@@ -0,0 +1,215 @@
...
+			defer wg.Done()
+
+			resp, err := broker.GetAvailableOffsets(req.request)
+			if err != nil {
+				results <- brokerOffsetResult{err: err}
+				return
+			}
+			broker.handleThrottledResponse(resp)
+
+			partitionResults := make(map[string]map[int32]*OffsetResult)
```
puellanivis: As a seeming reversal of my prior guidance; since this type is function internal, we could also use the `topicPartition` type as a key here, and avoid needing more than the one map.

(One of those cases where blackbox internals can choose different paradigms from the public API.)

puellanivis (commented):
Looks pretty good, just two things that are perhaps more style than substance.

admin_offsets.go
DCjanus: > [nitpick] We’re compiling this under `go.mod` `go version v1.24.0`, which should be well enough into the change of semantics for loopvars that removed the need to passing these variables through arguments, rather than accessing them by simple closure. 🤔 Though, I’m not sure how we would want to approach changing this common pattern, even if it is now unnecessary.

Thanks for the style note.

Personally, I prefer passing loop variables into goroutines so it is clear what is shared.

I also want to match the project style. If you think it is better to rely on the Go 1.24 behavior, I can change it — or you can push a commit since maintainer edits are enabled.

admin_offsets.go
DCjanus: > As a seeming reversal of my prior guidance; since this type is function internal, we could also use the `topicPartition` type as a key here, and avoid needing more than the one map.
> 
> (One of those cases where blackbox internals can choose different paradigms from the public API.)

Thanks! Updated in https://github.com/IBM/sarama/pull/3419/commits/494f7c37f0b11daada4feb3901422579bef7f399.

admin_offsets.go
puellanivis: The usual problem is that to know for sure which values are being put into these parameters, one has to drop all the way to the bottom of the `go func() { … }()` to check, (though good here, you’re also shadowing the originals, so one cannot accidentally access the wrong one; since whether one likes to be explicit or not, the implicit closure access is always there), and it requires repeating the type information of the parameters as well, while when accessed through closure.

So, if we refactor a type name of one of the variables, this is then Yet Another Place where the type-name must be refactored.

PS: To be more clear, because I forgot to mention it, this is fine. 🤷‍♀️

DCjanus: > The usual problem is that to know for sure which values are being put into these parameters, one has to drop all the way to the bottom of the `go func() { … }()` to check, (though good here, you’re also shadowing the originals, so one cannot accidentally access the wrong one; since whether one likes to be explicit or not, the implicit closure access is always there), and it requires repeating the type information of the parameters as well, while when accessed through closure.
> 
> So, if we refactor a type name of one of the variables, this is then Yet Another Place where the type-name must be refactored.
> 
> PS: To be more clear, because I forgot to mention it, this is fine. 🤷‍♀️

Thanks for the clarification — you convinced me. I switched this to simple closure capture in 3d11625.

In a language without a “no implicit capture” guarantee, explicit parameters don’t really reduce review burden, since you still have to check for other implicit captures.

admin_offsets.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	if conf.Version.IsAtLeast(V2_0_0_0) {
+		request.Version = 4
+	}
+	if conf.Version.IsAtLeast(V2_1_0_0) {
+		request.Version = 6
+	}
+	if conf.Version.IsAtLeast(V2_3_0_0) {
+		request.Version = 7
+	}
```
dnwe: Can we structure these in descending order like we do elsewhere in Sarama? e.g., https://github.com/IBM/sarama/blob/2fb8eaf4eca13fb44e4aac9b2ece0f53be988666/offset_fetch_request.go#L23-L46

It is somewhat stylistic (but also avoids multiple assignment). It is also useful to have the version-to-version additions (taken from Kafka's protocol json files) to show what changed between them too

You may also want to move this into offset_commit_request.go as a NewOffsetCommitRequest func that sets up the correct version etc.

dnwe (commented):
I did a quick first pass, one stylistic comment below, but also one main question here: is there a reason why we have the `ca.retryOnError` wrapper on AlterConsumerGroupOffsets, but not on any of the calls in ListOffsets?

admin_offsets.go
DCjanus: > Can we structure these in descending order like we do elsewhere in Sarama? e.g.,
> 
> https://github.com/IBM/sarama/blob/2fb8eaf4eca13fb44e4aac9b2ece0f53be988666/offset_fetch_request.go#L23-L46
> 
> It is somewhat stylistic (but also avoids multiple assignment). It is also useful to have the version-to-version additions (taken from Kafka's protocol json files) to show what changed between them too
> 
> You may also want to move this into offset_commit_request.go as a NewOffsetCommitRequest func that sets up the correct version etc.

- Switched to descending version checks with explicit version-diff comments.
- Moved the admin `OffsetCommitRequest` initializer to `offset_commit_request.go` as `NewOffsetCommitRequest` and used it from `admin_offsets.go`.

Done in commit c525eb08f0138d301c87dfb5a3c9d06a35b9bd63

DCjanus:
> I did a quick first pass, one stylistic comment below, but also one main question here: is there a reason why we have the `ca.retryOnError` wrapper on AlterConsumerGroupOffsets, but not on any of the calls in ListOffsets?

I missed the retry logic earlier; added it in commit 4446246.

admin_offsets.go
puellanivis: I wanna add as an aside, but we can also switch to a more go-style:
```go
switch {
case version.IsAtLeast(V2_5_0_0):
case version.IsAtLeast(V2_4_0_0):
...
}
```

But I’ve not mentioned this before, because the style already here is are `if … else if … else …` cascades.

DCjanus: > I wanna add as an aside, but we can also switch to a more go-style:
> 
> ```go
> switch {
> case version.IsAtLeast(V2_5_0_0):
> case version.IsAtLeast(V2_4_0_0):
> ...
> }
> ```
> 
> But I’ve not mentioned this before, because the style already here is are `if … else if … else …` cascades.

  Thanks for the suggestion! I agree it’s more Go‑style.

  To keep this PR focused and clean, I’d prefer not to change this here.

  @dnwe if you also prefer the `switch {}` style, I can follow up with a small PR to convert similar cascades across the project.

DCjanus:
Hi @dnwe,

I saw the new AI-assistance guidance in https://github.com/IBM/sarama/pull/3452, including the note about not putting AI tools/models in `Co-authored-by`.

Some commits in this PR include:`Co-authored-by: OpenAI Codex <codex@openai.com>`

I add that via a global `AGENTS.md` rule across repos to keep AI disclosure explicit and consistent. This also makes it easy for maintainers who prefer to avoid AI-assisted contributions to identify them quickly (I may have a different view, but I respect that preference).

I also understand why `Signed-off-by` should remain human-only, since AI agents do not have legal identity for attestation.

Before opening PRs, I still review and validate all changes myself.

Given the new policy, what would you prefer for this PR?
Should I rewrite history (force-push) to remove those `Co-authored-by` trailers, or keep existing commits as-is and follow the new rule from now on?

Happy to follow your preference.

(Also, this comment was lightly polished with AI for clarity because English is not my native language, and I want to avoid ambiguity in technical communication.)

DCjanus:
@dnwe Quick correction: for this PR, I had not enabled my AGENTS.md co-author rule yet, so these commits do not include an AI `Co-authored-by` trailer.

I still want to discuss the policy in general, but this specific PR is not an example of that case.

puellanivis:
So, it’s inappropriate to attribute co-authorship to anything that is not a person, and thus not able to hold any copyrights. Having random public domain uncopyrightable code mixed into your PR muddies the protections and/or requirements that may be necessary for license compliance.

admin.go
```diff
@@ -224,6 +238,17 @@ func isRetriableGroupCoordinatorError(err error) bool {
...
 
+// isRetriableBrokerError returns `true` if the given error is a retryable
+// transport error or a timeout.
+func isRetriableBrokerError(err error) bool {
+	return errors.Is(err, ErrNotConnected) || shouldCloseBrokerConn(err) || isTimeoutError(err)
+}
+
+func isTimeoutError(err error) bool {
+	var netErr net.Error
+	return errors.As(err, &netErr) && netErr.Timeout()
```
puellanivis, line 249: Ah, `errors.AsType` is in `v1.26` so not available yet for us.

puellanivis (approved):

dnwe (approved):
Thanks, LGTM - I added a couple of extra commits just to add a bit more test coverage and enhance the implementation a little

PR #3422, opened 2026-01-10
feat(producer): partition muting for msg ordering

dnwe:
This commit attempts to copy the Java client's concept of a `RecordAccumulator`, which "mutes" a partition when a batch is in-flight. This prevents a subsequent batch for the same partition from being sent before the first one has been acknowledged.

The `brokerProducer` now mutes a batch's partitions before sending and unmutes them only after the request is complete (including the queuing of any retries). A `partitionMuter` is introduced within `asyncProducer` to track and manage in-flight requests for each topic-partition and the `waitForSpace` logic is updated to respect partition mutes.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -127,23 +127,15 @@ func newPartitionMuter() *partitionMuter {
 // isMuted reports whether the partition has an in-flight batch.
 // Requires: m.mu held.
 func (m *partitionMuter) isMuted(topic string, partition int32) bool {
-	if partitions := m.inFlightCounts[topic]; partitions != nil {
-		return partitions[partition] > 0
-	}
-	return false
+	return m.inFlightCounts[topic][partition] > 0
 }
 
 // isAnyMuted reports whether any partition in the set has an in-flight batch.
 // Requires: m.mu held.
 func (m *partitionMuter) isAnyMuted(set *produceSet) bool {
-	muted := false
-	set.eachPartition(func(topic string, partition int32, _ *partitionSet) {
-		if muted {
-			return
-		}
-		muted = m.isMuted(topic, partition)
+	return set.anyPartition(func(topic string, partition int32, _ *partitionSet) bool {
+		return m.isMuted(topic, partition)
 	})
-	return muted
 }
 
 // mutePartition increments the in-flight count for a single partition.
@@ -1322,12 +1314,7 @@ func (bp *brokerProducer) waitForSpace(msg *ProducerMessage, forceRollover bool)
 
 func (bp *brokerProducer) rollOver() {
 	if bp.timer != nil {
-		if !bp.timer.Stop() {
-			select {
-			case <-bp.timer.C:
-			default:
-			}
-		}
+		bp.timer.Stop()
 	}
 	bp.timer = nil
 	bp.timerFired = false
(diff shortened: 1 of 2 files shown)


async_producer.go
```diff
@@ -100,9 +100,181 @@ type asyncProducer struct {
...
+	}
+	m.cond = sync.NewCond(&m.mu)
+	return m
+}
+
+// isMuted reports whether the partition has an in-flight batch.
+// Requires: m.mu held.
+func (m *partitionMuter) isMuted(topic string, partition int32) bool {
+	if partitions := m.inFlightCounts[topic]; partitions != nil {
+		return partitions[partition] > 0
```
puellanivis: If `partitions == nil` then `partitions[partition] > 0 == false`.

So this function is semantically identical to just `return m.inFlightCounts[topic][partition] > 0`

async_producer.go
```diff
@@ -100,9 +100,181 @@ type asyncProducer struct {
...
+	}
+	return false
+}
+
+// isAnyMuted reports whether any partition in the set has an in-flight batch.
+// Requires: m.mu held.
+func (m *partitionMuter) isAnyMuted(set *produceSet) bool {
+	muted := false
+	set.eachPartition(func(topic string, partition int32, _ *partitionSet) {
```
puellanivis: (We might want to consider reworking this function to work similar to an `iter.Seq2` like function, with an early return `bool`.)

async_producer.go
```diff
@@ -1046,31 +1271,67 @@ func (bp *brokerProducer) needsRetry(msg *ProducerMessage) error {
...
 }
 
 func (bp *brokerProducer) rollOver() {
 	if bp.timer != nil {
-		bp.timer.Stop()
+		if !bp.timer.Stop() {
+			select {
+			case <-bp.timer.C:
+			default:
+			}
```
puellanivis: As of go1.23, this receive is no longer necessary as the channel cannot hold a stale value.

produce_set.go
```diff
@@ -127,6 +131,52 @@ func (ps *produceSet) add(msg *ProducerMessage) error {
...
+			delete(ps.msgs, topic)
+		}
+	}
+	if out.empty() {
+		return nil
+	}
+	return out
+}
+
+func (ps *produceSet) filteredCopy(predicate func(topic string, partition int32) bool) *produceSet {
```
puellanivis: I usually find “filter” to be ambiguous, as after one has filtered something, one can keep the filtrate, or the residue.

In keeping with the `slices` package, `copyFunc` would be a parallel?

dnwe: Agreed, fixed

async_producer.go
dnwe: TIL! Thanks

async_producer.go
dnwe: Added an ‘anyPartition’ to the produceSet with that pattern and calling that here

async_producer.go
dnwe: Good catch!

dnwe:
> 😰 lotsa code, but looks great

😅 yeah I’ve written and thrown away so many branches where I’ve attempted this over the last couple of years because it ended up breaking more than it fixed. This is the first time where it seemed to have got a place that was reasonably understandable and seemed to behave across tests enough to actually consider merging it

puellanivis:
It definitely looks like you’ve spent a few prototypes figuring out how to not write it. It’s pretty solid. 👍 I don’t see anything to comment about anymore.

dnwe:
@puellanivis are you happy to re-review this since you became maintainer so that it counts as an approving review? 😅

puellanivis:
😅

PR #3423, opened 2026-01-11
test: expand Java interop tests to cover all compression codecs

dnwe:
- add gzip, lz4 and zstd to the existing snappy interop tests
- isolate each test's use of the shared topic
- support kafka <2.x (different cmdline options on console utilities)

### DIFF
--- a/functional_java_interop_test.go
+++ b/functional_java_interop_test.go
@@ -51,7 +51,9 @@ func produceWithJava(t *testing.T, topic string, codec CompressionCodec, message
 		if err != nil {
 			stdin.Close()
 			waitErr := cmd.Wait()
-			require.NoError(t, waitErr, "Java producer failed")
+			if waitErr != nil {
+				err = fmt.Errorf("failed to write message: %w; Java producer failed: %w", err, waitErr)
+			}
 		}
 		require.NoError(t, err)
 	}


functional_java_interop_test.go
```diff
@@ -200,25 +145,126 @@ func TestJavaConsumerSnappyRoundTrip(t *testing.T) {
 
 	var stderrOutput strings.Builder
 	go func() {
-		scanner := bufio.NewScanner(stderr)
-		for scanner.Scan() {
-			stderrOutput.WriteString(scanner.Text() + "\n")
+		s := bufio.NewScanner(stderr)
```
puellanivis, line 158: Shouldn’t we have a `s.Err()` test as a sanity check? (If it ever happens, and we’re not reporting it, it could lead to a difficult debug.)

functional_java_interop_test.go
```diff
@@ -200,25 +145,126 @@ func TestJavaConsumerSnappyRoundTrip(t *testing.T) {
...
+	if kafkaVersionAtLeast("0.10.0") {
+		args = append(args, "--bootstrap-server", brokerAddr)
+	} else {
+		args = append(args, "--zookeeper", zookeeperAddr)
+	}
+	return append(args,
+		"--topic", topic,
+		"--partition", "0",
+		"--offset", fmt.Sprintf("%d", startOffset),
+		"--max-messages", fmt.Sprintf("%d", count),
```
puellanivis: I’m not sure how much of a difference this would actually make though:
```suggestion
		"--offset", fmt.Sprint(startOffset),
		"--max-messages", fmt.Sprint(count),
```

functional_java_interop_test.go
dnwe: Sure, added

functional_java_interop_test.go
dnwe: Updated

functional_java_interop_test.go
```diff
@@ -17,208 +18,266 @@ import (
...
+	for _, msg := range messages {
 		_, err := fmt.Fprintln(stdin, msg)
 		if err != nil {
 			stdin.Close()
-			_ = cmd.Wait()
-			require.NoError(t, err, "Failed to write message")
+			waitErr := cmd.Wait()
+			require.NoError(t, waitErr, "Java producer failed")
 		}
+		require.NoError(t, err)
```
edoardocomar, line 58: if `waitErr` is not nil the test failure will hide the previous `err` - what about
```
-                       require.NoError(t, waitErr, "Java producer failed")
+                       if waitErr != nil {
+                               err = fmt.Errorf("failed to write message: %w; Java producer failed: %w", err, waitErr)
+                       }
+                       require.NoError(t, err)
                }
-               require.NoError(t, err)
```

dnwe: Added

edoardocomar (approved):

edoardocomar:
LGTM thanks @dnwe

functional_java_interop_test.go
puellanivis: I mean, if we already know we’re going to fail the `NoError` just put a `t.Fatal` in directly?

PS: I would recommend just using `assert.NoError(t, waittErr)` instead?

PR #3424, opened 2026-01-11
fix(test): resolve FVT issues in Kafka v2.x interop tests

edoardocomar:
use --bootstrap-server from version 2.5.0
capture stderr for better debugging

### DIFF
--- a/functional_java_interop_test.go
+++ b/functional_java_interop_test.go
@@ -44,6 +44,20 @@ func produceWithJava(t *testing.T, topic string, codec CompressionCodec, message
 	stdin, err := cmd.StdinPipe()
 	require.NoError(t, err)
 
+	stderr, err := cmd.StderrPipe()
+	require.NoError(t, err)
+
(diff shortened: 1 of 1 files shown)


functional_java_interop_test.go
```diff
@@ -44,22 +44,42 @@ func produceWithJava(t *testing.T, topic string, codec CompressionCodec, message
...
 	require.NoError(t, err)
 
+	stderr, err := cmd.StderrPipe()
+	require.NoError(t, err)
+
+	var stderrOutput strings.Builder
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
```
puellanivis, line 54: For single goroutine synchronization, I would usually recommend just using a simple channel:
```go
collectionDone := make(chan struct{})
go func() {
	defer close(collectionDone)
	…
}()
…
<-collectionDone
```

Not like using a WaitGroup is wrong.

Issue #3136, opened 2025-04-02
Support for other metrics providers (rcrowley/go-metrics is no longer maintained)


villevaltonen:
#### Description

Sarama relies on [rcrowley/go-metrics](https://github.com/rcrowley/go-metrics) for producing metrics and this library has been archived. Are there any plans for adding first-hand support for other common observability frameworks like OpenTelemetry or Prometheus?

#### Additional context

- [rcrowley/go-metrics](https://github.com/rcrowley/go-metrics)

vitaliyf:
Please review.

trever-atr:
@burke @bai @wvanbergen @dnwe @eapache

sorry to bother you, but can anyone have a look at this?
our SBOM tools flag this as an issue.

go-metrics is unmaintained and archived, the author recommends the [OpenTelemetry Go SDK](https://opentelemetry.io/docs/languages/go/instrumentation/#metrics) as a replacement.

this was mentioned once before: https://github.com/IBM/sarama/issues/2832

dnwe:
It’s a tricky one because it’s not something we can easily do in a v1 release. We’d have to jump to /v2 to make the breaking change and we’ve had requests for both minimal native Prometheus client metrics and fuller Opentelemetry metrics so there isn’t an obvious best option right now

vitaliyf:
Not stale.

vitaliyf:
Not stale.

mvanhorn:
I think this can be addressed incrementally within v1, without breaking the existing public API.

The idea: introduce a MetricsProvider interface alongside the existing MetricRegistry field. The interface defines only the metric-type factory methods sarama actually uses (Counter, Histogram, Meter, Gauge). A default GoMetricsProvider wraps the current rcrowley/go-metrics behavior. A new optional Config.MetricsProvider field lets users swap in OTel, Prometheus, or any other backend.

When MetricsProvider is nil (the default), the existing MetricRegistry path is used unchanged. When set, MetricsProvider takes precedence. This keeps full backward compatibility while unblocking alternative backends.

Scope for an initial PR: interface definition, default wrapper, Config field, and tests. No bulk migration of existing callsites - that can happen incrementally in follow-up PRs.

I'll put together a PR if there's interest.

PR #3481, opened 2026-03-22
feat: add MetricsProvider interface for pluggable metrics backends

mvanhorn:
## Summary

Introduces a `MetricsProvider` interface that wraps the current `rcrowley/go-metrics` usage, providing an incremental path toward replacing the archived dependency without breaking existing users.

- New `MetricsProvider` interface with `MetricsCounter`, `MetricsGauge`, `MetricsHistogram`, and `MetricsMeter` sub-interfaces matching only the methods sarama actually calls
- `GoMetricsProvider` wrapper that preserves current go-metrics behavior
- New optional `Config.MetricsProvider` field - nil by default, preserving full backward compatibility
- Comprehensive test suite including interface contract tests, concurrent access safety, and a stub provider demonstrating custom backend integration

## Motivation

`rcrowley/go-metrics` is archived and unmaintained. The author recommends OpenTelemetry as a replacement. This interface layer enables users to plug in OTel, Prometheus, or any other backend without waiting for a v2 release.

The approach is purely additive - the existing `Config.MetricRegistry` field and all current behavior is untouched. When `MetricsProvider` is nil (the default), nothing changes. Internal callsites can be migrated incrementally in follow-up PRs.

## Testing

`go test -race ./...` passes. New tests cover:
- GoMetricsProvider wrapper (counter, gauge, histogram, meter, unregister)
- Custom stub provider implementing the interface
- Config validation with MetricsProvider set
- Concurrent access safety with race detector

Addresses #3136

### DIFF
--- a/config.go
+++ b/config.go
@@ -530,6 +530,7 @@ type Config struct {
 	// like OpenTelemetry or Prometheus without depending on rcrowley/go-metrics.
 	//
 	// Defaults to nil, which preserves the existing MetricRegistry behavior.
+	// Call sites must nil-check before use: if conf.MetricsProvider != nil { ... }.
 	MetricsProvider MetricsProvider
 }
 
--- a/config.go
+++ b/config.go
@@ -524,6 +524,13 @@ type Config struct {
 	// prior to starting Sarama.
 	// See Examples on how to use the metrics registry
 	MetricRegistry metrics.Registry
+
+	// MetricsProvider is an optional pluggable metrics backend. When non-nil it
+	// takes precedence over MetricRegistry, allowing integration with systems
+	// like OpenTelemetry or Prometheus without depending on rcrowley/go-metrics.
+	//
+	// Defaults to nil, which preserves the existing MetricRegistry behavior.
+	MetricsProvider MetricsProvider
 }
 
 // NewConfig returns a new configuration instance with sane defaults.
--- a/metrics_provider.go
+++ b/metrics_provider.go
@@ -1,0 +1,85 @@
+package sarama
+
+import (
+	"github.com/rcrowley/go-metrics"
+)
+
+// MetricsProvider is an interface for pluggable metrics backends. Implementations
+// can bridge sarama's metrics to OpenTelemetry, Prometheus, or any other system.
+//
+// When Config.MetricsProvider is non-nil it takes precedence over Config.MetricRegistry.
+// The default value is nil, which preserves the existing rcrowley/go-metrics behavior.
+type MetricsProvider interface {
+	// NewCounter returns a named counter. Calling NewCounter with the same
+	// name should return the same instance.
+	NewCounter(name string) MetricsCounter
+
+	// NewGauge returns a named gauge.
+	NewGauge(name string) MetricsGauge
+
+	// NewHistogram returns a named histogram.
+	NewHistogram(name string) MetricsHistogram
+
+	// NewMeter returns a named meter.
+	NewMeter(name string) MetricsMeter
+
+	// UnregisterAll removes all metrics that have been registered.
+	UnregisterAll()
+}
+
+// MetricsCounter tracks a monotonically increasing value.
+type MetricsCounter interface {
+	Inc(int64)
+	Count() int64
+}
+
+// MetricsGauge tracks a single int64 value that can go up and down.
+type MetricsGauge interface {
+	Update(int64)
+	Value() int64
+}
+
+// MetricsHistogram tracks the distribution of a stream of int64 values.
+type MetricsHistogram interface {
+	Update(int64)
+	Count() int64
+}
+
+// MetricsMeter tracks the rate of events over time.
+type MetricsMeter interface {
+	Mark(int64)
+	Count() int64
+}
+
+// GoMetricsProvider wraps a rcrowley/go-metrics Registry to implement MetricsProvider.
+// This is the default implementation used when Config.MetricsProvider is nil.
+type GoMetricsProvider struct {
+	registry metrics.Registry
+}
+
+// NewGoMetricsProvider returns a MetricsProvider backed by the given go-metrics Registry.
+func NewGoMetricsProvider(registry metrics.Registry) *GoMetricsProvider {
+	return &GoMetricsProvider{registry: registry}
+}
+
+func (p *GoMetricsProvider) NewCounter(name string) MetricsCounter {
+	return metrics.GetOrRegisterCounter(name, p.registry)
+}
+
+func (p *GoMetricsProvider) NewGauge(name string) MetricsGauge {
+	return metrics.GetOrRegisterGauge(name, p.registry)
+}
+
+func (p *GoMetricsProvider) NewHistogram(name string) MetricsHistogram {
+	return p.registry.GetOrRegister(name, func() metrics.Histogram {
+		return metrics.NewHistogram(metrics.NewExpDecaySample(metricsReservoirSize, metricsAlphaFactor))
+	}).(metrics.Histogram)
+}
+
+func (p *GoMetricsProvider) NewMeter(name string) MetricsMeter {
+	return metrics.GetOrRegisterMeter(name, p.registry)
+}
+
+func (p *GoMetricsProvider) UnregisterAll() {
+	p.registry.UnregisterAll()
+}
--- a/metrics_provider.go
+++ b/metrics_provider.go
@@ -1,6 +1,8 @@
 package sarama
 
 import (
+	"sync"
+
 	"github.com/rcrowley/go-metrics"
 )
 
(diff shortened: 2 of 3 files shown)


metrics_provider.go
```diff
@@ -0,0 +1,85 @@
...
+func (p *GoMetricsProvider) NewCounter(name string) MetricsCounter {
+	return metrics.GetOrRegisterCounter(name, p.registry)
+}
+
+func (p *GoMetricsProvider) NewGauge(name string) MetricsGauge {
+	return metrics.GetOrRegisterGauge(name, p.registry)
+}
+
+func (p *GoMetricsProvider) NewHistogram(name string) MetricsHistogram {
+	return p.registry.GetOrRegister(name, func() metrics.Histogram {
```
dnwe, line 96: Not sure why Histogram one is p.registry.GetOrregister when the others are metrics.GetOrRegister which handles the nil case and won’t panic, was that a limitation in go-metrics or just a mistake here?

metrics_provider.go
```diff
@@ -0,0 +1,85 @@
...
+		return metrics.NewHistogram(metrics.NewExpDecaySample(metricsReservoirSize, metricsAlphaFactor))
+	}).(metrics.Histogram)
+}
+
+func (p *GoMetricsProvider) NewMeter(name string) MetricsMeter {
+	return metrics.GetOrRegisterMeter(name, p.registry)
+}
+
+func (p *GoMetricsProvider) UnregisterAll() {
+	p.registry.UnregisterAll()
```
dnwe: I wonder if we want to track which metrics we ourselves added and only unregister those, like we do with cleanupRegistry type today?

dnwe (commented):
Thanks for proposing this. Looks good overall, I’ve raised a couple of comments.

Currently this isn’t wired up in the metric call sites, were you planning to do that under this PR or with a follow up?

mvanhorn:
Thanks for the review, @dnwe.

To answer your question - the call site wiring is planned as a follow-up PR. This one establishes the interface and wrapper so we can migrate incrementally without a big-bang change.

Addressed both inline comments in 5077b34:

**Histogram (line 74):** go-metrics' `GetOrRegisterHistogram` requires a `Sample` parameter, so there's no zero-arg convenience function. The pattern matches sarama's existing `getOrRegisterHistogram` helper in `metrics.go` which uses `r.GetOrRegister(name, func() ...)` for the same reason. Left as-is since it's consistent with the codebase.

**UnregisterAll (line 84):** Fixed. `GoMetricsProvider` now tracks which metrics it registered and only unregisters those, matching the `cleanupRegistry` pattern. Added a test verifying that externally-registered metrics survive `UnregisterAll`.

puellanivis:
The second commit in the PR has: `Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>`

This is inconsistent with our https://github.com/IBM/sarama/blob/main/CONTRIBUTING.md#ai-assistance contributing guidelines.

PS: According to _Thaler v. Perlmutter_, under US law, AI cannot be an author in a joint work, because they cannot form legal intent, and this point is central to why they should not be purported as authors, or even as co-authors.

metrics_provider.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// GoMetricsProvider wraps a rcrowley/go-metrics Registry to implement MetricsProvider.
+// This is the default implementation used when Config.MetricsProvider is nil.
+// It tracks which metrics it registers so UnregisterAll only removes its own,
+// matching the cleanupRegistry pattern used elsewhere in sarama.
+type GoMetricsProvider struct {
+	registry metrics.Registry
+	names    map[string]struct{}
+	mu       sync.Mutex
```
puellanivis: Mutexes should head a paragraph of the values that it guards:
```go
registry metrics.Registry

mu    sync.Mutex
names map[string]struct{}
```

metrics_provider.go
```diff
@@ -0,0 +1,109 @@
...
+	return &GoMetricsProvider{
+		registry: registry,
+		names:    make(map[string]struct{}),
+	}
+}
+
+func (p *GoMetricsProvider) track(name string) {
+	p.mu.Lock()
+	p.names[name] = struct{}{}
+	p.mu.Unlock()
```
puellanivis: `defer` statements are now inlined, so there’s no reason to not use them:
```go
p.mu.Lock()
defer p.mu.Unlock()

if p.names == nil {
  // We’re already doing a full synchronous locking. Performing on-the-fly allocation is trivial compared.
  p.names = make(map[string]struct{})
}

p.names[name] = struct{}{}
```

metrics_provider.go
puellanivis: This is essentially already what it does internally: https://github.com/rcrowley/go-metrics/blob/65e299d6c5c9/histogram.go#L22-L27 Though, it would seem that the original could call into `NewExpDecaySample` every time… which is probably a bit of a misstep in the `go-metrics` library.

metrics_provider.go
```diff
@@ -0,0 +1,109 @@
...
+	return metrics.GetOrRegisterMeter(name, p.registry)
+}
+
+func (p *GoMetricsProvider) UnregisterAll() {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	for name := range p.names {
+		p.registry.Unregister(name)
+	}
+	p.names = make(map[string]struct{})
```
puellanivis: Why not `delete(p.names, name)` in the loop?

metrics_provider.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+}
+
+func (p *GoMetricsProvider) track(name string) {
+	p.mu.Lock()
+	p.names[name] = struct{}{}
+	p.mu.Unlock()
+}
+
+func (p *GoMetricsProvider) NewCounter(name string) MetricsCounter {
```
puellanivis: This isn’t a `NewCounter`… it’s a `GetOrNewCounter` right?

We might as well settle on `GetMetricType` as a name for these. Whether it’s allocating a new one under the hood is, essentially, an implementation detail.

config.go
```diff
@@ -524,6 +524,13 @@ type Config struct {
 	// prior to starting Sarama.
 	// See Examples on how to use the metrics registry
 	MetricRegistry metrics.Registry
+
+	// MetricsProvider is an optional pluggable metrics backend. When non-nil it
+	// takes precedence over MetricRegistry, allowing integration with systems
+	// like OpenTelemetry or Prometheus without depending on rcrowley/go-metrics.
+	//
+	// Defaults to nil, which preserves the existing MetricRegistry behavior.
```
puellanivis, line 532: … Are we sure about this assertion? `MetricsProvider(nil).NewCounter("foo")` will panic with a nil-pointer deref.

Every use of this field is going to require us to use a `if conf.MetricsProvider == nil  { … } else { … }` nil-guard before we can use it at all.

mvanhorn:
Removed the Co-Authored-By trailer in 00cc9cc. I missed that section of the contributing guide -- won't repeat it.

metrics_provider.go
mvanhorn: Addressed in a1d7f8c:
- Renamed `New{Counter,Gauge,Histogram,Meter}` to `Get{...}` on both interface and implementation
- Mutex now heads the guarded-values paragraph
- `track()` uses defer and handles lazy init
- `UnregisterAll` uses `delete()` in the loop
- Added nil-check documentation on `Config.MetricsProvider`

Re: the Histogram `GetOrRegister` (line 92) -- agreed it's a quirk of go-metrics. The wrapper matches the existing `getOrRegisterHistogram` helper in `metrics.go` so behavior stays consistent.

puellanivis: 👍 Yeah, I’m solid behind “I guess this is the necessary workaround.”

metrics_provider.go
```diff
@@ -0,0 +1,114 @@
...
+
+	mu    sync.Mutex
+	names map[string]struct{}
+}
+
+// NewGoMetricsProvider returns a MetricsProvider backed by the given go-metrics Registry.
+func NewGoMetricsProvider(registry metrics.Registry) *GoMetricsProvider {
+	return &GoMetricsProvider{
+		registry: registry,
+		names:    make(map[string]struct{}),
```
puellanivis: Now that we have lazy initialization, we can remove this line.

puellanivis (commented):
Just the one point.

metrics_provider.go
mvanhorn: Removed in 3527aba. Also rebased to add DCO sign-off on all commits.

Verified: `go test -race` passes — `TestGoMetricsProvider*` and `TestStubMetricsProvider` all green. `UnregisterAll` before any `track()` call is safe since range over nil map is a no-op in Go.

puellanivis (commented):
I don’t really see anything wrong with the code anymore, but I also wonder how useful it is that we’ve replaced go-metrics with interfaces that back in go-metrics’s APIs.

That is, this builds in a leverage point to replace go-metrics, but whatever replaces it, is going to have to just contort itself to fit the go-metrics API anyways.

Usually, the recommendation is to build on concrete types, and if we need to upgrade, write a refactor.

Nothing on the author or the design, this is pervasive in CompSci, when there’s only one implementation, it can be difficult to build a true abstraction layer that suits multiple implementations, and not one that just mirrors the only implementation.

But as said, the code is fine.

mvanhorn:
Fair point. The interface does mirror go-metrics closely, and you're right that a replacement backend would still need to map to those semantics. The main value I was targeting is testability - being able to swap in a no-op or recording implementation in tests without importing go-metrics. But I take your point that building a true abstraction ahead of a second implementation risks just mirroring the first.

If you'd rather keep concrete types and refactor when there's a real second backend, I'm happy to simplify. Otherwise the code is ready as-is.

puellanivis:
Agree. I think the code is ready as is. I’ll leave it to the maintainers (rather than me just a collaborator) to make a decision on this, if they’re good with the design decision proposed.

mvanhorn:
Thanks @puellanivis for the thorough review. @dnwe, whenever you get a chance to weigh in on the design approach, happy to adjust if needed.

dnwe:
I spent some more time thinking about this, and I think @puellanivis is correct here - assuming that we'd eventually support (for example) three possible backends of go-metrics, prometheus and opentelemetry, they have fundamentally different models for labels/attributes, meter vs query time rate, lifecycle etc. and if we don't reconcile those at the interface level then its going to be ugly making those work

I think for a pluggable interface we need to make it a purely _recording_ API, not a _reading_ API. So the interface defines how Sarama should report its metrics and in what shape, and then the various adapters will put that into the right native format for each backend. Currently in this PR we have the go-metrics shaped reads encoded.

As part of that change we'd need to make attributes/labels a first-class parameter. Today sarama builds names like "request-rate-for-broker-3-for-topic-foo". That's a workaround for go-metrics having no labels. Prometheus and OTel can't undo that, so end up with high-cardinality metric names instead of label sets.

mvanhorn:
@dnwe You're right. The current interface encodes go-metrics' read API. Translating names like `request-rate-for-broker-3-for-topic-foo` into Prometheus/OTel labels at the adapter layer is the high-cardinality footgun you'd expect.

Recording-only with first-class labels is a bigger change than this PR allows - every internal callsite touched, names split into `(name, attrs)` pairs. If that's the direction you want, the cleanest path is closing this and opening an RFC issue to design the interface, then migrating callsites in stages. I can drive that work if it fits sarama's roadmap, or close this and leave the design to whoever picks up #3136 next. Let me know which is useful.

PR #3485, opened 2026-03-28
fix: update API version URL as previous link was not working

MohishKhadse55:
While reviewing the Sarama library to understand its structure and behavior, 

I noticed that the referenced URL (https://kafka.apache.org/26/protocol.html#api_versions) was no longer working (returns 404), indicating it may be deprecated or archived. I compared it with the updated documentation and replaced it with ( https://kafka.apache.org/42/design/protocol/#retrieving-supported-api-versions) .

 Please let me know if this aligns with the requirements.


MohishKhadse55:
Wondering why changing comment is failing in this tests

hindessm (commented):
I re-ran the failing tests and they passed.

The new URL seems like a good choice to me. Thanks.

dnwe (approved):

PR #3486, opened 2026-03-30
fix(consumer): avoid broker race in response feeder

DCjanus:
## Background

A previous CI run in [PR #3419](https://github.com/IBM/sarama/pull/3419) exposed a `-race` failure in [`Unit Testing with Go oldstable`](https://github.com/IBM/sarama/actions/runs/23472786885/job/68298959022?pr=3419) caused by concurrent access to `partitionConsumer.broker`.

## Root Cause

`responseFeeder()` reads `child.broker`, but `dispatcher()` can change the same field during abort or redispatch. That is the race reported by `-race`.

The response path does not really need the current value of `child.broker`. It only needs to know which `brokerConsumer` produced the response that is being handled.

## Fix Approach

Pass that `brokerConsumer` together with the feeder response.

This removes the read of `child.broker` from `responseFeeder()` and keeps the ack / resubscribe path tied to the broker that produced the in-flight response.

## Testing

This PR was validated in three commits.

The first commit only adds the regression test plus a temporary workflow that runs just `TestPartitionConsumerBrokerRace` under `-race`, and is expected to fail. The second commit only contains the fix and is expected to make that workflow pass. The third commit removes the temporary workflow after the failure and recovery have both been demonstrated.

This sequence is intended to prove two things explicitly:

- the new regression test really detects the race in CI
- the follow-up fix really removes the race that the new test exercises

Relevant CI jobs:

- original motivating failure: [`Unit Testing with Go oldstable`](https://github.com/IBM/sarama/actions/runs/23472786885/job/68298959022?pr=3419) from [PR #3419](https://github.com/IBM/sarama/pull/3419)
- failing job: [Consumer Race Repro](https://github.com/IBM/sarama/actions/runs/23807818972/job/69386165640)
- passing job: [Consumer Race Repro](https://github.com/IBM/sarama/actions/runs/23807947844/job/69386644679)


DCjanus:
Leaving myself a short note for the actual fix later:

The race is on `partitionConsumer.broker`: `dispatcher()` writes it while `responseFeeder()` reads it. The plan is to switch that field over to atomic pointer access and then replace the direct reads/writes with `Load()` / `Store()`.

The main places to touch are the `partitionConsumer` field itself, `dispatch()`, `dispatcher()`, `responseFeeder()`, and any other `child.broker` access such as the logging paths in `parseResponse()`.

After that, rerun:
- `GOTOOLCHAIN=go1.25.8 go test -race -run ^TestPartitionConsumerBrokerRace .`
- `GOTOOLCHAIN=go1.25.8 go test -race -run ^TestConsumerExpiryTicker .`
- `GOTOOLCHAIN=go1.25.8 go test -race -run ^TestConsumerRebalancingMultiplePartitions .`

If that all looks good, the temporary `Consumer Race Repro` workflow/check should be removed before merge.

puellanivis (approved):

dnwe (approved):
Thanks! LGTM too

PR #3487, opened 2026-03-31
fix: add Unwrap() to DescribeConfigError and AlterConfigError

ShinThirty:
## Summary
`TopicError` and `TopicPartitionError` implement `Unwrap()` returning their `KError` field, enabling `errors.Is`/`errors.As` to traverse the error chain. However, `DescribeConfigError` and `AlterConfigError` have the same structure (`Err KError` + `ErrMsg`) but were missing `Unwrap()`, making `errors.Is`/`errors.As` unable to match the underlying `KError`.

This adds the missing `Unwrap()` methods for consistency, allowing callers to use `errors.Is(err, sarama.ErrInvalidConfig)` instead of type-asserting to extract the error code.

## Changes
- Add `Unwrap() error` to `*DescribeConfigError` (`describe_configs_response.go`)
- Add `Unwrap() error` to `*AlterConfigError` (`alter_configs_response.go`)
- Add tests for both, following the existing `TestTopicError` pattern

## Test plan
- [x] `go test -run 'TestDescribeConfigError|TestAlterConfigError|TestTopicError'` — all pass
- [x] `go vet ./...` — clean

### DIFF
--- a/alter_configs_response.go
+++ b/alter_configs_response.go
@@ -29,6 +29,10 @@ func (c *AlterConfigError) Error() string {
 	return text
 }
 
+func (c *AlterConfigError) Unwrap() error {
+	return c.Err
+}
+
 // AlterConfigsResourceResponse is a response type for alter config resource
 type AlterConfigsResourceResponse struct {
 	ErrorCode int16
--- a/alter_configs_response_test.go
+++ b/alter_configs_response_test.go
@@ -3,6 +3,7 @@
 package sarama
 
 import (
+	"errors"
 	"testing"
 )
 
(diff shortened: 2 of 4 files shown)


puellanivis (commented):
🤷‍♀️ I see nothing wrong with the code. The only concern here is if this is desired behavior.

ShinThirty:
I don't think we documented the expected behavior of the errors? Also this is a pure addition so it is backward compatible.

puellanivis:
Correct. My point is more: as a collaborator who does code reviews, nothing is wrong with the code.

My point is that it’s on the maintainers to decide if they’re good with the behavior change, which I agree is an improvement.

puellanivis (approved):

PR #3488, opened 2026-04-01
fix(consumer): model redispatch subscription lifecycle

DCjanus:
## Why

`#3486` removed the response-feeder race by carrying the broker that produced each fetch response. The remaining shutdown issue is a separate lifecycle problem: redispatch can abandon a broker assignment while the partition dispatcher is waiting to decide whether shutdown can finish.

The previous draft synchronized `child.broker`, but that made a shared broker pointer act as the source of truth for pending, active, and released subscriptions. That was hard to reason about and made abort/redispatch handoff depend on implicit state.

## What

- Model each broker assignment as a `brokerSubscription` token with exactly-once release.
- Pass subscription tokens through the broker input queue, pending batch, active subscription map, and feeder response path.
- Let redispatch, abort, timeout requeue, and dispatcher shutdown all observe the same assignment lifecycle instead of synchronizing on `child.broker`.
- Stop closing the broker input channel on final unref; close `stop` instead so late queue attempts can fail without racing a send on a closed channel.

## Validation

- `go test ./... -run 'TestPartitionConsumerResponseUsesFeederSubscription|TestPartitionConsumerAsyncCloseAfterRedispatchExitsDispatcher|TestConsumerAbortNoGoroutineLeak|TestBrokerSubscriptionReleaseIsIdempotent' -count=1`

## Notes

This is still intentionally a draft. The goal of this update is to make the redispatch/shutdown ownership model easier to review before running the full validation matrix.


DCjanus:
Temporarily closing this PR to reduce review noise while we wait for #3486 to merge first. This is not abandoned; we will reopen or resubmit it once #3486 is merged and continue from there.

DCjanus:
Latest update in `f1beaec` keeps `child.broker` publication inside `queueSubscription()` under `child.brokerMu`, so a queued subscription cannot become visible to abort handling before the matching broker ownership is visible too.

Keeping this PR in draft while I evaluate whether a follow-up cleanup around pending subscriptions is still warranted.

consumer.go
```diff
@@ -409,6 +409,7 @@ type partitionConsumer struct {
 	consumer *consumer
 	conf     *Config
 	broker   *brokerConsumer
+	brokerMu sync.RWMutex
```
puellanivis, line 412: Convention is usually to put everything protected by a mutex in a separate paragraph, with the mutex that protects them at the top of it.

consumer.go
```diff
@@ -494,6 +495,26 @@ func (child *partitionConsumer) stopDispatcher() {
...
+	return child.broker
+}
+
+func (child *partitionConsumer) releaseBrokerConsumer(expected *brokerConsumer) {
+	child.brokerMu.Lock()
+	current := child.broker
+	if current == nil || (expected != nil && current != expected) {
+		child.brokerMu.Unlock()
+		return
+	}
```
puellanivis, line 511: 🤔 I keep coming back to this. It feels overly complicated. Looking at it, and working it, I can see it works out to something like:
```go
if expected == nil {
  expected = child.broker.Swap(nil)
  if expected == nil {
    return
  }
} else {
  if !child.broker.CompareAndSwap(expected, nil) {
    // the only way this swaps is if expected was equal to child.broker,
    // so we don't need to care about loading the child.broker value,
    // we already have it.
    return
  }
}

child.consumer.unrefBrokerConsumer(expected)
```

I’m not sure what the tradeoff would be between mutexes vs `atomic.Pointer` though.

DCjanus:
Thanks for the help here, @puellanivis. I think I put this PR up too early, before I was really happy with the design, so sorry for taking up your review time on it.

I'm going to step back and think through the approach a bit more before I bring it back for review. For some personal reasons, I may not have much time to spend on this in the short term.

If anyone else wants to take this on and get a fix in sooner, I'd be very happy to see that.

puellanivis:
Don’t worry. If I didn’t want to spend the time on the review, I wouldn’t have done it. 😂 (I did see it is a WIP draft PR.)

DCjanus:
Closing this draft now that #3524 has landed.

The current shutdown/subscription ownership model is still hard to reason about, but cleaning it up properly would likely require a larger refactor than this PR should carry. I do not have a clear enough approach that would justify keeping this draft open, so I am going to close it for now.

PR #3492, opened 2026-04-10
fix: stop dispatcher for dying children in brokerConsumer.abort()

### LINKED ISSUE #3491: Bug: partition consumer dispatcher livelock after broker abort during shutdown causes offset commits to hang
### Summary

Commit 97621ecc68a89df41a52c5b1867982549ec00d75 ("fix: prevent goroutine leak and panic in brokerConsumer.abort()") introduced a regression where a broker error occurring while partition consumers are shutting down (e.g. during a consumer group rebalance) causes the partition consumer dispatcher goroutine to livelock. This prevents the consumer group session from releasing, which blocks offset commits indefinitely.

This is a regression from `v1.47.0`.

### Root Cause

The commit rewrote `brokerConsumer.abort()` to call `notifyError()` on each subscribed partition consumer instead of directly sending to `child.trigger`. The problem is that `notifyError()` silently skips dying children:

```go
// consumer.go:446
func (child *partitionConsumer) notifyError(err error) {
    // ...
    select {
    case <-child.dying:
        // shutdown is already in progress; skip error delivery
        return   // ← dying children get NO trigger, NO stopDispatcher
    default:
    }
    // ...
    child.triggerRedispatch()
}
```

Meanwhile, the new dispatcher (`consumer.go:500`) requires either a `trigger` signal or `dispatcherStop` to be closed in order to exit:

```go
// consumer.go:509
for {
    select {
    case <-child.dispatcherStop:
        return
    case <-child.trigger:       // ← blocks forever: nobody sends
    }
    // ...
}
```

And `abort()` (`consumer.go:1168`) iterates children calling only `notifyError`:

```go
func (bc *brokerConsumer) abort(err error) {
    bc.consumer.abandonBrokerConsumer(bc)
    bc.stopConsuming()
    _ = bc.broker.Close()

    for child := range bc.subscriptions {
        child.notifyError(err)    // skips dying children entirely
    }
    // ...
}
```

No code path calls `child.stopDispatcher()` for dying children during abort. The brokerConsumer's goroutines (`subscriptionManager`, `subscriptionConsumer`) have already exited, so the checks for `child.dying` in `updateSubscriptions`, `handleResponses`, `fetchNewMessages`, and `subscriptionConsumer` will never run again.

### Comparison with v1.47.0 behavior

In v1.47.0, `abort()` unconditionally sent `child.trigger <- none{}` for **all** children regardless of dying state:

```go
// v1.47.0 abort()
for child := range bc.subscriptions {
    child.sendError(err)
    child.trigger <- none{}   // always sent, even for dying children
}
```

And the v1.47.0 dispatcher would detect the dying state and self-terminate:

```go
// v1.47.0 dispatcher()
for range child.trigger {
    select {
    case <-child.dying:
        close(child.trigger)  // terminates the for-range loop → dispatcher exits
    // ...
    }
}
```

### Deadlock Chain

When a broker error occurs while partition consumers are shutting down (common during rebalance):

1. `brokerConsumer.abort()` calls `notifyError()` on a dying child → **returns immediately, does nothing**
2. The dispatcher is stuck at `case <-child.trigger:` — no trigger will ever arrive, `dispatcherStop` will never close
3. Dispatcher never exits → `close(child.feeder)` never called (deferred in dispatcher)
4. `responseFeeder` never exits → `close(child.errors)` never called
5. `consumerGroupClaim.waitClosed()` blocks forever waiting for the errors channel to close
6. `consumerGroupSession.consume()` never returns → `waitGroup.Done()` never called
7. `consumerGroupSession.release()` blocks at `waitGroup.Wait()` → never reaches `s.offsets.Close()`
8. **Offset commits never happen; consumer group hangs**

### Reproduction Scenario

1. Start a consumer group consuming from multiple partitions
2. Trigger a broker disconnect or network error while a rebalance is in progress (so `AsyncClose()` has been called on partition consumers, closing `child.dying`)
3. `brokerConsumer.subscriptionConsumer` detects the fetch error and calls `abort()`
4. `abort()` → `notifyError()` → early return for all dying children
5. Consumer group session hangs; no further offset commits occur

### Suggested Fix

In `abort()`, call `child.stopDispatcher()` for dying children so the dispatcher can exit:

```go
func (bc *brokerConsumer) abort(err error) {
    bc.consumer.abandonBrokerConsumer(bc)
    bc.stopConsuming()
    _ = bc.broker.Close()

    for child := range bc.subscriptions {
        select {
        case <-child.dying:
            child.stopDispatcher()
        default:
            child.notifyError(err)
        }
    }

    for newSubscriptions := range bc.newSubscriptions {
        for _, child := range newSubscriptions {
            select {
            case <-child.dying:
                child.stopDispatcher()
            default:
                child.notifyError(err)
            }
        }
    }
}
```

### Affected Versions

Introduced after `v1.47.0` by commit `97621ec`. Present on current `main`.  This has been observed in the wild by @honeycombio, and is not a theoretical bug.

lizthegrey:
## Summary

- Fix dispatcher livelock when `brokerConsumer.abort()` is called while partition consumers are dying (e.g. during rebalance)
- Call `stopDispatcher()` for dying children in `abort()`, matching the pattern used in every other subscription iteration (`subscriptionConsumer`, `updateSubscriptions`, `handleResponses`, `fetchNewMessages`)
- Add test verifying the dispatcher actually exits when abort runs on a dying child

Fixes #3491

## Test plan

- [x] `go test -race -run TestConsumerAbortNoGoroutineLeak -v -count=1` passes (all 5 subtests including new `stops_dispatcher_for_dying_child`)
- [x] `go vet ./...` clean
- [x] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -1171,12 +1171,22 @@ func (bc *brokerConsumer) abort(err error) {
 	_ = bc.broker.Close() // we don't care about the error this might return, we already have one
 
 	for child := range bc.subscriptions {
-		child.notifyError(err)
+		select {
+		case <-child.dying:
+			child.stopDispatcher()
+		default:
+			child.notifyError(err)
+		}
 	}
 
 	for newSubscriptions := range bc.newSubscriptions {
 		for _, child := range newSubscriptions {
-			child.notifyError(err)
+			select {
+			case <-child.dying:
+				child.stopDispatcher()
+			default:
+				child.notifyError(err)
+			}
 		}
 	}
 }
--- a/consumer.go
+++ b/consumer.go
@@ -1177,12 +1177,22 @@ func (bc *brokerConsumer) abort(err error) {
 	_ = bc.broker.Close() // we don't care about the error this might return, we already have one
 
 	for child := range bc.subscriptions {
-		child.notifyError(err)
+		select {
+		case <-child.dying:
+			child.stopDispatcher()
+		default:
+			child.notifyError(err)
+		}
 	}
 
 	for newSubscriptions := range bc.newSubscriptions {
 		for _, child := range newSubscriptions {
-			child.notifyError(err)
+			select {
+			case <-child.dying:
+				child.stopDispatcher()
+			default:
+				child.notifyError(err)
+			}
 		}
 	}
 }
--- a/consumer_test.go
+++ b/consumer_test.go
@@ -2328,6 +2328,34 @@ func TestConsumerAbortNoGoroutineLeak(t *testing.T) {
 		runAbort(t, newBrokerConsumer(child))
 	})
 
+	t.Run("stops dispatcher for dying child", func(t *testing.T) {
+		child := newChild(config.ChannelBufferSize)
+		child.consumer = &consumer{
+			client:          client,
+			conf:            config,
+			children:        make(map[string]map[int32]*partitionConsumer),
+			brokerConsumers: make(map[*Broker]*brokerConsumer),
+			metricRegistry:  newCleanupRegistry(config.MetricRegistry),
+		}
+
+		// Start the dispatcher goroutine (it will block on <-child.trigger).
+		go child.dispatcher()
+
+		// Mark the child as dying (simulates AsyncClose during rebalance).
+		close(child.dying)
+
+		bc := newBrokerConsumer(child)
+		runAbort(t, bc)
+
+		// The dispatcher should have exited via stopDispatcher, closing feeder.
+		select {
+		case _, ok := <-child.feeder:
+			require.False(t, ok, "feeder channel should be closed after dispatcher exits")
+		case <-time.After(5 * time.Second):
+			require.FailNow(t, "dispatcher did not exit after abort() on dying child")
+		}
+	})
+
 	t.Run("returns when redispatch is already pending", func(t *testing.T) {
 		child := newChild(config.ChannelBufferSize)
 		child.trigger <- none{}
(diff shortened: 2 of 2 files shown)


consumer.go
```diff
@@ -1177,12 +1177,22 @@ func (bc *brokerConsumer) abort(err error) {
 	_ = bc.broker.Close() // we don't care about the error this might return, we already have one
 
 	for child := range bc.subscriptions {
-		child.notifyError(err)
+		select {
+		case <-child.dying:
+			child.stopDispatcher()
+		default:
+			child.notifyError(err)
+		}
```
puellanivis, line 1185: This still has a race-condition in that `child.dying` is closed after the `default` path is taken, and before `notifyError(err)` is called.

I suppose that’s safe to do? Since we’re already doing it anyways. And the next call would then take care of `stopDispatcher` as a call.

I’m wondering why this is even necessary though… the dispatcher should already be listening to this signal already, and handling it?

I guess it’s on a separate `select` though, and the dispatcher could be waiting on `dispatcherStop` and `trigger` without listening to any `dying` signal… perhaps, we should put the fix there, and have the dispatcher properly handle upgrading from `dying` to dead.

puellanivis: 🤔 I guess this is right, and matches the other subscription iteration. So, I don’t see a reason to block the change.

But it still might be nice to extract out needing to call this `stopDispatcher` manually (error prone), and have the dispatcher properly handle dying → dead.

puellanivis (approved):

dnwe (approved):
LGTM Thanks!

lizthegrey:
> LGTM Thanks!

my pleasure!

PR #3497, opened 2026-04-18
fix: retry ListTopics on transient transport errors

### LINKED ISSUE #3125: ListTopics of clusterAdmin doesn't retry
##### Description

Other methods in ClusterAdmin appear to use the retryOnError mechanism, allowing them to retry upon failure (e.g., CreateTopic, DeleteTopic, CreatePartitions, etc.).
However, ListTopics does not implement any retry logic.

As a result, if the method fails due to reasons like the broker closing the connection (e.g., caused by connections.max.idle.ms), it does not retry. Instead, the client/broker seems to deregister and register again. Once a new broker is registered, the issue no longer occurs.

Is there a specific reason why ListTopics does not include retry logic? Was this behavior intentional?

##### Versions

| Sarama | Kafka | Go  |
| ------ | ----- | --- |
| 1.45.1 |  3.7 | 1.23.0 |

##### Configuration

```go

```

##### Logs

<p>

```
time="2025-03-17T17:14:43+09:00" level=debug msg="client/metadata fetching metadata for all topics from broker kafka-broker-test-0:19000\n" 
time="2025-03-17T17:14:43+09:00" level=debug msg="client/metadata got error from broker 0 while fetching metadata: EOF\n" 
time="2025-03-17T17:14:43+09:00" level=debug msg="Closed connection to broker kafka-broker-test-0:19000\n" 
time="2025-03-17T17:14:43+09:00" level=debug msg="client/brokers deregistered broker #0 at kafka-broker-test-0:19000" 
time="2025-03-17T17:14:43+09:00" level=debug msg="client/metadata fetching metadata for all topics from broker kafka-broker-test-1:19001\n" 
time="2025-03-17T17:14:43+09:00" level=debug msg="client/metadata got error from broker 1 while fetching metadata: write tcp 172.24.1.6:53158->10.161.82.44:19001: write: broken pipe\n" 
time="2025-03-17T17:14:43+09:00" level=debug msg="Error while closing connection to broker kafka-broker-test-1:19001: tls: failed to send closeNotify alert (but connection was closed anyway): write tcp 172.24.1.6:53158->10.161.82.44:19001: write: broken pipe\n" 
time="2025-03-17T17:14:43+09:00" level=debug msg="client/brokers deregistered broker #1 at kafka-broker-test-1:19001" 
time="2025-03-17T17:14:43+09:00" level=debug msg="client/metadata fetching metadata for all topics from broker kafka-broker-test-2:19002\n" 
time="2025-03-17T17:14:43+09:00" level=debug msg="client/brokers registered new broker #0 at kafka-broker-test-0:19000" 
time="2025-03-17T17:14:43+09:00" level=debug msg="client/brokers registered new broker #1 at kafka-broker-test-1:19001" 
time="2025-03-17T17:18:05+09:00" level=debug msg="client/metadata fetching metadata for all topics from broker kafka-broker-test-0:19000\n" 
time="2025-03-17T17:18:05+09:00" level=debug msg="Connected to broker at kafka-broker-test-0:19000 (registered as #0)\n" 
time="2025-03-17T17:23:16+09:00" level=debug msg="client/metadata fetching metadata for all topics from broker kafka-broker-test-0:19000\n" 
```

</p>
</details>

huynhanx03:
## Summary

`ClusterAdmin.ListTopics()` currently fails fast on transient connection failures (for example `EOF`, `broken pipe`, timeout) and does not retry like other admin paths.

This change adds retry behavior for `ListTopics` so temporary transport failures can recover automatically.

## Root Cause

`ListTopics` performs two sequential RPCs (`Metadata` then `DescribeConfigs`) but was not wrapped in `retryOnError`, so any transient network failure aborted the whole call immediately.

## Fix

- Wrap the full `ListTopics` flow in `retryOnError` (retry from the start).
- Add a retriable predicate for:
  - existing controller-retriable errors
  - transport-close errors
  - timeout network errors
- Add a regression test that forces one transient metadata request failure, then verifies `ListTopics` retries and succeeds.

## Why This Is Safe

- No behavior change on healthy paths.
- Scope is limited to `ListTopics`.
- Retrying the full flow keeps metadata/config data consistent per attempt.

## Testing

- `go test -run 'TestClusterAdminListTopics|TestClusterAdminListTopicsRetriesOnTransientConnectionError' -count=10 ./...`
- `go test -run TestClusterAdminListTopicsRetriesOnTransientConnectionError -count=20 ./...`
- `go test -run ClusterAdmin -count=1 ./...`

## Issue

Fixes #3125

### DIFF
--- a/admin_test.go
+++ b/admin_test.go
@@ -5,6 +5,7 @@ package sarama
 import (
 	"errors"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 )
@@ -167,20 +168,20 @@ func TestClusterAdminListTopics(t *testing.T) {
 	}
 
 	if len(entries) == 0 {
-		t.Fatal(errors.New("no resource present"))
+		t.Fatal("no resource present")
 	}
 
 	topic, found := entries["my_topic"]
 	if !found {
-		t.Fatal(errors.New("topic not found in response"))
+		t.Fatal("topic not found in response")
 	}
 	_, found = topic.ConfigEntries["max.message.bytes"]
 	if found {
-		t.Fatal(errors.New("default topic config entry incorrectly found in response"))
+		t.Fatal("default topic config entry incorrectly found in response")
 	}
 	value := topic.ConfigEntries["retention.ms"]
 	if value == nil || *value != "5000" {
-		t.Fatal(errors.New("non-default topic config entry not found in response"))
+		t.Fatal("non-default topic config entry not found in response")
 	}
 
 	err = admin.Close()
@@ -188,8 +189,154 @@ func TestClusterAdminListTopics(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if topic.ReplicaAssignment == nil || topic.ReplicaAssignment[0][0] != 1 {
-		t.Fatal(errors.New("replica assignment not found in response"))
+	assignment, found := topic.ReplicaAssignment[0]
+	if !found {
+		t.Fatal("replica assignment for partition 0 not found in response")
+	}
+	if len(assignment) == 0 {
+		t.Fatal("replica assignment for partition 0 was empty")
+	}
+	if assignment[0] != 1 {
+		t.Fatal("replica assignment not found in response")
+	}
+}
+
+func TestClusterAdminListTopicsRetriesOnTransientConnectionError(t *testing.T) {
+	seedBroker := NewMockBroker(t, 1)
+	defer seedBroker.Close()
+
+	metadataResponse := NewMockMetadataResponse(t).
+		SetController(seedBroker.BrokerID()).
+		SetBroker(seedBroker.Addr(), seedBroker.BrokerID()).
+		SetLeader("my_topic", 0, seedBroker.BrokerID())
+
+	var stateMu sync.Mutex
+	var injectTimeout, injected bool
+	var metadataAttempts int
+
+	seedBroker.SetHandlerFuncByMap(map[string]requestHandlerFunc{
+		"MetadataRequest": func(req *request) encoderWithHeader {
+			stateMu.Lock()
+			metadataAttempts++
+			shouldInject := injectTimeout && !injected
+			if shouldInject {
+				injected = true
+			}
+			stateMu.Unlock()
+			if shouldInject {
+				return nil
+			}
+			return metadataResponse.For(req.body)
+		},
+		"DescribeConfigsRequest": func(req *request) encoderWithHeader {
+			return NewMockDescribeConfigsResponse(t).For(req.body)
+		},
+	})
+
+	config := NewTestConfig()
+	config.Version = V1_1_0_0
+	config.Net.ReadTimeout = 100 * time.Millisecond
+	config.Admin.Retry.Max = 3
+	config.Admin.Retry.Backoff = 10 * time.Millisecond
+
+	admin, err := NewClusterAdmin([]string{seedBroker.Addr()}, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer safeClose(t, admin)
+
+	stateMu.Lock()
+	injectTimeout = true
+	metadataAttemptsBeforeList := metadataAttempts
+	stateMu.Unlock()
+
+	entries, err := admin.ListTopics()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	stateMu.Lock()
+	gotInjected := injected
+	gotMetadataAttempts := metadataAttempts
+	stateMu.Unlock()
+	if !gotInjected {
+		t.Fatal("expected metadata timeout to be injected during ListTopics")
+	}
+	if gotMetadataAttempts < metadataAttemptsBeforeList+2 {
+		t.Fatal("expected ListTopics to retry metadata after transient timeout")
+	}
+	if len(entries) == 0 {
+		t.Fatal("no resource present")
+	}
+}
+
+func TestClusterAdminListTopicsRetriesOnDescribeConfigsTimeout(t *testing.T) {
+	seedBroker := NewMockBroker(t, 1)
+	defer seedBroker.Close()
+
+	metadataResponse := NewMockMetadataResponse(t).
+		SetController(seedBroker.BrokerID()).
+		SetBroker(seedBroker.Addr(), seedBroker.BrokerID()).
+		SetLeader("my_topic", 0, seedBroker.BrokerID())
+
+	var stateMu sync.Mutex
+	var injectTimeout, injected bool
+	var describeConfigsAttempts int
+	describeConfigsResponse := NewMockDescribeConfigsResponse(t)
+
+	seedBroker.SetHandlerFuncByMap(map[string]requestHandlerFunc{
+		"MetadataRequest": func(req *request) encoderWithHeader {
+			return metadataResponse.For(req.body)
+		},
+		"DescribeConfigsRequest": func(req *request) encoderWithHeader {
+			stateMu.Lock()
+			describeConfigsAttempts++
+			shouldInject := injectTimeout && !injected
+			if shouldInject {
+				injected = true
+			}
+			stateMu.Unlock()
+			if shouldInject {
+				return nil
+			}
+			return describeConfigsResponse.For(req.body)
+		},
+	})
+
+	config := NewTestConfig()
+	config.Version = V1_1_0_0
+	config.Net.ReadTimeout = 100 * time.Millisecond
+	config.Admin.Retry.Max = 3
+	config.Admin.Retry.Backoff = 10 * time.Millisecond
+
+	admin, err := NewClusterAdmin([]string{seedBroker.Addr()}, config)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer safeClose(t, admin)
+
+	stateMu.Lock()
+	injectTimeout = true
+	describeConfigsAttemptsBeforeList := describeConfigsAttempts
+	stateMu.Unlock()
+
+	entries, err := admin.ListTopics()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	stateMu.Lock()
+	gotInjected := injected
+	gotDescribeConfigsAttempts := describeConfigsAttempts
+	stateMu.Unlock()
+	if !gotInjected {
+		t.Fatal("expected DescribeConfigs timeout to be injected during ListTopics")
+	}
+	if gotDescribeConfigsAttempts < describeConfigsAttemptsBeforeList+2 {
+		t.Fatal("expected ListTopics to retry DescribeConfigs after transient timeout")
+	}
+	if len(entries) == 0 {
+		t.Fatal("no resource present")
 	}
 }
 
--- a/admin.go
+++ b/admin.go
@@ -224,6 +224,21 @@ func isRetriableGroupCoordinatorError(err error) bool {
 	return errors.Is(err, ErrNotCoordinatorForConsumer) || errors.Is(err, ErrConsumerCoordinatorNotAvailable) || errors.Is(err, io.EOF)
 }
 
+// isRetriableListTopicsError returns true for controller errors and transient
+// transport failures where reconnecting and retrying can succeed.
+func isRetriableListTopicsError(err error) bool {
+	if isRetriableControllerError(err) || shouldCloseBrokerConn(err) {
+		return true
+	}
+
+	return isRetriableListTopicsTimeoutError(err)
+}
+
+func isRetriableListTopicsTimeoutError(err error) bool {
+	var netErr net.Error
+	return errors.As(err, &netErr) && netErr.Timeout()
+}
+
 // retryOnError will repeatedly call the given (error-returning) func in the
 // case that its response is non-nil and retryable (as determined by the
 // provided retryable func) up to the maximum number of tries permitted by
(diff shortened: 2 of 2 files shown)


admin_test.go
```diff
@@ -193,6 +194,64 @@ func TestClusterAdminListTopics(t *testing.T) {
...
+
+	if len(entries) == 0 {
+		t.Fatal(errors.New("no resource present"))
+	}
+
+	topic, found := entries["my_topic"]
+	if !found {
+		t.Fatal(errors.New("topic not found in response"))
+	}
+	if topic.ReplicaAssignment == nil || topic.ReplicaAssignment[0][0] != 1 {
```
puellanivis: `topic.ReplicaAssignment` is a `map`, so there’s no need to check if it’s `nil` before accessing any keys. A nil map is equivalent to an empty map for reads, which means, every key returns the zero value.

Unlike the map access though, the value of `ReplicaAssignment` is `[]int32`, which means `[0]` _can_ panic, if the slice has length 0.

```suggestion
	assignment := topic.ReplicaAssignment[0]
	if len(assignment) == 0 {
		t.Fatal("ReplicaAssignment[0] was empty")
	}
	if assignment[0] != 1 {
		t.Error("replica assignment was not found in response")
	}
```

puellanivis (commented):
Generally looks pretty fine.

huynhanx03:
Hi @puellanivis , thanks for catching that. You're right: the nil map check wasn't useful here, and the real risk was indexing into an empty assignment slice. I've updated the test to check for the partition entry first, then verify the slice is non-empty before asserting the replica ID.

admin_test.go
```diff
@@ -188,7 +189,79 @@ func TestClusterAdminListTopics(t *testing.T) {
...
-	if topic.ReplicaAssignment == nil || topic.ReplicaAssignment[0][0] != 1 {
+	assignment, found := topic.ReplicaAssignment[0]
+	if !found {
+		t.Fatal(errors.New("replica assignment for partition 0 not found in response"))
+	}
+	if len(assignment) == 0 {
+		t.Fatal(errors.New("replica assignment for partition 0 was empty"))
+	}
+	if assignment[0] != 1 {
+		t.Fatal(errors.New("replica assignment not found in response"))
```
puellanivis: The use of `errors.New` here is unnecessary. `t.Fatal` takes `...any`, and a simple string is appropriate, and formats identically.

admin_test.go
```diff
@@ -188,7 +189,79 @@ func TestClusterAdminListTopics(t *testing.T) {
...
+	assignment, found := topic.ReplicaAssignment[0]
+	if !found {
+		t.Fatal(errors.New("replica assignment for partition 0 not found in response"))
+	}
+	if len(assignment) == 0 {
+		t.Fatal(errors.New("replica assignment for partition 0 was empty"))
+	}
+	if assignment[0] != 1 {
 		t.Fatal(errors.New("replica assignment not found in response"))
 	}
```
puellanivis: Same as above, none of these need to be `errors.New`.

huynhanx03:
Hi @puellanivis , I’ve changed those to plain t.Fatal("...").

dnwe:
Hmm, with this PR the new `isRetriableListTopicsError` treats `net.Error.Timeout()` as retryable [here](https://github.com/huynhanx03/sarama/blob/1c31c2416c7e757585524e140ada962fce4dec82/admin.go#L229-L235). However, the retry body reuses the same broker and only calls `b.Open(...)`, ignoring `ErrAlreadyConnected` [here](https://github.com/huynhanx03/sarama/blob/1c31c2416c7e757585524e140ada962fce4dec82/admin.go#L436-L488)

I think this means timeout retries would keep using the same potentially broken connection as `sendAndReceive` only closes the broker when `shouldCloseBrokerConn(err)` [returns true](https://github.com/huynhanx03/sarama/blob/1c31c2416c7e757585524e140ada962fce4dec82/broker.go#L1132-L1149
), but `shouldCloseBrokerConn` explicitly does not close for [timeout]( https://github.com/huynhanx03/sarama/blob/1c31c2416c7e757585524e140ada962fce4dec82/broker.go#L1293-L1322) `net.Error`
 
Once `responseReceiver` hits a read error, it records that as `dead` and [returns the same error](https://github.com/huynhanx03/sarama/blob/1c31c2416c7e757585524e140ada962fce4dec82/broker.go#L1229-L1249) for later promises on that connection. Since `Open` returns `ErrAlreadyConnected` while `opened` is still true, the retry does not [establish a fresh connection](https://github.com/huynhanx03/sarama/blob/1c31c2416c7e757585524e140ada962fce4dec82/broker.go#L160-L168)

So it seems like the new timeout retry path wouldn't recover from the transient timeout case it is trying to handle?
  
I think we need to explicitly close/reopen the broker before retrying a timeout from `GetMetadata` / `DescribeConfigs`?

huynhanx03:
Hi @dnwe , I updated the timeout retry path so `ListTopics` now closes the broker before returning a timeout from both `GetMetadata` and `DescribeConfigs`, allowing the next retry to reopen a fresh connection instead of reusing the same dead one.

I also converted the previous retry test so the timeout is injected after admin initialization, meaning it now exercises the actual `ListTopics` path instead of the client bootstrap metadata refresh. Added coverage for the `DescribeConfigs` timeout case as well.

Tested with: `go test -run 'TestClusterAdminListTopics$|TestClusterAdminListTopicsRetriesOnTransientConnectionError|TestClusterAdminListTopicsRetriesOnDescribeConfigsTimeout' -count=1`

admin_test.go
```diff
@@ -167,29 +168,165 @@ func TestClusterAdminListTopics(t *testing.T) {
...
+	defer seedBroker.Close()
+
+	metadataResponse := NewMockMetadataResponse(t).
+		SetController(seedBroker.BrokerID()).
+		SetBroker(seedBroker.Addr(), seedBroker.BrokerID()).
+		SetLeader("my_topic", 0, seedBroker.BrokerID())
+
+	var metadataAttempts atomic.Int32
+	var injectMetadataTimeout atomic.Bool
+	var metadataTimeoutInjected atomic.Bool
```
puellanivis: Hm… I think the use of atomics here has gotten out of hand. While atomics are good at synchronizing single accesses, two atomic loads and stores do not happen all together atomically.

As a result, there’s a ton of race conditions between tracking number of attempts, and if an error should be injected or not.

The proper non-racey functionality of this code would be better done under a mutex, so that incrementing attempts, and testing for inject timeout can all happen entirely exclusive of setting the trigger to inject timeouts, and getting a baseline of how many times metadata attempts has been called.

huynhanx03:
Hi @puellanivis , I refactored this test to use a mutex-protected state instead of multiple atomics. The timeout injection state, attempt counters, and baseline reads are now synchronized under the same lock, so the test no longer relies on separate atomic operations happening in a particular order. I also cleaned up the retry coverage into a table-driven test for both `MetadataRequest` and `DescribeConfigsRequest` timeout paths.

`go test -run 'TestClusterAdminListTopics$|TestClusterAdminListTopicsRetriesAfterTransientTimeout' -count=1`

admin_test.go
```diff
@@ -167,29 +178,164 @@ func TestClusterAdminListTopics(t *testing.T) {
...
+			injectTimeout := false
+			injected := false
+			metadataAttempts := 0
+			describeConfigsAttempts := 0
+			describeConfigsResponse := NewMockDescribeConfigsResponse(t)
+
+			makeRequestHandler := func(kind adminRequestKind, attempts *int, response func(*request) encoderWithHeader) requestHandlerFunc {
+				return func(req *request) encoderWithHeader {
+					stateMu.Lock()
+					(*attempts)++
```
puellanivis: This function already has access to `metadataAttempts`, there is no need to pass this through a `*int` function parameter.

admin_test.go
```diff
@@ -167,29 +178,164 @@ func TestClusterAdminListTopics(t *testing.T) {
...
+				SetLeader("my_topic", 0, seedBroker.BrokerID())
+
+			var stateMu sync.Mutex
+			injectTimeout := false
+			injected := false
+			metadataAttempts := 0
+			describeConfigsAttempts := 0
+			describeConfigsResponse := NewMockDescribeConfigsResponse(t)
+
+			makeRequestHandler := func(kind adminRequestKind, attempts *int, response func(*request) encoderWithHeader) requestHandlerFunc {
```
puellanivis: `response` is a noun, and thus not well suited for naming a function argument. `respond` would be more appropriate.

admin_test.go
```diff
@@ -5,10 +5,21 @@ package sarama
 import (
 	"errors"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 )
 
+type adminRequestKind string
```
puellanivis: This type is unnecessary. It does not implement any receiver methods, and although there are constants established with this type, I suppose in an attempt to provide a sort of closed enum, that does not provide any type safety in Golang.

An untyped constant string can be assigned into any typed constant string without type conversion, and without syntax error. (Lines 16, and 17 both demonstrates this.)

I can use below `injectOnRequest: "foo",` and nothing will not compile.

admin_test.go
```diff
@@ -5,10 +5,21 @@ package sarama
...
+
+const (
+	metadataRequestKind        adminRequestKind = "MetadataRequest"
+	describeConfigsRequestKind adminRequestKind = "DescribeConfigsRequest"
+	// One injected timeout should force exactly one extra call to complete:
+	// failed call + successful retry.
+	minAttemptsIncreaseAfterInjectedTimeout = 2
+)
```
puellanivis: These constants should be defined within the function body that they are relevant to. (Semantic locality should be reflected in spatial locality.)

admin_test.go
```diff
@@ -5,10 +5,21 @@ package sarama
...
 )
 
+type adminRequestKind string
+
+const (
+	metadataRequestKind        adminRequestKind = "MetadataRequest"
+	describeConfigsRequestKind adminRequestKind = "DescribeConfigsRequest"
+	// One injected timeout should force exactly one extra call to complete:
+	// failed call + successful retry.
+	minAttemptsIncreaseAfterInjectedTimeout = 2
```
puellanivis: One should avoid defining two concepts of consts in one const body. It’s easy enough to specify a new `const` block, or just `const minAttemptsIncreaseAfterInjectedTimeout = 2̀  without a new block.

The idea here is that each definition block should have one conceptual purpose. Listing the request kinds is a group of related constants, and thus reasonable to group. But they’re a separate purpose to `minAttempts…` and so they should be separated syntactically. (Syntax should reaffirm semantics.)

admin_test.go
```diff
@@ -167,29 +178,164 @@ func TestClusterAdminListTopics(t *testing.T) {
...
+	}{
+		{
+			name:            "metadata request",
+			injectOnRequest: metadataRequestKind,
+		},
+		{
+			name:            "describe configs request",
+			injectOnRequest: describeConfigsRequestKind,
+		},
+	}
```
puellanivis: The name described here is merely a repetition of the `injectOnRequest`. Just get ride of it, and use the request kind string:
```suggestion
	tests := []string{
		"MetadataRequest",
		"DescribeConfigsRequest",
	}
```

admin_test.go
```diff
@@ -5,10 +5,21 @@ package sarama
...
+	"sync"
 	"testing"
 	"time"
 )
 
+type adminRequestKind string
+
+const (
+	metadataRequestKind        adminRequestKind = "MetadataRequest"
+	describeConfigsRequestKind adminRequestKind = "DescribeConfigsRequest"
```
puellanivis: These consts serve little purpose. Their value is literally a part of the identifier. This means the identifier is just a longer way to type the same string.

admin_test.go
```diff
@@ -167,29 +178,164 @@ func TestClusterAdminListTopics(t *testing.T) {
...
+			metadataResponse := NewMockMetadataResponse(t).
+				SetController(seedBroker.BrokerID()).
+				SetBroker(seedBroker.Addr(), seedBroker.BrokerID()).
+				SetLeader("my_topic", 0, seedBroker.BrokerID())
+
+			var stateMu sync.Mutex
+			injectTimeout := false
+			injected := false
+			metadataAttempts := 0
+			describeConfigsAttempts := 0
```
puellanivis: There is no need to assign zero values, this is the default initialization value:
```go
// Defining values in this way make explicit the important information: the type, not the irrelevant information: the zero value.
var injectTimeout, injected bool
var metadataAttempts, describeConfigsAttempts int
```

admin_test.go
```diff
@@ -167,29 +178,164 @@ func TestClusterAdminListTopics(t *testing.T) {
...
+					metadataRequestKind,
+					&metadataAttempts,
+					func(req *request) encoderWithHeader { return metadataResponse.For(req.body) },
+				),
+				string(describeConfigsRequestKind): makeRequestHandler(
+					describeConfigsRequestKind,
+					&describeConfigsAttempts,
+					func(req *request) encoderWithHeader { return describeConfigsResponse.For(req.body) },
+				),
+			})
```
puellanivis: I’m unsure why we’re needing to setup these two operations?

We’ve complicated everything by making both tests setup all the test harnesses for the other test at the same time.

In attempting to remove repeating ourselves, we’re instead complicating the tests, and making it less clear what the tests are specifically testing.

We are allowed to go ahead and repeat ourselves a little bit… as a treat.

huynhanx03:
Hi @puellanivis , I simplified the retry tests based on the feedback.

I removed the extra request-kind type, constants, table-driven wrapper, and shared handler helper. The retry coverage is now split into two explicit tests: one for `MetadataRequest` timeout and one for `DescribeConfigsRequest` timeout.

I also switched the test state to idiomatic zero-value declarations under a mutex, so the timeout injection and attempt counters remain synchronized without the extra abstraction.

puellanivis (commented):
Looks fine on the code side, I’m unsure if the actual logic fits in with the rest of the functionality. (That is, in the concerns about reusing brokers and under what conditions.)

dnwe (approved):
LGTM and good enough to merge now. Thanks for working on this and updating it based on the feedback 👍🏻

PR #3498, opened 2026-04-18
perf: cache topic batch-size metric lookup

huynhanx03:
# Overview

Optimize `ProduceRequest.encode` by caching the topic-level `batch-size` histogram once per topic, instead of looking it up for every partition.

# Changes

- Cache `topicBatchSizeMetric` once per topic in `ProduceRequest.encode`.
- Reuse cached metric in the partition loop.
- Add focused benchmarks:
  - `BenchmarkProduceRequestEncodeMetrics1Partition`
  - `BenchmarkProduceRequestEncodeMetrics32Partitions`
  - `BenchmarkProduceRequestEncodeMetrics128Partitions`

## Benchmark
```text
goos: darwin
goarch: arm64
cpu: Apple M1

name                                        old ns/op   new ns/op   delta
ProduceRequestEncodeMetrics1Partition       1977        2040        +3.2%
ProduceRequestEncodeMetrics32Partitions     26250       24600       -6.3%
ProduceRequestEncodeMetrics128Partitions    100070      92300       -7.8%
```

# Testing

- `go test -run 'TestProduceRequest|TestAllocateBodyProtocolVersions' .`
- `go test -run '^$' -bench 'BenchmarkProduceRequestEncodeMetrics(1Partition|32Partitions|128Partitions)$' -benchmem -count=8 .`

### DIFF
--- a/produce_request.go
+++ b/produce_request.go
@@ -102,8 +102,10 @@ func (r *ProduceRequest) encode(pe packetEncoder) error {
 		}
 		topicRecordCount := int64(0)
 		var topicCompressionRatioMetric metrics.Histogram
+		var topicBatchSizeMetric metrics.Histogram
 		if metricRegistry != nil {
 			topicCompressionRatioMetric = getOrRegisterTopicHistogram("compression-ratio", topic, metricRegistry)
+			topicBatchSizeMetric = getOrRegisterTopicHistogram("batch-size", topic, metricRegistry)
 		}
 		for id, records := range partitions {
 			startOffset := pe.offset()
@@ -125,7 +127,7 @@ func (r *ProduceRequest) encode(pe packetEncoder) error {
 				}
 				batchSize := int64(pe.offset() - startOffset)
 				batchSizeMetric.Update(batchSize)
-				getOrRegisterTopicHistogram("batch-size", topic, metricRegistry).Update(batchSize)
+				topicBatchSizeMetric.Update(batchSize)
 			}
 		}
 		if topicRecordCount > 0 {
--- a/produce_request_test.go
+++ b/produce_request_test.go
@@ -109,3 +109,45 @@ func TestProduceRequest(t *testing.T) {
 	batch.compressedRecords = nil
 	testRequestDecode(t, "one record", request, packet)
 }
+
+func benchmarkProduceRequestEncodeMetrics(b *testing.B, partitions int) {
+	b.Helper()
+
+	produceReq := &ProduceRequest{
+		RequiredAcks: WaitForLocal,
+		Timeout:      1_000,
+	}
+	for partition := range partitions {
+		produceReq.AddMessage("bench.topic", int32(partition), &Message{
+			Codec: CompressionNone,
+			Value: []byte("payload"),
+		})
+	}
+
+	req := &request{
+		correlationID: 1,
+		clientID:      "bench",
+		body:          produceReq,
+	}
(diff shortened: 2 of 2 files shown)


produce_request.go
```diff
@@ -125,7 +127,7 @@ func (r *ProduceRequest) encode(pe packetEncoder) error {
 				}
 				batchSize := int64(pe.offset() - startOffset)
 				batchSizeMetric.Update(batchSize)
-				getOrRegisterTopicHistogram("batch-size", topic, metricRegistry).Update(batchSize)
+				topicBatchSizeMetric.Update(batchSize)
```
puellanivis, line 130: :squint: wouldn’t this panic if this were untyped nil? (That is, when `metricRetgistery == nil`.)

produce_request_test.go
```diff
@@ -109,3 +109,46 @@ func TestProduceRequest(t *testing.T) {
...
+
+	req := &request{
+		correlationID: 1,
+		clientID:      "bench",
+		body:          produceReq,
+	}
+	metricRegistry := NewConfig().MetricRegistry
+
+	b.ReportAllocs()
+	b.ResetTimer()
```
puellanivis: > Loop resets the benchmark timer the first time it is called in a benchmark, so any setup performed prior to starting the benchmark loop does not count toward the benchmark measurement. Likewise, when it returns false, it stops the timer so cleanup code is not measured.


https://pkg.go.dev/testing@go1.26.2#B.Loop

produce_request_test.go
```diff
@@ -109,3 +109,46 @@ func TestProduceRequest(t *testing.T) {
...
 }
+
+func benchmarkProduceRequestEncodeMetrics(b *testing.B, partitions int) {
+	b.Helper()
+
+	produceReq := &ProduceRequest{
+		RequiredAcks: WaitForLocal,
+		Timeout:      1_000,
+	}
+	for partition := 0; partition < partitions; partition++ {
```
puellanivis:
```suggestion
	for partition := range partitions {
```

huynhanx03:
Hi @puellanivis , thanks for the review. I’ve addressed your comments below:

1. Addressed the benchmark-loop style suggestion:
- changed `for partition := 0; partition < partitions; partition++` to `for partition := range partitions` in `benchmarkProduceRequestEncodeMetrics`.

2. Addressed the `b.Loop()` timer suggestion:
- removed `b.ResetTimer()` since `b.Loop()` already handles timer start/stop around the measured loop.

3. For the `metricRegistry == nil` concern:
- the cached metric lookup and the `Update(...)` call are both still under the same `if metricRegistry != nil` guard, so this change does not introduce a new plain-nil panic path.
- if the concern is typed-nil interface behavior, that risk already existed in the previous `getOrRegisterTopicHistogram(...)` path as well (not introduced by this optimization).

Verification run:
- `go test -run 'TestProduceRequest|TestAllocateBodyProtocolVersions' .`
- `go test -run '^$' -bench 'BenchmarkProduceRequestEncodeMetrics(1Partition|32Partitions|128Partitions)$' -benchmem .`

puellanivis (approved):
> * the cached metric lookup and the `Update(...)` call are both still under the same `if metricRegistry != nil` guard, so this change does not introduce a new plain-nil panic path.

Ah. I see, there are separate metrics nil guards.

dnwe (approved):
LGTM, thanks!

PR #3506, opened 2026-05-01
feat: add SubscriptionUserDataProvider hook for BalanceStrategy

### LINKED ISSUE #3505: Add per-cycle SubscriptionUserData hook for BalanceStrategy
## Summary

Sarama is missing the per-cycle subscription-metadata hook that the Java reference Kafka client exposes. This makes it impossible to write `BalanceStrategy` implementations whose `UserData` reflects per-cycle local state — e.g., load-aware partition assignors that report observed work per partition to the leader at JoinGroup time. Today, Go consumers using sarama can balance partitions by *count* but not by *work*, leading to fleets where some pods sit at 70% CPU while others sit at 10%.

## Spec context

The Kafka consumer protocol (KIP-54, KIP-394, KIP-462) defines `ConsumerGroupMemberMetadata.UserData` as opaque assignor-defined bytes and explicitly permits the assignor to populate it with arbitrary per-cycle content.

The Java reference client implements this via [`ConsumerPartitionAssignor.subscriptionUserData(Set<String> topics)`](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerPartitionAssignor.java#L51-L61):

```java
/**
 * Return serialized data that will be included in the {@link Subscription} sent to the leader
 * and can be leveraged in {@link #assign(Cluster, GroupSubscription)} ((e.g. local host/rack information)
 *
 * @param topics Topics subscribed to through {@link org.apache.kafka.clients.consumer.KafkaConsumer#subscribe(java.util.Collection)}
 * and variants
 * @return nullable subscription user data
 */
default ByteBuffer subscriptionUserData(Set<String> topics) {
    return null;
}
```

Java's [`ConsumerCoordinator`](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java) calls this method before each JoinGroup, allowing the assignor to inject fresh per-cycle bytes into `Subscription.userData`.

## Today's gap in sarama

Sarama populates `ConsumerGroupMemberMetadata.UserData` from `c.userData` ([`consumer_group.go`](https://github.com/IBM/sarama/blob/main/consumer_group.go) around line 482), which is set in only two places:

1. `Config.Consumer.Group.Member.UserData` — static, set once at startup.
2. `BalanceStrategy.AssignmentData()` — runs only on the **leader** during SyncGroup; bytes are echoed back by each member on subsequent JoinGroups.

Neither path lets a member-side strategy inject its own per-cycle local observations. A `BalanceStrategy` can only see what the leader assigned to that member in the previous cycle, which is sufficient for sticky-style "remember my partitions" but not for load-aware reporting where each member needs to tell the leader what it has observed locally.

## Use case (and why the worked example matters)

Building a load-aware sticky partition assignor: each member observes per-partition work (in-scope events/sec, processing time, etc.) and reports those observations to the leader, which makes load-balanced assignment decisions instead of partition-count-balanced ones. This is exactly what `subscriptionUserData` enables in Java; without it in sarama, the pattern is structurally impossible.

Will fast-follow with a PR that adds:
- The hook (described below)
- Tests covering both the implemented and not-implemented paths
- A worked example: a load-aware sticky `BalanceStrategy` showing the pattern, so other people hitting the same problem can build off it

## Proposed API (backward-compatible)

A new optional interface — does **not** modify `BalanceStrategy`:

```go
// SubscriptionUserDataProvider is an optional interface that BalanceStrategy
// implementations may satisfy. When implemented, the consumer calls
// SubscriptionUserData immediately before each JoinGroup to populate
// ConsumerGroupMemberMetadata.UserData with assignor-specific per-cycle data.
// This mirrors the Java client's
// ConsumerPartitionAssignor.subscriptionUserData(Set<String> topics).
//
// Strategies that do not implement this interface fall back to the existing
// static Config.Consumer.Group.Member.UserData field — today's behavior is
// preserved exactly.
//
// This is V1 of the hook contract. Its signature is intentionally narrow,
// matching Java's subscriptionUserData(Set<String> topics). If sarama later
// needs to expose additional pre-JoinGroup state to assignors (e.g. previous
// memberID, previous generationID), do NOT widen this interface — that would
// silently un-call existing V1 implementations because the type assertion
// would quietly start returning false on a sarama upgrade. Add a separate
// V2 interface (e.g. SubscriptionUserDataProviderV2) and have sarama type-
// assert both, preferring V2 when present and falling back to V1 otherwise.
type SubscriptionUserDataProvider interface {
    SubscriptionUserData(topics []string) ([]byte, error)
}
```

Wire-up in `consumer_group.go:joinGroupRequest`, just before `meta := &ConsumerGroupMemberMetadata{...}`:

```go
userData := c.userData
if p, ok := strategy.(SubscriptionUserDataProvider); ok {
    if data, err := p.SubscriptionUserData(topics); err == nil && data != nil {
        userData = data
    }
    // on error or nil result, fall through to existing c.userData
}
```

## Backward compatibility

- `BalanceStrategy` interface unchanged. Every existing implementation compiles and runs unmodified.
- Default behavior unchanged. Strategies that don't implement the new interface use today's static path.
- Strategies that implement the new interface but return `(nil, err)` fall back to `c.userData` — no surprises.
- The new interface's signature is treated as V1 (see doc comment) so we don't quietly break adopters of *this* interface on a future sarama release.

## Next steps

If the API direction looks acceptable, I'll send the PR with implementation, tests, and the worked-example load-aware sticky assignor.

lizthegrey:
Add an optional SubscriptionUserDataProvider interface that a BalanceStrategy may implement. When present, sarama invokes SubscriptionUserData(topics) immediately before each JoinGroup, and the returned bytes replace the statically configured Consumer.Group.Member.UserData for that strategy's subscription metadata only. A nil result or non-nil error falls back to the static UserData (errors are logged, not fatal). This mirrors the Java ConsumerPartitionAssignor.subscriptionUserData hook and enables load-aware assignors that report fresh per-cycle observations (CPU, lag, latency, etc.) to the group leader on every rebalance cycle.

Existing BalanceStrategy implementations are unaffected: the interface is optional and the behavior for strategies that do not implement it is unchanged. The interface is treated as V1; future revisions should be introduced as separate interfaces (e.g., SubscriptionUserDataProviderV2) to preserve backward compatibility.

Includes a unit test of the per-strategy metadata builder and a worked example under examples/consumer_load_aware/ that wraps the built-in sticky strategy with a load-reporting SubscriptionUserDataProvider.

Fixes #3505


balance_strategy.go
```diff
@@ -57,6 +57,29 @@ type BalanceStrategy interface {
...
+// may implement to inject per-cycle metadata into the ConsumerGroupMemberMetadata
+// UserData field on each JoinGroup. When a strategy implements this interface,
+// Sarama invokes SubscriptionUserData immediately before each JoinGroup, passing
+// the member's currently subscribed topics. The returned bytes replace the
+// statically configured Consumer.Group.Member.UserData for the metadata of that
+// strategy only.
+//
+// Returning a nil byte slice or a non-nil error causes Sarama to fall back to
+// the statically configured UserData; an error is logged but does not fail the
+// JoinGroup.
```
puellanivis: It is considered inappropriate to treat `nil` slices differently from zero-length slices. (That is: either we ignore all zero-length slices, or we ignore neither.)

Additionally, a general good API guideline is that one should either return a valid value, or a non-nil error, never both, and even more so, never _neither_.

If the `SubscriptionUserDataProvider` returns `nil, nil` that should mean that they intend to clear any userdata, because they have returned a `nil` error, the other value should be considered valid.

balance_strategy.go
```diff
@@ -57,6 +57,29 @@ type BalanceStrategy interface {
...
+//
+// This mirrors Java's ConsumerPartitionAssignor.subscriptionUserData and enables
+// load-aware assignors that report fresh observations (e.g., CPU, lag, latency)
+// to the group leader on every rebalance cycle.
+//
+// The interface is treated as V1; future revisions should be introduced as
+// separate interfaces (e.g., SubscriptionUserDataProviderV2) to preserve
+// backward compatibility with existing implementations.
+type SubscriptionUserDataProvider interface {
```
puellanivis: If this is an extension interface, it would be better to embed the `BalanceStrategy` in the `interface` making it clear that anything implementing this interface must also be a ̀ BalanceStrategy`. Additionally extension interfaces usually end with the type they are extending.

```suggestion
type SubscriptionUserDataBalanceStrategy interface {
	BalanceStrategy

```

balance_strategy.go
```diff
@@ -57,6 +57,29 @@ type BalanceStrategy interface {
...
+	SubscriptionUserData(topics []string) ([]byte, error)
```
puellanivis, line 79: This receiver method is named as a noun. This is usually non-ideal, as the whole idiomatic use of Golang naming is to name the interface as a `Doer`, that has receiver method `Do`, or `Handler` that has receiver method `Handle`.

The receiver method is a verb, and the interface name is then the agent nominalization of that verb.

balance_strategy.go
```diff
@@ -57,6 +57,29 @@ type BalanceStrategy interface {
...
+// the statically configured UserData; an error is logged but does not fail the
+// JoinGroup.
+//
+// This mirrors Java's ConsumerPartitionAssignor.subscriptionUserData and enables
+// load-aware assignors that report fresh observations (e.g., CPU, lag, latency)
+// to the group leader on every rebalance cycle.
+//
+// The interface is treated as V1; future revisions should be introduced as
+// separate interfaces (e.g., SubscriptionUserDataProviderV2) to preserve
+// backward compatibility with existing implementations.
```
puellanivis: YAGNI. We do not need to explain how a future extension to this interface should work. This is already common practice, and the naming and handling of it could be different.

For instance, we could decide on providing a completely different mutually-exclusive signature, and balance strategies can implement either or. c.f. in the standard library `errors` package, an error may implement either `Unwrap() error` or `Unwrap() []errors`, and the package will handle whichever is given.

consumer_group.go
```diff
@@ -477,18 +477,17 @@ func (c *consumerGroup) joinGroupRequest(coordinator *Broker, topics []string) (
 		}
 	}
 
-	meta := &ConsumerGroupMemberMetadata{
-		Topics:   topics,
-		UserData: c.userData,
+	addProtocol := func(strategy BalanceStrategy) error {
+		return req.AddGroupProtocolMetadata(strategy.Name(), c.subscriptionMetadata(strategy, topics))
 	}
```
puellanivis: Joining this code makes the code less readible. We are allowed to repeat ourselves a little, as a treat. And especially so, when repeating ourselves makes the code more clear.

Extracting a single line of common code out into an “convenience” function does not improve readability, and at the point of usage, replaces a one-line function call with… a one-line function call. That’s why “convenience” functions like these are rarely actually convenient, and should instead simply be inlined, even if that involves repeating ourselves a bit.

consumer_group.go
```diff
@@ -497,6 +496,31 @@ func (c *consumerGroup) joinGroupRequest(coordinator *Broker, topics []string) (
...
+	if p, ok := strategy.(SubscriptionUserDataProvider); ok {
+		data, err := p.SubscriptionUserData(topics)
+		switch {
+		case err != nil:
+			Logger.Printf(
+				"consumergroup/%s strategy %q SubscriptionUserData failed: %v; falling back to static UserData\n",
+				c.groupID, strategy.Name(), err,
+			)
+		case data != nil:
+			userData = data
```
puellanivis: We could just return here directly, rather than mutating `userData`, and giving us a proper early return design:
```go
if p, ok := strategy.(SubscriptionUserDataProvider); ok {
	userData, err := p.SubscriptionUserData(topics)
	if err == nil {
		return &ConsumerGroupMemberMetadata{
			Topics:   topics,
			UserData: userData,
		}
	}

	Logger.Printf(
		"consumergroup/%s strategy %q SubscriptionUserData failed: %v; falling back to static UserData\n",
		c.groupID, strategy.Name(), err,
	)
}
return &ConsumerGroupMemberMetadata{
	Topics:   topics,
	UserData: c.userData,
}
```

Ideally, log lines should also group a constant value that is significant and unique in order to make finding that log line in the code easier. `"falling back to standard user data for strategy %q consumergroup/%s: %v"`

consumer_group_test.go
```diff
@@ -252,3 +253,77 @@ func TestConsumerShouldNotRetrySessionIfContextCancelled(t *testing.T) {
...
+// assert per-cycle invocation.
+type strategyWithSubscriptionUserData struct {
+	BalanceStrategy
+	data   []byte
+	err    error
+	called [][]string
+}
+
+func (s *strategyWithSubscriptionUserData) SubscriptionUserData(topics []string) ([]byte, error) {
+	s.called = append(s.called, append([]string(nil), topics...))
```
puellanivis: `slices.Clone` has been available since `go1.21`.
```suggestion
	s.called = append(s.called, slices.Clone(topics))
```

consumer_group_test.go
```diff
@@ -252,3 +253,77 @@ func TestConsumerShouldNotRetrySessionIfContextCancelled(t *testing.T) {
...
+		meta := c.subscriptionMetadata(strategy, topics)
+		assert.Equal(t, staticUserData, meta.UserData)
+	})
+
+	t.Run("provider returning error falls back to static user data", func(t *testing.T) {
+		c := &consumerGroup{userData: staticUserData}
+		strategy := &strategyWithSubscriptionUserData{
+			BalanceStrategy: NewBalanceStrategyRange(),
+			data:            []byte("ignored"),
+			err:             fmt.Errorf("boom"),
```
puellanivis: We can also simply use `errors.New` considering we’ve already imported that package, and we’re not even using any of the formatting potential of `fmt`.

consumer_group_test.go
```diff
@@ -252,3 +253,77 @@ func TestConsumerShouldNotRetrySessionIfContextCancelled(t *testing.T) {
...
+
+	t.Run("provider returning nil falls back to static user data", func(t *testing.T) {
+		c := &consumerGroup{userData: staticUserData}
+		strategy := &strategyWithSubscriptionUserData{
+			BalanceStrategy: NewBalanceStrategyRange(),
+			data:            nil,
+		}
+		meta := c.subscriptionMetadata(strategy, topics)
+		assert.Equal(t, staticUserData, meta.UserData)
+	})
```
puellanivis: We have here a test for `data: nil`, but none for `data: []byte{}̀ .

If we were handling all zero-length slices the same, then we wouldn’t need this separate-yet-identical test case. However, as designed, we’re defining different behavior between nil slices, and empty slices, so there should be a separate test case.

(This comment can be discarded if the API changes to treat all zero-length slices the same.)

puellanivis (approved):
Code seems to look fine, though I’m unsure of the integration into the larger logic though.

consumer_group.go
```diff
@@ -497,6 +492,31 @@ func (c *consumerGroup) joinGroupRequest(coordinator *Broker, topics []string) (
...
 }
 
+// subscriptionMetadata builds the ConsumerGroupMemberMetadata for a single
+// strategy in a JoinGroup request. If the strategy implements
+// SubscriptionUserDataBalanceStrategy, its SubscriptionUserData hook is invoked
+// to obtain per-cycle UserData; on error the statically configured
+// Consumer.Group.Member.UserData is used and the error is logged.
+func (c *consumerGroup) subscriptionMetadata(strategy BalanceStrategy, topics []string) *ConsumerGroupMemberMetadata {
+	if p, ok := strategy.(SubscriptionUserDataBalanceStrategy); ok {
+		userData, err := p.SubscriptionUserData(topics)
```
dnwe: Do we want a read-only sort of approach to the topics slice here, else the user's func could modify it before we later pass it to the JoinGroup? Wondering if we should be defensively passing `slices.Clone(topics)` to give them a throwaway copy

consumer_group.go
```diff
@@ -497,6 +492,31 @@ func (c *consumerGroup) joinGroupRequest(coordinator *Broker, topics []string) (
...
+		if err == nil {
+			return &ConsumerGroupMemberMetadata{
+				Topics:   topics,
+				UserData: userData,
+			}
+		}
+		Logger.Printf(
+			"falling back to static user data for strategy %q consumergroup/%s: %v\n",
+			strategy.Name(), c.groupID, err,
+		)
```
dnwe, line 515: nit: the rest of the log lines usually have `consumergroup/%s` as the prefix, so probably better to follow that convention even if the sentence flow isn't quite as good

`"consumergroup/%s: falling back to static user data for strategy %q due to %v", c.groupID, strategy.Name(), err`

dnwe (commented):
LGTM, couple of minor comments:

consumer_group.go
puellanivis: I’ll defer to consistency, yes.

lizthegrey: done.

consumer_group.go
lizthegrey: done.

balance_strategy.go
dnwe: Whilst I sort of agree, we do already have the `AssignmentData` noun in the `BalanceStrategy` interface, so there is some existing precedent for noun-named methods that return bytes and it does match Java's `subscriptionUserData()` so I think we're probably best to accept this as-is

dnwe (approved):
LGTM, thanks for those updates

balance_strategy.go
puellanivis: Indeed, I haven’t further advanced the thread, because it’s not really a big deal, and we do have noun-named receiver methods elsewhere.

PR #3507, opened 2026-05-01
perf(zstd): scale idle zstd encoder cap to GOMAXPROCS

lizthegrey:
## Summary

Raise the per-`CompressionLevel` encoder cache size from 1 to GOMAXPROCS, and lazy-init each channel under a mutex. Under concurrent producer load this eliminates a hot path where any goroutine that arrived while the one cached encoder was checked out would allocate (and immediately discard) a fresh encoder on every batch.

## Motivation

`getZstdEncoder` previously read from a buffered channel of capacity 1 (`zstdMaxBufferedEncoders = 1`) per `ZstdEncoderParams`. When two or more goroutines compressed at the same level concurrently, the channel was empty for the second arrival, the `default` branch fired, and a new `zstd.Writer` was constructed for every batch. Encoder construction is non-trivial (it allocates internal buffers and starts goroutines for parallel encoding), so under steady contention the cache effectively did nothing for all but the first goroutine in flight.

Sizing the channel to GOMAXPROCS lets steady-state per-CPU concurrency reuse encoders, while still bounding per-level memory on high-core hosts. Each retained encoder is constructed with `WithEncoderConcurrency(1)` (unchanged) so its own footprint stays small.

## Changes

- `zstd.go`: raise the `chan *zstd.Encoder` capacity to `max(runtime.GOMAXPROCS(0), 1)`, and serialize lazy channel creation behind a mutex with a double-checked `sync.Map` load so a stampede of goroutines cannot create multiple competing channels for the same params.
- `zstd_test.go`: update existing benchmarks and add a `RunParallel` benchmark to demonstrate the contention case the size-1 cache could not handle.

Public API and wire format are unchanged.

## Test plan

- [x] `go test ./...` passes
- [x] Existing zstd round-trip tests pass
- [x] `go test -race -run TestZstd ./...` passes
- [ ] Reviewers: please run benchmarks locally to confirm the parallel case improves on your hardware: `go test -run=^$ -bench=BenchmarkZstd -benchmem ./...`

## DCO
Co-author has authorized me (committer) to sign off on their behalf; the work is Honeycomb-copyrighted.

### DIFF
--- a/zstd.go
+++ b/zstd.go
@@ -6,10 +6,6 @@ import (
 	"github.com/klauspost/compress/zstd"
 )
 
-// zstdMaxBufferedEncoders maximum number of not-in-use zstd encoders
-// If the pool of encoders is exhausted then new encoders will be created on the fly
-const zstdMaxBufferedEncoders = 1
-
 type ZstdEncoderParams struct {
 	Level int
 }
@@ -20,35 +16,35 @@ var zstdDecMap sync.Map
 
 var zstdAvailableEncoders sync.Map
 
-func getZstdEncoderChannel(params ZstdEncoderParams) chan *zstd.Encoder {
+func getZstdEncoderPool(params ZstdEncoderParams) *sync.Pool {
 	if c, ok := zstdAvailableEncoders.Load(params); ok {
-		return c.(chan *zstd.Encoder)
+		return c.(*sync.Pool)
 	}
-	c, _ := zstdAvailableEncoders.LoadOrStore(params, make(chan *zstd.Encoder, zstdMaxBufferedEncoders))
-	return c.(chan *zstd.Encoder)
+
+	encoderLevel := zstd.SpeedDefault
+	if params.Level != CompressionLevelDefault {
+		encoderLevel = zstd.EncoderLevelFromZstd(params.Level)
+	}
+
+	pool := &sync.Pool{
+		New: func() any {
+			enc, _ := zstd.NewWriter(nil, zstd.WithZeroFrames(true),
+				zstd.WithEncoderLevel(encoderLevel),
+				zstd.WithEncoderConcurrency(1))
+			return enc
+		},
+	}
+
+	c, _ := zstdAvailableEncoders.LoadOrStore(params, pool)
+	return c.(*sync.Pool)
 }
 
 func getZstdEncoder(params ZstdEncoderParams) *zstd.Encoder {
-	select {
-	case enc := <-getZstdEncoderChannel(params):
-		return enc
-	default:
-		encoderLevel := zstd.SpeedDefault
-		if params.Level != CompressionLevelDefault {
-			encoderLevel = zstd.EncoderLevelFromZstd(params.Level)
-		}
-		zstdEnc, _ := zstd.NewWriter(nil, zstd.WithZeroFrames(true),
-			zstd.WithEncoderLevel(encoderLevel),
-			zstd.WithEncoderConcurrency(1))
-		return zstdEnc
-	}
+	return getZstdEncoderPool(params).Get().(*zstd.Encoder)
 }
 
 func releaseEncoder(params ZstdEncoderParams, enc *zstd.Encoder) {
-	select {
-	case getZstdEncoderChannel(params) <- enc:
-	default:
-	}
+	getZstdEncoderPool(params).Put(enc)
 }
 
 func getDecoder(params ZstdDecoderParams) *zstd.Decoder {
--- a/zstd.go
+++ b/zstd.go
@@ -1,10 +1,24 @@
 package sarama
 
 import (
+	"runtime"
 	"sync"
 
 	"github.com/klauspost/compress/zstd"
 )
+
+// ZstdMaxBufferedEncoders bounds the number of idle zstd encoders Sarama
+// retains per ZstdEncoderParams. The previous bound of 1 forced any
+// concurrent caller after the first to allocate (and discard) a fresh
+// encoder per batch; defaulting to runtime.GOMAXPROCS lets steady-state
+// per-CPU concurrency reuse encoders while still bounding per-level memory
+// on high-core hosts. Each retained encoder is constructed with
+// WithEncoderConcurrency(1) to keep its own memory footprint small.
+//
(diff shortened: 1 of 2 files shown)


dnwe:
Hey Liz - thanks for putting up this PR for review. Our zstd encoding has bounced around a few different implementations over time 

  - c9e4131 introduced zstdEncMap: one cached encoder per compression level.
  - 38204cb deliberately replaced that with a channel-backed encoder pool of size 1, plus WithEncoderConcurrency(1) because the old/default-concurrency encoder
   reserved about 1GB memory for every 32 cores
  - 779fb1f changed decoder concurrency to account for an upstream change in compress
  - this PR replaces the bounded channel pool with sync.Pool.
  
  So it does partially undo the bounded-memory property that 38204cb added and will retain multiple idle encoders until GC decides to drop them. Do we need to make the idle count configurable for high core machines?

zstd.go
```diff
@@ -20,35 +16,35 @@ var zstdDecMap sync.Map
...
 var zstdAvailableEncoders sync.Map
 
-func getZstdEncoderChannel(params ZstdEncoderParams) chan *zstd.Encoder {
+func getZstdEncoderPool(params ZstdEncoderParams) *sync.Pool {
 	if c, ok := zstdAvailableEncoders.Load(params); ok {
-		return c.(chan *zstd.Encoder)
+		return c.(*sync.Pool)
+	}
+
+	encoderLevel := zstd.SpeedDefault
```
puellanivis: I’m thinking maybe we might want to put a mutex on this instead of letting a stampede of calls all create a bunch of garbage that only the first to hit `LoadOrStore` will end up saving.

I mean, we’re not expecting this to generally be in any hot path, right? Except for a stampede, all this code would otherwise be expected to run uncontented.

zstd_test.go
```diff
@@ -16,16 +16,39 @@ func BenchmarkZstdMemoryConsumption(b *testing.B) {
...
+		}
+		runtime.GOMAXPROCS(gomaxprocsBackup)
+
+	})
+
+	// Concurrent encodes, which historically would allocate excessively
+	b.Run("concurrent", func(b *testing.B) {
+		b.ReportAllocs()
+
+		gomaxprocsBackup := runtime.GOMAXPROCS(cpus)
```
puellanivis: We should be using [B.SetParallelism](https://pkg.go.dev/testing@go1.26.2#B.SetParallelism) instead.

zstd_test.go
```diff
@@ -16,16 +16,39 @@ func BenchmarkZstdMemoryConsumption(b *testing.B) {
...
 	cpus := 96
 
-	gomaxprocsBackup := runtime.GOMAXPROCS(cpus)
-	b.ReportAllocs()
-	for b.Loop() {
-		for j := 0; j < 2*cpus; j++ {
+	b.Run("no_drain", func(b *testing.B) {
+		b.ReportAllocs()
+
+		gomaxprocsBackup := runtime.GOMAXPROCS(cpus)
```
puellanivis: I don’t see how setting `GOMAXPROCS` is useful here. We don’t seem to be doing anything parallel?

zstd.go
```diff
@@ -18,36 +28,63 @@ type ZstdDecoderParams struct {
...
+	defer zstdEncoderInitMu.Unlock()
+
+	if c, ok := zstdAvailableEncoders.Load(params); ok {
+		return c.(chan *zstd.Encoder)
+	}
+
+	size := ZstdMaxBufferedEncoders
+	if size < 1 {
+		size = 1
+	}
```
puellanivis:
```suggestion
	size := max(ZstdMaxBufferedEncoders, 1)
```

zstd.go
```diff
@@ -1,14 +1,24 @@
...
+// concurrent caller after the first to allocate (and discard) a fresh
+// encoder per batch; defaulting to runtime.GOMAXPROCS lets steady-state
+// per-CPU concurrency reuse encoders while still bounding per-level memory
+// on high-core hosts. Each retained encoder is constructed with
+// WithEncoderConcurrency(1) to keep its own memory footprint small.
+//
+// The value is read on first use of each ZstdEncoderParams; changes after
+// that point only affect compression levels not yet seen. To override the
+// default, set this variable before any zstd compression takes place.
+var ZstdMaxBufferedEncoders = max(runtime.GOMAXPROCS(0), 1)
```
puellanivis: I’m unsure about exposing this knob… it’s unlikely that callers would know that its here, or even what a useful value would be. Plus, the `GOMAXPROCS` is set during `init` not during runtime, where that value (which is a more well-known concurrency control knob) could be changed before execution.

zstd.go
```diff
@@ -1,14 +1,24 @@
...
-// zstdMaxBufferedEncoders maximum number of not-in-use zstd encoders
-// If the pool of encoders is exhausted then new encoders will be created on the fly
-const zstdMaxBufferedEncoders = 1
+// ZstdMaxBufferedEncoders bounds the number of idle zstd encoders Sarama
+// retains per ZstdEncoderParams. The previous bound of 1 forced any
+// concurrent caller after the first to allocate (and discard) a fresh
+// encoder per batch; defaulting to runtime.GOMAXPROCS lets steady-state
+// per-CPU concurrency reuse encoders while still bounding per-level memory
+// on high-core hosts. Each retained encoder is constructed with
+// WithEncoderConcurrency(1) to keep its own memory footprint small.
```
puellanivis: The detail of the `WithEncoderConcurrency` is an internal implementation detail, and shouldn’t be exposed in documentation.

zstd.go
dnwe: Has been addressed

zstd_test.go
dnwe: Has been addressed

zstd_test.go
dnwe: Has been addressed

zstd.go
dnwe: Has been addressed

dnwe:
Thanks for the PR! Update version looks good to me. I also updated the PR summary and description to match the final state of the PR prior to merge

dnwe (approved):

Issue #2116, opened 2022-01-20
How to consume all messages in a topic without hanging there


magicLian:
##
I'm trying to consume all messages in a topic from the oldest to the newest.
but when it runs to the end, it will hang there unless new data coming.

How to stop it? 

`message, ok := <-partitionConsumer.Messages()` 
The code above will hang there and I can't return it.

Issue #848, opened 2017-03-15
Stop at end of topic/partition


arnaud-lb:
It doesn't seem to be possible to stop consuming when the end of a partition is reached.

I'm using compacted topics, and one common use-case I have is to start by reading them entirely from the beginning to the end. However, there is no easy way to stop consuming when the end is reached.

I tried a few things, but the library is not making that easy.

Basically, here is what I have to do currently:

 1. Read the newest offset: `client.GetOffset(topic, partition, sarama.OffsetNewest)`
 2. Consume every message
 3. If a message with offset >= newest offset is seen, stop consuming
 4. If topic is empty, _3_ may never happen, so I have to also produce a dummy message before step 1, for instance

It would be awesome if Sarama provided a way to stop consuming when the end of a partition is reached and/or to be notified about it.

eapache:
Similar to https://github.com/Shopify/sarama/issues/666 and https://github.com/Shopify/sarama/issues/628, this is not something we provide because kafka does not really have the concept; the entire design is based on the fact that streams are never-ending.

arnaud-lb:
I believe that the Kafka protocol notifies the client when the end of the partition is reached ?

librdkafka and librdkafka-based clients also notify the user when the end of the partition is reached.

eapache:
> I believe that the Kafka protocol notifies the client when the end of the partition is reached ?

Not to my knowledge.

wvanbergen:
It tells you the latest available offset while consuming: https://github.com/Shopify/sarama/pull/339, which you could use for this purpose.

Issue #628, opened 2016-03-12
How to consume until the end?


k2xl:
Are there any examples of consuming until reaching the end of a kafka partition? Right now I'm having to implement a timeout (if I don't get a message after X seconds I assume I got to the end)

wvanbergen:
There is no direct way to do this, because Kafka/sarama is built around the concept of never ending streams.

Two ways to accomplish this:
- Timeouts - but this can be tricky when the partition is still receiving new events.
- Get the last offset, start consuming and stop when you have consumed that offset. `Client.GetOffset` can be used to find the latest offset. https://godoc.org/github.com/Shopify/sarama#Client

PR #3516, opened 2026-05-06
fix: Make InitialOffset available also in PartitionConsumer

fredde-fisk:
Store the initial offset in the `partitionConsumer`
and make it available via the `InitialOffset`
method. The same method is then also reused when
embedded in a `consumerGroupClaim`.

I also added tests for the `InitialOffset` method since it did not exist before.

This change will align the `ConsumerGroupClaim` and `PartitionConsumer` interfaces, and will also make it possible to implement a proper end of partition detection that is not timer based or has to access internal fields, something that has been requested several times (https://github.com/IBM/sarama/issues/628, https://github.com/IBM/sarama/issues/848, https://github.com/IBM/sarama/issues/2116).

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -116,6 +116,7 @@ func NewConsumer(addrs []string, config *Config) (Consumer, error) {
 	}
 	c, err := newConsumer(client)
 	if err != nil {
+		// Return nil rather than a Consumer interface with a nil value
 		return nil, err
 	}
 
@@ -130,6 +131,7 @@ func NewConsumerFromClient(client Client) (Consumer, error) {
 	cli := &nopCloserClient{client}
 	c, err := newConsumer(cli)
 	if err != nil {
+		// Return nil rather than a Consumer interface with a nil value
 		return nil, err
 	}
 
@@ -169,12 +171,14 @@ func (c *consumer) Partitions(topic string) ([]int32, error) {
 func (c *consumer) ConsumePartition(topic string, partition int32, offset int64) (PartitionConsumer, error) {
 	pc, err := c.consumePartition(topic, partition, offset)
 	if err != nil {
+		// Return nil rather than a PartitionConsumer interface with a nil value
 		return nil, err
 	}
 
 	return pc, nil
 }
 
+// consumePartition returns the concrete partitionConsumer struct used internally in the ConsumerGroup implementation
 func (c *consumer) consumePartition(topic string, partition int32, offset int64) (*partitionConsumer, error) {
 	child := &partitionConsumer{
 		consumer:             c,
--- a/consumer.go
+++ b/consumer.go
@@ -114,7 +114,12 @@ func NewConsumer(addrs []string, config *Config) (Consumer, error) {
 	if err != nil {
 		return nil, err
 	}
-	return newConsumer(client)
+	c, err := newConsumer(client)
+	if err != nil {
+		return nil, err
+	}
+
+	return c, nil
 }
 
 // NewConsumerFromClient creates a new consumer using the given client. It is still
@@ -123,10 +128,15 @@ func NewConsumerFromClient(client Client) (Consumer, error) {
 	// For clients passed in by the client, ensure we don't
 	// call Close() on it.
 	cli := &nopCloserClient{client}
-	return newConsumer(cli)
+	c, err := newConsumer(cli)
+	if err != nil {
+		return nil, err
+	}
+
+	return c, nil
 }
 
-func newConsumer(client Client) (Consumer, error) {
+func newConsumer(client Client) (*consumer, error) {
 	// Check that we are not dealing with a closed Client before processing any other arguments
 	if client.Closed() {
 		return nil, ErrClosedClient
@@ -157,6 +167,15 @@ func (c *consumer) Partitions(topic string) ([]int32, error) {
 }
 
 func (c *consumer) ConsumePartition(topic string, partition int32, offset int64) (PartitionConsumer, error) {
+	pc, err := c.consumePartition(topic, partition, offset)
+	if err != nil {
+		return nil, err
+	}
+
+	return pc, nil
+}
+
+func (c *consumer) consumePartition(topic string, partition int32, offset int64) (*partitionConsumer, error) {
 	child := &partitionConsumer{
 		consumer:             c,
 		conf:                 c.conf,
@@ -380,10 +399,6 @@ type PartitionConsumer interface {
 	// Consumer.Return.Errors setting to true, and read from this channel.
 	Errors() <-chan *ConsumerError
 
-	// InitialOffset returns the initial offset that was used as a starting
-	// point for this partition.
-	InitialOffset() int64
-
 	// HighWaterMarkOffset returns the high water mark offset of the partition,
 	// i.e. the offset that will be used for the next message that will be produced.
 	// You can use this to determine how far behind the processing is.
@@ -402,6 +417,16 @@ type PartitionConsumer interface {
 
 	// IsPaused indicates if this partition consumer is paused or not
 	IsPaused() bool
+}
+
+// InitialOffsetPartitionConsumer is an extension of PartitionConsumer that adds InitialOffset. The object returned
+// from the default Consumer.ConsumePartition implements this interface.
+type InitialOffsetPartitionConsumer interface {
+	PartitionConsumer
+
+	// InitialOffset returns the initial offset that was used as a starting
+	// point for this partition.
+	InitialOffset() int64
 }
 
 type partitionConsumerResponse struct {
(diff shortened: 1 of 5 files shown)


consumer_test.go
```diff
@@ -68,6 +68,9 @@ func TestConsumerOffsetManual(t *testing.T) {
 	}
 
 	// Then
+	if io := consumer.InitialOffset(); io != manualOffset {
```
puellanivis: I would suggest avoiding `io` as an abbreviated name, since it’s such a well known abbreviation for `input/output`.

consumer.go
```diff
@@ -378,6 +378,10 @@ type PartitionConsumer interface {
 	// Consumer.Return.Errors setting to true, and read from this channel.
 	Errors() <-chan *ConsumerError
 
+	// InitialOffset returns the initial offset that was used as a starting
+	// point for this partition.
+	InitialOffset() int64
```
puellanivis: As we’ve pointed out in other PRs, adding new methods to an `interface` is technically a breaking change. 🤔

puellanivis (commented):
Contains a breaking change 🤔 is this unavoidable?

consumer.go
fredde-fisk: I couldn't find how/if you have solved similar things before, so I don't know if you have a general strategy, but here are some alternatives that I see:

1. Add helper function that returns the `partitionConsumer.initialOffset`, for example:
   ```
   func PartitionConsumerInitialOffset(pc PartitionConsumer) int64 {
       if p, ok := pc.(*partitionConsumer); ok {
           return p.initialOffset
       }
   
       return -1 // or panic
   }
   ```

   This will however clutted the package namespace.

2. Keep `partitionConsumer.InitialOffset()` but not in the `PartitionConsumer` interface. To access the method the user has to define a new interface with `InitialOffset()`, for example:
   ```
   pc.(interface{ InitialOffset() int64 }).InitialOffset()
   ```

   This will however hide the method from the actual public API, making it semi-internal.

3. Add a new interface with `InitialOffset()` available under a new name (e.g. with a version number). For example:
   ```
   type PartitionConsumerV2 interface {
       PartitionConsumer
       InitialOffset()
   }
   
   // To use the new `PartitionConsumerV2` interface
   pc := consumer.ConsumePartition(...).(sarama.PartitionConsumerV2)
   // or through a new set of constructors
   pc := consumer.ConsumePartitionV2(...)

   pc.InitialOffset()
   ```

   This will however become a bit messy if there are a lot of new versioned interfaces. At a major release, when breaking backward compatibility, the versioned interfaces can be flattened to get a single interface again.

puellanivis: Indeed, this basically why there’s the general guideline “return concrete types, and accept interfaces”. If the `PartitionConsumer` were actually the unexported `partitionConsumer` type, then we could easily expand the receiver methods all we want. But we’re stuck at this version with what we have.

Given the lock in, the preferred way to expand an interface is with an expansion interface example:
```go
type InitialOffsetPartitionConsumer interface{
	// N.B. Do not use "V2", instead be descriptive about what's being added.
	PartitionConsumer
	InitialOffset() int64
}
```

As a good stdlib example see: [fs.ReadDirFS](https://pkg.go.dev/io/fs@go1.26.3#ReadDirFS), an expansion interface on ̀ fs.FS` for things that also implement a `ReadDir` method. This would mostly align with your option 2, but there’s an important distinction: each extension should supply the minimal amount of expansion for that particular feature, and each extension should build _separately_ off of the base interface†, and not just keep aggregating up more and more interfaces.

†: unless it depends on an earlier feature, and thus would need to be an extension interface on that interface.

> This will however become a bit messy if there are a lot of new versioned interfaces.

Yes, this can build up quite a lot of extension interfaces, and that can get a little messy. But this is also basically the best way to provide selective feature implementation on interfaces.

> At a major release, when breaking backward compatibility, the versioned interfaces can be flattened to get a single interface again.

No, no ~~capes~~ interfaces. On a major release, any library that is returning interfaces with unexported implementations should switch to returning exported concrete types, which _can_ expand their receiver method coverage at will without it being a breaking change.

fredde-fisk: I moved `InitialOffset` to a `InitialOffsetPartitionConsumer` interface. I also changed the internal `consumerGroupClaim.PartitionConsumer` from an interface to the concrete type that it will actually contain, which will then handle the implementation of `InitialOffset` in the `ConsumerGroupClaim` interface as well.

consumer_group.go
```diff
@@ -1174,16 +1130,15 @@ type ConsumerGroupClaim interface {
 type consumerGroupClaim struct {
 	topic     string
 	partition int32
-	offset    int64
-	PartitionConsumer
+	*partitionConsumer
```
puellanivis: Usually embedded types are placed at the top of a type.

consumer.go
```diff
@@ -157,6 +167,15 @@ func (c *consumer) Partitions(topic string) ([]int32, error) {
...
 func (c *consumer) ConsumePartition(topic string, partition int32, offset int64) (PartitionConsumer, error) {
+	pc, err := c.consumePartition(topic, partition, offset)
+	if err != nil {
+		return nil, err
+	}
+
+	return pc, nil
+}
+
+func (c *consumer) consumePartition(topic string, partition int32, offset int64) (*partitionConsumer, error) {
```
puellanivis, line 187: I do not understand why we would want to move this function into an unexported function, with an exported wrapper.

This serves no purpose, everyone who can can `ConsumePartition` can then effectively also call `consumePartition` and that’s the whole usage of receiver methods, meaning we’ve effectively unexported `ConsumePartition`.

fredde-fisk: > This serves no purpose, everyone who can can ConsumePartition can then effectively also call consumePartition and that’s the whole usage of receiver methods, meaning we’ve effectively unexported ConsumePartition.

Not sure what you mean with `ConsumePartition` being unexported.

The reason for this change was to make it possible to use the concrete `partitionConsumer` struct internally (in `consumerGroupClaim`) while keeping the public API unchanged, still returning the `PartitionConsumer` interface.

An alternative is to do a type assertion on the returned `PartitionConsumer` interface in `consumerGroupClaim` to convert it back to the concrete type, like this:
```
consumerGroupClaim{
	partitionConsumer: pcm.(*partitionConsumer),
	...
}
```

puellanivis: OH!!! I see what’s going on now. I then would recommend putting an explanation somewhere (probably in the `ConsumePartition` wrapper) explaining that we’re converting from the concrete type to the interface.

This would also help explain why we need the `if err != nil { return nil, err }` so that we return an untyped nil, rather a typed nil. Otherwise one might be tempted to “optimize” the code by simply converting it to a simple return.

fredde-fisk: I added some comments to explain why it is done this way, and yes, the urge to optimize this is strong :)

If you think that it is ok with a minor API change, you could remove this error handling. The caller will probably check the error anyway.

consumer_group.go
fredde-fisk: Moved to the top

consumer.go
```diff
@@ -157,6 +169,17 @@ func (c *consumer) Partitions(topic string) ([]int32, error) {
 }
 
 func (c *consumer) ConsumePartition(topic string, partition int32, offset int64) (PartitionConsumer, error) {
+	pc, err := c.consumePartition(topic, partition, offset)
+	if err != nil {
+		// Return nil rather than a PartitionConsumer interface with a nil value
```
puellanivis: I would adjust the language here that we’re returning an “untyped nil” and “rather than… a typed pointer nil”. (The disambiguation of these two conflated terms being the whole point of the comment here.)

consumer.go
puellanivis: Indeed. Callers are just as likely to throw away the typed pointer just like we’re doing here for them.

But the number of cases where I’ve seen tests testing that the other return in the case of a `err != nil` is `nil`… 😬 Such tests would likely break.

consumer.go
fredde-fisk: So "Return untyped nil rather than a typed pointer nil"?

puellanivis: Sorry, the additional context you have/had is great. To be more clear: `Return untyped nil rather than a PartitionConsumer interface with a typed nil pointer.` is great.

(Now that I type it out, and read it, “typed nil pointer” sounds better than “typed pointer nil”.)

fredde-fisk: Changed the wording in the comments

fredde-fisk:
The PR has been rebased and all comments has already been addressed. Waiting for review/approval from @puellanivis.

puellanivis (approved):
Code seems good for me now. Leaving for the team for fit to the design.

Issue #3502, opened 2026-04-23
flaky test: TestFuncTxnProduceAndCommitOffset


dnwe:
TestFuncTxnProduceAndCommitOffset times out periodically

DCjanus:
Looks like this was addressed by #3517, so this can probably be closed.

PR #3517, opened 2026-05-07
fix: stabilise TestFuncTxnProduceAndCommitOffset flakes

dnwe:
Trying to resolve the flaky failures in `TestFuncTxnProduceAndCommitOffset` tracked in #3502. Interestingly it actually looks like it caught a genuine issue!

Two independent issues seemed to be contributing:

1. When the consumer group rebalances at session start, `newConsumerGroupClaim` can return a transient leader/metadata error (`ErrNotLeaderForPartition`, `ErrFencedLeaderEpoch`, `ErrLeaderNotAvailable`, etc.) before metadata has caught up. That error went straight to `handleError` and the partition stayed unclaimed for the rest of the session, which in the test manifested as the consumer never seeing the produced message.

2. The handler used `started.Add(4)` and called `Done()` from `ConsumeClaim`, but `ConsumeClaim` can fire more than four times when a rebalance re-invokes it for the same partition, pushing the counter negative and panicking.

With this PR, `consumerGroupSession.consume` now calls a small retry wrapper that backs off, refreshes metadata, and retries up to `Metadata.Retry.Max` times before giving up, and the test uses a simple `chan struct{}` closed once via `sync.Once` on the first claim.

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -946,9 +946,9 @@ func (s *consumerGroupSession) newClaimWithRetry(topic string, partition int32, 
 		if retries <= 0 || !isRetriableClaimError(err) {
 			return nil, err
 		}
-		backoff := computeMetadataBackoff(s.parent.config, retries)
 		retries--
 
+		backoff := computeMetadataBackoff(s.parent.config, retries)
 		Logger.Printf(
 			"consumer-group/claim %s/%d retrying after %dms... (%d attempts remaining): %v\n",
 			topic, partition, backoff/time.Millisecond, retries, err)
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -933,6 +933,50 @@ func (s *consumerGroupSession) Context() context.Context {
 	return s.ctx
 }
 
+// newClaimWithRetry calls newConsumerGroupClaim, retrying transient errors so
+// that brief leader/metadata desync around a rebalance doesn't leave a
+// partition permanently unclaimed for the lifetime of this session
+func (s *consumerGroupSession) newClaimWithRetry(topic string, partition int32, offset int64) (*consumerGroupClaim, error) {
+	retries := s.parent.config.Metadata.Retry.Max
+	for {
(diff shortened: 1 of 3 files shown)


functional_producer_test.go
```diff
@@ -428,13 +430,12 @@ func TestFuncTxnProduceAndCommitOffset(t *testing.T) {
 		require.NoError(t, err)
 	}
 
-	handler.started.Add(4)
 	go func() {
 		err = cg.Consume(ctx, []string{"test.4"}, handler)
 		require.NoError(t, err)
```
puellanivis: FYI, we shouldn’t be calling `FailNow` in a goroutine other than the main one. We should probably switch this to `assert.NoError`.

consumer_group.go
```diff
@@ -933,6 +933,50 @@ func (s *consumerGroupSession) Context() context.Context {
...
+	for {
+		claim, err := newConsumerGroupClaim(s, topic, partition, offset)
+		if err == nil {
+			return claim, nil
+		}
+		if retries <= 0 || !isRetriableClaimError(err) {
+			return nil, err
+		}
+		backoff := computeMetadataBackoff(s.parent.config, retries)
+		retries--
```
puellanivis, line 949: [style] It might be good to group the retries stuff together into a paragraph, or at least one after the other, rather than putting the backoff calculation in the middle of them?

puellanivis (commented):
Wow, that’s some linty lint in my comments. 🙈

functional_producer_test.go
dnwe: Good spot - I'll also add the testifylint check in golangci-lint in a separate PR (as there's probably other instances of this we should fix) to catch these in the future too

consumer_group.go
dnwe: Agreed and fixed

dnwe:
@puellanivis thanks for the review, hopefully addressed your comments in the follow-up commit

puellanivis (approved):
:shipit:

PR #3523, opened 2026-05-09
test: relax producer batch metrics assertions

### LINKED ISSUE #3496: flaky test: TestFuncProducingFlushing
**Summary**

`TestFuncProducingFlushing` hit an intermittent FVT failure on 2026-04-13 in PR [#3486](https://github.com/IBM/sarama/pull/3486), run [24371018231](https://github.com/IBM/sarama/actions/runs/24371018231), job [71174464166](https://github.com/IBM/sarama/actions/runs/24371018231/job/71174464166).

The failing assertion was:

```text
functional_producer_test.go:1198: Expected histogram metric 'compression-ratio' count >= 3, got 2
--- FAIL: TestFuncProducingFlushing/TestFuncProducingFlushing-v4.2.0
```

In the same run, the other Kafka matrix jobs passed: `1.0.2`, `2.6.3`, `3.6.2`, `3.9.2`, and `4.1.1`.

PR `#3486` only changes `consumer.go` and `consumer_test.go`. This failure is in `functional_producer_test.go`, so it does not look directly related to the PR changes.

**Why This Looks Flaky**

From a quick read of the code, this looks more like a timing issue in the test than a clear producer bug.

- The `compression-ratio` histogram seems to be updated once per encoded batch, not once per message.
- This test uses one topic/partition with `Flush.Messages = TestBatchSize / 8` and `Flush.Frequency = 250ms`.
- In the async producer path, later messages can keep building up while an earlier batch is still in flight. So this setup does not guarantee 3 or more batches.
- One possible case is one earlier batch plus one larger final batch. Then the histogram count would be `2`, and this assertion would fail.
- The functional environment also adds toxiproxy latency, so timing can affect the result.

I could not find an existing issue specifically tracking this test or this assertion.

Sharing this here because I am not sure what the right fix for this test is.

DCjanus:
## Why

`TestFuncProducingFlushing` currently assumes the RecordBatch path emits at least three `compression-ratio` samples. That is not guaranteed: `compression-ratio` is recorded once per encoded RecordBatch, and `Flush.Messages` only makes a batch eligible to flush. It does not guarantee a fixed number of batches.

This is the same boundary that [#2603](https://github.com/IBM/sarama/pull/2603) started to address when it split the old MessageSet metric expectations from the batched produce path. [#2616](https://github.com/IBM/sarama/pull/2616) lowered the RecordBatch sample-count floor from 4 to 3, but the remaining fixed floor can still fail when the test produces fewer batches.

Fixes #3496.

## What

- Keep the test focused on the stable flushing contract: all records are delivered, and producer metrics are emitted.
- Keep exact `compression-ratio` sample-count validation only for the legacy MessageSet path.
- For the RecordBatch path, validate metric values without assuming a fixed number of encoded batches; also allow `record-send-rate` to exceed delivered records when retries occur.

## Validation

I used a temporary focused workflow to make the flaky test visible in PR CI before keeping the fix. The workflow first runs a fail-fast `TestFuncProducingFlushing` check on Kafka 2.8.2, then runs the same test with `-count=100` after the single check passes. After the fix was verified, I removed the workflow so the final diff only keeps the assertion change.

- Before the fix, the focused workflow reproduced the failure: [run](https://github.com/IBM/sarama/actions/runs/25597180721).
- After the fix, the same workflow passed the single check and `-count=100` stress check: [run](https://github.com/IBM/sarama/actions/runs/25597232623).
- After removing the temporary workflow, final PR CI passed: [run](https://github.com/IBM/sarama/actions/runs/25597462181).


dnwe (approved):

dnwe:
Thanks!

PR #3524, opened 2026-05-09
fix: prevent race during partition consumer close

dnwe:
When AsyncClose ran concurrently with subscriptionConsumer dispatching a fetch response, the dispatcher could exit and close(feeder) while subscriptionConsumer still had a write in flight, panicking on a send to a closed channel.

Introduce brokerSubscription as an explicit handover handle: each subscription carries a released channel that the brokerConsumer closes when it drops the subscription from its live set. The dispatcher waits on this channel before closing the feeder, so the brokerConsumer is guaranteed to have stopped writing first.

Also restructure the dispatcher so dying is a top-level select case rather than nested under trigger.

This _should_ prevent a bunch of the remaining FVT flakes from failing periodically.

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -192,10 +192,12 @@ func (c *consumer) ConsumePartition(topic string, partition int32, offset int64)
 	child.leaderEpoch = epoch
 	for {
 		child.broker = c.refBrokerConsumer(leader)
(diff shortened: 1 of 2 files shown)


hindessm (approved):
LGTM

Issue #3514, opened 2026-05-06
Permanent application stall after broker reboot


tilinna:
##### Description

A client application got stuck forever when the two brokers it was connected to were rebooted.

Last log line seen: "client/metadata found some partitions to be leaderless" — application stalls permanently with no further output or errors

##### Versions

github.com/IBM/sarama v1.48.0

##### Configuration

```go
	sc.ApiVersionsRequest = true
	sc.Version = sarama.MaxVersion
	sc.Net.DialTimeout = 10 * time.Second
	sc.Net.ReadTimeout = 25 * time.Second
	sc.Net.WriteTimeout = 15 * time.Second
	sc.Metadata.RefreshFrequency = 5 * time.Minute
	sc.Metadata.Timeout = 60 * time.Second
```

Analyzing the code in client.go suggests the following three bugs interacting together.

Bug 1 — updateMetadataMs timestamp recorded before the fetch

updateMetadataMs.Store was called before broker.GetMetadata, so it records "a fetch was attempted" rather than "a clean response was received."

The retry closure uses this timestamp to skip retrying if another goroutine refreshed metadata during the backoff sleep:

When a broker reboots, many goroutines flood into tryRefreshMetadata simultaneously. Each one stores a fresh timestamp before its own GetMetadata call, keeping the timestamp perpetually recent. Every goroutine's retry() call sees a fresh timestamp and short-circuits — no actual retry ever executes.

Proposed fix: move updateMetadataMs.Store to fire only after a complete, non-leaderless response:

```go
// after updateMetadata returns shouldRetry=false:
client.updateMetadataMs.Store(time.Now().UnixMilli())
return err
```

Bug 2 — updateMetadata returns err=nil when partitions are leaderless

ErrLeaderNotAvailable at topic or partition level sets retry=true but leaves err=nil. This means retry(nil) is called, and when the timestamp short-circuit fires (Bug 1), it returns nil to the caller.

Proposed fix: set err when retry is set, guarded by err == nil to avoid overwriting a higher-priority error:

```go
case ErrLeaderNotAvailable:
    retry = true
    if err == nil {
        err = ErrLeaderNotAvailable
    }
```

```go
if errors.Is(partition.Err, ErrLeaderNotAvailable) {
    retry = true
    if err == nil {
        err = ErrLeaderNotAvailable
    }
}
```

Bug 3 — getPartitions returns ([]int32{}, nil) for all-leaderless topics, causing an infinite producer loop

[...]

PR #3525, opened 2026-05-09
fix: return leaderless errors in metadata refresh

dnwe:
When every partition is leaderless (e.g. all brokers rebooting), updateMetadata signalled retry=true but left err=nil, so RefreshMetadata returned nil and async producers spun on WritablePartitions. The fetch timestamp was also stamped _before_ the broker call, so concurrent goroutines short-circuited each other's retries indefinitely. Propagate the leaderless KError and only stamp updateMetadataMs after a clean response.

Contributes-to: #3514

### DIFF
--- a/client.go
+++ b/client.go
@@ -1011,7 +1011,6 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int,
 
 		req := NewMetadataRequest(client.conf.Version, topics)
 		req.AllowAutoTopicCreation = allowAutoTopicCreation
-		client.updateMetadataMs.Store(time.Now().UnixMilli())
 
 		response, err := broker.GetMetadata(req)
 		var kerror KError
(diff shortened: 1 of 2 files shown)


hindessm (approved):
LGTM

client.go
```diff
@@ -1029,8 +1028,11 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int,
 			shouldRetry, err := client.updateMetadata(response, allKnownMetaData)
 			if shouldRetry {
 				Logger.Println("client/metadata found some partitions to be leaderless")
-				return retry(err) // note: err can be nil
+				return retry(err)
```
puellanivis, line 1031: If we want to remove this comment, then we intend that `err != nil`. But if that is the case, then we should probably be testing for it explicitly.

PR #3528, opened 2026-05-11
test(fvt): speedup functional test runs

dnwe:
A few commits to try and see if we can bring down the execution times for our FVTs as they consistently exceed 10-12 minutes each nowadays.


puellanivis (approved):
We could probably change some of the `for init ; test ; next` loops into `for range int` loops, but otherwise LGTM.

I’m doubting any of the tests could really be changed to using [synctest](https://pkg.go.dev/testing/synctest@go1.26.3)… it would likely require significant refactoring to reevaluate all the synchronization to ensure it can run essentially deterministically.

PR #3529, opened 2026-05-11
fix: scope metadata refresh topic errors

DCjanus:
## Background

`TestClientRefreshesMetadataConcurrently` can fail because topic-level metadata errors from one caller can leak to other callers sharing the same metadata refresh.

This has already been observed on `main`: [Unit Testing with Go stable](https://github.com/IBM/sarama/actions/runs/25613507187/job/75187553329).

Concurrent `RefreshMetadata` calls can be batched into the same metadata request, and that batching should be preserved. The issue is that the old shared result path can return one topic's error to every waiter. As a result, a caller that only requested a healthy topic can receive an unrelated `ErrUnknownTopicOrPartition` from another topic in the same refresh.

## Fix Approach

Keep the metadata refresh singleflight behavior, but make topic-level errors visible by topic.

The fix introduces an internal `topicErrorSet` error type that carries metadata errors keyed by topic. `updateMetadata` can then keep all topic-level errors from a shared refresh together instead of collapsing them into one arbitrary error.

When `Client.RefreshMetadata` returns, Sarama filters that internal error set by the topics requested by the current caller. This prevents unrelated topic errors from leaking across callers that share one metadata refresh, while non-topic refresh errors such as broker or network failures are returned unchanged.

## Validation

This PR uses staged commits to prove the fix:

- Before the fix, a temporary workflow reran the existing `TestClientRefreshesMetadataConcurrently` test 100 times and reproduced the failure: [Client Metadata Concurrent Refresh Repro](https://github.com/IBM/sarama/actions/runs/25667570545/job/75343978158)
- After the fix, the same temporary workflow passed: [Client Metadata Concurrent Refresh Repro](https://github.com/IBM/sarama/actions/runs/25667743862/job/75344572251)
- The final commit removes the temporary workflow so the PR diff stays focused.

Local validation:

```bash
go test -race -count=1 -run '^(TestClientReceivingUnknownTopic|TestClientRefreshesMetadataConcurrently)$' -v .
golangci-lint run
```

## Follow-up

This PR keeps Sarama's existing single-error return shape: when one caller requests multiple topics and several of them have topic-specific errors, Sarama still returns one relevant error.

A possible follow-up would be to use the standard library's `errors.Join` to aggregate these topic errors, so a multi-topic refresh can preserve more error information. However, directly joining raw `KError` values would still not tell callers which topic produced which error.

If we want to add that aggregation later, a more complete approach would likely encode the topic into the error value first. Then these topic-aware errors could be combined with `errors.Join`, preserving `errors.Is` checks for the underlying `KError` values while also keeping topic ownership visible in the error message.

That would change the returned text and error shape for multi-error cases. To keep this flaky-test fix focused, this PR does not introduce error aggregation behavior.


client.go
```diff
@@ -1146,6 +1147,7 @@ func (client *client) updateMetadata(data *MetadataResponse, allKnownMetaData bo
 		partitionCache[writablePartitions] = client.setPartitionCache(topic.Name, writablePartitions)
 		client.cachedPartitionsResults[topic.Name] = partitionCache
 	}
+	err = topicErrors.errOrNil()
```
puellanivis: Usually this is an `err()` or `Err()` receiver method. Saying “OrNil` is basically superfluous, because we should all know that `error` can be `nil` when no error.

Also, we can go ahead and `return retry, topicErrors.Err()` (This effectively does the assignment to the named returned values and then returns. Even better, even if we’ve somehow accidentally shadowed one of the values, the return will assign to the actual return values.)

dnwe:
Thanks. There's definitely an issue here that I introduced when I added the per-topic error returns, so thanks for raising it. I'm not 100% sure on this PR as the fix though, let me have a deeper look into this one and come back to you

client.go
DCjanus: Adjusted in [9c2fe7dc](https://github.com/IBM/sarama/pull/3529/commits/9c2fe7dc136587dc07c1e288000116ac01cd4fec). Renamed `errOrNil` to `Err` and removed the named returns from `updateMetadata`.

DCjanus:
> Thanks. There's definitely an issue here that I introduced when I added the per-topic error returns, so thanks for raising it. I'm not 100% sure on this PR as the fix though, let me have a deeper look into this one and come back to you

Feel free to treat this PR as a concrete repro and one possible fix direction, not necessarily the final shape.

I initially tried to write this up as an issue, but it was hard to keep the explanation both precise and concise. A PR seemed clearer because it shows the failure mode directly in code and CI.

metadata.go
```diff
@@ -1,11 +1,69 @@
...
 
+type topicErrorSet struct {
+	topicErrors map[string]error
+}
+
+func (e *topicErrorSet) Error() string {
+	if err := e.firstVisibleErrorFor(nil); err != nil {
+		return err.Error()
+	}
+	return ""
```
puellanivis: 🤔 We might want some sort of non-empty return here? So that if it ever (accidentally) comes up, one won’t be confused about how there’s an empty string error message.

puellanivis (commented):
I thought it would be interesting to note, this was essentially the code as it was before it was switched from `(retry []string, err error)` to `(retry bool, err error)`. It used a temporary map of `map[topic]none` to collect which topics that should be retried, and then did what would now be a `slices.Collect(maps.Keys(retryTopics))`.

metadata.go
DCjanus: Adjusted in [ea2bdd44](https://github.com/IBM/sarama/pull/3529/commits/ea2bdd443ef2cdde4e0f1a5834b0736bddaeaa1b). This now returns an explicit `unexpected empty metadata topic error set` fallback.

dnwe:
The alternative that I came up with is in https://github.com/IBM/sarama/pull/3532 - but I'm not sure if I'm happy with that or not either at the moment, particular the way multiple topic errors in the `.Error()` case just give a joined slice of the errors without referencing the topic names in the error string

puellanivis:
It’s now even more hilarious that both solutions converge on the same map solution that was discarded before. 😂 

Both are pretty solid. 🤷‍♀️ Either would be good.

DCjanus:
I'm going to close this in favor of #3532.

This PR helped make the issue concrete, and #3532 now covers the same problem with a more complete fix. Thanks for the review and discussion here.

PR #3530, opened 2026-05-11
test(fvt): only safeClose if we created

dnwe:
Move the defer safeClose after the err check else we risk a panic


puellanivis (approved):
Ooo! good catch!

dnwe:
:) thanks, happened to see it fail in one of my other PRs - I think it was masking another problem in that test though

PR #3532, opened 2026-05-11
fix(client): scope metadata refresh errors to requested topics

dnwe:
When single-flight is enabled, concurrent callers can piggy-back on a shared metadata refresh. Previously a topic-level error from any topic in the batch was returned to every caller, even those that did not ask about the failing topic. Track errors per topic and filter the returned error to the caller's topic set.

Thanks to @DCjanus for discovering the original problem and identifying the fix

### DIFF
--- a/client.go
+++ b/client.go
@@ -1073,9 +1073,9 @@ func (client *client) tryRefreshMetadata(topics []string, attemptsRemaining int,
 }
 
 // if no fatal error, returns a list of topics that need retrying due to ErrLeaderNotAvailable
-func (client *client) updateMetadata(data *MetadataResponse, allKnownMetaData bool) (retry bool, err error) {
+func (client *client) updateMetadata(data *MetadataResponse, allKnownMetaData bool) (bool, error) {
 	if client.Closed() {
-		return
+		return false, nil
 	}
 
 	client.lock.Lock()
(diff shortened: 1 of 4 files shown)


dnwe:
This is an alternative fix I came up with for a similar happy, but I'm not sure if I'm happy with it at the moment, particular the way multiple topic errors in the `.Error()` case just give a joined slice of the errors without referencing the topic names in the error string

client.go
```diff
@@ -1103,6 +1103,8 @@ func (client *client) updateMetadata(data *MetadataResponse, allKnownMetaData bo
 		client.metadataTopics = make(map[string]none)
 		client.cachedPartitionsResults = make(map[string][maxPartitionIndex][]int32)
 	}
+	topicErrs := refreshError{}
```
puellanivis: [style] I generally discourage using empty maps like this. An empty map/struct/slice body says “something will go here… maybe not now, but maybe eventually.” The idea is to align syntax with semantics. This format invites expansion.

However, if one truly wants an empty mutable zero value, often a better fit would be `make(map[K]V)`, `var slice []T`, or `s := new(STRUCT)` . These don’t suggest that any keys, values, or fields will be set in the future, because they are by their nature closed. It reduplicates that we want an empty thing, to be filled it or mutated later, but not now, and not ever now.

(No action necessary. It’s fine. Just noting an idea that has been kicking around a lot in my head.)

metadata.go
```diff
@@ -1,11 +1,65 @@
...
+	"fmt"
+	"maps"
+	"slices"
 	"sync"
 )
 
 type metadataRefresh func(topics []string) error
 
+type refreshError map[string]error
+
```
hindessm, line 14: I'd be tempted to include an `add` method:
```go
// addError adds an error for a topic to the set. The wrap
// preserves the underlying error for errors.Is/As.
func (e refreshError) addError(topic string, err error) {
	e[topic] = fmt.Errorf("%s: %w", topic, err)
}
```
Then you would avoid wrapping the errors on every use - i.e. wrapping again with each call to `errorForTopics`?

dnwe: Something like [178bac0](https://github.com/IBM/sarama/pull/3532/commits/178bac0c04c41bddbc7adf5ca1d78682fc5d02aa)?

hindessm (approved):

PR #3535, opened 2026-05-13
feat(consumer): warn on sustained partition retries

dnwe:
After a broker disconnect a partition consumer can sit retrying indefinitely. Each attempt's error is logged but nothing tells the operator the same partition is still failing ten retries later, so it's hard to tell a transient blip from a stuck partition.

Bump child.retries on every backoff (it was only incremented when a BackoffFunc was configured) and log every 10 consecutive failures.

Also note on ConsumeClaim that handlers should return when the session context is done; reading Messages() alone hangs while a partition keeps retrying.

---

We've had a few users complain about scenarios like this "I get EOF errors and then my consumers go silent and stop consuming any message until I restart my whole app", so try and make this clearer when its happening

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -455,6 +455,9 @@ type partitionConsumer struct {
 
 var errTimedOut = errors.New("timed out feeding messages to the user") // not user-facing
 
+// log every Nth consecutive failure (~20s apart at the default backoff)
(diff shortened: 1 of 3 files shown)


hindessm (approved):
LGTM

PR #3538, opened 2026-05-14
fix(consumer): decouple FetchRequest.MaxBytes from MaxResponseSize

### LINKED ISSUE #1370: Sarama Consumer should not check for MaxResponseSize in FetchResponse >= v3
##### Versions

Sarama Version: 35324cf48e33d8260e1c7c18854465a904ade249
Kafka Version: 1.1.1
Go Version: 1.12.0

##### Configuration

Not relevant 

##### Logs

```
kafka: error while consuming foo/61: kafka: error decoding packet: message of length 104857857 too large or too small
```

##### Problem Description

Per Kafka [KIP 74](https://cwiki.apache.org/confluence/display/KAFKA/KIP-74%3A+Add+Fetch+Response+Size+Limit+in+Bytes), Sarama should not [check for FetchResponse size](https://github.com/Shopify/sarama/blob/master/response_header.go#L18) when sending FetchRequest >= V3. Kafka broker may return up to 1 RecordBatch more messages than set in [FetchRequest](https://github.com/Shopify/sarama/blob/master/consumer.go#L871).

dnwe:
Per KIP-74, a v3+ broker may return up to one extra RecordBatch beyond the requested MaxBytes to guarantee progress when a single record exceeds the limit. Sending request.MaxBytes = MaxResponseSize left no headroom, so legitimate responses tripped the response-header size check with "message of length N too large or too small".

Introduce Consumer.Fetch.MaxBytes (default 50 MiB, matching the JVM client's fetch.max.bytes) for the request limit, leaving MaxResponseSize as the OOM safety net.

Fixes #1370

### DIFF
--- a/config.go
+++ b/config.go
@@ -405,6 +405,14 @@ type Config struct {
 			// (no limit). Similar to the JVM's `fetch.message.max.bytes`. The
 			// global `sarama.MaxResponseSize` still applies.
 			Max int32
+			// MaxBytes is the maximum number of bytes the broker should return for
+			// a fetch request across all partitions. Defaults to 50 MiB, matching
+			// the JVM client's `fetch.max.bytes`. This is the value sent as
+			// FetchRequest.MaxBytes for v3+ (KIP-74); the broker may return up to
+			// one extra RecordBatch beyond it to guarantee progress when a single
+			// record exceeds the limit, so keep this comfortably below the global
+			// `sarama.MaxResponseSize` safety net. Only used for Kafka >= 0.10.1.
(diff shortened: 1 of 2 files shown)


hindessm (approved):
LGTM

dnwe:
I'm going to add a 'breaking' label to this one so its flagged up in the release notes. It's not technically a _breaking_ change, but it is a change in behaviour that by default we'll only fetch up to 50MB in a single fetch request where before we could have fetched up to 100MB

PR #3539, opened 2026-05-14
feat(consumer): cap partition consumer retries

dnwe:
A partition consumer currently retries forever after a broker disconnect or other dispatch failure. A single stuck partition can thus keep the consumer-group session alive (no rebalance) while the application sees no messages.

Add Consumer.Retry.Max as an opt-in maximum on the number of dispatch failures. When the max is hit the partition consumer sends ErrConsumerRetriesExhausted on its Errors channel and closes itself; inside a consumer group the deferred sess.cancel() then ends the session and triggers a fresh rejoin, picking up new metadata and a healthy broker. Defaults to 0 (unlimited) to keep existing behaviour.

### DIFF
--- a/config.go
+++ b/config.go
@@ -382,6 +382,12 @@ type Config struct {
 			// more sophisticated backoff strategies. This takes precedence over
 			// `Backoff` if set.
 			BackoffFunc func(retries int) time.Duration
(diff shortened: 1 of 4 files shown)


hindessm (approved):
LGTM

PR #3540, opened 2026-05-14
feat(producer): add Produce v8 request/response support

dnwe:
Adds protocol support for KIP-467 (Kafka 2.4.0) which extends the Produce response with per-record errors and a batch-level error message, and bumps the negotiated Produce version to 8 when the configured Kafka version is at least 2.4.0.

Also makes MockBroker's default request handler match the queued response version to the request version, so tests using broker.Returns() don't have to manually bump version numbers.

### DIFF
--- a/produce_response.go
+++ b/produce_response.go
@@ -22,10 +22,20 @@ import (
 
 // partition_responses in protocol
 type ProduceResponseBlock struct {
-	Err         KError    // v0, error_code
-	Offset      int64     // v0, base_offset
-	Timestamp   time.Time // v2, log_append_time, and the broker is configured with `LogAppendTime`
-	StartOffset int64     // v5, log_start_offset
+	Err          KError                       // v0, error_code
+	Offset       int64                        // v0, base_offset
+	Timestamp    time.Time                    // v2, log_append_time, and the broker is configured with `LogAppendTime`
+	StartOffset  int64                        // v5, log_start_offset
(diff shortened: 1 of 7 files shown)


produce_response.go
```diff
@@ -22,10 +22,20 @@ import (
...
+	Timestamp    time.Time                    // v2, log_append_time, and the broker is configured with `LogAppendTime`
+	StartOffset  int64                        // v5, log_start_offset
+	RecordErrors []ProduceResponseRecordError // v8, record_errors (KIP-467)
+	ErrorMessage *string                      // v8, error_message (KIP-467)
+}
+
+// ProduceResponseRecordError identifies a record within a produced batch that
+// caused the whole batch to be dropped, along with the per-record error
+// message. Added in Produce response v8 (KIP-467).
+type ProduceResponseRecordError struct {
```
hindessm, line 36: I was a little surprised that you didn't create `encode` and `decode` methods for this struct. I thought that was the typical pattern followed by the protocol code. (However, 20% of the similar structs don't follow this pattern either so perhaps it is already too late for consistency here.)

hindessm (approved):
LGTM. Minor comment about something that surprised me reading the code.

produce_response.go
dnwe: Yeah I did wonder whether to or not, but with only two fields it was easier to just do inline. Maybe I’ll do a follow-up PR to do some of that

PR #3541, opened 2026-05-14
fix(consumer): size partial-batch retry correctly

### LINKED ISSUE #1657: Exact fetch size to retry partial record set delivery
The handling of partial delivery of a record set can be improved.
instead of trying to double the fetch size and try again: 
https://github.com/Shopify/sarama/blob/v1.26.1/consumer.go#L602-L609
we can make use of the size returned from the partial delivery:
https://github.com/Shopify/sarama/blob/v1.26.1/fetch_response.go#L86
i.e. in addition to the `Partial` boolean flag
https://github.com/Shopify/sarama/blob/v1.26.1/fetch_response.go#L40
we can add a `RawSize` field and use it in the next fetch request.

Each fetch should request at least 12 bytes to ensure `RawSize` is included in the fetch response.

dnwe:
When a fetch response contains a partial trailing record batch, the broker already encodes the batch's full size in its length field. Capture it during decode and use it to size the follow-up fetch instead of just doubling Consumer.Fetch.Default until the batch fits.

Fixes #1657

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -866,7 +866,13 @@ func (child *partitionConsumer) parseResponse(response *FetchResponse) ([]*Consu
 				child.sendError(ErrMessageTooLarge)
 				child.offset++ // skip this one so we can keep processing future messages
 			} else {
-				child.fetchSize *= 2
+				// if the broker told us the exact size of the partial batch, request that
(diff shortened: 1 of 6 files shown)


hindessm (approved):
LGTM

PR #3542, opened 2026-05-14
test(fvt): register restart cleanup before stop

dnwe:
Even if stop itself fails, ensure the broker is restarted again before the next test runs - otherwise subsequent setupFunctionalTest calls sit in ensureFullyReplicated waiting on a dead broker.


hindessm (approved):
LGTM

PR #3543, opened 2026-05-15
fix: flexible decoder out-of-bounds panic

hindessm:
Using same check as non-flexible decoder and add a regression test

Thanks to Dominik Blain of Cobalt AI for reporting


dnwe (approved):
Thanks!

PR #3545, opened 2026-05-16
feat(consumer): add OffsetCommit v8 request/response support

dnwe:
OffsetCommit v8 is the first flexible version of the API (KIP-482), introduced in Kafka 2.4.0. Wire encoding switches to compact strings, compact arrays, and per-struct tagged-fields trailers; request and response headers gain a tagged-fields trailer too (request header v2, response header v1).

Bumps both call sites (NewOffsetCommitRequest for the admin path and offsetManager.constructRequest for the consumer-group path) to negotiate v8 when Config.Version is at least V2_4_0_0.

### DIFF
--- a/offset_commit_request.go
+++ b/offset_commit_request.go
@@ -30,7 +30,11 @@ func (b *offsetCommitRequestBlock) encode(pe packetEncoder, version int16) error
 		pe.putInt32(b.committedLeaderEpoch)
 	}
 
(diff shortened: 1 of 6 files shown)


hindessm (approved):
LGTM. Funnily enough I had started this too. My code was identical except that I handled the `if count == 0` cases by just added the empty tagged field handling and still returning early.

Issue #2464, opened 2023-04-14
sarama consumer client can not return to leader when the client fetch from follower which is outed of isr


mtj075:
now, there maybe have a problem, when the  sarama client configured the rackid, it will fetch message from the matched follower replica.  When the consumer After the replica subscribed by the end falls out of the ISR, the sarama client can not notice. Once the replica  falls out of the ISR and does not return to the ISR for a long time, resulting in a backlog of client lag.

I think we can remove followers who have fallen out of ISR from the subscription by listening to ISR on the client side.When the dropped replica returns to the ISR, the leader will return to the corresponding follower for the client to pull data.

What needs to be guaranteed is that the client version > 3.2.3.  This version guarantees that the replica returned by the leader is in the isr.  https://github.com/apache/kafka/pull/12877/files

mtj075:
收到

dnwe:
This should be fixed by #3553 which is included in the latest release (v1.50.0)

mtj075:
收到

PR #3553, opened 2026-05-17
fix(consumer): lease preferred read replicas

dnwe:
The Java client treats a preferred read replica as a lease: a consumer may fetch from the follower for a bounded period, then it goes back to the leader so the broker can re-evaluate the replica choice.

Do the same in Sarama. Expire the preferred replica after Metadata.RefreshFrequency, and use the default metadata refresh interval as the lease when periodic metadata refresh is disabled. That avoids the stale-follower case without throwing away the benefit of preferred replica reads.

Refs: https://github.com/IBM/sarama/issues/2464

### DIFF
--- a/config.go
+++ b/config.go
@@ -13,7 +13,10 @@ import (
 	"golang.org/x/net/proxy"
 )
 
-const defaultClientID = "sarama"
+const (
+	defaultClientID                 = "sarama"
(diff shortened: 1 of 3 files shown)


hindessm (approved):
LGTM

Issue #2735, opened 2023-11-28
We probably need an example of how to use the AlterPartitionReassignments function


VinGarcia:
#### Description

Right now I needed to change the replication factor of an existing topic, and I found this issue here that discuss this issue:

https://github.com/IBM/sarama/issues/1238

This issue culminated on a [PR](https://github.com/IBM/sarama/pull/1617) that added the function:

- [AlterPartitionReassignments](https://pkg.go.dev/github.com/IBM/sarama#Broker.AlterPartitionReassignments)

But the API is not intuitive since the request object has a `AddBlock` function that expects an argument `replicas []int32`, and I was (naively) expecting a function that would receive just the number of desired replicas and update the topic accordingly.

So what I really need right now is to know how to use this function for changing the number of replicas, but we could take this opportunity to also add an example of how to do this to either the example directory or the GoDoc.

sanathkumarbs:
+1

stephanpelikan:
+1

dnwe:
Example has been added to [Version 1.49.0 (2026-05-18)](https://github.com/IBM/sarama/releases/tag/v1.49.0) - hopefully that meets your needs - please re-open the issue if not

PR #3556, opened 2026-05-18
docs: add AlterPartitionReassignments example and functional test

dnwe:
Demonstrates how to grow a topic's replication factor by constructing the per-partition replica list and polling ListPartitionReassignments until the move completes.

Refs #2735


hindessm (approved):
LGTM

PR #3560, opened 2026-05-18
refactor: replace eapache/queue with generic ring buffer

dnwe:
Drop the github.com/eapache/queue dep for an internal generic Queue[T] of the same shape. The async producer's two queues now hold typed pointers instead of interface{}, halving the backing buffer footprint and dropping four type assertions.

benchstat on an M1 Pro, n=12, -benchtime=1s:

| benchmark      | eapache      | generic      | delta             |
|----------------|--------------|--------------|-------------------|
| SteadyState-10 | 38.39n ± 0%  | 37.92n ± 0%  | -1.24% (p=0.000)  |
| Burst64-10     | 694.3n ± 1%  | 502.6n ± 1%  | -27.61% (p=0.000) |
| Burst1024-10   | 11.368µ ± 6% | 7.956µ ± 0%  | -30.02% (p=0.000) |
| Burst16384-10  | 292.3µ ± 1%  | 191.7µ ± 1%  | -34.40% (p=0.000) |
| Window1024-10  | 38.56n ± 1%  | 38.31n ± 4%  | ~ (p=0.225)       |
| geomean        | 1.278µ       | 1.022µ       | -20.08%           |

| benchmark      | eapache     | generic     | delta             |
|----------------|-------------|-------------|-------------------|
| SteadyState-10 | 82.00 B     | 82.00 B     | ~ (p=1.000)       |
| Burst64-10     | 2.375 KiB   | 1.125 KiB   | -52.63% (p=0.000) |
| Burst1024-10   | 54.00 KiB   | 27.12 KiB   | -49.77% (p=0.000) |
| Burst16384-10  | 776.0 KiB   | 392.4 KiB   | -49.44% (p=0.000) |
| Window1024-10  | 82.00 B     | 82.00 B     | ~ (p=1.000)       |
| geomean        | 3.639 KiB   | 2.383 KiB   | -34.53%           |

### DIFF
--- a/internal/queue/queue.go
+++ b/internal/queue/queue.go
@@ -3,17 +3,13 @@
 package queue
 
 // power of two so wrap-around can use bitwise AND
-const minLen = 16
+const minLen = 1 << 4
 
-// Queue is a generic FIFO ring buffer.
+// Queue is a generic FIFO ring buffer. The zero value is an empty queue ready
+// for use.
 type Queue[T any] struct {
 	buf               []T
 	head, tail, count int
-}
-
-// New returns an empty Queue.
-func New[T any]() *Queue[T] {
-	return &Queue[T]{buf: make([]T, minLen)}
 }
 
 // Length returns the number of elements in the queue.
@@ -21,9 +17,10 @@ func (q *Queue[T]) Length() int {
 	return q.count
 }
 
-// resize rebuilds buf at 2*count with head back at index 0.
+// resize rebuilds buf with head back at index 0, sized to 2*count but never
+// below minLen so the zero value of Queue grows on first Add.
 func (q *Queue[T]) resize() {
-	newBuf := make([]T, q.count<<1)
+	newBuf := make([]T, max(minLen, q.count<<1))
 	if q.tail > q.head {
 		copy(newBuf, q.buf[q.head:q.tail])
 	} else {
--- a/internal/queue/queue.go
+++ b/internal/queue/queue.go
@@ -1,0 +1,71 @@
+// Package queue is a generic FIFO ring buffer adapted from
+// github.com/eapache/queue. It is not safe for concurrent use.
+package queue
+
+// power of two so wrap-around can use bitwise AND
+const minLen = 16
+
+// Queue is a generic FIFO ring buffer.
+type Queue[T any] struct {
+	buf               []T
+	head, tail, count int
+}
+
+// New returns an empty Queue.
+func New[T any]() *Queue[T] {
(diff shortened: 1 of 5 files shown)


internal/queue/queue.go
```diff
@@ -0,0 +1,71 @@
+// Package queue is a generic FIFO ring buffer adapted from
+// github.com/eapache/queue. It is not safe for concurrent use.
+package queue
+
+// power of two so wrap-around can use bitwise AND
+const minLen = 16
```
puellanivis: 🤔 I think expressing this as a `1 << n` value might make the syntax more reflective of the need to be a power of two.

internal/queue/queue.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Length returns the number of elements in the queue.
+func (q *Queue[T]) Length() int {
+	return q.count
+}
+
+// resize rebuilds buf at 2*count with head back at index 0.
+func (q *Queue[T]) resize() {
+	newBuf := make([]T, q.count<<1)
```
puellanivis: 🤔 We could replace this with `max(minLen, q.count<<1)` and get auto-initialization from plain zero value, and then we wouldn’t need a `New[T]` function. Just an idea.

puellanivis (approved):
Just lint

internal/queue/queue.go
dnwe: Good idea!

internal/queue/queue_test.go
```diff
@@ -0,0 +1,105 @@
...
+	t.Run("works with pointer types and clears slot on remove", func(t *testing.T) {
+		type box struct{ v int }
+		var q Queue[*box]
+		a, b := &box{v: 1}, &box{v: 2}
+		q.Add(a)
+		q.Add(b)
+		got := q.Remove()
+		require.NotNil(t, got)
+		assert.Equal(t, 1, got.v)
+		// the just-vacated slot must hold the zero value so the referent can
```
hindessm: I'm not sure I understand this test. If I delete the lines:
```go
	var zero T
	q.buf[q.head] = zero // clear slot so the popped element can be GC'd
```
then the test still passes.

You write that "we can't inspect head directly" which would be true if we were black box testing but it's our test and we can do white box testing if we want and could just test it directly with:
```go
		head := q.head
		got := q.Remove()
		require.NotNil(t, got)
		assert.Equal(t, 1, got.v)
		require.Nil(t, q.buf[head])
```

hindessm (approved):

hindessm (approved):

PR #3562, opened 2026-05-19
test(fvt): use a per-message timeout in follower failover test

dnwe:
Add a 30s timeout on each pc.Messages() read so the test fails fast rather than running out the 15m binary timeout, which panics and also defeats gotestsum's --rerun-fails.

Also wait for the follower to drop out of the ISR after we stop it. Kafka can be slow about this for some reason (saw 14+ minutes once) and the leader won't advance HWM until it happens, so the consumer has nothing to make progress on.

### DIFF
--- a/functional_consumer_follower_fetch_test.go
+++ b/functional_consumer_follower_fetch_test.go
@@ -5,10 +5,14 @@ package sarama
 import (
 	"context"
 	"fmt"
+	"slices"
 	"strconv"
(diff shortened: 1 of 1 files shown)


hindessm (approved):
LGTM

puellanivis (commented):
Wondering which tests could benefit from being rewritten with `synctest`.

PR #3565, opened 2026-05-20
feat(admin): support OffsetFetchRequest v8

### LINKED ISSUE #2827: Support batch offset fetch requests
Since OffsetFetch API version 8, it supports fetching offsets for multiple consumer groups in a single request as introduced in [KIP-709](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=173084258).
It seems like Sarama currently does not support fetching offsets for multiple consumer groups (<= OffsetFetch API version 7).

We need to regularly fetch current offsets for multiple consumer groups for monitoring purposes, and it looks like this could help reduce the overhead of making a separate request for each consumer group every time.

Would it be possible to add support for the recent version of the OffsetFetch API?

dnwe:
This is quite tricky to support in a backwards compatible way because the previous API only supported fetching offsets for a single group and had top-level constructs for that, but from v8 onwards you can fetch for multiple groups in one request. We want to support that, but need to continue to support the old method and expectations around single group usage too, so we map to/from the fields on the fly, using Groups[0] in the v8 case. Eventually we'd want users to migrate entirely to the new APIs and we could remove the old ones.

Fixes #2827

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -115,6 +115,17 @@ type ClusterAdmin interface {
 	// List the consumer group offsets available in the cluster.
 	ListConsumerGroupOffsets(group string, topicPartitions map[string][]int32) (*OffsetFetchResponse, error)
 
+	// ListConsumerGroupOffsetsBatch fetches committed offsets for multiple consumer groups
+	// in a single round trip per coordinator using OffsetFetch v8+ (KIP-709). A nil
+	// partitions map fetches offsets for all topics in that group.
+	//
+	// Returns ErrUnsupportedVersion if any coordinator advertises < v8. Requires
+	// Config.ApiVersionsRequest to be enabled for the version check to take effect.
+	//
+	// On a retriable per-group error all coordinators are re-resolved and every group
+	// is re-batched. Non-retriable per-group errors are returned in the per-group Err.
+	ListConsumerGroupOffsetsBatch(groupTopics map[string]map[string][]int32) (map[string]*OffsetFetchResponseGroup, error)
+
 	// ListOffsets lists offsets for the specified topic partitions.
 	// Each value is OffsetNewest, OffsetOldest, or a timestamp in milliseconds.
 	// Results are keyed by topic/partition and include per-partition errors.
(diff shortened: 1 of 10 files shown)


broker.go
```diff
@@ -1172,6 +1172,34 @@ func (b *Broker) sendAndReceive(req protocolBody, res protocolBody) error {
...
+
+	r := b.brokerAPIVersions[pb.key()]
+	if r == nil {
+		pb.setVersion(minVersion)
+		return minVersion, true
+	}
+	if r.maxVersion < minVersion {
+		return 0, false
+	}
+	for v := r.maxVersion; v >= minVersion; v-- {
```
DCjanus: Small question: is `negotiateApiVersion` meant to be a generic helper, or only for this OffsetFetch path?

If it is meant to be generic, maybe the lower bound should also include the broker-advertised `minVersion`.

For example, current Kafka trunk has [Fetch `[4,18]`](https://github.com/apache/kafka/blob/f7dbf0bf3bbbf7d0e035f72b3370d347361e53ae/clients/src/main/resources/common/message/FetchRequest.json#L61). If the same logic were used for Fetch with a client that only supports up to v3, it could still pick v3, even though a Kafka 4.x broker no longer supports it.

This is unlikely in practice since we probably would not use such an old request version, but accounting for the broker `minVersion` would make the helper match the expected range-overlap behavior.

dnwe: Let me rework this a little as I also realised we could probably re-purpose the existing 'restrictApiVersion' code that already comes from the ApiVersionsResponse we have in broker.go

dnwe:
@puellanivis / @hindessm please could one of you take a look at the latest state of the PR here?

I was planning to cut another version once a few more PRs have landed, but wanted to include this one ideally

admin.go
```diff
@@ -1215,6 +1226,68 @@ func (ca *clusterAdmin) ListConsumerGroupOffsets(group string, topicPartitions m
...
+			}
+		}()
+
+		// re-resolve coordinators each attempt; key by broker id to coalesce groups
+		// sharing a coordinator
+		batches := make(map[int32]*brokerBatch)
+		for group, partitions := range groupTopics {
+			coordinator, cerr := ca.client.Coordinator(group)
+			if cerr != nil {
+				return cerr
```
puellanivis: Using `cerr` here doesn’t seem to be necessary? We don’t use any `err` value in the body that would be hurt by just using `err` alone?

admin.go
```diff
@@ -1215,6 +1226,68 @@ func (ca *clusterAdmin) ListConsumerGroupOffsets(group string, topicPartitions m
...
+			}
+			batch, ok := batches[coordinator.ID()]
+			if !ok {
```
puellanivis: [Testing for validity also tests for existence] Testing for `batch == nil ̀  might be the better choice here?

admin.go
```diff
@@ -1215,6 +1226,68 @@ func (ca *clusterAdmin) ListConsumerGroupOffsets(group string, topicPartitions m
...
+				batch = &brokerBatch{broker: coordinator}
+				batches[coordinator.ID()] = batch
+			}
+			batch.groups = append(batch.groups,
+				OffsetFetchRequestGroup{GroupId: group, Partitions: partitions})
+		}
+
+		clear(result)
```
puellanivis, line 1262: Recently saw some bugs on this. The function `clear` can take a while to run if there’s high bucket cardinality. Though, I imagine we’re not in any particular danger here? So, we get the advantage of reusing maps on any errors.

admin.go
```diff
@@ -1215,6 +1226,68 @@ func (ca *clusterAdmin) ListConsumerGroupOffsets(group string, topicPartitions m
...
+		for _, batch := range batches {
+			req := NewOffsetFetchRequest(ca.conf.Version, "", nil)
+			req.Groups = batch.groups
+			if _, ok := batch.broker.negotiateApiVersion(req, 8); !ok {
+				return ErrUnsupportedVersion
+			}
+			resp, ferr := batch.broker.FetchOffset(req)
+			if ferr != nil {
+				return ferr
+			}
```
puellanivis: Same as above, there does not appear to be any need to avoid shadowing/stomping `err` here.

offset_fetch_request.go
```diff
@@ -192,15 +315,47 @@ func (r *OffsetFetchRequest) requiredVersion() KafkaVersion {
...
 
 	r.partitions[topic] = append(r.partitions[topic], partitionID)
 }
+
+// AddGroupPartition adds a partition under group in a v8+ batch, creating the
+// group entry if absent.
+func (r *OffsetFetchRequest) AddGroupPartition(group, topic string, partitionID int32) {
+	for i := range r.Groups {
+		if r.Groups[i].GroupId == group {
+			r.Groups[i].AddPartition(topic, partitionID)
```
puellanivis, line 361: [just an idea] The indexing here into the slice rather than simply using a `for _, group := range r.Groups` suggests that maybe we want to use a slice off pointers, rather than a slice of values. (In another place in the PR, we’re converting a slice of values into a slice of pointers by referencing the individual elements of the slice of values, this is definitely a good way to allocate a whole block of same-lifetime objects in bulk.)

offset_fetch_response.go
```diff
@@ -58,11 +58,41 @@ func (b *OffsetFetchResponseBlock) encode(pe packetEncoder, version int16) (err
...
+	partitions[partition] = block
+}
+
+func (g *OffsetFetchResponseGroup) GetBlock(topic string, partition int32) *OffsetFetchResponseBlock {
+	if g.Blocks == nil {
+		return nil
+	}
+	if g.Blocks[topic] == nil {
+		return nil
+	}
```
puellanivis: [unnecessary(?) nil-guards] since a `nil` map access is the same as an empty map access, these nil guards don’t change any significant behavior, they just short-circuit empty entries… but empty entries would still cause the same issue.

What with branch prediction and all, the plain `return g.Blocks[topic][partition]` might even be faster, since it has less branches.

offset_fetch_response.go
```diff
@@ -234,7 +409,28 @@ func (r *OffsetFetchResponse) GetBlock(topic string, partition int32) *OffsetFet
...
+	return group.GetBlock(topic, partition)
+}
+
+// AddBlock adds a block for topic/partition. On v8+ it appends under Groups[0],
+// creating it if needed.
 func (r *OffsetFetchResponse) AddBlock(topic string, partition int32, block *OffsetFetchResponseBlock) {
+	if r.Version >= 8 {
+		if len(r.Groups) == 0 {
+			r.Groups = []OffsetFetchResponseGroup{{}}
+		}
```
puellanivis, line 424: 😂 It looks kind of silly, but I get that it’s correct.

puellanivis (approved):
It’s all just lint.

hindessm (commented):
LGTM.

However, given that we have errors like `PacketEncodingError{"requireStable is not supported. use version 7 or later"}`, I wonder if we should also add some errors for some of the other unexpected combinations - such as `len(r.Groups) > 1 && r.Version < 8`, `len(r.Groups) > 0 && r.ConsumerGroup != ""`, `r.ConsumerGroup != "" && r.Version >= 8` (implied by the comment at line 19), etc. It might make it easier to transition

dnwe:
> LGTM.
> 
> However, given that we have errors like `PacketEncodingError{"requireStable is not supported. use version 7 or later"}`, I wonder if we should also add some errors for some of the other unexpected combinations - such as `len(r.Groups) > 1 && r.Version < 8`, `len(r.Groups) > 0 && r.ConsumerGroup != ""`, `r.ConsumerGroup != "" && r.Version >= 8` (implied by the comment at line 19), etc. It might make it easier to transition

Thanks, good idea - have added those

hindessm (approved):

PR #3566, opened 2026-05-20
fix(client): don't log ErrNoTopicsToUpdateMetadata every tick

### LINKED ISSUE #2630: backgroundMetadataUpdater can log ErrNoTopicsToUpdateMetadata repetitively if client.conf.Metadata.Full is false

dnwe:
With Metadata.Full=false and no topics registered yet, the background updater spams the log on every refresh. Skip that error and only log real ones.

Fixes #2630

### DIFF
--- a/client.go
+++ b/client.go
@@ -939,6 +939,10 @@ func (client *client) backgroundMetadataUpdater() {
 		select {
 		case <-ticker.C:
 			if err := client.refreshMetadata(); err != nil {
+				// expected no-op, not worth logging every tick
+				if errors.Is(err, ErrNoTopicsToUpdateMetadata) {
+					continue
+				}
 				Logger.Println("Client background metadata update:", err)
 			}
 		case <-client.closer:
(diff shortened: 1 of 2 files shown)


puellanivis (approved):
I suppose, since it’s in a `<-ticker.C` there’s no worries about a tight spinloop if the channel ever closes.

Looks good. 👍

hindessm (approved):
LGTM

PR #3567, opened 2026-05-21
feat: add Java-compatible murmur2 partitioner

### LINKED ISSUE #3537: Java-compatible murmur2 partitioner to ensure cross-language partition affinity
#### Summary
We are using Sarama as the Kafka producer in a Go service that runs alongside a Java-based Kafka producer (Apache Kafka client). Both producers publish messages to the same topic using a partition key. We need messages with the same key to always land on the same partition regardless of which producer sends them.

#### Problem
Sarama's default partitioner (NewHashPartitioner) and even NewReferenceHashPartitioner use FNV-1a as the underlying hash function. The Apache Kafka Java client uses murmur2 by default. This means the same key produces a different partition number depending on which producer publishes the message, breaking consumer affinity.
Current workaround
We implemented a custom partitioner that replicates Java's Utils.murmur2() algorithm and applies the same (hash & 0x7fffffff) % numPartitions calculation. While this works, it introduces a maintenance burden — any change to the hashing logic in the Java client would need to be manually mirrored in our codebase. During any such transition window, partition affinity would silently break.
Request
Add a built-in NewJavaCompatiblePartitioner (or equivalent) to Sarama that implements murmur2 matching the Apache Kafka Java client's DefaultPartitioner behaviour, so cross-language partition consistency is guaranteed by the library without requiring custom implementations.
References
	Apache Kafka Java client: org.apache.kafka.common.utils.Utils.murmur2()
	Sarama: partitioner.go — NewReferenceHashPartitioner uses fnv.New32a()

dnwe:
Add NewMurmur2Partitioner which replicates the Apache Kafka Java client's DefaultPartitioner hashing algorithm (Utils.murmur2), using the formula (murmur2(key) & 0x7fffffff) % numPartitions.

This allows Go producers to have partition affinity with Java producers without requiring users to supply their own custom partitioner.

Also adds unit tests covering raw hash values and known partition assignments, and a functional interop test that verifies Sarama's murmur2 partitioner routes keyed messages to the same partitions as the Java console producer.

Closes #3537

### DIFF
--- a/murmur2.go
+++ b/murmur2.go
@@ -1,0 +1,42 @@
+package sarama
+
+// murmur2 implements the same hashing algorithm used by the Apache Kafka Java
+// client's DefaultPartitioner (org.apache.kafka.common.utils.Utils.murmur2).
+// It is a variant of MurmurHash2 with a fixed seed of 0x9747b28c and
+// little-endian byte ordering.
+//
+// Reference:
+//   - https://github.com/apache/kafka/blob/4.3.0/clients/src/main/java/org/apache/kafka/common/utils/Utils.java#L496-L541
+//   - https://github.com/aappleby/smhasher/blob/master/src/MurmurHash2.cpp
+func murmur2(b []byte) uint32 {
+	const (
+		seed uint32 = 0x9747b28c
+		m    uint32 = 0x5bd1e995
+		r           = 24
+	)
+	h := seed ^ uint32(len(b))
+	for len(b) >= 4 {
+		k := uint32(b[3])<<24 + uint32(b[2])<<16 + uint32(b[1])<<8 + uint32(b[0])
+		b = b[4:]
+		k *= m
+		k ^= k >> r
+		k *= m
+		h *= m
+		h ^= k
+	}
+	switch len(b) {
+	case 3:
+		h ^= uint32(b[2]) << 16
+		fallthrough
+	case 2:
+		h ^= uint32(b[1]) << 8
+		fallthrough
+	case 1:
+		h ^= uint32(b[0])
+		h *= m
+	}
+	h ^= h >> 13
+	h *= m
+	h ^= h >> 15
+	return h
+}
--- a/murmur2.go
+++ b/murmur2.go
@@ -1,5 +1,7 @@
 package sarama
 
+import "encoding/binary"
+
 // murmur2 implements the same hashing algorithm used by the Apache Kafka Java
 // client's DefaultPartitioner (org.apache.kafka.common.utils.Utils.murmur2).
 // It is a variant of MurmurHash2 with a fixed seed of 0x9747b28c and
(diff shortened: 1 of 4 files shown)


puellanivis (commented):
Curious why it’s implemented as a `func([]byte) uint32` rather than as a `hash.Hash32`?

I suppose for simplisity of the implementation… but it’s just kind of odd to not have it as a `hash.Hash` compatible type.

murmur2.go
```diff
@@ -0,0 +1,42 @@
...
+//   - https://github.com/aappleby/smhasher/blob/master/src/MurmurHash2.cpp
+func murmur2(b []byte) uint32 {
+	const (
+		seed uint32 = 0x9747b28c
+		m    uint32 = 0x5bd1e995
+		r           = 24
+	)
+	h := seed ^ uint32(len(b))
+	for len(b) >= 4 {
+		k := uint32(b[3])<<24 + uint32(b[2])<<16 + uint32(b[1])<<8 + uint32(b[0])
```
hindessm: It doesn't really matter much here but the compiler does a better job of optimising this (at least on amd64) if you do:
```suggestion
		k := binary.LittleEndian.Uint32(b)
```

partitioner_test.go
```diff
@@ -318,6 +318,113 @@ func TestWithCustomFallbackPartitioner(t *testing.T) {
...
+	}{
+		{"foo", 597841616},
+		{"bar", 988958145},
+		{"baz", 1973573280},
+		{"", 275646681},
+		{"kafka", 3496464228},
+		{"sarama", 770765511},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.input, func(t *testing.T) {
```
hindessm: The empty string doesn't make a good test name - `TestMurmur2#00` - so I'd do something like:
```suggestion
		t.Run(fmt.Sprintf("input=%q", tc.input), func(t *testing.T) {
```

hindessm (commented):
Looks good but I added a couple of minor comments for your consideration.

partitioner_test.go
puellanivis: The name gets converted into a Go-compatible-type identifier. I would generally recommend against using a name like this, instead of a more meaningful one.

If we want to record the input, we can always use `t.Logf("input=%q", input)`.

dnwe:
> Curious why it’s implemented as a `func([]byte) uint32` rather than as a `hash.Hash32`?
> 
> I suppose for simplisity of the implementation… but it’s just kind of odd to not have it as a `hash.Hash` compatible type.

Yeah, just that the simpler `func([]byte) uint32` signature avoids the    Write/Sum32/Reset ceremony for what is a single one-shot hash per partition call. I’m happy to see what that would look like in a follow-on PR thoigh

dnwe:
Actually thinking about it some more, hash.Hash32 is really meant for streaming data, but with murmur2 you need the full length up front to commence the hash, so you can never truly stream it, you have to cache the content in memory. So for the Write interface of hash.Hash32 we’d just end up alloc’ing and copying the short []byte we already had for no obvious benefit

hindessm (approved):

puellanivis:
I mean, the most it would ever have to store is a `[4]byte`… your own implementation is throwing away the buffer as it iterates through it.

Also, the `Sum` receiver method of the `hash.Hash` is there for reusable hashing off a common salt… each execution can run in parallel, and does not change the underlying hash state. Of course, it still returns the result as a `[]byte` that we would have to convert to an `uintXY` anyways.

Not sure why they didn’t make `hash.Hash32` take a byte-slice the same as `Sum`… it would help in cases like this, where we just need an arbitrary type to hold the implementation.

P.S.: Yeah, though. It would be better as a further PR.

PR #3569, opened 2026-05-21
fix(admin): cap batched OffsetFetch version

DCjanus:
This is a draft helper PR for #3565, not a replacement for it. Feel free to copy or adapt the test/fix into #3565 and close this PR.

## Problem

`ListConsumerGroupOffsetsBatch` currently picks the broker's maximum OffsetFetch version. Kafka 4.2 reports OffsetFetch versions above v8, while #3565 only implements v8. The batch request can then fail during encoding before it is sent.

## Change

The commits are intentionally split into three steps:

- fba831c755b494cdc055c68ba071028f870cfc36 adds the FVT regression test and a temporary focused workflow. This shows the new test can fail before the fix.
- 950ca2643107eefa5e067b6a4e9839ce73b08e15 adds only the production fix, without changing the test.
- 76d6e3c1de65fb1315a20b11ffc194f9c6bb4830 removes the temporary workflow, so the final diff keeps only the regression test and the fix.

This keeps the repro, fix, and cleanup separate while leaving no temporary workflow in the final patch.

## Validation

GitHub Actions was unreliable while preparing this draft PR, so I am not using workflow job URLs as evidence here.

Local validation:

- before the fix, the focused Kafka 4.2 FVT failed with `invalid or unsupported OffsetFetchRequest version field`
- after the fix, the same focused Kafka 4.2 FVT passed
- `go test ./...` passed


dnwe:
Thanks for spotting this. I’ve added a couple of tests and plugged it in a similar way to how we choose the apiVersions elsewhere by walking down IsValidVersion checks 

see https://github.com/IBM/sarama/pull/3565/changes/4e035f3fb8c389f5ea531bcd9b7c11c01dde189d

PR #3575, opened 2026-05-22
feat(consumer_group): set cancellation cause on session context

prakhar7651:
## Motivation

When a consumer group session context is cancelled, consumers cannot distinguish
between a rebalance, a heartbeat failure, a fenced instance, or a ConsumeClaim
exit — all they see is `context.Canceled`. This makes it difficult to implement
graceful handling (e.g. aborting in-flight work quickly on rebalance vs. logging
an error on heartbeat failure) and limits observability.

## Changes

Replace `context.WithCancel` with `context.WithCancelCause` in
`newConsumerGroupSession`, and pass a descriptive cause at each call site that
cancels the session. Consumers can now call `context.Cause(sess.Context())` to
determine why the session ended.

### Cancellation causes

| Cause | When set |
|---|---|
| `ErrRebalanceInProgress` | Coordinator heartbeat response |
| `ErrUnknownMemberId` | Coordinator heartbeat response |
| `ErrIllegalGeneration` | Coordinator heartbeat response |
| `ErrFencedInstancedId` | Coordinator heartbeat response, or offset commit |
| `ErrSessionHeartbeatFailed` | Heartbeat loop exited after retries exhausted (new sentinel) |
| `ErrSessionPartitionCountChanged` | Leader detects partition count change for a subscribed topic (new sentinel) |
| `ErrSessionConsumeClaimExited` | A ConsumeClaim goroutine returned (new sentinel) |
| `nil` | Normal session release |

### Additional fix

`MockHeartbeatResponse.For()` was not propagating the `Err` field set via
`SetError()` — fixed here since the new test relies on it.

### Test plan

- Added `TestConsumerGroupSessionCancelCause_Rebalance` — verifies
  `context.Cause()` returns `ErrRebalanceInProgress` when heartbeat receives
  `ErrRebalanceInProgress`
- `go vet`, `go fmt`, `go test -race` clean

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -1112,6 +1112,7 @@ func (s *consumerGroupSession) heartbeatLoop() {
 		if err != nil {
 			if retries <= 0 {
 				s.parent.handleError(err, "", -1)
+				s.cancel(err)
 				return
 			}
 			retryBackoff.Reset(s.parent.config.Metadata.Retry.Backoff)
@@ -1130,6 +1131,7 @@ func (s *consumerGroupSession) heartbeatLoop() {
 
 			if retries <= 0 {
 				s.parent.handleError(err, "", -1)
+				s.cancel(err)
 				return
 			}
 
@@ -1155,6 +1157,7 @@ func (s *consumerGroupSession) heartbeatLoop() {
 			return
 		default:
 			s.parent.handleError(err, "", -1)
+			s.cancel(err)
 			return
 		}
 
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -15,6 +15,20 @@ import (
 // ErrClosedConsumerGroup is the error returned when a method is called on a consumer group that has been closed.
 var ErrClosedConsumerGroup = errors.New("kafka: tried to use a consumer group that was closed")
 
+// ErrSessionPartitionCountChanged is set as the cancellation cause of a consumer group
+// session context when the leader detects that the partition count for a subscribed topic
+// has changed, requiring a new session.
+var ErrSessionPartitionCountChanged = errors.New("kafka: partition count changed for subscribed topic")
+
+// ErrSessionConsumeClaimExited is set as the cancellation cause of a consumer group session
+// context when a ConsumeClaim goroutine exits, triggering the end of the session.
+var ErrSessionConsumeClaimExited = errors.New("kafka: ConsumeClaim goroutine exited")
+
+// ErrSessionHeartbeatFailed is set as the cancellation cause of a consumer group session
+// context when the heartbeat loop exits due to an unrecoverable error (e.g. coordinator
+// unreachable after retries).
+var ErrSessionHeartbeatFailed = errors.New("kafka: heartbeat loop failed")
+
 // ConsumerGroup is responsible for dividing up processing of topics and partitions
 // over a collection of processes (the members of the consumer group).
 type ConsumerGroup interface {
@@ -731,7 +745,7 @@ func (c *consumerGroup) loopCheckPartitionNumbers(allSubscribedTopicPartitions m
 		return
 	}
 
-	defer session.cancel()
+	defer session.cancel(ErrSessionPartitionCountChanged)
 
 	oldTopicToPartitionNum := make(map[string]int, len(allSubscribedTopicPartitions))
 	for topic, partitions := range allSubscribedTopicPartitions {
@@ -838,7 +852,7 @@ type consumerGroupSession struct {
 	claims  map[string][]int32
 	offsets *offsetManager
 	ctx     context.Context
-	cancel  func()
+	cancel  context.CancelCauseFunc
 
 	waitGroup       sync.WaitGroup
 	releaseOnce     sync.Once
@@ -847,7 +861,7 @@ type consumerGroupSession struct {
 
 func newConsumerGroupSession(ctx context.Context, parent *consumerGroup, claims map[string][]int32, memberID string, generationID int32, handler ConsumerGroupHandler) (*consumerGroupSession, error) {
 	// init context
-	ctx, cancel := context.WithCancel(ctx)
+	ctx, cancel := context.WithCancelCause(ctx)
 
 	// init offset manager
 	offsets, err := newOffsetManagerFromClient(parent.groupID, memberID, generationID, parent.client, cancel)
@@ -903,7 +917,7 @@ func newConsumerGroupSession(ctx context.Context, parent *consumerGroup, claims 
 			go func(topic string, partition int32) {
 				defer sess.waitGroup.Done()
 				// cancel the group session as soon as any of the consume calls return
-				defer sess.cancel()
+				defer sess.cancel(ErrSessionConsumeClaimExited)
 
 				// if partition not currently readable, wait for it to become readable
 				if sess.parent.client.PartitionNotReadable(topic, partition) {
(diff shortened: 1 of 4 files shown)


dnwe (commented):
Thanks - I think it is a good idea to include a cause in the context cancel

However, that is quite a number of added errors, for now I wonder if we can should reduce the number by re-using as many of the existing `err` types as possible? For example, rather than introducing a new ErrRebalance can we just call s.cancel with the existing ErrRebalanceInProgress type?

I'm also wondering about JoinGroup/LeaveGroup which in [KIP-800](https://cwiki.apache.org/confluence/display/KAFKA/KIP-800:+Add+reason+to+JoinGroupRequest+and+LeaveGroupRequest) added support for telling Kafka _why_ the consumer is (re-)joining/leaving and that gets logged server side, so we probably want to be thinking about that when we choose the content of these particular errors, so we can flow the same across once we add the protocol support for those

consumer_group.go
```diff
@@ -1128,14 +1142,16 @@ func (s *consumerGroupSession) heartbeatLoop() {
...
+			s.cancel(ErrRebalanceInProgress)
 		case ErrUnknownMemberId, ErrIllegalGeneration:
+			s.cancel(resp.Err)
 			return
 		case ErrFencedInstancedId:
 			if s.parent.groupInstanceId != nil {
 				Logger.Printf("JoinGroup failed: group instance id %s has been fenced\n", *s.parent.groupInstanceId)
 			}
 			s.parent.handleError(resp.Err, "", -1)
+			s.cancel(ErrFencedInstancedId)
```
puellanivis: On the line above, we use `resp.Err` rather than repeating `ErrFencedInstancedId`, we should probably follow suit here for consistency.

consumer_group.go
```diff
@@ -1128,14 +1142,16 @@ func (s *consumerGroupSession) heartbeatLoop() {
 			retries = s.parent.config.Metadata.Retry.Max
 		case ErrRebalanceInProgress:
 			retries = s.parent.config.Metadata.Retry.Max
-			s.cancel()
+			s.cancel(ErrRebalanceInProgress)
```
puellanivis: As below, we could use `resp.Err` here.

Heck, we could just use `switch err := resp.Err; err { … }` and then use the shorter `err` directly.

consumer_group_test.go
```diff
@@ -338,3 +338,100 @@ func TestSubscriptionMetadata(t *testing.T) {
...
+	h.causeCh <- context.Cause(sess.Context())
+	return nil
+}
+
+// mockHeartbeatRebalanceResponse returns ErrNoError for the first N heartbeats,
+// then ErrRebalanceInProgress for all subsequent heartbeats.
+type mockHeartbeatRebalanceResponse struct {
+	t           TestReporter
+	successLeft int
+	mu          sync.Mutex
```
puellanivis: Convention is to place mutex protect fields in a separate paragraph headed by the mutex that is protecting them. This lets spatial locality reflect their critical section relationship.
```suggestion

	mu          sync.Mutex
	successLeft int
```

consumer_group_test.go
```diff
@@ -338,3 +338,100 @@ func TestSubscriptionMetadata(t *testing.T) {
...
+
+	select {
+	case cause := <-h.causeCh:
+		assert.True(t, errors.Is(cause, ErrRebalanceInProgress),
+			"expected ErrRebalanceInProgress as cause, got: %v", cause)
+	case <-time.After(5 * time.Second):
+		t.Fatal("timed out waiting for session context cancellation")
+	}
+
+	cancel()
```
puellanivis: `defer` this and put it on line 422. (Golang now inlines deferred functions to every exit point if it can be deferred, so there’s no extra cost for using a `defer`.)

consumer_group_test.go
```diff
@@ -338,3 +338,100 @@ func TestSubscriptionMetadata(t *testing.T) {
...
+	_ = group.Close()
```
puellanivis: As above, this should be defered, but placed on line 419

consumer_group_test.go
```diff
@@ -338,3 +338,100 @@ func TestSubscriptionMetadata(t *testing.T) {
...
+		_ = group.Consume(ctx, []string{"my-topic"}, h)
+	}()
+
+	select {
+	case cause := <-h.causeCh:
+		assert.True(t, errors.Is(cause, ErrRebalanceInProgress),
+			"expected ErrRebalanceInProgress as cause, got: %v", cause)
+	case <-time.After(5 * time.Second):
+		t.Fatal("timed out waiting for session context cancellation")
+	}
```
puellanivis: I’m thinking it might be more robust to put a `close(h.causeCh)` somewhere, and turn this select into a `for range h.causeCh` with a check to ensure at least (or only) one error was sent over the channel.

We should be able to easily control timeout via a `context.WithTimeout(t.Context(), 5 * time.Second)` somewhere above.

In fact, if such a timeout ever fails, then that alone itself is a bunch that we need to catch.

consumer_group.go
prakhar7651: Made the change.

consumer_group_test.go
prakhar7651: I've closed the channel in cleanup and am checking that we receive only 1 cause.

puellanivis (approved):

consumer_group_test.go
```diff
@@ -338,3 +338,101 @@ func TestSubscriptionMetadata(t *testing.T) {
 		assert.Equal(t, []byte("from-s2"), c.subscriptionMetadata(s2, topics).UserData)
 	})
 }
+
+// causeHandler is a ConsumerGroupHandler that captures the context.Cause when
+// the session context is canceled.
```
puellanivis, line 343: FYI: neither “cancelled” nor “canceled” are misspelled, or misspelt. 

P.S.: Though, I suppose “cancelled” is at least inconsistent, since the standard library `context` package uses `Canceled` 😂 https://pkg.go.dev/context@go1.26.3#pkg-variables

consumer_group.go
```diff
@@ -1123,22 +1137,24 @@ func (s *consumerGroupSession) heartbeatLoop() {
...
 			if s.parent.groupInstanceId != nil {
 				Logger.Printf("JoinGroup failed: group instance id %s has been fenced\n", *s.parent.groupInstanceId)
 			}
-			s.parent.handleError(resp.Err, "", -1)
+			s.parent.handleError(err, "", -1)
+			s.cancel(err)
 			return
 		default:
-			s.parent.handleError(resp.Err, "", -1)
+			s.parent.handleError(err, "", -1)
```
dnwe, line 1159: Any particular reason that we don't `s.cancel(err)` in the `default` case too? At the moment we'll just fall through to the deferred cancel with `ErrSessionHeartbeatFailed`, but seems like we may as well cancel with the actual kafka `err` here?

Similarly slightly above this block (not modified in this PR) we currently call handleError when retries are exhausted (`retries <= 0`), but we don't call `s.cancel(err)` so it just gets the generic deferred err

dnwe (commented):
Thanks, overall looks good, one comment below

Also note that prior to reviewing I rebased and squashed your branch down to a single commit to avoid the merge commits etc. and to update the commit body to match the refined implementation, so make sure you've re-synced before making any further changes

consumer_group.go
prakhar7651: The deferred s.cancel(ErrSessionHeartbeatFailed) now only applies if the function exits via <-s.hbDying (the session was released externally) — in all other exit paths we now set the actual error as the cause.

dnwe: Thanks!

dnwe (approved):
LGTM

PR #3584, opened 2026-05-26
feat(consumer_group): send KIP-800 reason on JoinGroup and LeaveGroup

dnwe:
Mirror the Java client's KIP-800 implementation by populating the reason field on JoinGroup v8+ and LeaveGroup v5+ requests. Session cancellation causes are mapped to descriptive strings matching the Java client conventions, and LeaveGroup always sends "the consumer is being closed".


puellanivis (approved):
Looks fine.

PR #3585, opened 2026-05-26
fix: correct requiredVersion for V8 JoinGroup and add protocol version placeholders

dnwe:
Correct the requiredVersion check for V8 JoinGroupRequest+Response

Also fill out the TestAllocateBodyProtocolVersions table to outline the protocol versions required to reach parity which each major kafka client version

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -181,6 +181,12 @@ func allocateResponseBody(req protocolBody) protocolBody {
 	return nil
 }
 
+// TestAllocateBodyProtocolVersions tests two related version expectations:
+//  1. uncommented entries are protocol versions Sarama currently supports with
+//     parity to a given Kafka release version
+//  2. commented TODO entries are the max versions supported by the given Kafka
+//     release that Sarama still needs to implement for parity, this is for dev
+//     tracking.
 func TestAllocateBodyProtocolVersions(t *testing.T) {
 	type test struct {
 		version     KafkaVersion
@@ -422,7 +428,7 @@ func TestAllocateBodyProtocolVersions(t *testing.T) {
 				// TODO: ListOffsetsRequest v6 is not supported, but expected for KafkaVersion 2.8.0
 				// apiKeyListOffsets:          6, // up from 5
 				// TODO: MetadataRequest v11 is not supported, but expected for KafkaVersion 2.8.0
-				// apiKeyMetadata:             11, // up from 10
+				// apiKeyMetadata:             11, // up from 9
 				// TODO: AddPartitionsToTxnRequest v3 is not supported, but expected for KafkaVersion 2.8.0
 				// apiKeyAddPartitionsToTxn:   3, // up from 2
 				// TODO: AddOffsetsToTxnRequest v3 is not supported, but expected for KafkaVersion 2.8.0
@@ -438,7 +444,7 @@ func TestAllocateBodyProtocolVersions(t *testing.T) {
 				// TODO: DeleteTopicsRequest v6 is not supported, but expected for KafkaVersion 2.8.0
 				// apiKeyDeleteTopics:         6, // up from 5
 				// TODO: DescribeProducersRequest v0 is not supported, but expected for KafkaVersion 2.8.0
-				// 61:                         0, // new in 2.8
+				// apiKeyDescribeProducers /* (61) */: 0, // new in 2.8
 			},
 		},
 		{
@@ -467,7 +473,7 @@ func TestAllocateBodyProtocolVersions(t *testing.T) {
 				apiKeyLeaveGroup:      5, // up from 4
 				apiKeyDescribeLogDirs: 3, // up from 2
 				// TODO: JoinGroupRequest v9 is not supported, but expected for KafkaVersion 3.2.0
-				// apiKeyJoinGroup:    9, // up from 8
+				// apiKeyJoinGroup:    9, // up from 7
 			},
 		},
 		{
--- a/request_test.go
+++ b/request_test.go
@@ -301,9 +301,9 @@ func TestAllocateBodyProtocolVersions(t *testing.T) {
 				apiKeyMetadata:             7,  // up from 6
 				apiKeyOffsetCommit:         6,  // up from 4
 				apiKeyOffsetFetch:          5,  // up from 4
+				apiKeyDeleteTopics:         3,  // up from 2
 				apiKeyOffsetForLeaderEpoch: 2,  // up from 1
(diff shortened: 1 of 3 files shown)


request_test.go
```diff
@@ -379,54 +379,182 @@ func TestAllocateBodyProtocolVersions(t *testing.T) {
...
+				// TODO: DeleteTopicsRequest v5 is not supported, but expected for KafkaVersion 2.7.0
+				// apiKeyDeleteTopics:     5, // up from 4
+				// TODO: CreatePartitionsRequest v3 is not supported, but expected for KafkaVersion 2.7.0
+				// apiKeyCreatePartitions: 3, // up from 2
+			},
+		},
+		{
+			V2_8_0_0,
+			map[int16]int16{
+				apiKeyMetadata:             10, // up from 9
```
hindessm, line 423: Why is this here and in the TODO comments?

request_test.go
```diff
@@ -379,54 +379,182 @@ func TestAllocateBodyProtocolVersions(t *testing.T) {
...
+				apiKeyDescribeClientQuotas: 1,  // up from 0
+				apiKeyDescribeCluster:      0,  // new in 2.8
+				// TODO: ProduceRequest v9 is not supported, but expected for KafkaVersion 2.8.0
+				// apiKeyProduce:              9, // up from 8
+				// TODO: ListOffsetsRequest v6 is not supported, but expected for KafkaVersion 2.8.0
+				// apiKeyListOffsets:          6, // up from 5
+				// TODO: MetadataRequest v11 is not supported, but expected for KafkaVersion 2.8.0
+				// apiKeyMetadata:             11, // up from 10
```
hindessm: I always read the `up from X` comments as meaning the previous kafka version used version `X` so I was expecting this to be `up from 9` since v2.7 used metadata v9?

request_test.go
```diff
@@ -379,54 +379,182 @@ func TestAllocateBodyProtocolVersions(t *testing.T) {
...
+				// apiKeyAddPartitionsToTxn:   3, // up from 2
+				// TODO: AddOffsetsToTxnRequest v3 is not supported, but expected for KafkaVersion 2.8.0
+				// apiKeyAddOffsetsToTxn:      3, // up from 2
+				// TODO: EndTxnRequest v3 is not supported, but expected for KafkaVersion 2.8.0
+				// apiKeyEndTxn:               3, // up from 2
+				// TODO: AlterConfigsRequest v2 is not supported, but expected for KafkaVersion 2.8.0
+				// apiKeyAlterConfigs:         2, // up from 1
+				// TODO: AlterClientQuotasRequest v1 is not supported, but expected for KafkaVersion 2.8.0
+				// apiKeyAlterClientQuotas:    1, // up from 0
+			},
```
hindessm, line 448: Don't we need `CreateTopics v7` and `DeleteTopics v6` too?
And maybe `DescribeProducers v0`?

request_test.go
```diff
@@ -379,54 +379,182 @@ func TestAllocateBodyProtocolVersions(t *testing.T) {
...
+		},
 		{
 			V3_1_0_0,
 			map[int16]int16{
 				apiKeyJoinGroup: 8, // up from 7
```
hindessm: I only see v7 at https://kafka.apache.org/31/design/protocol/#The_Messages_JoinGroup ?
The same page also has Fetch v13 but you've included that TODO under v3.3?

hindessm (commented):
I'm seeing quite a few issues with these?

dnwe:
> I'm seeing quite a few issues with these?

That’s what happens when you do something by hand 😂

request_test.go
dnwe: Yeah this was a grey area, because we currently do support V10 which was technically never a released version, but was a stepping stone between V9 and V11 - I understand why it's confusing though, the intent was that we'd delete when we bumped to V11

request_test.go
dnwe: Same reason as above

request_test.go
```diff
@@ -379,60 +385,196 @@ func TestAllocateBodyProtocolVersions(t *testing.T) {
...
 		{
 			V3_2_0_0,
 			map[int16]int16{
-				apiKeyLeaveGroup: 5, // up from 4
+				apiKeyJoinGroup:       8, // up from 7
+				apiKeyLeaveGroup:      5, // up from 4
+				apiKeyDescribeLogDirs: 3, // up from 2
+				// TODO: JoinGroupRequest v9 is not supported, but expected for KafkaVersion 3.2.0
+				// apiKeyJoinGroup:    9, // up from 7
+			},
```
dnwe, line 477: This was a mistake in my earlier KIP-800 PR when I implemented JoinGroup V8 and put it in V3_1_0_0, but similar to V10+V11 Metadata, we've implemented one of the intermediary versions that isn't quite a whole version number

hindessm:
Out of curiosity, I tried adding protocol version placeholders and ended up with slightly different [results](https://github.com/IBM/sarama/commit/b2b1946a63874a20984cee3b097503ea588dd946).

dnwe:
> Out of curiosity, I tried adding protocol version placeholders and ended up with slightly different [results](https://github.com/IBM/sarama/commit/b2b1946a63874a20984cee3b097503ea588dd946).

Where are you sourcing your versions from? I’ve only been using the in-repo json, so for example you seem to have no 3.9.0 and have a bunch at 4.0.0 that actually showed up in 3.9.0 according to the json, for example FetchRequest v17 https://github.com/apache/kafka/blob/3.9.0/clients/src/main/resources/common/message/FetchRequest.json and FindCoordinatorRequest v6 https://github.com/apache/kafka/blob/3.9.0/clients/src/main/resources/common/message/FindCoordinatorRequest.json

hindessm:
> > Out of curiosity, I tried adding protocol version placeholders and ended up with slightly different [results](https://github.com/IBM/sarama/commit/b2b1946a63874a20984cee3b097503ea588dd946).
> 
> Where are you sourcing your versions from? I’ve only been using the in-repo json, so for example you seem to have no 3.9.0 and have a bunch at 4.0.0 that actually showed up in 3.9.0 according to the json, for example FetchRequest v17 https://github.com/apache/kafka/blob/3.9.0/clients/src/main/resources/common/message/FetchRequest.json and FindCoordinatorRequest v6 https://github.com/apache/kafka/blob/3.9.0/clients/src/main/resources/common/message/FindCoordinatorRequest.json

Human-error caused my process to skip 3.9 (as well as correctly skipping 2.9) ... https://github.com/IBM/sarama/commit/fa095a3 should be better

PR #3591, opened 2026-05-27
feat(consumer): add support for SyncGroupRequest/Response v5 (KIP-559)

dnwe:
Add ProtocolType and ProtocolName nullable string fields to both request and response, bump isValidVersion to 5

https://cwiki.apache.org/confluence/display/KAFKA/KIP-559%3A+Make+the+Kafka+Protocol+Friendlier+with+L7+Proxies

Also tidyup the syncGroupRequest version ladder to be descending, using a separate req.Version check for additional fields.

### DIFF
--- a/sync_group_request.go
+++ b/sync_group_request.go
@@ -44,6 +44,10 @@ type SyncGroupRequest struct {
 	MemberId string
 	// GroupInstanceId contains the unique identifier of the consumer instance provided by end user.
 	GroupInstanceId *string
+	// ProtocolType contains the group protocol type.
+	ProtocolType *string
+	// ProtocolName contains the group protocol name.
+	ProtocolName *string
 	// GroupAssignments contains each assignment.
 	GroupAssignments []SyncGroupRequestAssignment
 }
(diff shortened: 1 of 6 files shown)


sync_group_request.go
```diff
@@ -69,6 +73,15 @@ func (s *SyncGroupRequest) encode(pe packetEncoder) (err error) {
 		}
 	}
 
+	if s.Version >= 5 {
```
hindessm, line 76: I'd have probably nested this in the `s.Version >= 3` block.

sync_group_request.go
```diff
@@ -102,6 +115,15 @@ func (s *SyncGroupRequest) decode(pd packetDecoder, version int16) (err error) {
...
```
hindessm, line 118: Same here.

hindessm (approved):
LGTM but made minor comments.

sync_group_request.go
dnwe: Yeah I did consider it, and from a code logic perspective it makes sense, but as a discrete feature of V5+ I ended up putting it standalone rather than nested. I’m not especially wed to either form, but that was my 50/50 choice at the time 😅

PR #3595, opened 2026-05-30
feat: support transactions v2 (KIP-890)

omerzamir:
Closes #3470.

Kafka 4.0 ships `transaction.version=2` (TV2) from [KIP-890: Transactions Server-Side Defense](https://cwiki.apache.org/confluence/display/KAFKA/KIP-890%3A+Transactions+Server-Side+Defense), which fixes a correctness bug where `Produce` requests from one transaction could leak into another after a producer epoch bump. The fix reworks the protocol so the broker bumps the epoch on every `EndTxn` and registers partitions/offsets implicitly, removing the explicit `AddPartitionsToTxn`/`AddOffsetsToTxn` round-trips the bug came from. Sarama was capped at the TV1 versions and inherited the bug; this PR brings it to parity with the Java client.

## Negotiation

TV2 is auto-discovered from the coordinator broker's finalized features (KIP-584) in `ApiVersionsResponse v3+`. When `transaction.version >= 2` is advertised, the producer uses the v2 flow; otherwise it behaves exactly as before. No new config knob, no behaviour change on pre-4.0 brokers.

## What changes under TV2

- `AddPartitionsToTxn` / `AddOffsetsToTxn` are no longer sent — the broker registers partitions/offsets implicitly.
- `EndTxn` is sent at v5; the bumped `ProducerId`/`ProducerEpoch` in the response are adopted directly, with no extra `InitProducerId` round-trip.
- `InitProducerId` v5 and `Produce` v12 are sent as the TV2 capability marker.
- `TRANSACTION_ABORTABLE` (error 120) is surfaced as `ErrTransactionAbortable` and handled as a retriable abortable error.

## Protocol support added

`ApiVersionsResponse` v3+ finalized/supported features; `EndTxn` v3–v5; `InitProducerId` v5; `AddPartitionsToTxn` v3–v5 (incl. the v4 batched form); `TxnOffsetCommit` v4–v5 (v3 already landed in #3592); `Produce` v9–v12 (flexible encoding). `requiredVersion` for every bumped API was set from the Kafka release branch that introduced it.

## Compatibility

Additive only — `apidiff -incompatible` reports nothing against `main`. New exported fields/types: `ApiVersionsResponse.{SupportedFeatures,FinalizedFeatures,FinalizedFeaturesEpoch}`, `EndTxnResponse.{ProducerID,ProducerEpoch}`, `AddPartitionsToTxn{Request.Transactions,Response.ErrorCode,Response.ResultsByTransaction}` plus their element types, and `ErrTransactionAbortable`.

## Tests

Unit coverage for each protocol bump (byte-level round-trips) and the TV2 manager behaviour: epoch adoption from `EndTxn v5`, skipping `AddPartitionsToTxn`/`AddOffsetsToTxn`, and a TV1 regression guard. The version matrix in `request_test.go` is extended. `go test -race ./...` passes. The FVT matrix already runs Kafka 4.1/4.2, exercising the path against a real broker; TV1 behaviour is unchanged when the broker doesn't advertise


dnwe:
@omerzamir welcome! Thanks for looking into this capability, that’s really helpful

Would it be possible to submit the commits as a PR each so they can be reviewed and merged in isolation? Just looking over the diff stats it feels like they could be easier to review that way:

```
a62a715f feat(protocol): expose finalized features in ApiVersionsResponse v3+
 6 files changed, 256 insertions(+), 4 deletions(-)
08a022b9 feat(protocol): support EndTxnRequest/Response v3-v5 (KIP-890)
 5 files changed, 192 insertions(+), 17 deletions(-)
0f9f024e feat(protocol): support InitProducerIdRequest/Response v5 (KIP-890)
 5 files changed, 18 insertions(+), 3 deletions(-)
335849a9 feat(protocol): support AddPartitionsToTxnRequest/Response v3-v5 (KIP-890)
 5 files changed, 433 insertions(+), 18 deletions(-)
62c3a518 feat(protocol): support TxnOffsetCommitRequest/Response v3-v5 (KIP-890)
 5 files changed, 201 insertions(+), 12 deletions(-)
da35bc11 feat(protocol): support ProduceRequest/Response v9-v12 (KIP-890)
 6 files changed, 261 insertions(+), 33 deletions(-)
eb64e5f7 feat: implement transactions v2 producer behavior 
(KIP-890)
 3 files changed, 387 insertions(+), 103 deletions(-)
 ```

File: add_partitions_to_txn_request.go [hunk 1]
```diff
@@ -30,12 +61,59 @@ func (a *AddPartitionsToTxnRequest) encode(pe packetEncoder) error {
...
+		}
 	}
 
+	if a.isFlexible() {
+		pe.putEmptyTaggedFieldArray()
+	}
+	return nil
+}
+
+func encodeAddPartitionsToTxnTransaction(pe packetEncoder, t *AddPartitionsToTxnTransaction) error {
```
hindessm, line 75: For consistency with existing code, I'd expect this to be a method:
```go
func (t *AddPartitionsToTxnTransaction) encode(pe packetEncoder) error {
```
rather than a function.

add_partitions_to_txn_request.go
```diff
@@ -64,11 +142,58 @@ func (a *AddPartitionsToTxnRequest) decode(pd packetDecoder, version int16) (err
...
 
+	if a.isFlexible() {
+		if _, err := pd.getEmptyTaggedFieldArray(); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
+func decodeAddPartitionsToTxnTransaction(pd packetDecoder, t *AddPartitionsToTxnTransaction) (err error) {
```
hindessm, line 160: Similarly, I'd have expected this to be a `decode` method with `t *AddPartitionsToTxnTransaction` as receiver.

add_partitions_to_txn_response.go
```diff
@@ -33,8 +59,39 @@ func (a *AddPartitionsToTxnResponse) encode(pe packetEncoder) error {
...
+		}
 	}
 
+	if a.isFlexible() {
+		pe.putEmptyTaggedFieldArray()
+	}
+	return nil
+}
+
+func encodeAddPartitionsToTxnTransactionResult(pe packetEncoder, r *AddPartitionsToTxnTransactionResult) error {
```
hindessm, line 73: Similarly, I'd expect this to be an `encode` method with `r *AddPartitionsToTxnTransactionResult` receiver.

add_partitions_to_txn_response.go
```diff
@@ -70,11 +150,61 @@ func (a *AddPartitionsToTxnResponse) decode(pd packetDecoder, version int16) (er
...
 
+	if a.isFlexible() {
+		if _, err := pd.getEmptyTaggedFieldArray(); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
+func decodeAddPartitionsToTxnTransactionResult(pd packetDecoder, r *AddPartitionsToTxnTransactionResult) (err error) {
```
hindessm, line 168: Similarly, I'd expect this to be an `decode` method with `r *AddPartitionsToTxnTransactionResult` receiver.

hindessm (commented):
Thanks for tackling this.

This isn't a full review as I'll wait for the split out PRs that @dnwe requested but I thought I'd comment on a couple of minor issues I noticed that you might want to consider while preparing the PRs.

add_partitions_to_txn_request.go
```diff
@@ -30,12 +61,59 @@ func (a *AddPartitionsToTxnRequest) encode(pe packetEncoder) error {
 		if err := pe.putInt32Array(partitions); err != nil {
 			return err
 		}
+		if a.isFlexible() {
```
puellanivis, line 64: I could have sworn we setup the `pe` implementations so that the non-flexible one is a no-op https://github.com/IBM/sarama/blob/45b4840aa51af8207a07292512e9afca92f45206/prep_encoder.go#L164 so that this sort of conditional isn’t necessary. 🤔

add_partitions_to_txn_request.go
```diff
@@ -64,11 +142,58 @@ func (a *AddPartitionsToTxnRequest) decode(pd packetDecoder, version int16) (err
 		}
 
 		a.TopicPartitions[topic] = partitions
+		if a.isFlexible() {
+			if _, err := pd.getEmptyTaggedFieldArray(); err != nil {
+				return err
+			}
+		}
```
puellanivis, line 149: Again, I thought we make this conditional unnecessary.

broker.go
```diff
@@ -438,6 +445,21 @@ func (b *Broker) Rack() string {
...
 
+func (b *Broker) finalizedFeatureLevel(name string) int16 {
+	if b == nil {
+		return 0
+	}
+	b.lock.Lock()
+	defer b.lock.Unlock()
+	if b.brokerFinalizedFeatures == nil {
+		return 0
+	}
```
puellanivis, line 456: [unnecessary guarding] If `b.brokerFinalizedFeatures == nil` then every access of it will return the zero-value and false in the ok pattern, so the following line 457 implicitly contains a test for `map == nil`, and returns `ok == false`

prep_encoder.go
```diff
@@ -256,3 +260,22 @@ func (pe *prepFlexibleEncoder) putNullableInt32Array(in []int32) error {
...
+		inner := &prepFlexibleEncoder{prepEncoder: &prepEncoder{}}
+		if err := enc(inner); err != nil {
+			return err
+		}
+		pe.putUVarint(id)
+		pe.putUVarint(uint64(inner.length))
+		pe.length += inner.length
+	}
+	return nil
+}
```
puellanivis, line 281: I can’t follow why any of this is necessary.

real_encoder.go
```diff
@@ -266,3 +270,37 @@ func (re *realFlexibleEncoder) putNullableInt32Array(in []int32) error {
...
+	return nil
+}
+
+func sortUint64Slice(s []uint64) {
+	for i := 1; i < len(s); i++ {
+		for j := i; j > 0 && s[j-1] > s[j]; j-- {
+			s[j-1], s[j] = s[j], s[j-1]
+		}
+	}
+}
```
puellanivis, line 306: Use `slices.Sort(s)` instead.

PR #3596, opened 2026-05-30
feat(protocol): add support for DescribeConfigs v3 and v4

dnwe:
v3 (KIP-569) adds the IncludeDocumentation field to the request and ConfigType plus Documentation to each config entry on the response, from Kafka 2.6.0.

v4 makes the API flexible (compact types and tagged fields), from Kafka 2.8.0.

Note: this is protocol only for now, not yet exposing the extra fields in admin.go, that’ll come in a follow up PR

### DIFF
--- a/describe_configs_response.go
+++ b/describe_configs_response.go
@@ -34,6 +34,21 @@ const (
 	SourceDefault
 )
 
+type ConfigType int8
+
+const (
+	UnknownConfigType ConfigType = iota
+	BooleanConfigType
+	StringConfigType
+	IntConfigType
+	ShortConfigType
+	LongConfigType
+	DoubleConfigType
+	ListConfigType
+	ClassConfigType
+	PasswordConfigType
+)
+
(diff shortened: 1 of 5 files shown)


describe_configs_response.go
```diff
@@ -70,13 +85,15 @@ type ResourceResponse struct {
...
-	Synonyms  []*ConfigSynonym
+	Name          string
+	Value         string
+	ReadOnly      bool
+	Default       bool
+	Source        ConfigSource
+	Sensitive     bool
+	Synonyms      []*ConfigSynonym
+	Type          ConfigType // v3
+	Documentation string     // v3
```
hindessm: Why is this a `string` rather than a `*string`? It represents a nullable string so surely we'd want to represent the `nil` case?

dnwe: Thanks, fixed!

hindessm (approved):
LTGM

PR #3598, opened 2026-06-01
fix(admin): retry ACL and SCRAM ops on stale controller

dnwe:
CreateACL, CreateACLs, ListAcls, DeleteACL and DescribeUserScramCredentials issued their request once against the cached controller and returned without retrying when the broker reported it was no longer the controller. Wrap each in retryOnError and refresh the cached controller on the not-controller code so the request is re-sent to the new controller.

Add controllerError to map a response-level error code to a returned error, refreshing the controller on retriable codes, and inspect the per-filter error in DeleteACL responses that previously went unchecked.

Thanks to @christos-spearbit for reporting the issue

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -288,6 +288,19 @@ func (ca *clusterAdmin) retryOnError(retryable func(error) bool, fn func() error
 	}
 }
 
+func (ca *clusterAdmin) controllerError(code KError, msg *string) error {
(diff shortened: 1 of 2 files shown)


hindessm (approved):
LGTM. Though I can't help feeling there is a better way to structure the retry logic to avoid having repeated calls to `isRetriableControllerError` on the same error.

PR #3599, opened 2026-06-01
fix: handle errors gracefully instead of panicking in async producer

willy182:
This PR replaces a panic in produce_set.go with a proper error return when message set encoding fails during produce request building

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -988,17 +988,11 @@ func (p *asyncProducer) newBrokerProducer(broker *Broker) *brokerProducer {
 			if err != nil {
 				// Count the in flight requests to know when we can close the pending channel safely
 				wg.Add(1)
-				// capture the muted set. unmuting is deferred to handleResponse.
-				mutedSet := set
-				sendResponse := func(response *ProduceResponse, err error) {
-					pending <- &brokerProducerResponse{
-						set: mutedSet,
-						err: err,
-						res: response,
-					}
-					wg.Done()
+				pending <- &brokerProducerResponse{
+					set: set,
+					err: err,
 				}
-				sendResponse(nil, err)
+				wg.Done()
 				continue
 			}
 
--- a/async_producer.go
+++ b/async_producer.go
@@ -986,13 +986,10 @@ func (p *asyncProducer) newBrokerProducer(broker *Broker) *brokerProducer {
 		for set := range bridge {
 			request, err := set.buildRequest()
 			if err != nil {
-				// Count the in flight requests to know when we can close the pending channel safely
-				wg.Add(1)
 				pending <- &brokerProducerResponse{
 					set: set,
 					err: err,
 				}
-				wg.Done()
 				continue
 			}
 
--- a/async_producer.go
+++ b/async_producer.go
@@ -984,7 +984,23 @@ func (p *asyncProducer) newBrokerProducer(broker *Broker) *brokerProducer {
 		var wg sync.WaitGroup
 
 		for set := range bridge {
-			request := set.buildRequest()
+			request, err := set.buildRequest()
+			if err != nil {
+				// Count the in flight requests to know when we can close the pending channel safely
+				wg.Add(1)
+				// capture the muted set. unmuting is deferred to handleResponse.
+				mutedSet := set
+				sendResponse := func(response *ProduceResponse, err error) {
+					pending <- &brokerProducerResponse{
+						set: mutedSet,
+						err: err,
+						res: response,
+					}
+					wg.Done()
+				}
+				sendResponse(nil, err)
+				continue
+			}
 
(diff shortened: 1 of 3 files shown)


async_producer.go
```diff
@@ -1002,7 +984,23 @@ func (p *asyncProducer) newBrokerProducer(broker *Broker) *brokerProducer {
...
+				mutedSet := set
+				sendResponse := func(response *ProduceResponse, err error) {
+					pending <- &brokerProducerResponse{
+						set: mutedSet,
+						err: err,
+						res: response,
+					}
+					wg.Done()
+				}
+				sendResponse(nil, err)
```
puellanivis: It makes no sense to pass `response = nil` as a parameter here, nor set it as a field in the `&brokerProducerRespnse` as `nil` is going to be the automatic zero value.

Additionally, there’s no reason to pass `err` as a parameter either, it’s already available by closure, and we’re not mutating it anyways.

puellanivis: One can simply call an anonymous function without needing to pass it through a temporary named variable.
```suggestion
				func(response *ProduceResponse, err error) {
					pending <- &brokerProducerResponse{
						set: mutedSet,
						err: err,
						res: response,
					}
					wg.Done()
				}(nil, err)
```

async_producer.go
```diff
@@ -1002,7 +984,23 @@ func (p *asyncProducer) newBrokerProducer(broker *Broker) *brokerProducer {
 		var wg sync.WaitGroup
 
 		for set := range bridge {
-			request := set.buildRequest()
+			request, err := set.buildRequest()
+			if err != nil {
+				// Count the in flight requests to know when we can close the pending channel safely
+				wg.Add(1)
+				// capture the muted set. unmuting is deferred to handleResponse.
+				mutedSet := set
```
puellanivis: We’re using go1.25 https://github.com/IBM/sarama/blob/main/go.mod#L3

Capturing a loop variable in an iteration variable if it might escape isn’t necessary.

produce_set_test.go
```diff
@@ -158,7 +158,7 @@ func TestProduceSetRequestBuilding(t *testing.T) {
 		safeAddMessage(t, ps, msg)
 	}
 
-	req := ps.buildRequest()
+	req, _ := ps.buildRequest()
```
puellanivis: Do not throw away errors in tests. This is the fastest way to have an issue that could have been caught in tests get missed until it’s released to a deployment.

puellanivis (commented):
We can have a little bit of copying… as a treat.

But if we copy some code, and then mutate it, we should also look into how that mutation means we might be able to simplify things.

willy182:
> We can have a little bit of copying… as a treat.
> 
> But if we copy some code, and then mutate it, we should also look into how that mutation means we might be able to simplify things.

I already changed, please review again

dnwe:
I’m curious when and how you’ve hit this panic?

async_producer.go
```diff
@@ -1002,7 +984,17 @@ func (p *asyncProducer) newBrokerProducer(broker *Broker) *brokerProducer {
...
-			request := set.buildRequest()
+			request, err := set.buildRequest()
+			if err != nil {
+				// Count the in flight requests to know when we can close the pending channel safely
+				wg.Add(1)
+				pending <- &brokerProducerResponse{
+					set: set,
+					err: err,
+				}
+				wg.Done()
```
puellanivis: I spot now that the `wg.Add` and `wg.Done` are unnecessary. The `WaitGroup` is defined on line 984, so this wait group control is definitely per-function-call, and since the send on `pending` blocks execution of this goroutine, and so the `for` loop cannot exit while this is executing, there’s no need to use wait group coördination.

You can spot below that the `wg.Done` is executed at an indeterminate point in the future by the async producer, but we’re in a different situation in this block.

puellanivis (commented):
Other than this last lint, the code seems reasonable. The larger scope I leave to the team.

puellanivis (approved):
Approving just the code. The logic and fit with design lies with the main contributors.

dnwe:
I’m still curious when and how you’ve hit this panic, because we should never fail to encode requests, or whether this is just a speculative PR because you spotted panic in the code?

willy182:
@dnwe Honestly, I haven't hit this panic in production. I spotted it while reviewing the codebase and realized that a panic here could crash the entire application unnecessarily. Since there's already a working error propagation path., it felt like a straightforward improvement to replace the panic with a proper error return - defensinve programming rather than a bug fix. But if I guess, maybe this Panic case could happen because the encoding/compression process failed.

dnwe:
OK cool thanks - I was just concerned if we'd accidentally regressed something as I hadn't seen this panic happening

PR #3600, opened 2026-06-02
feat(admin): add DescribeConfigs for multiple resources

dnwe:
Add DescribeConfigs, which describes several resources in one call, returns per-resource errors, and takes options to request synonyms (Kafka 1.1.0+) and documentation (Kafka 2.6.0+). Resources are grouped by the broker that must serve them, so broker and broker-logger configs still reach the right node.

Negotiate DescribeConfigs v3 and v4 from the configured version, and deprecate the singular DescribeConfig in favour of the new method.

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -64,7 +64,22 @@ type ClusterAdmin interface {
 	// The value of config entries where Sensitive is true is always nil so
 	// sensitive information is not disclosed.
 	// This operation is supported by brokers with version 0.11.0.0 or higher.
+	//
+	// Deprecated: use DescribeConfigs, which describes multiple resources in a
+	// single call, returns per-resource errors, and accepts options to request
+	// synonyms and documentation.
 	DescribeConfig(resource ConfigResource) ([]ConfigEntry, error)
+
+	// Get the configuration for the specified resources.
+	// The returned configuration includes default values and the Default is true
+	// can be used to distinguish them from user supplied values.
+	// Config entries where ReadOnly is true cannot be updated.
+	// The value of config entries where Sensitive is true is always nil so
+	// sensitive information is not disclosed.
+	// Use the options to request the synonyms (Kafka 1.1.0+) or the type and
+	// documentation (Kafka 2.6.0+) for each config entry.
+	// This operation is supported by brokers with version 0.11.0.0 or higher.
+	DescribeConfigs(resources []ConfigResource, options DescribeConfigsOptions) ([]ConfigResourceResult, error)
 
 	// Update the configuration for the specified resources with the default options.
 	// This operation is supported by brokers with version 0.11.0.0 or higher.
@@ -526,12 +541,14 @@ func (ca *clusterAdmin) ListTopics() (map[string]TopicDetail, error) {
 			Resources: describeConfigsResources,
 		}
 
-		if ca.conf.Version.IsAtLeast(V1_1_0_0) {
-			describeConfigsReq.Version = 1
-		}
-
-		if ca.conf.Version.IsAtLeast(V2_0_0_0) {
+		if ca.conf.Version.IsAtLeast(V2_8_0_0) {
+			describeConfigsReq.Version = 4
+		} else if ca.conf.Version.IsAtLeast(V2_6_0_0) {
+			describeConfigsReq.Version = 3
+		} else if ca.conf.Version.IsAtLeast(V2_0_0_0) {
 			describeConfigsReq.Version = 2
+		} else if ca.conf.Version.IsAtLeast(V1_1_0_0) {
+			describeConfigsReq.Version = 1
 		}
 
 		describeConfigsResp, err := b.DescribeConfigs(describeConfigsReq)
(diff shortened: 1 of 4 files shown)


admin.go
```diff
@@ -815,60 +832,130 @@ func dependsOnSpecificNode(resource ConfigResource) bool {
...
-	var resources []*ConfigResource
-	resources = append(resources, &resource)
+	results, err := ca.DescribeConfigs([]ConfigResource{resource}, DescribeConfigsOptions{})
+	if err != nil {
+		return nil, err
+	}
 
-	request := &DescribeConfigsRequest{
-		Resources: resources,
+	var entries []ConfigEntry
```
puellanivis: Maybe we should pre-alloc `len(results)` items? Since this is not a slice of pointers, and thus reallocation is relatively expensive.

Or, we could at least count the number of `result.Name == resource.Name` entries, and pre-allocate that.

admin.go
```diff
@@ -815,60 +832,130 @@ func dependsOnSpecificNode(resource ConfigResource) bool {
...
+			return nil, &DescribeConfigError{Err: result.ErrorCode, ErrMsg: result.ErrorMsg}
+		}
+		entries = append(entries, result.Configs...)
 	}
+	return entries, nil
+}
 
-	if ca.conf.Version.IsAtLeast(V1_1_0_0) {
-		request.Version = 1
+func (ca *clusterAdmin) DescribeConfigs(resources []ConfigResource, options DescribeConfigsOptions) ([]ConfigResourceResult, error) {
```
puellanivis: We could take the chance to change the parameters and returns here to slices of pointers?

Concerns about mutability of parameters and returns is a no issue, because we’re already passing around a pointer to the allocated underlying array anyways.

Obviously, we cannot change and of the fields _in_ the type to slices of pointers, but we could at least use slices to pointers for the parameters and returns.

admin.go
```diff
@@ -815,60 +832,130 @@ func dependsOnSpecificNode(resource ConfigResource) bool {
...
+
+	// broker and broker-logger configs must be described on the specific node;
+	// everything else can be served by any broker, so group the resources by the
+	// broker that must handle them and send one request per broker
+	groups := make(map[*Broker][]*ConfigResource)
+	var anyBroker *Broker
+	for i := range resources {
+		resource := &resources[i]
+		var b *Broker
+		if dependsOnSpecificNode(*resource) {
```
puellanivis: We could change the signature of this function to accept a pointer?

It’s only used in 4 places, including the one we’re modifying here.

admin.go
```diff
@@ -815,60 +832,130 @@ func dependsOnSpecificNode(resource ConfigResource) bool {
...
+		} else {
+			if anyBroker == nil {
+				broker, err := ca.findAnyBroker()
+				if err != nil {
+					return nil, err
+				}
+				anyBroker = broker
+			}
+			b = anyBroker
+		}
```
puellanivis: Maybe:
```go
switch {
case dependsOnSpecificNode(*resource):
	…
case anyBroker == nil:
	…
	fallthrough
default:
	b = anyBroker
}
```

We could also in the first case write the broker we found into `anyBroker` if `anyBroker == nil` at least. Actually, do we have to even keep `anyBroker` and `b` separate? Can we distribute all the `anyBrokers` across whatever broker was last used?

admin.go
```diff
@@ -815,60 +832,130 @@ func dependsOnSpecificNode(resource ConfigResource) bool {
...
 	}
 
-	if ca.conf.Version.IsAtLeast(V2_0_0_0) {
-		request.Version = 2
+	type resourceKey struct {
+		Type ConfigResourceType
+		Name string
 	}
+	resultByKey := make(map[resourceKey]ConfigResourceResult)
+	for b, grouped := range groups {
```
puellanivis: Is there a reason to avoid the name `group` as the singular from `groups`?

admin.go
```diff
@@ -815,60 +832,130 @@ func dependsOnSpecificNode(resource ConfigResource) bool {
...
+			request.Version = 2
+		} else if ca.conf.Version.IsAtLeast(V1_1_0_0) {
+			request.Version = 1
+		}
 
-	// DescribeConfig of broker/broker logger must be sent to the broker in question
-	if dependsOnSpecificNode(resource) {
-		var id int64
-		id, err = strconv.ParseInt(resource.Name, 10, 32)
+		_ = b.Open(ca.client.Config())
```
puellanivis, line 930: Do we want to throw away the `error` here? I suppose the `DesribeConfig` will throw an error if it’s not open?

admin.go
```diff
@@ -815,60 +832,130 @@ func dependsOnSpecificNode(resource ConfigResource) bool {
...
-	rsp, err := b.DescribeConfigs(request)
-	if err != nil {
-		return nil, err
-	}
-
-	for _, rspResource := range rsp.Resources {
-		if rspResource.Name == resource.Name {
-			if rspResource.ErrorCode != 0 {
-				return nil, &DescribeConfigError{Err: KError(rspResource.ErrorCode), ErrMsg: rspResource.ErrorMsg}
+		for _, rspResource := range rsp.Resources {
```
puellanivis: I don’t think we need to avoid the variable name `resource` anymore?

admin.go
```diff
@@ -815,60 +832,130 @@ func dependsOnSpecificNode(resource ConfigResource) bool {
...
+			result := ConfigResourceResult{
+				Type:      rspResource.Type,
+				Name:      rspResource.Name,
+				ErrorCode: KError(rspResource.ErrorCode),
+				ErrorMsg:  rspResource.ErrorMsg,
 			}
 			for _, cfgEntry := range rspResource.Configs {
```
puellanivis: I think we can switch from `cfgEntry` to `config`?

admin.go
```diff
@@ -815,60 +832,130 @@ func dependsOnSpecificNode(resource ConfigResource) bool {
...
 			}
+			resultByKey[resourceKey{rspResource.Type, rspResource.Name}] = result
 		}
 	}
-	return entries, nil
+
+	// return the results in the order the resources were requested
+	results := make([]ConfigResourceResult, 0, len(resources))
+	for i := range resources {
+		key := resourceKey{resources[i].Type, resources[i].Name}
```
puellanivis:
```suggestion
	for i, resource := range resources {
		key := resourceKey{resources.Type, resources.Name}
```

admin.go
dnwe: Yeah, _ = b.Open is a pattern we use everywhere, and it might be that we still establish connection within the request timeout

dnwe:
@puellanivis thanks! Very useful observations as always - addressed

puellanivis (approved):

PR #3604, opened 2026-06-02
feat(consumer): option to cap decompressed batch size

dnwe:
Add MaxDecompressedBatchSize to bound how far a single compressed record batch may inflate during decompression, guarding against compression bombs that turn a small payload into a large allocation.

The bound is applied while decompressing rather than after, so an oversized batch is rejected with ErrDecompressedBatchTooLarge before its output is fully allocated. The default of 0 leaves decompression unbounded, preserving existing behaviour.

### DIFF
--- a/decompress.go
+++ b/decompress.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"slices"
 	"sync"
 
 	"github.com/klauspost/compress/gzip"
@@ -45,25 +46,24 @@ func newDecompressedBatchTooLargeError(cc CompressionCodec, limit int) error {
 // fully inflates. A non-positive limit disables the cap.
 func boundedDecompress(cc CompressionCodec, reader io.Reader, limit int) ([]byte, error) {
 	buffer := bufferPool.Get().(*bytes.Buffer)
-	var src io.Reader = reader
+	defer func() {
+		buffer.Reset()
+		bufferPool.Put(buffer)
+	}()
+
+	src := reader
 	if limit > 0 {
 		src = io.LimitReader(reader, int64(limit)+1)
 	}
+
 	n, err := buffer.ReadFrom(src)
-
-	var res []byte
-	switch {
-	case err != nil:
-	case limit > 0 && n > int64(limit):
-		err = newDecompressedBatchTooLargeError(cc, limit)
-	default:
-		res = make([]byte, buffer.Len())
-		copy(res, buffer.Bytes())
+	if err != nil {
+		return nil, err
 	}
-
-	buffer.Reset()
-	bufferPool.Put(buffer)
-	return res, err
+	if limit > 0 && n > int64(limit) {
+		return nil, newDecompressedBatchTooLargeError(cc, limit)
+	}
+	return slices.Clone(buffer.Bytes()), nil
 }
 
 // boundedSnappyDecode decodes xerial/snappy into a capped buffer grown on
@@ -77,11 +77,12 @@ func boundedSnappyDecode(cc CompressionCodec, data []byte, limit int) ([]byte, e
 
 	bufp := bytesPool.Get().(*[]byte)
 	buf := *bufp
+	defer func() {
+		*bufp = buf[:0]
+		bytesPool.Put(bufp)
+	}()
 
-	size := cap(buf)
-	if size < 1024 {
-		size = 1024
-	}
+	size := max(cap(buf), 1024)
 	var out []byte
 	var err error
 	for {
@@ -98,19 +99,14 @@ func boundedSnappyDecode(cc CompressionCodec, data []byte, limit int) ([]byte, e
 		size *= 2
 	}
 
-	var res []byte
 	switch {
 	case errors.Is(err, snappy.ErrDstTooSmall):
-		err = newDecompressedBatchTooLargeError(cc, limit)
+		return nil, newDecompressedBatchTooLargeError(cc, limit)
 	case err != nil:
+		return nil, err
 	default:
-		res = make([]byte, len(out))
-		copy(res, out)
+		return slices.Clone(out), nil
 	}
-
-	*bufp = buf[:0]
-	bytesPool.Put(bufp)
-	return res, err
 }
 
 // boundedZstdDecode decodes into a pooled buffer using a decoder whose max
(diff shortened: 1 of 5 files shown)


decompress.go
```diff
@@ -33,7 +36,107 @@ var (
...
+	case err != nil:
+	case limit > 0 && n > int64(limit):
+		err = newDecompressedBatchTooLargeError(cc, limit)
+	default:
+		res = make([]byte, buffer.Len())
+		copy(res, buffer.Bytes())
+	}
+
+	buffer.Reset()
+	bufferPool.Put(buffer)
```
puellanivis: Maybe `defer`? If these are in a deferred func, then we could make the `switch` the final function statement, and thus ensure that each `case` has to have a proper `return` statement.

I find that forcing a `switch` statement to have to have explicit return on each case, it’s easier to reason about, because we don’t end up with a bunch of separate code paths joining back up for a single return.

Actually, we could then just convert this `switch` block to a simple early return series, and leave the non-error case unindented.

decompress.go
```diff
@@ -33,7 +36,107 @@ var (
...
+	if limit > 0 {
+		src = io.LimitReader(reader, int64(limit)+1)
+	}
+	n, err := buffer.ReadFrom(src)
+
+	var res []byte
+	switch {
+	case err != nil:
+	case limit > 0 && n > int64(limit):
+		err = newDecompressedBatchTooLargeError(cc, limit)
```
puellanivis: Do we know that if we hit the limit that we will get back an `err == nil`? Otherwise we would return that error instead, right?

decompress.go
```diff
@@ -33,7 +36,107 @@ var (
...
+	var err error
+	for {
+		if size > limit {
+			size = limit
+		}
+		if cap(buf) < size {
+			buf = make([]byte, 0, size)
+		}
+		out, err = snappy.DecodeCapped(buf[:0:size], data)
+		if !errors.Is(err, snappy.ErrDstTooSmall) || size >= limit {
```
puellanivis, line 96: Because of the conditional on lines 88–90, I don’t think we could get to `size >= limit` unless `size == limit`.

I guess we’re just sanity checking the value?

decompress.go
```diff
@@ -33,7 +36,107 @@ var (
...
+		return snappy.Decode(data)
+	}
+
+	bufp := bytesPool.Get().(*[]byte)
+	buf := *bufp
+
+	size := cap(buf)
+	if size < 1024 {
+		size = 1024
+	}
```
puellanivis:
```go
size := min(cap(buf), 1024)
```

decompress.go
```diff
@@ -33,7 +36,107 @@ var (
...
+
+	var res []byte
+	switch {
+	case errors.Is(err, snappy.ErrDstTooSmall):
+		err = newDecompressedBatchTooLargeError(cc, limit)
+	case err != nil:
+	default:
+		res = make([]byte, len(out))
+		copy(res, out)
```
puellanivis: Same as above, `slices.Clone`

decompress.go
```diff
@@ -33,7 +36,107 @@ var (
...
+	n, err := buffer.ReadFrom(src)
+
+	var res []byte
+	switch {
+	case err != nil:
+	case limit > 0 && n > int64(limit):
+		err = newDecompressedBatchTooLargeError(cc, limit)
+	default:
+		res = make([]byte, buffer.Len())
+		copy(res, buffer.Bytes())
```
puellanivis: We should have `slices.Clone` right?

decompress.go
```diff
@@ -33,7 +36,107 @@ var (
...
+}
+
+// boundedZstdDecode decodes into a pooled buffer using a decoder whose max
+// memory is capped at limit, so an oversized frame fails without allocating its
+// output. ErrDecoderSizeExceeded is translated into a too-large error.
+func boundedZstdDecode(cc CompressionCodec, data []byte, limit int) ([]byte, error) {
+	buffer := *bytesPool.Get().(*[]byte)
+	buffer, err := zstdDecompress(ZstdDecoderParams{}, buffer, data, limit)
+	if errors.Is(err, zstd.ErrDecoderSizeExceeded) {
+		err = newDecompressedBatchTooLargeError(cc, limit)
```
puellanivis: If we `defer` the `bytesPool.Put`, then we can just early return here instead.

decompress.go
```diff
@@ -33,7 +36,107 @@ var (
...
+	case errors.Is(err, snappy.ErrDstTooSmall):
+		err = newDecompressedBatchTooLargeError(cc, limit)
+	case err != nil:
+	default:
+		res = make([]byte, len(out))
+		copy(res, out)
+	}
+
+	*bufp = buf[:0]
+	bytesPool.Put(bufp)
```
puellanivis: As with the other functions, `defer`ing this lets us use early returns.

decompress_test.go
```diff
@@ -0,0 +1,99 @@
...
+		CompressionLZ4,
+		CompressionZSTD,
+	}
+
+	const limit = 64 * 1024
+
+	withLimit := func(t *testing.T, v int32) {
+		t.Helper()
+		old := MaxDecompressedBatchSize
+		MaxDecompressedBatchSize = v
```
puellanivis, line 25: Usually its a good idea to put something like this together on one line, so it at least “looks” atomic.
```suggestion
		old, MaxDecompressedBatchSize  := MaxDecompressedBatchSize, v
```

zstd.go
```diff
@@ -73,19 +73,32 @@ func releaseEncoder(params ZstdEncoderParams, enc *zstd.Encoder) {
...
+	if maxDecodedSize > 0 {
+		opts = append(opts, zstd.WithDecoderMaxMemory(uint64(maxDecodedSize)))
+	}
 	// It's possible to race and create multiple new readers.
 	// Only one will survive GC after use.
-	zstdDec, _ := zstd.NewReader(nil, zstd.WithDecoderConcurrency(0))
-	zstdDecMap.Store(params, zstdDec)
+	zstdDec, _ := zstd.NewReader(nil, opts...)
+	zstdDecMap.Store(key, zstdDec)
 	return zstdDec
```
puellanivis, line 106: 🤔 usually, we want to mutex lock the creation here… to prevent stampedes throwing a bunch of `zstd.Decoder`s on the floor.

decompress.go
dnwe: Yeah io.LimitReader caps the read and hitting that limit isn’t an error, so an oversized batch comes back as n>limit and err==nil

decompress.go
dnwe: Yeah

decompress_test.go
dnwe: That suggestion would give a different/wrong behaviour though right? You’d shadow and update the global MaxDecompressedSize locally rather than updating the global and then setting it back. We could declare the old var, but not sure it’s worthwhile in a test 

Aside: in general I hate globals, but this was the simplest way to add a hook like MaxResponseSize without having to thread it through from config to the encoder/decoder

puellanivis: Oh yeah… 🤦‍♀️

decompress.go
puellanivis: 👍

hindessm (approved):
LGTM

PR #3605, opened 2026-06-03
feat(client): Re-use already connected broker without pending request to fetch metadata

lbcjbb:
When several brokers are not connected or connected and idle, `Client.LeastLoadedBroker()` could establish a new connection to a broker rather than favouring an already connected broker. This happends depending on the iteration order of the brokers's map which is not predictable.

This change makes a slight adjustment of this method to return the first connected and idle broker encountered, which may drastically reduce the total number of connection per broker when a same Kafka cluster is used by several thousand applications.

### DIFF
--- a/client.go
+++ b/client.go
@@ -788,21 +788,50 @@ func (client *client) resurrectDeadBrokers() {
 }
 
 // LeastLoadedBroker returns the broker with the least pending requests.
-// Firstly, choose the broker from cached broker list. If the broker list is empty, choose from seed brokers.
+//
+// The aim is to choose the broker, from the cached broker list, that offers
+// the best balance between the lowest current load and the cost of a new connection,
+// in accordance with the following list of priorities:
+//
+//  1. Returns an already connected broker if and only if it is idle.
+//  2. Returns a broker that is not yet connected. Etablish a new connection
+//     before to return it.
+//  3. Returns the currently connected broker with the least number of pending requests.
+//  4. If the cached list of brokers is empty, returns a seed broker.
+//
+// As the order in which brokers are iterated is not stable, the broker
+// returned may vary from one call to the next if several brokers of the list
+// share the same selection criteria.
 func (client *client) LeastLoadedBroker() *Broker {
 	client.lock.RLock()
 	defer client.lock.RUnlock()
 
-	var leastLoadedBroker *Broker
-	pendingRequests := math.MaxInt
-	for _, broker := range client.brokers {
-		if pendingRequests > broker.ResponseSize() {
-			pendingRequests = broker.ResponseSize()
-			leastLoadedBroker = broker
+	var leastLoadedBroker, notConnectedBroker *Broker
+	leastPendingRequests := math.MaxInt
+
+	for _, b := range client.brokers {
+		isConnected, pendingRequests := b.status()
+		if !isConnected {
+			notConnectedBroker = b
+			continue
+		}
+
+		if pendingRequests == 0 {
+			return b
+		}
+
+		if pendingRequests < leastPendingRequests {
+			leastPendingRequests = pendingRequests
+			leastLoadedBroker = b
 		}
 	}
+
+	if notConnectedBroker != nil {
+		_ = notConnectedBroker.Open(client.conf)
+		return notConnectedBroker
+	}
+
 	if leastLoadedBroker != nil {
-		_ = leastLoadedBroker.Open(client.conf)
 		return leastLoadedBroker
 	}
 
@@ -811,7 +840,7 @@ func (client *client) LeastLoadedBroker() *Broker {
 		return client.seedBrokers[0]
 	}
 
-	return leastLoadedBroker
+	return nil
 }
 
 // private caching/lazy metadata helpers
--- a/broker.go
+++ b/broker.go
@@ -346,6 +346,14 @@ func (b *Broker) Connected() (bool, error) {
 	return b.conn != nil, b.connErr
 }
 
+// status is the short and one lock version of (Connected(), ResponseSize()).
+func (b *Broker) status() (bool, int) {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	return b.conn != nil, len(b.responses)
+}
+
 // TLSConnectionState returns the client's TLS connection state. The second return value is false if this is not a tls connection or the connection has not yet been established.
 func (b *Broker) TLSConnectionState() (state tls.ConnectionState, ok bool) {
 	b.lock.Lock()
--- a/client_test.go
+++ b/client_test.go
@@ -699,6 +699,152 @@ func TestClientGetBroker(t *testing.T) {
 	}
 }
 
+func TestClientLeastLoadedBroker(t *testing.T) {
+	createFakeBroker := func(t *testing.T, brokerID int32, connected bool, pendingRequests int) *Broker {
+		t.Helper()
+
+		mb := NewMockBroker(t, brokerID)
+		t.Cleanup(mb.Close)
+
+		broker := NewBroker(mb.Addr())
+		broker.id = brokerID
+
+		if connected {
+			conn, err := net.Dial("tcp", mb.listener.Addr().String())
+			require.NoErrorf(t, err, "Broker %d", brokerID)
+
+			broker.conn = conn.(*net.TCPConn)
+			broker.metricRegistry = metrics.NewRegistry()
+			broker.opened.Store(true)
+
+			if pendingRequests > 0 {
+				broker.responses = make(chan *responsePromise, pendingRequests)
+				for range pendingRequests {
+					broker.responses <- new(responsePromise)
+				}
+			}
+		}
+
+		return broker
+	}
+
+	createTestClient := func(t *testing.T, seedBrokers []*Broker, brokers []*Broker) Client {
+		t.Helper()
+
+		conf := NewTestConfig()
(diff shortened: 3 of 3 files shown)


client.go
```diff
@@ -788,19 +788,36 @@ func (client *client) resurrectDeadBrokers() {
 }
 
 // LeastLoadedBroker returns the broker with the least pending requests.
-// Firstly, choose the broker from cached broker list. If the broker list is empty, choose from seed brokers.
+//
+// Firstly, choose the broker from cached broker list.
+// If the broker list is empty, choose from seed brokers.
+//
+// If an already connected broker has no pending requests, it will be returned
+// to avoid to create a new connection to another broker.
```
puellanivis:
```suggestion
// to avoid creating a new connection to another broker.
```

client.go
```diff
@@ -788,19 +788,36 @@ func (client *client) resurrectDeadBrokers() {
...
-	var leastLoadedBroker *Broker
-	pendingRequests := math.MaxInt
-	for _, broker := range client.brokers {
-		if pendingRequests > broker.ResponseSize() {
-			pendingRequests = broker.ResponseSize()
-			leastLoadedBroker = broker
+	var (
+		leastLoadedBroker *Broker
+		pendingRequests   = math.MaxInt
+	)
```
puellanivis: This code didn’t need to change. It was fine the way it was:

```suggestion
		var leastLoadedBroker *Broker
		pendingRequests := math.MaxInt
```

client.go
```diff
@@ -788,19 +788,36 @@ func (client *client) resurrectDeadBrokers() {
...
+
+	// map iteration is not stable, the chosen broker
+	// may change from one call to the next.
+	for _, b := range client.brokers {
+		rs := b.ResponseSize()
+
+		if isConnected, _ := b.Connected(); isConnected && rs == 0 {
+			return b
+		}
```
puellanivis: Another option might be to early continue if not `b.Connected()`, this would then leave `leastLoadedBroker` as nil if all clients are disconnected. At which point, we can just choose the zeroth `client.brokers[0]` if `len(client.brokers) > 0` before moving on to the seedBroker. But now that I think about it, there’s not really a good reason to keep going if we’re just going to return one of these brokers that are connected and `rs == 0`.

Second though, we can wrap the connected test in a simple `if rs == 0 { }` that way we only check the connected status of any brokers that are idle or disconnected, which would give us a short-circuit out of a mutex lock to see if we’re connected… alternatively, we could add a new unexported function that does `return b.conn != nil, len(b.responses)` all in one mutex lock.

lbcjbb: `client.brokers` could be splitted to two list:
- List of connected brokers:
  - [**S**] If one broker is idle, return it. It's the best choice.
  - [**B**] Otherwise choose the one with the least load.
- List of not connected brokers:
  - [**A**] Any broker of this list is a candidate.

In terms of priority:
  **S** > **A** > **B** > _(fallback to the first broker of `client.seedBrokers`)_

We could have something like:

``` golang
func (client *client) LeastLoadedBroker() *Broker {
	client.lock.RLock()
	defer client.lock.RUnlock()

	var leastLoadedBroker, notConnectedBroker *Broker
	leastPendingRequests := math.MaxInt

	// map iteration is not stable, the chosen broker
	// may change from one call to the next.
	for _, b := range client.brokers {
		connected, pendingRequests := b.status()
		if !connected {
			notConnectedBroker = b
			continue
		}

		if pendingRequests == 0 {
			return b
		}

		if pendingRequests < leastPendingRequests {
			leastPendingRequests = pendingRequests
			leastLoadedBroker = b
		}
	}

	if notConnectedBroker != nil {
		_ = notConnectedBroker.Open(client.conf)
		return notConnectedBroker
	}

	if leastLoadedBroker != nil {
		return leastLoadedBroker
	}

	if len(client.seedBrokers) > 0 {
		_ = client.seedBrokers[0].Open(client.conf)
		return client.seedBrokers[0]
	}

	return nil
}

// ...

// status is the short and one lock version of (Connected(), ResponseSize()).
func (b *Broker) status() (bool, int) {
	b.lock.Lock()
	defer b.lock.Unlock()

	return b.conn != nil, len(b.responses)
}
```

Have I understood your point correctly?

puellanivis: Yes. 👍

dnwe:
Thanks, seems like a good refinement - can you add a simple TestLeastLoadedBroker unittest just to lock in the deterministic behaviour - with connected idle broker and a couple of disconnected brokers assert that no connection is opened on the others?

lbcjbb:
@puellanivis @dnwe Thank you for your review :pray:.
I've just updated the pull request. Please let me know if you spot anything else that needs improving or correcting.

dnwe:
Hmm, I'm not sure about these [most recent updates ](https://github.com/IBM/sarama/compare/d5f32a6e482db48abcb53ffd907f31537a87f2be..239728c3e33dd75b339b7b6fe95cd67ff5483fb1)

Refactoring to add `brokerConnectionState` and `getConnectionState(withTLS bool)` just to support this selection logic feels like too much abstraction for a very small feature? The `ResponseSize()` and `Connected()` funcs were simple and direct before; now they route through this struct-returning helper that seemingly also optionally inspects TLS state?

The `TestClientLeastLoadedBroker` is also a lot larger than I was expecting and tightly coupled to implementation details rather than just the observable behaviour

Regarding the change in general, the new priority order seems to be:

1. prefer idle connected broker
2. then prefer any disconnected broker
3. only then use least-loaded connected broker

That means a healthy connected broker with 1 pending request now loses to a completely disconnected broker, forcing a new connection? That doesn't seem to meet the goal that your original issue description was aiming for?

lbcjbb:
> Refactoring to add brokerConnectionState and getConnectionState(withTLS bool) just to support this selection logic feels like too much abstraction for a very small feature? The ResponseSize() and Connected() funcs were simple and direct before; now they route through this struct-returning helper that seemingly also optionally inspects TLS state?

I've added this new func `getConnectionState()` to avoid to duplicate the logic under `ResponseSize()` and `Connected()` inside a new func with a single lock to return these required info, as proposed by @puellanivis. I extended it to `TLSConnectionState()` just to make all these getter funcs consistent.
If calling `ResponseSize()` **and** `Connected()` is not a problem, I can revert this part.

> The TestClientLeastLoadedBroker is also a lot larger than I was expecting and tightly coupled to implementation details rather than just the observable behaviour

I can't see how we could do it any other way, or perhaps by introducing brokers with fake connection when necessary.

> That means a healthy connected broker with 1 pending request now loses to a completely disconnected broker, forcing a new connection? That doesn't seem to meet the goal that your original issue description was aiming for?

I’ve been wondering about that. Is it better to prioritise a healthy connected broker, even if this broker is still processing a request? Or assume the costs involve in establishing a new connection to another broker?

puellanivis:
I was proposing an equally simple and direct third function that returned two pieces of information that would otherwise be acquired through two mutex locks.

I did not suggest that there should be a One Function To Rule Them All™ that each of the simple functions should call into and then return specific values from. We’re allowed to have a little bit of copying… as a treat. Especially, when that little bit of copying keeps everything simple.

Three simple and direct functions _heavily_ outweighs one complicated function called by three functions that then prune out the results they care about.

puellanivis:
> That means a healthy connected broker with 1 pending request now loses to a completely disconnected broker, forcing a new connection? That doesn't seem to meet the goal that your original issue description was aiming for?

Current existing behavior did not account for if the broker was connected or not. So the previous behavior was to prefer the first broker that was either idle, or disconnected. (Both would have a RequestSize of zero.)

We probably do want to switch to preferring the least loaded connected broker, especially if it’s under some magic number of requests. Like you say, a broker with a RequestSize of 1 is probably better than connecting out to another broker. But if all the connected Brokers are all swamped, then we probably really do want to connect out to a disconnected broker.

lbcjbb:
> I was proposing an equally simple and direct third function that returned two pieces of information that would otherwise be acquired through two mutex locks.

> I did not suggest that there should be a One Function To Rule Them All™ that each of the simple functions should call into and then return specific values from. We’re allowed to have a little bit of copying… as a treat. Especially, when that little bit of copying keeps everything simple.

> Three simple and direct functions heavily outweighs one complicated function called by three functions that then prune out the results they care about.

It's fixed

lbcjbb:
Do you know how it was implemented in the [confluent-kafka-go](https://github.com/confluentinc/confluent-kafka-go), or a Kafka library in Python or Java?

lbcjbb:
In the _librdkafka_ library, [rd_kafka_metadata()](https://github.com/confluentinc/librdkafka/blob/master/src/rdkafka_metadata.c#L87) calls [rd_kafka_broker_weight_usable](https://github.com/confluentinc/librdkafka/blob/master/src/rdkafka_broker.c#L1574).

```
 * @brief Weighing function to select a usable broker connections,
 *        promoting connections according to the scoring below.
 *
 * Priority order:
 *  - is not a bootstrap broker
 *  - least idle last 10 minutes (unless blocking)
 *  - least idle hours (if above 10 minutes idle)
 *  - is not a logical broker (these connections have dedicated use and should
 *                             preferably not be used for other purposes)
 *  - is not blocking
 *
 * Will prefer the most recently used broker connection for two reasons:
 * - this connection is most likely to function properly.
 * - allows truly idle connections to be killed by the broker's/LB's
 *   idle connection reaper.
 *
 * Connection must be up.
 ```

PR #3608, opened 2026-06-05
chore: update default Kafka version to 2.6

dnwe:
Now we support the full set of 2.6 protocols we can make it the default pin


hindessm (approved):
LGTM

puellanivis (approved):

PR #3609, opened 2026-06-05
feat: support Fetch v12

dnwe:
Add Kafka 2.7 Fetch protocol support with flexible encoding, request last-fetched epoch, and response diverging-epoch/current-leader tagged fields.

Also fix nil handling in the flexible bytes codec: a nil slice now encodes as a compact null (0x00) rather than an empty string (0x01), and zero-length compact bytes decode to nil. This affects all flexible messages passing nil through putBytes, not just Fetch.


hindessm (approved):
LGTM

PR #3612, opened 2026-06-07
fix: avoid makeslice "len out of range" panics on invalid response

hindessm:
A few cases permit nullable arrays but most should be `errInvalidArrayLength` errors.

### DIFF
--- a/create_topics_response.go
+++ b/create_topics_response.go
@@ -64,6 +64,9 @@ func (c *CreateTopicsResponse) decode(pd packetDecoder, version int16) (err erro
 	if err != nil {
 		return err
 	}
+	if n < 0 {
+		return errInvalidArrayLength
+	}
 
 	c.TopicErrors = make(map[string]*TopicError, n)
 	if version >= 5 {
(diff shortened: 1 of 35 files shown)


create_topics_response.go
```diff
@@ -245,15 +248,19 @@ func (r *CreatableTopicResult) decode(pd packetDecoder, version int16) (err erro
...
-	for i := 0; i < n; i++ {
-		name, err := pd.getString()
-		if err != nil {
-			return err
-		}
-		r.Configs[name] = &CreatableTopicConfigs{}
-		if err := r.Configs[name].decode(pd, version); err != nil {
-			return err
+	if n < 0 {
+		// null allowed
```
dnwe, line 252: Interesting, I’m mildly surprised the linter didn’t complain about the empty branch here and force you to do a single if n > 0 for r.Configs instead

dnwe (approved):
LGTM

create_topics_response.go
hindessm: In order to keep the API the same, it would need to be `n >= 0`.
There is a little bit of inconsistency here across the protocol implementation. Some empty maps are returned as `nil` maps - even when the API defines them as non-nullable - and some are really empty maps.
I'm not decided on what the correct choice would be for a v2. Go is not java so `nil` maps aren't as problematic as `null` in java so I'd be inclined to just accept it and return nil for empty map cases.

puellanivis (commented):
We could have maybe also introduced a `getNonNullableArrayLength` 🤔 And the only difference is that it doesn’t accept any negative lengths at all.

PR #3613, opened 2026-06-08
fix: avoid makeslice panic in request decode too

dnwe:
Commit 0931fbcf added this robustness check to the response decode, which are what sarama as a client to decode over-the-wire replies. We do currently have decode implemented also for our request structs, even though Sarama itself never needs to decode them outside of unittests, it makes sense to add the same robustness fix just for feature parity and for any 3rdparty using the decodes in wireshark or similar.

### DIFF
--- a/acl_create_request.go
+++ b/acl_create_request.go
@@ -31,6 +31,9 @@ func (c *CreateAclsRequest) decode(pd packetDecoder, version int16) (err error) 
 	if err != nil {
 		return err
 	}
+	if n < 0 {
+		return errInvalidArrayLength
+	}
 
 	c.AclCreations = make([]*AclCreation, n)
 
--- a/acl_delete_request.go
+++ b/acl_delete_request.go
@@ -32,6 +32,9 @@ func (d *DeleteAclsRequest) decode(pd packetDecoder, version int16) (err error) 
 	if err != nil {
 		return err
 	}
+	if n < 0 {
+		return errInvalidArrayLength
+	}
 
 	d.Filters = make([]*AclFilter, n)
 	for i := 0; i < n; i++ {
--- a/add_partitions_to_txn_request.go
+++ b/add_partitions_to_txn_request.go
@@ -50,6 +50,9 @@ func (a *AddPartitionsToTxnRequest) decode(pd packetDecoder, version int16) (err
 	if err != nil {
 		return err
 	}
+	if n < 0 {
+		return errInvalidArrayLength
+	}
 
 	a.TopicPartitions = make(map[string][]int32)
 	for i := 0; i < n; i++ {
--- a/alter_client_quotas_request.go
+++ b/alter_client_quotas_request.go
@@ -55,6 +55,9 @@ func (a *AlterClientQuotasRequest) decode(pd packetDecoder, version int16) error
 	if err != nil {
 		return err
 	}
+	if entryCount < 0 {
+		return errInvalidArrayLength
+	}
 	if entryCount > 0 {
 		a.Entries = make([]AlterClientQuotasEntry, entryCount)
 		for i := range a.Entries {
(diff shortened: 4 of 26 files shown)


hindessm (approved):
LGTM but is it complete? For instance, I'd have expected a guard [here](https://github.com/IBM/sarama/blob/0931fbcfcac895bc1631c6118f13559fcc1c6690/add_partitions_to_txn_request.go#L53).
And, for example, [here](https://github.com/IBM/sarama/blob/0931fbcfcac895bc1631c6118f13559fcc1c6690/alter_configs_request.go#L101) we are currently incorrectly permitting null for a non-nullable array.

I was planning to do this shortly so feel free to merge if you like and I'll create PR with any you missed.

puellanivis (commented):
I guess I could have raised the point of a `getNonNullableArrayLength` function here. 😂

hindessm:
> I guess I could have raised the point of a `getNonNullableArrayLength` function here. 😂

I was a little tempted when writing the `_response.go` PR that prompted this PR. However, given that the most common is non-nullable I'd be more tempted to make `getArrayLength` have the non-null check and create a new `getNullableArrayLength` for the "special"/less-common case.

I will have a look at refactoring if/when Dom takes me up on my offer to create a follow up to this PR.

dnwe:
> > I guess I could have raised the point of a `getNonNullableArrayLength` function here. 😂
> 
> I was a little tempted when writing the `_response.go` PR that prompted this PR. However, given that the most common is non-nullable I'd be more tempted to make `getArrayLength` have the non-null check and create a new `getNullableArrayLength` for the "special"/less-common case.
> 

The slight nuance there is that it would invert the failure mode. Currently if we forget a guard in a particular decoder, the worst thing that happens is malformed input slips through and gets silently accepted or causes a panic. A refactor to getArrayLength vs getNullableArrayLength could mean that sarama rejects a valid null array that a real broker sent because we chose the wrong getArray func when we implemented the protocol?

dnwe:
@hindessm FYI I added a commit which I _think_ now covers all the _request.go decoders

hindessm:
> > > I guess I could have raised the point of a `getNonNullableArrayLength` function here. 😂
> > 
> > 
> > I was a little tempted when writing the `_response.go` PR that prompted this PR. However, given that the most common is non-nullable I'd be more tempted to make `getArrayLength` have the non-null check and create a new `getNullableArrayLength` for the "special"/less-common case.
> 
> The slight nuance there is that it would invert the failure mode. Currently if we forget a guard in a particular decoder, the worst thing that happens is malformed input slips through and gets silently accepted or causes a panic. A refactor to getArrayLength vs getNullableArrayLength could mean that sarama rejects a valid null array that a real broker sent because we chose the wrong getArray func when we implemented the protocol?

"silently accepted or causes a panic" vs a well-defined error doesn't seem like a difficult choice to me. Besides, hasn't this choice already been made in the `getString` / `getNullableString` case? What I am suggesting is consistent with that?

hindessm (approved):
That looks better though there are still a few edge cases - like describe_configs_request.go which correctly copes with `confLength == -1` but not with other negative `confLength` values, similarly in describe logdirs and describe user scram creds, etc.

hindessm (approved):
LGTM. I can't even think of anything else to complain about. (Except perhaps that you've not [yet] replied to my [comment](https://github.com/IBM/sarama/pull/3613#issuecomment-4650647564) :-)

dnwe:
> LGTM. I can't even think of anything else to complain about. (Except perhaps that you've not [yet] replied to my [comment](https://github.com/IBM/sarama/pull/3613#issuecomment-4650647564) :-)

I'll save any possible nitpicking for your PR that refactors all of this into neat concise code :-)

hindessm:
> > LGTM. I can't even think of anything else to complain about. (Except perhaps that you've not [yet] replied to my [comment](https://github.com/IBM/sarama/pull/3613#issuecomment-4650647564) :-)
> 
> I'll save any possible nitpicking for your PR that refactors all of this into neat concise code :-)

That PR is being held hostage until my comment is answered :-)

PR #3614, opened 2026-06-08
chore: extend fuzz testing to more types

dnwe:
Now we have proper guard rails on our decoders/encoders add more fuzzing coverage

### DIFF
--- a/request.go
+++ b/request.go
@@ -239,3 +239,99 @@ func allocateBody(key, version int16) protocolBody {
 	}
 	return nil
 }
+
+// allocateResponseBody returns a fresh response struct for the given api key
+// and version, or nil when the key is unknown. Not used at runtime, but
+// mirrors allocateBody so the unittests and fuzz harness can drive every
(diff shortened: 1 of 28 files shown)


request.go
```diff
@@ -239,3 +239,99 @@ func allocateBody(key, version int16) protocolBody {
 	}
 	return nil
 }
+
+// allocateResponseBody returns a fresh response struct for the given api key
+// and version, or nil when the key is unknown. Not used at runtime, but
+// mirrors allocateBody so the unittests and fuzz harness can drive every
+// supported response decoder uniformly.
+//
+//nolint:unused
```
hindessm, line 248: Staticcheck needs pragma?
```
//lint:ignore U1000 ...
```

dnwe: Bah, yeah mine was fixing golangci-lint when functional tag is on, but I forgot we run staticcheck separately in CI too

hindessm (approved):
LGTM

PR #3619, opened 2026-06-08
chore: clean up Int32 array encode/decode

hindessm:
Add explicit nullable version and ensure we use the helpers in a few more cases.

Compared to a recent PR, I made a different choice about lengths less than -1. I chose to make them an error rather than assuming the sender "meant" -1. There are lots of values of int16 < -1 giving lots of ways to accidentally be accepted if we don't treat them as errors and supposing we have a corrupt/incorrect packet, it will likely still error on subsequent data anyway so it is less confusing to fail on the first error and not accept the length < -1.

### DIFF
--- a/real_decoder.go
+++ b/real_decoder.go
@@ -233,12 +233,15 @@ func (rd *realDecoder) getNullableString() (*string, error) {
 	return &tmpStr, err
 }
 
-func (rd *realDecoder) getInt32Array() ([]int32, error) {
+func (rd *realDecoder) getNullableInt32Array() ([]int32, error) {
 	n, err := rd.getArrayLength()
 	if err != nil {
 		return nil, err
 	}
-	if n <= 0 {
+	if n < -1 {
+		return nil, errInvalidArrayLength
+	}
+	if n == -1 {
 		return nil, nil
 	}
 
@@ -253,6 +256,14 @@ func (rd *realDecoder) getInt32Array() ([]int32, error) {
 		rd.off += 4
 	}
 	return ret, nil
+}
+
+func (rd *realDecoder) getInt32Array() ([]int32, error) {
+	ret, err := rd.getNullableInt32Array()
+	if ret == nil && err == nil {
+		return nil, errInvalidArrayLength
+	}
+	return ret, err
 }
 
 func (rd *realDecoder) getInt64Array() ([]int64, error) {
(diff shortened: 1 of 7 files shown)


dnwe:
LGTM, I know it gets tested indirectly, but is it worth maybe adding a few direct real_decoder unit cases (null -1, < -1, empty 0, populated) for both getInt32Array and getNullableInt32Array, plus the flexible equivalents just to get this contract laid out in test form?

dnwe (approved):
Thanks!

real_decoder.go
```diff
@@ -255,6 +258,14 @@ func (rd *realDecoder) getInt32Array() ([]int32, error) {
 	return ret, nil
 }
 
+func (rd *realDecoder) getInt32Array() ([]int32, error) {
+	ret, err := rd.getNullableInt32Array()
+	if ret == nil && err == nil {
+		return nil, errInvalidArrayLength
+	}
+	return ret, err
```
puellanivis, line 266: Hm… quite a clever (derogatory) solution here. I probably would have built this as:
```go
if err != nil {
	return nil, err
}
if ret == nil { // Since this is a non-nullable array, this is an error.
	return nil, errInvalidArrayLength
}
return ret, nil
```

Although, I suspect this is still just as brittle, because Golang handles nil and empty slices so similarly… so the urge to “fix” the allocation of a zero-length slice to a nil slice is strong. So, it probably would make sense to have a comment somewhere in/on the `getNullableInt32Array` that calls out that it distinctly returns a zero-length (zero-capacity) non-nil slice vs a nil slice based on the encoding.

hindessm: > Hm… quite a clever (derogatory) solution here.

:joy:

I wasn't trying for (any interpretation of) "clever" but rather trying to write a condition to handle precisely the "only different state from the nullable case" and pass through the result values otherwise. So in my head (and possibly only _my_ head), it seemed simple and logical. :-)

Further to my defence, I point out that overly-clever code is quite prevalent here already - e.g. [exhibit 1](https://github.com/IBM/sarama/blob/1ae399f54b576613e7bff6bebb448cfd55b094aa/real_decoder.go#L135), [exhibit 2](https://github.com/IBM/sarama/blob/1ae399f54b576613e7bff6bebb448cfd55b094aa/real_decoder.go#L227), etc.

The decoding in different message types is a little inconsistent about returning `nil` vs empty slices. As you suggest, the go semantics (being different from java) make the relevance of the two rather less significant. I haven't checked if any code actually makes use of the difference. We should probably decide what the correct behaviour should be and implement it consistently though.

puellanivis: Yeah, it’s a bit weird when the obvious direct solution might be a little more clever than breaking up the two other cases where the behavior isn’t going to change.

It’s fine, and works, and it’s too difficult to understand. It’s just that by breaking up the two unchanged cases, we can get a little less mental Boolean algebra expression evaluation. :)

PR #3621, opened 2026-06-09
chore: apply Go's modernize fixes

dnwe:
See individual commits for which category was applied in each

### DIFF
--- a/admin.go
+++ b/admin.go
@@ -693,7 +693,7 @@ func (ca *clusterAdmin) AlterPartitionReassignments(topic string, assignment [][
 		Version:   int16(0),
 	}
 
-	for i := 0; i < len(assignment); i++ {
+	for i := range assignment {
 		request.AddBlock(topic, int32(i), assignment[i])
 	}
 
--- a/async_producer_test.go
+++ b/async_producer_test.go
@@ -156,10 +156,10 @@ func TestAsyncProducer(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	for i := 0; i < 10; i++ {
+	for i := range 10 {
 		producer.Input() <- &ProducerMessage{Topic: "my_topic", Key: nil, Value: StringEncoder(TestMessage), Metadata: i}
 	}
-	for i := 0; i < 10; i++ {
+	for i := range 10 {
 		select {
 		case msg := <-producer.Errors():
 			t.Error(msg.Err)
@@ -207,8 +207,8 @@ func TestAsyncProducerMultipleFlushes(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	for flush := 0; flush < 3; flush++ {
-		for i := 0; i < 5; i++ {
+	for range 3 {
+		for range 5 {
 			producer.Input() <- &ProducerMessage{Topic: "my_topic", Key: nil, Value: StringEncoder(TestMessage)}
 		}
 		expectResults(t, producer, 5, 0)
@@ -248,7 +248,7 @@ func TestAsyncProducerMultipleBrokers(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	for i := 0; i < 10; i++ {
+	for range 10 {
 		producer.Input() <- &ProducerMessage{Topic: "my_topic", Key: nil, Value: StringEncoder(TestMessage)}
 	}
 	expectResults(t, producer, 10, 0)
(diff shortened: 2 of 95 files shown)


hindessm (approved):
Not the most exciting PR but LGTM

dnwe:
> Not the most exciting PR but LGTM

😆 - I didn’t previously like the range stuff, but it has grown on me now

mocks/async_producer_test.go
```diff
@@ -33,12 +33,12 @@ func newTestReporterMock() *testReporterMock {
...
+func (trm *testReporterMock) Errorf(format string, args ...any) {
 	trm.errors = append(trm.errors, fmt.Sprintf(format, args...))
 }
 
 func TestMockAsyncProducerImplementsAsyncProducerInterface(t *testing.T) {
-	var mp interface{} = &AsyncProducer{}
+	var mp any = &AsyncProducer{}
 	if _, ok := mp.(sarama.AsyncProducer); !ok {
 		t.Error("The mock producer should implement the sarama.Producer interface.")
 	}
```
puellanivis, line 44: I usually recommend switching these to:
```go
var _ sarama.AsyncProducer = &AsyncProducer{}
```

Then we get compilation errors rather than simple test failures. (Though, I do also put them in `_test.go` files, because they’re tests, not functional code.)

tools/kafka-producer-performance/main.go
```diff
@@ -388,11 +388,10 @@ func runSyncProducer(topic string, partition, messageLoad, messageSize, routines
 	var wg gosync.WaitGroup
 	if throughput > 0 {
 		for _, messages := range messages {
```
puellanivis, line 390: Putting eyes on it, I think this shadowing actually results in some confusion about the logic. One of these should probably be renamed something else. Like “messageChunks”.

admin.go
```diff
@@ -693,7 +693,7 @@ func (ca *clusterAdmin) AlterPartitionReassignments(topic string, assignment [][
 		Version:   int16(0),
 	}
 
-	for i := 0; i < len(assignment); i++ {
+	for i := range assignment {
 		request.AddBlock(topic, int32(i), assignment[i])
 	}
```
puellanivis, line 698: The function parameter should probably be named “assignments”, in addition to:
```suggestion
	for i, assignment := range assignments {
		request.AddBlock(topic, int32(i), assignment)
	}
```

async_producer_test.go
```diff
@@ -207,8 +207,8 @@ func TestAsyncProducerMultipleFlushes(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	for flush := 0; flush < 3; flush++ {
-		for i := 0; i < 5; i++ {
```
puellanivis, line 211: 🎉 No names are best names!†

†: exceptions apply.

balance_strategy_test.go
```diff
@@ -2261,9 +2261,9 @@ func verifyValidityAndBalance(t *testing.T, consumers map[string]ConsumerGroupMe
 
 // Produces the intersection of two slices
 // From https://github.com/juliangruber/go-intersect
-func intersection(a interface{}, b interface{}) []interface{} {
-	set := make([]interface{}, 0)
-	hash := make(map[interface{}]bool)
+func intersection(a any, b any) []any {
+	set := make([]any, 0)
```
puellanivis, line 2265:
```suggestion
	vart set []any
```

logger_test.go
```diff
@@ -14,23 +14,23 @@ type testLogger struct {
 }
 
 // nolint:unused
-func (l *testLogger) Print(v ...interface{}) {
+func (l *testLogger) Print(v ...any) {
 	if l.t != nil {
 		l.t.Helper()
 		l.t.Log(v...)
 	}
```
puellanivis, line 21: [super tiny lint] Should probably be inverted into a guard condition:
```go
if l.t == nil {
	return
}
l.t.Helper()
l.t.Log(v...)
```

metadata_response.go
```diff
@@ -245,7 +245,7 @@ func (r *MetadataResponse) decode(pd packetDecoder, version int16) (err error) {
 	}
 
 	r.Brokers = make([]*Broker, brokerArrayLen)
-	for i := 0; i < brokerArrayLen; i++ {
+	for i := range brokerArrayLen {
```
puellanivis, line 248: Fun: some of these could be “simplified” more with for example: `for i := range r.Brokers`.

At least, they would be more direct about what they’re iterating over. But there would be a lot of these to fix, and no automation for it, I believe.

PR #3623, opened 2026-06-10
fix(test): wait for leaders after topic create

dnwe:
TestFuncAdminListOffsets could fail with ErrNotLeaderForPartition or ErrUnknownTopicOrPartition because topicWithEvenLeaders returned as soon as CreateTopic succeeded, racing the broker-side leader election for the new partitions. Poll metadata until every partition has an elected leader before producing.


hindessm (approved):
LGTM

PR #3629, opened 2026-06-11
feat: support Produce v9

dnwe:
Add the first flexible version of Produce (KIP-482, Kafka 2.8.0): compact encoding, tagged fields and header versions 2/1. Records keep the batch format, framed as compact bytes. Negotiate v9 for brokers at 2.8.0 or later.


encoder_decoder.go
```diff
@@ -132,3 +132,13 @@ func downgradeFlexibleDecoder(pd packetDecoder) packetDecoder {
...
 }
+
+func downgradeFlexibleEncoder(pe packetEncoder) packetEncoder {
+	switch e := pe.(type) {
+	case *prepFlexibleEncoder:
+		return e.prepEncoder
+	case *realFlexibleEncoder:
+		return e.realEncoder
+	}
+	return pe
```
puellanivis: I recommend putting the default return on an actual `default:` clause.

This way, if something happens to one of the other codepaths, where somehow no longer end in a return, the code will cease to compile, with the syntax error that there is a missing return.

encoder_decoder.go
```diff
@@ -132,3 +132,13 @@ func downgradeFlexibleDecoder(pd packetDecoder) packetDecoder {
 	}
 	return pd
 }
+
+func downgradeFlexibleEncoder(pe packetEncoder) packetEncoder {
+	switch e := pe.(type) {
```
puellanivis: Unless there’s a need to retain access to the original variable, I usually recommend that an assigning type-switch should shadow the original interface variable:

```go
	switch pe := pe.(type) {
```

(The general principle is that if accessing the to-be-shadowed variable should be considered an error, then shadowing the variable makes that error impossible.)

produce_request.go
```diff
@@ -169,47 +185,50 @@ func (r *ProduceRequest) decode(pd packetDecoder, version int16) error {
...
-		}
-		if partitionCount < 0 {
-			return errInvalidArrayLength
-		}
-		r.records[topic] = make(map[int32]Records)
 
-		for range partitionCount {
-			partition, err := pd.getInt32()
+	if topicCount > 0 {
+		r.records = make(map[string]map[int32]Records)
```
puellanivis: Maybe we could add a size hint here? I mean, we do have the expected size count.

produce_request.go
```diff
@@ -169,47 +185,50 @@ func (r *ProduceRequest) decode(pd packetDecoder, version int16) error {
...
 			}
-			recordsDecoder, err := pd.getSubset(int(size))
-			if err != nil {
-				return err
+			if partitionCount < 0 {
+				return errInvalidArrayLength
 			}
-			var records Records
-			if err := records.decode(recordsDecoder); err != nil {
+			r.records[topic] = make(map[int32]Records)
```
puellanivis: Same here, size hint?

puellanivis (approved):
Looks acceptable. Just some linty notes.

encoder_decoder.go
lizthegrey: ooh thanks TIL

puellanivis (approved):
LGTM

PR #3632, opened 2026-06-12
feat: support UpdateFeatures 

dnwe:
Add request/response types for the UpdateFeatures API (key 57, KIP-584, Kafka 2.7.0) along with a typed Broker method. Expose the UpdateFeatures API via ClusterAdmin, mirroring the Java AdminClient's updateFeatures

PR #3633, opened 2026-06-12
feat: expose broker features from ApiVersions v3

dnwe:
Parse the `SupportedFeatures`, `FinalizedFeaturesEpoch` and `FinalizedFeatures` tagged fields on the v3 ApiVersionsResponse, and re-encode them on round-trips. This gives clients the same view of broker features as the Java client's `describeFeatures` (KIP-584).

- `FinalizedFeaturesEpoch` defaults to `-1` (unknown) when the tag is absent, matching the Java client
- unknown tags (e.g. `ZkMigrationReady`) continue to be skipped
- adds a small `flexibleEncode` helper to size and emit variable-length tagged-field values through the existing two-pass encoder machinery

Since sarama already negotiates ApiVersions v3 on connection handshake, feature data from KRaft brokers flows into the response struct with no further changes. Follow-up: the `TestFuncAdminUpdateFeatures` FVT from #3632 can switch its version-window gate to a client-side precondition check using these fields.

### DIFF
--- a/api_versions_response.go
+++ b/api_versions_response.go
@@ -49,6 +49,86 @@ func (a *ApiVersionsResponseKey) decode(pd packetDecoder, version int16) (err er
 	return err
 }
 
+// SupportedFeatureKey contains a feature supported by the broker.
+type SupportedFeatureKey struct {
+	// Name contains the name of the feature.
+	Name string
+	// MinVersion contains the minimum supported version for the feature.
+	MinVersion int16
+	// MaxVersion contains the maximum supported version for the feature.
+	MaxVersion int16
+}
+
+func (f *SupportedFeatureKey) encode(pe packetEncoder) error {
+	if err := pe.putString(f.Name); err != nil {
+		return err
+	}
+
+	pe.putInt16(f.MinVersion)
+
+	pe.putInt16(f.MaxVersion)
+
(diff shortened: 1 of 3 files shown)


api_versions_response.go
```diff
@@ -81,9 +167,79 @@ func (r *ApiVersionsResponse) encode(pe packetEncoder) (err error) {
...
+		value, err := encode(taggedFieldValue(func(pe packetEncoder) error {
+			if err := pe.putArrayLength(len(r.FinalizedFeatures)); err != nil {
+				return err
+			}
+			for i := range r.FinalizedFeatures {
+				if err := r.FinalizedFeatures[i].encode(pe); err != nil {
+					return err
+				}
+			}
+			return nil
```
puellanivis, line 222: [out-of-scope] 🤔 We could probably build a generic `encodeArray[E selfEncoder](pe packetEncoder, s []E)` ?

Only, the type inference won’t let a pointer receiver resolve from a value receiver.
```
filename.go:$lineno:$colno: in call to encodeArray, E (type SupportedFeatureKey) does not satisfy selfEncoder (method encode has pointer receiver)
```

So, we would also have to use slices of pointers, which we could start doing for new stuff, but not for older stuff…

puellanivis (approved):

PR #3634, opened 2026-06-12
feat: support ApiVersions V4 and discover upgradable features in test

dnwe:
Rather than hardcoding share.version on a 4.1 version window, upgrade whichever features the controller advertises below their supported max and verify the finalized levels afterwards.


hindessm (approved):
LGTM

PR #3636, opened 2026-06-15
feat: support Metadata V11

dnwe:
Bump MetadataRequest/Response to v11, which deprecates the cluster-authorized-operations fields (now served by DescribeCluster) per KIP-700. Negotiate v11 at Kafka 2.8.0.


puellanivis (approved):
Neat new version change. Having something deprecated/removed. :)

PR #3637, opened 2026-06-16
feat: support Produce V10

dnwe:
Bump ProduceRequest/Response to v10 (KIP-951). The request is unchanged from v9; the response gains CurrentLeader (per-partition) and NodeEndpoints (top-level) as tagged fields, letting the broker return the new leader on NOT_LEADER_OR_FOLLOWER. Negotiate v10 at Kafka 3.7.0.


hindessm (approved):
LGTM

PR #3638, opened 2026-06-16
fix: correct fetch v12 tagged struct encoding

dnwe:
The DivergingEpoch and CurrentLeader tagged fields are structs, so in a flexible response each needs its own trailing empty tagged-field array. Without it the encoder wrote a length one byte short, so the bytes didn't match what a broker sends.

Only encoding was affected; decoding slices by the declared length and ignored the missing byte.


hindessm (approved):
LGTM

PR #1203, opened 2018-10-30
Revert to individual msg retries for non-idempotent

mimaison:
This should fix #1187 

Also do idempotent batch retries in a goroutine. This avoids the deadlock but does not
solve a race condition that can lead to out-of-sequence errors.
Idempotent producer is at proof-of-concept quality

Co-authored-by: Edoardo Comar <ecomar@uk.ibm.com>
Co-authored-by: Mickael Maison <mickael.maison@gmail.com>

PR #3639, opened 2026-06-17
producer: keep non-idempotent connection retries ordered

3AceShowHand:
## Problem

Ref #2860, #2619, #3422, #1203.

On `main`, the producer muter prevents concurrent in-flight requests for the same topic/partition. That keeps later batches from passing an earlier batch while the earlier batch is still owned by a broker request.

The remaining gap is the broker-level error path in `brokerProducer.handleError`. When a produce request fails without a per-partition `ProduceResponse` (for example EOF, broken pipe, or a read timeout), non-idempotent producers currently move the sent messages to the retry queue and release the sent partition mute. A later same-partition batch can then become flushable before the failed batch is handed back to a broker.

A possible ordering on `main` is:

1. batch A is sent and the broker connection fails
2. batch A is moved to the retry queue
3. the partition mute is released
4. batch B for the same partition flushes first
5. batch A is retried later

That breaks producer-side ordering for the same topic/partition. Enabling idempotence is not always available to users because some Kafka deployments require additional permissions such as `IdempotentWrite`.

## Change

This PR changes the broker-level error path for sent batches:

- `handleError` keeps sent topic/partition batches muted while they still have retry attempts left.
- Retryable sent partition batches are passed to one retry worker for the failed produce request.
- The retry worker refreshes metadata once, resolves refreshed leaders, groups retry batches by target broker, and hands one grouped `produceSet` to each broker.
- Unsent `accumulatingBatch` messages stay on the existing `retryMessages` path.
- Sent batches that cannot retry are unmuted and failed.

This keeps the failed sent batch ahead of later same-partition work without blocking the `brokerProducer` response loop.

## Retry Semantics

The direct retry worker uses the same batch boundary as `retryBatch`: if any message in a same-partition batch has exhausted retries, the whole partition batch fails. That avoids splitting one broker request's partition batch into partly failed and partly retried messages, which would weaken the ordering boundary this PR is preserving.

The worker also uses the producer retry backoff path and exits on producer shutdown. If shutdown or broker handoff prevents ownership transfer, it releases the partition mute and returns errors for the affected messages.

## What Is Not Changed

This PR does not change the non-idempotent `ProduceResponse` retry path in `handleSuccess`.

That path still uses `retryMessages` and goes back through `partitionProducer`, where the retry high watermark and per-partition dispatch state are maintained. The new direct retry worker is only for broker-level failures where there is no per-partition `ProduceResponse` to feed back through `partitionProducer`.

## Retry Buffer Accounting

`Producer.Retry.MaxBufferLength` and `Producer.Retry.MaxBufferBytes` now apply to producer retry buffering as a whole:

- the existing bridge between `input` and `retries` in `retryHandler`
- direct retry batches waiting for metadata refresh, backoff, leader lookup, partition mute, or broker handoff

Both limits remain disabled by default. Small positive limits still use the existing minimum functional thresholds.

## Tests

Added or updated coverage for:

- keeping non-idempotent broker-error retries muted
- failing a whole same-partition batch when any message has exhausted retries
- keeping metadata refresh out of the broker response loop
- refreshing metadata once per failed produce request
- grouping direct retry batches by refreshed leader
- keeping non-idempotent `ProduceResponse` retries on the `partitionProducer` path
- releasing mutes on leader lookup failure, shutdown, or broker handoff failure
- applying shared retry buffer length and byte budgets to direct retries and `retryHandler`
- waking `brokerProducer.run` after an external unmute

Local verification:

```sh
git diff --check
go test -run 'Test(HandleErrorNonIdempotentRetry|RetryBatchesAfterRefresh|RetryBatch|ProducerRetryBufferLimits|BrokerProducerRunFlushesAfterExternalUnmute|AsyncProducerRetryOrdering|AsyncProducerPartitionUnmuting)' .
go test ./...
```

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/eapache/go-resiliency/breaker"
@@ -103,6 +104,10 @@ type asyncProducer struct {
 	errors                    chan *ProducerError
 	input, successes, retries chan *ProducerMessage
 	inFlight                  sync.WaitGroup
+	// shutdownCh is closed before waiting on in-flight messages so retryBatch
+	// goroutines can stop waiting on broker handoff and release partition mutes.
+	shutdownCh       chan struct{}
+	shutdownChClosed atomic.Bool
 
 	brokers    map[*Broker]*brokerProducer
 	brokerRefs map[*brokerProducer]int
@@ -314,6 +319,7 @@ func newAsyncProducer(client Client) (AsyncProducer, error) {
 		input:           make(chan *ProducerMessage),
 		successes:       make(chan *ProducerMessage),
 		retries:         make(chan *ProducerMessage),
+		shutdownCh:      make(chan struct{}),
 		brokers:         make(map[*Broker]*brokerProducer),
 		brokerRefs:      make(map[*brokerProducer]int),
 		txnmgr:          txnmgr,
@@ -991,6 +997,7 @@ func (p *asyncProducer) newBrokerProducer(broker *Broker) *brokerProducer {
 		input:             input,
 		output:            bridge,
 		responses:         responses,
+		done:              make(chan struct{}),
 		accumulatingBatch: newProduceSet(p),
 		currentRetries:    make(map[string]map[int32]error),
 	}
@@ -1104,6 +1111,11 @@ type brokerProducer struct {
 	output    chan<- *produceSet
 	responses <-chan *brokerProducerResponse
 	abandoned chan struct{}
+	// done is closed before output is closed. Direct retry handoff waits on
+	// output while the batch still owns its partition mute; if the brokerProducer
+	// is shutting down, that send may never be received and the mute would block
+	// later same-partition messages and producer shutdown.
+	done chan struct{}
 
 	accumulatingBatch *produceSet
 	flushingBatch     *produceSet // batch that has been muted and is ready to send
@@ -1246,6 +1258,9 @@ func (bp *brokerProducer) tryBuildFlushingBatch() bool {
 }
 
 func (bp *brokerProducer) shutdown() {
+	if bp.done != nil {
+		close(bp.done)
+	}
 	// flush any ready buffer
 	for bp.flushingBatch != nil {
 		select {
@@ -1476,12 +1491,21 @@ func (p *asyncProducer) retryBatch(topic string, partition int32, pSet *partitio
 	produceSet.msgs[topic][partition] = pSet
 	produceSet.bufferBytes += pSet.bufferBytes
 	produceSet.bufferCount += len(pSet.msgs)
+	muted := alreadyMuted
+	failBatch := func(err error) {
+		// Release the partition before reporting errors so a blocked Errors
+		// consumer cannot keep later same-partition messages muted.
+		if muted {
+			p.muter.unmute(produceSet)
+			muted = false
+		}
+		for _, msg := range pSet.msgs {
+			p.returnError(msg, err)
+		}
+	}
 	for _, msg := range pSet.msgs {
 		if msg.retries >= p.conf.Producer.Retry.Max {
-			p.returnErrors(pSet.msgs, retryErr)
-			if alreadyMuted {
-				p.muter.unmute(produceSet)
-			}
+			failBatch(retryErr)
 			return
 		}
 		msg.retries++
@@ -1497,25 +1521,88 @@ func (p *asyncProducer) retryBatch(topic string, partition int32, pSet *partitio
 	leader, leaderErr := p.client.Leader(topic, partition)
 	if leaderErr != nil {
 		Logger.Printf("Failed retrying batch for %v-%d because of %v while looking up for new leader\n", topic, partition, leaderErr)
-		for _, msg := range pSet.msgs {
-			p.returnError(msg, retryErr)
-		}
-		if alreadyMuted {
-			p.muter.unmute(produceSet)
-		}
+		failBatch(retryErr)
 		return
 	}
 	if !alreadyMuted {
 		if !p.muter.waitUntilMuted(produceSet) {
-			for _, msg := range pSet.msgs {
-				p.returnError(msg, retryErr)
-			}
+			failBatch(retryErr)
 			return
 		}
+		muted = true
 	}
 	bp := p.getBrokerProducer(leader)
-	bp.output <- produceSet
-	p.unrefBrokerProducer(leader, bp)
+	defer p.unrefBrokerProducer(leader, bp)
+	if !p.sendRetryBatch(bp, produceSet) {
+		failBatch(ErrShuttingDown)
+		return
+	}
+}
+
+type partitionBatchRetry struct {
+	topic     string
+	partition int32
+	pSet      *partitionSet
+}
+
+// retryBatchesAfterRefresh keeps metadata refresh out of brokerProducer.handleError.
+// Connection errors already hold partition mutes; doing the refresh in the
+// response loop can block all progress for that broker while the cluster is
+// unstable. Refresh once for the failed request so multi-partition batches do
+// not amplify controller-fail metadata traffic, then retry partitions
+// independently so one blocked broker handoff cannot hold up the rest.
+func (p *asyncProducer) retryBatchesAfterRefresh(topics []string, batches []partitionBatchRetry, retryErr error) {
+	if p.shutdownCh != nil {
+		select {
+		case <-p.shutdownCh:
+			for _, batch := range batches {
+				go p.retryBatch(batch.topic, batch.partition, batch.pSet, retryErr, true)
+			}
+			return
+		default:
+		}
+	}
+	if len(topics) > 0 {
+		if err := p.client.RefreshMetadata(topics...); err != nil {
+			Logger.Printf("Failed refreshing metadata because of %v\n", err)
+		}
+	}
+	for _, batch := range batches {
+		go p.retryBatch(batch.topic, batch.partition, batch.pSet, retryErr, true)
+	}
+}
+
+// sendRetryBatch transfers ownership of a muted retry batch to the broker
+// bridge. A false result means no owner accepted the batch, so the caller must
+// release the mute and fail the messages instead of leaving a retry goroutine
+// blocked while it still owns the partition mute.
+func (p *asyncProducer) sendRetryBatch(bp *brokerProducer, set *produceSet) bool {
+	var done <-chan struct{}
+	if bp.done != nil {
+		done = bp.done
+	}
+	select {
+	case <-done:
+		return false
+	default:
+	}
+	select {
+	case bp.output <- set:
+		return true
+	default:
+	}
+	var shutdownCh <-chan struct{}
+	if p.shutdownCh != nil {
+		shutdownCh = p.shutdownCh
+	}
+	select {
+	case bp.output <- set:
+		return true
+	case <-done:
+		return false
+	case <-shutdownCh:
+		return false
+	}
 }
 
 func (bp *brokerProducer) handleError(sent *produceSet, err error) {
@@ -1530,22 +1617,10 @@ func (bp *brokerProducer) handleError(sent *produceSet, err error) {
 		bp.parent.abandonBrokerConnection(bp.broker)
 		_ = bp.broker.Close()
 		bp.closing = err
+		keepMuted := make(map[string]map[int32]struct{})
 		var retryTopics []string
 		retryTopicSeen := make(map[string]struct{})
-		sent.eachPartition(func(topic string, partition int32, pSet *partitionSet) {
-			if _, ok := retryTopicSeen[topic]; ok {
-				return
-			}
-			retryTopicSeen[topic] = struct{}{}
-			retryTopics = append(retryTopics, topic)
-		})
-		if len(retryTopics) > 0 {
-			refreshErr := bp.parent.client.RefreshMetadata(retryTopics...)
-			if refreshErr != nil {
-				Logger.Printf("Failed refreshing metadata because of %v\n", refreshErr)
-			}
-		}
-		keepMuted := make(map[string]map[int32]struct{})
+		var retryBatches []partitionBatchRetry
 		sent.eachPartition(func(topic string, partition int32, pSet *partitionSet) {
 			// Connection failures have no ProduceResponse to feed back through
 			// partitionProducer. Keep the sent batch muted and retry it asynchronously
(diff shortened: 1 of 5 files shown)


async_producer.go
```diff
@@ -804,15 +820,51 @@ func (pp *partitionProducer) backoff(retries int) {
...
+	backoff := p.retryBackoff(retries)
+	if backoff <= 0 {
+		return !p.shuttingDown()
+	}
+	timer := time.NewTimer(backoff)
+	defer timer.Stop()
+	if p.shutdownCh == nil {
+		<-timer.C
+		return true
+	}
```
puellanivis: There’s no reason to check for `shutdownCh == nil` since a receive on a nil channel will not panic, it will just forever block.

This means if `p.shutdownCh == nil` then the `select` will just become a `<-time.C` alone.

async_producer.go
```diff
@@ -1227,17 +1292,21 @@ func (bp *brokerProducer) tryBuildFlushingBatch() bool {
...
+	if partial != nil {
+		bp.flushingBatch = partial
+		if bp.accumulatingBatch.empty() {
+			bp.rollOver()
+		}
+		return true
 	}
-	return true
+
+	return false
```
puellanivis: This change does not have any change in behavior, and shouldn’t be in scope of this PR.

(It’s just adding noise about changes, when there’s no change in behavior. Cleanup operations or reflowing code should all be done in a PR targeting that specific activity.)

async_producer.go
```diff
@@ -1461,51 +1533,318 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
...
+	}
+	select {
+	case bp.output <- set:
+		return true
+	default:
+	}
+	var shutdownCh <-chan struct{}
+	if p.shutdownCh != nil {
+		shutdownCh = p.shutdownCh
+	}
```
puellanivis: This nil guard is ineffective. If `p.shutdownCh` is nil we skip this assignment, and then `shutdownCh` is… nil.

Just unconditionally perform the assignment.

async_producer.go
```diff
@@ -1461,51 +1533,318 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
...
+	select {
+	case bp.output <- set:
+		return true
+	case <-done:
+		return false
+	case <-shutdownCh:
```
puellanivis: Actually, no I don’t even know why this is being put into a temp variable.

We can just use `case <-p.shutdownCh` directly.

produce_set_test.go
```diff
@@ -25,6 +25,24 @@ func safeAddMessage(t *testing.T, ps *produceSet, msg *ProducerMessage) {
...
+		{retries: 0},
+	}}).shouldKeepMuted(1) {
+		t.Fatal("expected batch to be retryable when every message is below retry max")
+	}
+	if (&partitionSet{msgs: []*ProducerMessage{
+		{retries: 0},
+		{retries: 1},
+	}}).shouldKeepMuted(1) {
+		t.Fatal("expected batch not to be retryable when any message has exhausted retries")
+	}
```
puellanivis: These are grammatically overly complicated. Instead of trying to shove everything into a single line, we can just assign the partition sets to a temp variable, and operate on that directly:
```go
pset := &partitionSet{}
if !pset.shouldKeepMuted(1) {
	t.Error(…) // none of these test failures preclude testing the other test cases.
}
pset = &partitionSet{
	msg: []*ProducerMessage{
		{retries: 0},
		{retries: 1},
	},
}
if pset.shouldKeyMuted(1) {
	t.Error(…)
}
…
```

async_producer.go
```diff
@@ -103,11 +104,17 @@ type asyncProducer struct {
 	errors                    chan *ProducerError
 	input, successes, retries chan *ProducerMessage
 	inFlight                  sync.WaitGroup
+	// shutdownCh is closed before waiting on in-flight messages so retryBatch
+	// goroutines can stop waiting on broker handoff and release partition mutes.
+	shutdownCh       chan struct{}
+	shutdownChClosed atomic.Bool
```
puellanivis: I’d say we probably want to just group these in their own paragraph. Since it has its own leading comment describing the block, and is aligned separately from the block above it.

async_producer.go
```diff
@@ -118,6 +125,14 @@ type asyncProducer struct {
 	metricsRegistry metrics.Registry
 }
 
+type retryBufferQuota struct {
+	// messages and bytes track producer-level retry buffer occupancy when
+	// Producer.Retry.MaxBufferLength or MaxBufferBytes is bounded.
+	messages int64
+	bytes    int64
+	mu       sync.Mutex
```
puellanivis: Convention is to put mutexes at the top of the paragraph for the variables that it is protecting:
```go
	mu       sync.Mutex
	messages int64
	bytes    int64
```

async_producer.go
```diff
@@ -804,15 +820,51 @@ func (pp *partitionProducer) backoff(retries int) {
 }
 
 func (p *asyncProducer) backoff(retries int) {
+	backoff := p.retryBackoff(retries)
+	if backoff > 0 {
+		time.Sleep(backoff)
+	}
```
puellanivis: This guard is actually unnecessary.

> A negative or zero duration causes Sleep to return immediately.
https://pkg.go.dev/time@go1.26.4#Sleep

async_producer.go
```diff
@@ -804,15 +820,51 @@ func (pp *partitionProducer) backoff(retries int) {
...
+}
+
+func (p *asyncProducer) retryBackoff(retries int) time.Duration {
```
puellanivis: This feels like it should be named either `backoffValue` (as a noun) or `calcBackoff` (as a verb).

async_producer.go
```diff
@@ -804,15 +820,51 @@ func (pp *partitionProducer) backoff(retries int) {
...
 	}
-	if backoff > 0 {
-		time.Sleep(backoff)
+	return backoff
+}
+
+func (p *asyncProducer) shuttingDown() bool {
+	if p.shutdownCh == nil {
+		return false
+	}
```
puellanivis: This guard is unnecessary.

If `shutdownCh == nil` then the ´case <-p.shutdown` would block forever, meaning the default clause is taken, and we return `false` the same as this guard here.

async_producer.go
```diff
@@ -1227,17 +1292,21 @@ func (bp *brokerProducer) tryBuildFlushingBatch() bool {
...
 	}
-	return true
+
+	return false
 }
 
 func (bp *brokerProducer) shutdown() {
+	if bp.done != nil {
+		close(bp.done)
+	}
```
puellanivis, line 1276: This operation is not idempotent, and could potentially cause a close on a closed channel panic. What are the concerns about us potentially calling `shutdown` multiple times?

async_producer.go
```diff
@@ -1461,51 +1533,318 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
...
-	produceSet.bufferCount += len(pSet.msgs)
+	produceSet.addPartitionSet(topic, partition, pSet)
+	bufferMessages, bufferBytes := int64(produceSet.bufferCount), int64(produceSet.bufferBytes)
+	bufferReserved := false
+	releaseBuffer := func() {
+		if bufferReserved {
+			p.releaseBuffer(bufferMessages, bufferBytes)
+			bufferReserved = false
+		}
+	}
```
puellanivis: I believe that this is called in every code path… it may be more appropriate to put this in a `defer` rather than requiring developers to remember to put it into the execution?

async_producer.go
```diff
@@ -1461,51 +1533,318 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
...
+	for _, batch := range batches {
+		p.failMutedBatch(batch, err)
+	}
+}
+
+func (p *asyncProducer) retryBufferLimits() (int64, int64) {
+	maxBufferLength := p.conf.Producer.Retry.MaxBufferLength
+	if 0 < maxBufferLength && maxBufferLength < minFunctionalRetryBufferLength {
+		maxBufferLength = minFunctionalRetryBufferLength
+	}
```
puellanivis: I think this could just be
```go
	maxBufferLength = p.conf.Producer.Retry.MaxBufferLength
	if maxBufferLength > 0 {
		maxBufferLength = max(maxBufferLength, minFunctionalRetryBufferLength)
	}
```

async_producer.go
```diff
@@ -1461,51 +1533,318 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
...
+
+	maxBufferBytes := p.conf.Producer.Retry.MaxBufferBytes
+	if 0 < maxBufferBytes && maxBufferBytes < minFunctionalRetryBufferBytes {
+		maxBufferBytes = minFunctionalRetryBufferBytes
```
puellanivis: Same as above, we should be able to use  `max` here.

async_producer.go
```diff
@@ -1461,51 +1533,318 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
...
+		return true
+	}
+	p.retryBuffer.mu.Lock()
+	defer p.retryBuffer.mu.Unlock()
+
+	if p.bufferWouldOverflow(p.retryBuffer.messages+messages, p.retryBuffer.bytes+bytes) {
+		return false
+	}
+	p.retryBuffer.messages += messages
+	p.retryBuffer.bytes += bytes
```
puellanivis: Maybe precalculate the additions… we do even have them already in local variables (as parameters):
```go
messages += p.retryBuffer.messages
bytes += p.retryBuffer.bytes
if p.bufferWouldOverflow(messages, bytes) {
	return false
}

p.retryBuffer.messages = messages
p.retryBuffer.bytes = bytes
return true
```

async_producer.go
```diff
@@ -1461,51 +1533,318 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
...
+
+func (p *asyncProducer) addToBuffer(messages, bytes int64) bool {
+	if messages == 0 && bytes == 0 {
+		return false
+	}
+	if !p.retryBufferBounded() {
+		return false
+	}
+	p.retryBuffer.mu.Lock()
+	defer p.retryBuffer.mu.Unlock()
```
puellanivis: I would recommend putting the locking here into its own paragraph. (As it’s not logically related to the guard conditions.)

async_producer.go
```diff
@@ -1461,51 +1533,318 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
...
+
+func (p *asyncProducer) reserveBuffer(messages, bytes int64) bool {
+	if messages == 0 && bytes == 0 {
+		return true
+	}
+	if !p.retryBufferBounded() {
+		return true
+	}
+	p.retryBuffer.mu.Lock()
+	defer p.retryBuffer.mu.Unlock()
```
puellanivis: I would recommend putting the locking here into its own paragraph. (As it’s not logically related to the guard conditions.)

async_producer.go
```diff
@@ -1461,51 +1533,318 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
...
+	for _, batch := range retryable {
+		if p.shuttingDown() {
+			batchMessages, batchBytes := int64(len(batch.pSet.msgs)), int64(batch.pSet.bufferBytes)
+			p.releaseBuffer(batchMessages, batchBytes)
+			p.failMutedBatch(batch, ErrShuttingDown)
+			continue
+		}
+		leader, leaderErr := p.client.Leader(batch.topic, batch.partition)
+		if leaderErr != nil {
+			Logger.Printf("Failed retrying batch for %v-%d because of %v while looking up for new leader\n", batch.topic, batch.partition, leaderErr)
```
puellanivis: I’m not seeing any reason why we need to disambiguate `leaderErr` and cannot just shadow `err` here.

async_producer.go
```diff
@@ -1461,51 +1533,318 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
...
+
+// sendRetryBatch transfers ownership of a muted retry batch to the broker
+// bridge. A false result means no owner accepted the batch, so the caller must
+// release the mute and fail the messages instead of leaving a retry goroutine
+// blocked while it still owns the partition mute.
+func (p *asyncProducer) sendRetryBatch(bp *brokerProducer, set *produceSet) bool {
+	var done <-chan struct{}
+	if bp.done != nil {
+		done = bp.done
+	}
```
puellanivis: Same as below, this nil guard is unnecessary. Just unconditionally perform the assignment.

I’m not sure why we don’t just use `bp.done` directly instead of going through a local variable, because there’s no synchronization going on, so we would probably race even if we go through a local variable.

puellanivis (commented):
I know it’s a draft, but here’s some things I spotted already.

3AceShowHand:
Thank you so much for the review, I will fix issues above and test it .

3AceShowHand:
@puellanivis @dnwe 

Can you take a review about this PR

async_producer.go
```diff
@@ -800,19 +812,37 @@ func (p *asyncProducer) newPartitionProducer(topic string, partition int32) chan
...
+
+func (p *asyncProducer) backoffOrDone(retries int) bool {
+	backoff := p.calcBackoff(retries)
+	if backoff <= 0 {
+		return !p.shuttingDown()
+	}
+	timer := time.NewTimer(backoff)
+	defer timer.Stop()
+	select {
+	case <-timer.C:
```
puellanivis: > As of Go 1.23, the garbage collector can recover unreferenced, unstopped timers. There is no reason to prefer NewTimer when After will do.

```suggestion
	select {
	case <-time.After(backoff):
```

async_producer.go
```diff
@@ -1238,6 +1282,12 @@ func (bp *brokerProducer) tryBuildFlushingBatch() bool {
 }
 
 func (bp *brokerProducer) shutdown() {
+	if bp.closed.Swap(true) {
+		return
+	}
```
puellanivis: [nothing to change] Huh… I guess this does work… I’m just used to the `CompareAndSwap(false, true)` which guarantees that we only make the assignment if the previous value is `false`… but yeah, I guess this works just as well.

Although, since below there is a `CompareAndSwap` usage for this, for consistency’s sake, we should probably choose one or the other.

async_producer.go
```diff
@@ -1461,51 +1514,233 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
...
+
+	if !p.backoffOrDone(maxRetryAttempt) {
+		p.retryBufferQuota.release(bufferMessages, bufferBytes)
+		p.failMutedBatches(retryable, ErrShuttingDown)
+		return
+	}
+
+	brokerSets := make(map[*Broker]*produceSet)
+	for _, batch := range retryable {
+		if p.shuttingDown() {
```
puellanivis: Do we want to potentially change behavior in the middle of processing these batches?

async_producer.go
```diff
@@ -1461,51 +1514,233 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
...
+		}
+	}
+}
+
+// handoffRetryBatch transfers ownership of a muted retry batch to the broker
+// bridge. On false, the caller still owns the mute and must fail the batch.
+func (p *asyncProducer) handoffRetryBatch(bp *brokerProducer, set *produceSet) bool {
+	if bp.closed.Load() {
+		return false
+	}
```
puellanivis: This creates a race condition. This `bp.closed.Load` can return `false` just before we do the `bp.closed.Swap` is happening above. And then we ende up doing the `close` and receive at the same time.

It’s probably better to just rely on `<-bp.done` alone receiving below… we don’t really gain anything from performing this atomic test here.

async_producer.go
```diff
@@ -1778,3 +2004,71 @@ func (p *asyncProducer) abandonBrokerConnection(broker *Broker) {
...
+	if !q.shouldTrack(messages, bytes) {
+		return false
+	}
+
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	q.messages += messages
+	q.bytes += bytes
+	return q.wouldOverflow(q.messages, q.bytes)
```
puellanivis: 🤔 This “would” subjunctive is out of place. Here, we’ve made the change, and we care if the values have overflowed, not if they would overflow.

Perhaps the name we want is just `overflows`?

The “would” makes at least some sense in the `tryReserve`, but the non-subjunctive also makes sense. “I’ve calculated these values if they overflow, then return.”

3AceShowHand:
@puellanivis @dnwe 

Can you take a review about this PR, thx.

dnwe:
Thanks for digging into this @3AceShowHand, and for the thorough write-up 

My main hesitation is the size and surface area. This is ~1300 lines and several new moving parts (a dedicated retry worker, per-leader regrouping, a shared retry-buffer quota, plus broker- and producer-level shutdown signalling) to fix what is ultimately a narrow hole in a best-effort guarantee.

I've been experimenting with whether the same guarantee can be reached far more cheaply. The key observation you made above is that the idempotent path already does the right thing on a connection error: it keeps the sent batch muted and retries it directly via retryBatch, instead of dropping it onto retryMessages and releasing the mute. Extending that same branch to the non-idempotent case is only a few lines and reuses a flow we already exercise, with no new worker or regrouping https://github.com/IBM/sarama/pull/3644 and then pull in your shutdown fix as https://github.com/IBM/sarama/pull/3645 – what do you think?

3AceShowHand:
@dnwe 

Agreed，Your solution is more simple and can fix the issue, so let's merge that 2 PRs.

PR #3642, opened 2026-06-24
fix: map new metadata request v8 to kafka 2.3.0

fereidani:
Hi,

I've built a static analyzer for Go called [Ghoul](https://fereidani.com/ghoul), and I ran it against the Sarama codebase. It flagged a duplicate condition in `NewMetadataRequest`, where two consecutive `else if` branches both check `version.IsAtLeast(V2_4_0_0)`, which makes the second branch unreachable.

Ghoul related reports:
```txt
metadata_request.go:34:12: [duplicate-elseif-condition] this condition already appeared earlier in the if/else-if chain, so this branch is unreachable (confidence: 8.0/10, CWE-571)
```

Because the second `V2_4_0_0` check can never be reached, `MetadataRequest` version 8 is never selected. Kafka metadata v8 corresponds to Kafka 2.3.0, so clients configured for `V2_3_0_0` silently fell through to v7. This matches `MetadataRequest.requiredVersion()`, which maps version 8 back to `V2_3_0_0`.

The fix changes the duplicate check to `V2_3_0_0`, restoring the intended mapping. I also added a table test, `TestNewMetadataRequestVersionMapping`, that pins each Kafka version to the metadata request version it selects, so this cannot regress unnoticed.

### DIFF
--- a/metadata_request.go
+++ b/metadata_request.go
@@ -31,7 +31,7 @@ func NewMetadataRequest(version KafkaVersion, topics []string) *MetadataRequest 
 		m.Version = 11
 	} else if version.IsAtLeast(V2_4_0_0) {
 		m.Version = 9
-	} else if version.IsAtLeast(V2_4_0_0) {
+	} else if version.IsAtLeast(V2_3_0_0) {
 		m.Version = 8
 	} else if version.IsAtLeast(V2_1_0_0) {
 		m.Version = 7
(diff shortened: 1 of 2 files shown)


dnwe (approved):
Thanks, good catch! The fix looks good, I tweaked the unittest a little to still catch the original bug, but avoid us duplicating and maintaining the implementation in the test

puellanivis:
I suppose that would be an advantage of the `switch ... case ...` block over an `if ... else if ...` cascade. Go will itself check for duplicate cases, but not for duplicate conditions.

PR #3643, opened 2026-06-24
fix: handle zstd encoder and decoder construction errors

fereidani:
In continuation to #3642, while reviewing the compression path I noticed that the errors returned by `zstd.NewWriter` and `zstd.NewReader` are discarded. Both are `(*T, error)` constructors that return a nil value together with an error, so ignoring the error means a later `enc.EncodeAll` or `dec.DecodeAll` call could dereference a nil pointer and panic in the produce and consume hot path.

This change propagates the construction errors out of `newZstdEncoder`, `getZstdEncoder`, and `getDecoder`, and returns them from `zstdCompress` and `zstdDecompress`. A failed decoder is no longer cached. The signatures of `zstdCompress` and `zstdDecompress` are unchanged, and their callers in `compress.go` and `decompress.go` already handle the returned error, so there is no behaviour change on the success path.

### DIFF
--- a/zstd.go
+++ b/zstd.go
@@ -47,22 +47,21 @@ func getZstdEncoderChannel(params ZstdEncoderParams) chan *zstd.Encoder {
 	return ch
 }
 
-func newZstdEncoder(params ZstdEncoderParams) *zstd.Encoder {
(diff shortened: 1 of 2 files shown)


dnwe (approved):
LGTM, thanks!

PR #3644, opened 2026-06-25
fix: order non-idempotent connection retries

dnwe:
A connection error has no per-partition response to route back through partitionProducer, so the non-idempotent path moved the batch to retryMessages and released its mute, letting a later batch on the same partition flush ahead of the failed one.

Retry directly via retryBatch with the mute held, like the idempotent path, and refresh metadata for both.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -1543,7 +1543,7 @@ func (bp *brokerProducer) handleError(sent *produceSet, err error) {
 			retryTopicSeen[topic] = struct{}{}
 			retryTopics = append(retryTopics, topic)
 		})
-		if bp.parent.conf.Producer.Idempotent && len(retryTopics) > 0 {
+		if len(retryTopics) > 0 {
 			refreshErr := bp.parent.client.RefreshMetadata(retryTopics...)
 			if refreshErr != nil {
 				Logger.Printf("Failed refreshing metadata because of %v\n", refreshErr)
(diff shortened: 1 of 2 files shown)


3AceShowHand:
```
  On the producer side, the ordering issue is in `handleSuccess` when a ProduceResponse contains a retriable partition-level error for a non-idempotent producer.

  Previously, for non-idempotent producers, the failed sent batch was retried via `retryMessages(pSet.msgs, block.Err)`, but the partition was not kept muted. At the end of
  `handleSuccess`, the original sent produceSet was unmuted. This creates a window where a later batch for the same topic/partition can acquire the mute and be sent before the
  original failed batch is retried.

  Example:

  1. The application sends records A then B to the same topic/partition.
  2. A is sent first.
  3. The broker returns a retriable partition error for A, for example `ErrNotLeaderForPartition`.
  4. The producer schedules A for retry, but releases the partition mute.
  5. B is already buffered and now acquires the mute.
  6. B is sent successfully before A is retried.
  7. Metadata refresh finishes, then A is retried after B.

  So the producer can send B before A to the same Kafka partition, even though A was accepted by the producer first.

  This is not specific to `ErrNotLeaderForPartition`; any retriable partition error handled by the same `handleSuccess` branch can expose the same ordering window. The fix is to keep
  the partition muted for response-level retriable errors and retry the failed sent batch as a batch after metadata refresh, instead of releasing the mute and routing the messages
  back through the normal retry queue.
```

@dnwe  I have found a new potential out of order when meet ErrNotLeaderForPartition, Please take a look.

hindessm (approved):
LGTM

PR #3645, opened 2026-06-25
fix: release retry mute on shutdown handoff

dnwe:
retryBatch handed the muted batch off with a blocking send. On shutdown, with nothing reading, it blocked forever and the partition stayed muted, so later messages and Close() hung too.

Close a done channel on shutdown and select on it in the handoff, releasing the mute and failing the batch with ErrShuttingDown.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"sync"
+	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


puellanivis (approved):
Seems like this has some overlap with https://github.com/IBM/sarama/pull/3639 (corrected) I guess that one got closed.

But this one is much more narrow, and easier to comprehend what changes.

PR #3646, opened 2026-06-29
sample: Use SASL/OAUTHBEARER for Event Streams authentication

edoardocomar:
Sample code demonstrating how to use SASL/OAUTHBEARER for authentication with IBM Event Streams.


edoardocomar:
sample log of the reauthentication demo 
[log.txt](https://github.com/user-attachments/files/29502126/log.txt)

It's interesting that on the first reauth the client got a Session expiration in 161338 ms — ~2m41s, not 1 hour. 
Kafka set the session lifetime to min(connections.max.reauth.ms, time until the presented token expires). 
As the client re-authenticated with a token that only had ~1m45s (+60s window) of life left, the broker capped the session to that remaining life.

PR #3656, opened 2026-07-08
fix: correct typos in comments and godoc

maxtaran2010:
## Summary

Fix five typos in Go source comments and godoc strings.

| File | Typo | Correction |
|------|------|------------|
| `broker.go` | `implementors` | `implementers` |
| `client.go` | `superceded` | `superseded` |
| `consumer.go` | `requestor` | `requester` |
| `partitioner.go` | `execpt` | `except` |
| `transaction_manager.go` | `ot` | `to` |

No logic changes — comments and godoc only.

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -101,7 +101,7 @@ type AccessToken struct {
 	Extensions map[string]string
 }
 
-// AccessTokenProvider is the interface that encapsulates how implementors
+// AccessTokenProvider is the interface that encapsulates how implementers
 // can generate access tokens for Kafka broker authentication.
 type AccessTokenProvider interface {
 	// Token returns an access token. The implementation should ensure token
--- a/client.go
+++ b/client.go
@@ -275,7 +275,7 @@ func (client *client) Broker(brokerID int32) (*Broker, error) {
 }
 
 func (client *client) InitProducerID() (*InitProducerIDResponse, error) {
-	// FIXME: this InitProducerID seems to only be called from client_test.go (TestInitProducerIDConnectionRefused) and has been superceded by transaction_manager.go?
+	// FIXME: this InitProducerID seems to only be called from client_test.go (TestInitProducerIDConnectionRefused) and has been superseded by transaction_manager.go?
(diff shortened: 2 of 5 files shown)


consumer.go
```diff
@@ -1311,7 +1311,7 @@ func (bc *brokerConsumer) fetchNewMessages() (*FetchResponse, error) {
 	// partitions are processed in the order they appear in the request.
 	case bc.consumer.conf.Version.IsAtLeast(V0_10_1_0):
 		request.Version = 3
-	// Starting in version 2, the requestor must be able to handle Kafka log
+	// Starting in version 2, the requester must be able to handle Kafka log
```
puellanivis, line 1314: This is considered a valid alternative, but the more common spelling is probably the better idea. (Thus recommend taking the change.)

broker.go
```diff
@@ -101,7 +101,7 @@ type AccessToken struct {
 	Extensions map[string]string
 }
 
-// AccessTokenProvider is the interface that encapsulates how implementors
+// AccessTokenProvider is the interface that encapsulates how implementers
```
puellanivis, line 104: This is considered a valid alternative, but the more common spelling is probably the better idea. (Thus recommend taking the change.)

client.go
```diff
@@ -275,7 +275,7 @@ func (client *client) Broker(brokerID int32) (*Broker, error) {
 }
 
 func (client *client) InitProducerID() (*InitProducerIDResponse, error) {
-	// FIXME: this InitProducerID seems to only be called from client_test.go (TestInitProducerIDConnectionRefused) and has been superceded by transaction_manager.go?
+	// FIXME: this InitProducerID seems to only be called from client_test.go (TestInitProducerIDConnectionRefused) and has been superseded by transaction_manager.go?
```
puellanivis, line 278: The spelling with ⟨c⟩ dates back to Middle French, and thus is a valid variant spelling. However, it is non-standard, and considered by many to be a misspelling, so I recommend taking the change.

puellanivis (approved):

PR #3664, opened 2026-07-13
feat: more flexible protocols for kafka28 parity

dnwe:
In the remaining protocols for Kafka 2.8 parity, this batch only add flexible encoding and no additional fields or behavioural changes so can be grouped as one PR. The create and delete topics add topic uuids so will be done separately


hindessm (approved):
LGTM

PR #3669, opened 2026-07-14
feat: add TransactionClusterAdmin for KIP-664 detection APIs

### LINKED ISSUE #3655: Support KIP-664 transaction-detection admin APIs (DescribeProducers, DescribeTransactions, ListTransactions)
#### Description

Sarama's `ClusterAdmin` currently has no way to inspect transactional state in a cluster. Operators cannot enumerate in-flight transactions, inspect a specific transactional ID's coordinator state, or see the active producers on a topic-partition. This makes it hard to observe transactional activity or diagnose common production problems such as hanging/long-running transactions that hold back the last stable offset (LSO) and block consumers using `read_committed`.

[KIP-664](https://cwiki.apache.org/confluence/display/KAFKA/KIP-664%3A+Provide+tooling+to+detect+and+abort+hanging+transactions) adds three read-only admin protocol APIs to cover exactly this. I'd like to add support for all three to `ClusterAdmin`:

- **DescribeProducers** (API key 61, v0) — active producers per topic-partition, queried per partition leader. Requires broker `2.8.0.0+`.
- **DescribeTransactions** (API key 65, v0) — transaction details for given transactional IDs, queried per transaction coordinator. Requires broker `3.0.0.0+`.
- **ListTransactions** (API key 66, v0 and v1) — enumerate transactional IDs by fanning out to every broker and merging the results; v1 (Kafka `3.6`) adds a duration filter. Requires broker `3.0.0.0+` (v1 requires `3.6.0.0+`).

Proposed signatures:

```go
DescribeProducers(topicPartitions map[string][]int32) (map[string]map[int32]DescribeProducersResponsePartition, error)
DescribeTransactions(transactionalIDs []string) (map[string]DescribeTransactionsResponseState, error)
ListTransactions(stateFilters []string, producerIDFilters []int64, durationFilterMs int64) ([]ListTransactionsResponseState, error)
```

#### Why the current workaround is not enough

Without these APIs, the only way to detect transactional activity with sarama today is a **data-plane tail probe** — no coordinator involvement. For each topic, walk every partition and fetch a bounded slice of the tail (e.g. the last ~100 records) with a `FetchRequest` (v4) under `sarama.ReadCommitted` isolation, then classify the returned block for four signals:

| Signal | Evidence in the fetch block |
| ------ | --------------------------- |
| transactional-record-batch | `RecordBatch.IsTransactional` set on a batch |
| control-batch | `RecordBatch.Control` — a commit/abort marker |
| aborted-transaction | `block.AbortedTransactions` non-empty (broker reports these under `read_committed`) |
| open-transaction | `HighWaterMarkOffset > LastStableOffset` — records exist that a `read_committed` consumer can't see yet |

Any signal on any partition flags the topic as transactional. This works, but it is fundamentally **presence detection**: a positive flag is a definite positive, but a "clean" result only samples the recent tail — it is not whole-log proof of absence. A producer that wrote transactionally millions of records ago and then went quiet won't show up, and it tells you nothing about *which* transactional IDs / producers are involved or whether a transaction is currently hanging. The KIP-664 APIs answer those questions authoritatively at the coordinator, which is what this feature is really after.

#### Backwards compatibility

There are two ways to expose these three methods, and I'd like maintainer input on which to take. The functional implementation (per-leader / per-coordinator fan-out, retry-on-stale-metadata, partial-result semantics) is identical in both — the only difference is *where* the methods live in the type system.

##### Option A — grow the `ClusterAdmin` interface

Add the three methods directly to the existing `ClusterAdmin` interface:

```go
type ClusterAdmin interface {
    // ... existing methods ...

    DescribeProducers(topicPartitions map[string][]int32) (map[string]map[int32]DescribeProducersResponsePartition, error)
    DescribeTransactions(transactionalIDs []string) (map[string]DescribeTransactionsResponseState, error)
    ListTransactions(stateFilters []string, producerIDFilters []int64, durationFilterMs int64) ([]ListTransactionsResponseState, error)

    Close() error
}
```

Callers use them directly on the value returned by `NewClusterAdmin` / `NewClusterAdminFromClient`:

```go
admin, err := sarama.NewClusterAdmin(addrs, conf)
txns, err := admin.ListTransactions(nil, nil, -1)
```

- **Pro:** most discoverable and ergonomic — no type assertion, the methods sit alongside every other admin operation, and it matches how `ClusterAdmin` has grown before (e.g. KIP-394 added `RemoveMemberFromConsumerGroup`).
- **Con:** source-compatible for *callers*, but a **breaking change for anyone who implements the interface directly** (custom mocks/wrappers, generated gomock/mockery/counterfeiter fakes) rather than embedding it — they must add the three methods before their code compiles again. Would need labelling as a breaking change in the release notes.

##### Option B — separate optional `TransactionAdmin` interface (non-breaking)

Leave `ClusterAdmin` untouched and put the new methods on a separate interface that the concrete `*clusterAdmin` also satisfies. Callers reach it via a type assertion — the same idiom the stdlib uses for `http.Flusher` / `http.Hijacker` on `http.ResponseWriter`:

```go
// TransactionAdmin exposes the KIP-664 transaction-detection APIs. The value
// returned by NewClusterAdmin / NewClusterAdminFromClient implements it; it is
// kept separate from ClusterAdmin so that adding these methods does not break
// external types that implement ClusterAdmin directly.
type TransactionAdmin interface {
    DescribeProducers(topicPartitions map[string][]int32) (map[string]map[int32]DescribeProducersResponsePartition, error)
    DescribeTransactions(transactionalIDs []string) (map[string]DescribeTransactionsResponseState, error)
    ListTransactions(stateFilters []string, producerIDFilters []int64, durationFilterMs int64) ([]ListTransactionsResponseState, error)
}

var _ TransactionAdmin = (*clusterAdmin)(nil)
```

```go
admin, err := sarama.NewClusterAdmin(addrs, conf)
if txAdmin, ok := admin.(sarama.TransactionAdmin); ok {
    txns, err := txAdmin.ListTransactions(nil, nil, -1)
}
```

(The assertion works even though `clusterAdmin` is unexported — asserting to an interface doesn't depend on the concrete type's visibility.)

- **Pro:** fully backwards compatible — no existing implementer of `ClusterAdmin` breaks. Also forward-compatible: the methods can be promoted into `ClusterAdmin` proper at the next major/`v2`, where the break is expected.
- **Con:** slightly less discoverable and requires a type assertion at the call site.

##### Summary

| | Option A: grow `ClusterAdmin` | Option B: `TransactionAdmin` interface |
| --- | --- | --- |
| Breaks direct implementers | **Yes** (release-note label) | No |
| Call-site ergonomics | Direct method call | Type assertion first |
| Discoverability | High (on main interface) | Lower (separate type) |
| Precedent in sarama | Yes (KIP-394 etc.) | New pattern for this repo |
| Path forward | Already the target shape | Promote to `ClusterAdmin` at v2 |

Per CONTRIBUTING ("Don't break backwards compatibility... open an issue to discuss this first"), I'd default to **Option B** to keep the release non-breaking, but I'm happy to go with **Option A** if maintainers consider interface growth acceptable here and prefer the ergonomics — flagging for your call on which to merge and how to label it.

#### Additional context

- KIP: [KIP-664: Provide tooling to detect and abort hanging transactions](https://cwiki.apache.org/confluence/display/KAFKA/KIP-664%3A+Provide+tooling+to+detect+and+abort+hanging+transactions)

- Existing implementations in other Kafka clients — a goal here is **parity with [franz-go](https://github.com/twmb/franz-go)**, whose [`kadm`](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kadm) admin package already exposes all three (verified against the current package docs):

  ```go
  func (cl *Client) DescribeProducers(ctx context.Context, s TopicsSet) (DescribedProducersTopics, error)
  func (cl *Client) DescribeTransactions(ctx context.Context, txnIDs ...string) (DescribedTransactions, error)
  func (cl *Client) ListTransactions(ctx context.Context, producerIDs []int64, filterStates []string) (ListedTransactions, error)
  ```

  (kadm additionally offers a `ListTransactionsByTxPattern` variant.)

- Scope note: this issue covers only the read-only *detection* half of KIP-664. The write side (aborting hanging transactions, i.e. an admin over the WriteTxnMarkers / producer-fencing path) is intentionally out of scope and could be a follow-up.

AI Disclaimer: This issue was written with the help of Claude.

adrian-januzi:
Implement the admin-client portion of https://github.com/IBM/sarama/issues/3655: a `TransactionClusterAdmin` expansion interface embedding `ClusterAdmin`, with three read-only detection methods on `*clusterAdmin`.

- `DescribeProducers` lists active producers per topic-partition, fanning out to each partition leader (Kafka 2.8.0.0+)
- `DescribeTransactions` returns transaction state for the given transactional IDs, fanning out to each coordinator and retrying on coordinator-moved/loading errors (Kafka 3.0.0.0+)
- `ListTransactions` lists transactions from all brokers in parallel and merges the results, with an optional duration filter on Kafka 3.8.0.0+ (request v1) (Kafka 3.0.0.0+)
- add mock response helpers and unit tests covering happy paths, multi-broker fan-out, empty input, coordinator-move retries and partial failures

---

The interface is kept separate from the `ClusterAdmin` to avoid breaking changes for anything currently implementing `ClusterAdmin` as discussed in the initial issue. 

The decision was taken to follow a similar pattern for `admin_offsets.go` in creating a `admin_transactions.go` to make the code itself self-contained and avoid growing the `admin.go` file. However, tests were kept in `admin_tests.go`.

All three methods return partial results, following the same convention as `admin_offsets.go`: per-item error codes are carried through in the returned structs and RPC-level failures are aggregated with `errors.Join`.

---

KIP-664: https://cwiki.apache.org/confluence/display/KAFKA/KIP-664%3A+Provide+tooling+to+detect+and+abort+hanging+transactions

Closes: [#3655](https://github.com/IBM/sarama/issues/3655)


admin_transactions.go
```diff
@@ -0,0 +1,227 @@
...
+	"sync"
+)
+
+// TransactionClusterAdmin extends ClusterAdmin with the read-only KIP-664
+// transaction-detection APIs (DescribeProducers, DescribeTransactions and
+// ListTransactions). The value returned by NewClusterAdmin and
+// NewClusterAdminFromClient implements it. It is kept separate from ClusterAdmin
+// so that adding these methods does not break external types that implement
+// ClusterAdmin directly; callers reach the new methods with a single type
+// assertion and can then use the result in place of ClusterAdmin.
```
puellanivis: > It is kept separate from ClusterAdmin so that adding these methods does not break external types that implement ClusterAdmin directly; callers reach the new methods with a single type assertion and can then use the result in place of ClusterAdmin.

This sort of thing does not need to be explained. This is all already the inherent purpose of extension interfaces.

admin_transactions.go
```diff
@@ -0,0 +1,227 @@
...
+	ListTransactions(stateFilters []string, producerIDFilters []int64, durationFilterMs int64) ([]ListTransactionsResponseTransactionState, error)
+}
+
+var _ TransactionClusterAdmin = (*clusterAdmin)(nil)
+
+func (ca *clusterAdmin) DescribeProducers(topicPartitions map[string][]int32) (map[string]map[int32]DescribeProducersResponsePartition, error) {
+	result := make(map[string]map[int32]DescribeProducersResponsePartition)
+	if len(topicPartitions) == 0 {
+		return result, nil
+	}
```
puellanivis, line 40: Returning a `nil` map should be an allowed, and expected operation, especially when the results can be consider to be read-only.

That is, do `results := make(…)` as late as possible, creating it only when it ends up being necessary to exist so that it can be filled in. (This would be just before line 64.)

(This is mostly because it is not an error to access a `nil` map. It operates identically to an empty map, except that it cannot be assigned to. Thus, returning a nil map is generally preferred over returning an empty map.)

admin_transactions.go
```diff
@@ -0,0 +1,227 @@
...
+			leader, err := ca.client.Leader(topic, partition)
+			if err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			partitionsPerBroker[leader] = append(partitionsPerBroker[leader], topicPartition{topic, partition})
+		}
+	}
+
+	for broker, tps := range partitionsPerBroker {
```
puellanivis: `tps` is not really a clear enough abbreviation… `topicPartitions` or `topicParts`?

admin_transactions.go
```diff
@@ -0,0 +1,227 @@
...
+// isRetriableTransactionCoordinatorError reports whether the given error is a
+// transaction-coordinator error that refreshing the coordinator and retrying can
+// resolve. It mirrors the set the transaction manager treats as retriable
+// (COORDINATOR_NOT_AVAILABLE, NOT_COORDINATOR, COORDINATOR_LOAD_IN_PROGRESS) plus
+// EOF for a dropped connection.
+func isRetriableTransactionCoordinatorError(err error) bool {
+	return errors.Is(err, ErrConsumerCoordinatorNotAvailable) ||
+		errors.Is(err, ErrNotCoordinatorForConsumer) ||
+		errors.Is(err, ErrOffsetsLoadInProgress) ||
+		errors.Is(err, io.EOF)
```
puellanivis: A `switch` statement can often be more legible than a long boolean expression.

admin_transactions.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+func (ca *clusterAdmin) ListTransactions(stateFilters []string, producerIDFilters []int64, durationFilterMs int64) ([]ListTransactionsResponseTransactionState, error) {
+	// Transactions may be listed by any broker, so query all brokers in parallel
+	// and merge the results.
+	brokers := ca.client.Brokers()
+	results := make(chan []ListTransactionsResponseTransactionState, len(brokers))
+	errChan := make(chan error, len(brokers))
+	wg := sync.WaitGroup{}
```
puellanivis: [style] Declared variables do not need to be explicitly assigned a zero value. They can simply be declared.

```suggestion
	var wg sync.WaitGroup
```

admin_transactions.go
```diff
@@ -0,0 +1,227 @@
...
+
+	for _, b := range brokers {
+		wg.Add(1)
+		go func(b *Broker) {
+			defer wg.Done()
```
puellanivis: We are using [go1.25](https://github.com/IBM/sarama/blob/main/go.mod#L3), use `wg.Go` instead. We also already have access to `b` through closure, and now that `loopvar` is the standard and not an experiment, we do not have to even worry about doing a `b := b` to get a loop-local variable.

```suggestion
		wg.Go(func() {
```

admin_transactions.go
```diff
@@ -0,0 +1,227 @@
...
+			if err != nil {
+				errChan <- err
+				return
+			}
+			if !errors.Is(response.ErrorCode, ErrNoError) {
+				errChan <- response.ErrorCode
+				return
+			}
+
+			results <- response.TransactionStates
```
puellanivis: Prefer sending all results through a single channel, that has a `func`-scoped type that is `struct { result ListTransactionsResponseTransactionState, err error }`.

Then you can have just the single `for … range chan` collector, rather than needing two separate channels collected separately.

This also means you don’t have to use a buffered channel to store all the results until they’re complete. You can then collect results in parallel to a `go func() { wg.Wait(); close(results) }` that will handle the channel closure for us, and then the collecting `for … range chan` collector ends per norrmal.

admin_transactions.go
adrian-januzi: That makes sense, thanks.

Just a question on placement preference - I see now and agree that we could place it just before `for broker, tps := range partitionsPerBroker {` loop. However, we could still end up with an empty map if the request was to error.

Instead, I was thinking of moving it into the `for _, partition := range topic.Partitions {` loop instead, something like:
```go
if result == nil {
        result = make(map[string]map[int32]DescribeProducersResponsePartition)
}
```

admin_transactions.go
adrian-januzi: I chose to go for:
```go
func isRetriableTransactionCoordinatorError(err error) bool {
	switch {
	case errors.Is(err, ErrConsumerCoordinatorNotAvailable):
		return true
	case errors.Is(err, ErrNotCoordinatorForConsumer):
		return true
	case errors.Is(err, ErrOffsetsLoadInProgress):
		return true
	case errors.Is(err, io.EOF):
		return true
	default:
		return false
	}
}
```

instead of:
```go
func isRetriableTransactionCoordinatorError(err error) bool {
	switch {
	case errors.Is(err, ErrConsumerCoordinatorNotAvailable),
		errors.Is(err, ErrNotCoordinatorForConsumer),
		errors.Is(err, ErrOffsetsLoadInProgress),
		errors.Is(err, io.EOF):
		return true
	default:
		return false
	}
}
```

as I didn't think we gained much readability from the second option compared to what was there previously.

admin_transactions.go
puellanivis: Putting the allocation into the loop is indeed an acceptable answer.

It’s just usually easier to unconditionally create the result map, if you’re sure at least one element will be assign.

admin_transactions.go
```diff
@@ -36,7 +32,7 @@ type TransactionClusterAdmin interface {
 var _ TransactionClusterAdmin = (*clusterAdmin)(nil)
 
 func (ca *clusterAdmin) DescribeProducers(topicPartitions map[string][]int32) (map[string]map[int32]DescribeProducersResponsePartition, error) {
-	result := make(map[string]map[int32]DescribeProducersResponsePartition)
+	var result map[string]map[int32]DescribeProducersResponsePartition
 	if len(topicPartitions) == 0 {
 		return result, nil
```
puellanivis: It is clearer to use `return nil, nil` since that is the only value `result` can have at this point. (That is, if we can make a stronger statement about what the value being returned is than a variable, then we should use that stronger statement.)

That also means that the declaration of the variable should still be moved to just above the for loop on line 60.

admin_transactions.go
```diff
@@ -95,7 +94,7 @@ func (ca *clusterAdmin) DescribeProducers(topicPartitions map[string][]int32) (m
 }
 
 func (ca *clusterAdmin) DescribeTransactions(transactionalIDs []string) (map[string]TransactionState, error) {
-	result := make(map[string]TransactionState)
+	var result map[string]TransactionState
 	if len(transactionalIDs) == 0 {
 		return result, nil
```
puellanivis: Same as above, move the `result` definition to as late a point as possible, and use a more direct `return nil, nil`.

puellanivis (commented):
Just some style lint now.

admin_transactions.go
adrian-januzi: > (That is, if we can make a stronger statement about what the value being returned is than a variable, then we should use that stronger statement.)

That's a fair point. To me both read much the same but I get your meaning and agree.

puellanivis (approved):
Code looks good now. :)

Up to the rest of the contributors for design fit.

adrian-januzi:
Thanks for all your help @puellanivis!

admin_transactions.go
```diff
@@ -0,0 +1,260 @@
...
+			err := ca.retryOnError(isRetriableTransactionCoordinatorError, func() (err error) {
+				defer func() {
+					if err != nil && isRetriableTransactionCoordinatorError(err) {
+						for _, id := range ids {
+							_ = ca.client.RefreshTransactionCoordinator(id)
+						}
+					}
+				}()
+
+				coordinator, err := ca.client.TransactionCoordinator(ids[0])
```
dnwe: On the first attempt we group the transaction IDs by coordinator to send the requests to the correct one, but on a retry it seems like we just hardcode the first coordinator and send all of the requests to that one, which seems like it can never succeed as we'll just get NOT_COORDINATOR for those?

Looking at the Java admin code equivalent it re-groups per retry

admin_transactions.go
```diff
@@ -0,0 +1,260 @@
...
+			request := &ListTransactionsRequest{
+				StateFilters:      stateFilters,
+				ProducerIDFilters: producerIDFilters,
+				DurationFilter:    -1,
+			}
+			if ca.conf.Version.IsAtLeast(V3_8_0_0) {
+				// Version 1 adds the DurationFilter field.
+				request.Version = 1
+				request.DurationFilter = durationFilterMs
+			}
```
dnwe: If the user gave us a durationFilterMs parameter but ca.conf.Version is not at least V3_8_0_0 we just silently ignore it - we probably want to return an unsupported err instead? Although the sendAndReceive downgrade due to ApiVersionsResponse happens outside of this, so not sure where we'd handle that one

admin_transactions.go
```diff
@@ -0,0 +1,260 @@
...
+				Name:             topic,
+				PartitionIndexes: partitions,
+			})
+		}
+
+		response, err := broker.DescribeProducers(request)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
```
dnwe: Minor, but I noticed that DescribeProducers we seem to submit the requests serially, but those other transaction describer funcs use a WaitGroup to do them in parallel - was that an explicit choice or just overlooked?

admin_test.go
```diff
@@ -2676,3 +2676,556 @@ func TestClusterAdminUpdateFeatures(t *testing.T) {
 		require.Equal(t, []UpdatableFeatureResult{{Feature: "metadata.version"}}, results)
 	})
 }
+
+func TestClusterAdminDescribeProducers(t *testing.T) {
```
dnwe: minor, but as we spun the transactions admin code off into its own admin_transactions.go then maybe we should have these tests in admin_transactions_test.go too?

dnwe (commented):
Thanks for working on this, a few comments below

admin_test.go
adrian-januzi: I thought about that too but the `admin_offsets.go` tests were thrown in `admin_test.go` so I just went with the existing precedent. Though personally I do agree that I like the separation too. You're call :)

admin_transactions.go
adrian-januzi: Yea you're right, I missed that a retry only re-resolves `ids[0]`'s coordinator, sending the whole group there. I'll look into the Java implementation and see how it can be replicated here.

admin_test.go
dnwe: Ah good spot, yes we should probably move those too now that admin_test.go is a bit bigger

admin_transactions.go
puellanivis: 😅 This is why we need the design-fit code review, too. Things I tend to miss.

admin_transactions.go
adrian-januzi: What do you think about the following approach - A version + `durationFilterMs` filter like below to catch the version/duration set mismatch early:
```golang
durationFilterMs >= 0 && Version < V3_8_0_0
```

Then for the `sendAndReceive` downgrade you mentioned, we could add a second guard around `ListTransactionsRequest.encode()` modelling similar to what [offset_fetch_requests.go](https://github.com/IBM/sarama/blob/main/offset_fetch_request.go#L77-L79) does? Which if I'm understanding the process correctly will ensure `restrictApiVersion` runs just before `encode`, `r.Version` there already reflects the negotiated version.

dnwe: Yep that’s exactly right 👌🏻

admin_test.go
adrian-januzi: I've created a `admin_transactions_test.go` file for these tests to begin the separation.

admin_transactions.go
adrian-januzi: Thanks for this and the crux of the change is [here](https://github.com/adrian-januzi/sarama/blob/519d0899f75ea986e410dd6e9668e33ce7d3a2aa/admin_transactions.go#L156-L163) so that we are no longer stuck asking the old `ids[0]`'s coordinator for the entire batch.

dnwe:
@adrian-januzi one general question - I haven’t seen what sort of error response detail we get from Kafka here, but at the moment we just append the error we got back as-is with no wrapping. Are we confident that they include enough context information around broker/topic/partition or do we need to consider wrapping them with that?

e.g.,

```go
errs = append(errs, fmt.Errorf(
    "describe producers for %s/%d: %w",
    topic,
    partition,
    err,
))
```

adrian-januzi:
@dnwe the errors look like this when I tested it against one of my Kafka clusters:

- Here at least we can see the topic name that it is referring to that doesn't exist on the broker but not fantastic.
```shell
DescribeProducers, unknown topic:
  this-topic-does-not-exist: kafka server: Request was for a topic or partition that does not exist on this broker
```

- This one is a basic error that really isn't helpful at all.
```shell
DescribeProducers, test-topic-4 partition 9999:
  kafka server: Request was for a topic or partition that does not exist on this broker
```

After wrapping, the `DescribeProducers` call for example looks like this:
```shell
DescribeProducers, unknown topic:
  describe producers: find leader for partition 0 of topic this-topic-does-not-exist: this-topic-does-not-exist: kafka server: Request was for a topic or partition that does not exist on this broker

DescribeProducers, test-topic-3 partition 9999:
  describe producers: find leader for partition 9999 of topic test-topic-3: kafka server: Request was for a topic or partition that does not exist on this broker
```

Let me know if that's what you are expecting.

dnwe:
Yeah so something like that would probably be handy because otherwise you have to determine which of the TPs failed based on which weren’t in the results map, so including in the err is worthwhile

adrian-januzi:
Added a fix for the failing tests that led to`DurationFilter` being interpreted as 0ms by `Encode` which was a v1 API only field instead of the correct 'no filter' request. Fix adds a constructor that defaults `DurationFilter` to -1 (no filter), which is legal to encode at any version. This is also how the Java client handles this.

admin_transactions.go
```diff
@@ -0,0 +1,294 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+	"sync"
+)
+
+// TransactionClusterAdmin extends ClusterAdmin with the read-only KIP-664
+// transaction-detection APIs (DescribeProducers, DescribeTransactions and
+// ListTransactions). The value returned by NewClusterAdmin and
```
dnwe: One small nit, this comment finishes mid sentence?
 
You probably meant something like:

> // TransactionClusterAdmin extends ClusterAdmin with the read-only KIP-664
// transaction-detection APIs: DescribeProducers, DescribeTransactions, and
// ListTransactions. Values returned by NewClusterAdmin and
// NewClusterAdminFromClient implement TransactionClusterAdmin.

admin_transactions.go
```diff
@@ -0,0 +1,294 @@
...
+		states []TransactionState
+		err    error
+	}
+	results := make(chan queryResult)
+
+	var wg sync.WaitGroup
+	for _, ids := range idsPerCoordinator {
+		wg.Go(func() {
+			var states []TransactionState
+			err := ca.retryOnError(isRetriableTransactionCoordinatorError, func() (err error) {
```
dnwe, line 150: It’s not blocking for this PR but might be something we want to think about in a follow-up. Currently if one describe fails all of its retries then we return the errs and discard the successful results we had collected from the other coordinators. Perhaps we’d want to return partial results in that case. Something to consider

dnwe (commented):
Looks ready now. One small nit on a code comment that seems to be cut off and a future thought for follow-up consideration of the “partial success” case

admin_transactions.go
adrian-januzi: Yup, found the old comment in the commit history - no idea how I managed that.

>   // TransactionClusterAdmin extends ClusterAdmin with the read-only KIP-664
  // transaction-detection APIs (DescribeProducers, DescribeTransactions and
  // ListTransactions). The value returned by NewClusterAdmin and
  // NewClusterAdminFromClient implements it. It is kept separate from ClusterAdmin
  // so that adding these methods does not break external types that implement
  // ClusterAdmin directly; callers reach the new methods with a single type
  // assertion and can then use the result in place of ClusterAdmin.

Gonna go with yours as it is much less wordy and straight to the point.

admin_transactions.go
adrian-januzi: We already are returning partial results as well as the joined errors further down [here](https://github.com/adrian-januzi/sarama/blob/feat/transaction-cluster-admin/admin_transactions.go#L201-L215) and a [test too](https://github.com/adrian-januzi/sarama/blob/feat/transaction-cluster-admin/admin_transactions_test.go#L413-L445).

I've went and added another test closer to the situation you described where two transactional ids sit on separate coordinators and one coordinator returns a retriable code on every attempt, exhausting all its retries, while the other answers cleanly and the end result is a partial successful result + errors.

dnwe: Ah that’s right when split across multiple goroutines, it’s the regrouping that can then leave an edge case I think. So if the initial coordinator lookup groups multiple transaction IDs into a single goroutine, but a retry (which correctly looks up coordinators again) splits them across more than one coordinator. Then if one coordinator succeeds and the other returns a retriable error, the successful results are lost because they’re only kept for the current retry attempt in that loop I think. 

Let’s not worry about it and get this merged though as we've covered the realistic scenarios

dnwe (approved):

admin_transactions.go
adrian-januzi: Ah I think I get your meaning now. A better test would have been to have a single goroutine split going to two different coordinators with one returning a result and the other returning an error.

Thanks for your patience and merging the PR, happy Friday!

PR #3670, opened 2026-07-15
feat(consumer): add rebalance protocol plumbing

dnwe:
Select the most preferred protocol shared by all configured balance strategies. Strategies that do not declare protocol support remain eager-only.

Include owned partitions and the assignment generation in subscription metadata when supported by the configured Kafka version. Assignment behaviour remains unchanged.

### DIFF
--- a/rebalance_protocol.go
+++ b/rebalance_protocol.go
@@ -1,0 +1,84 @@
+package sarama
+
+import "fmt"
+
+// RebalanceProtocol identifies a rebalance protocol supported by a BalanceStrategy
+//
+// The zero value is RebalanceProtocolEager, so a strategy that does not declare
+// its supported protocols is treated as eager-only.
+type RebalanceProtocol int8
+
+const (
+	// RebalanceProtocolEager revokes every assignment before a group rejoins
+	RebalanceProtocolEager RebalanceProtocol = iota
+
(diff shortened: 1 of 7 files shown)


rebalance_protocol.go
```diff
@@ -0,0 +1,86 @@
...
+			}
+		}
+	}
+
+	if _, ok := common[RebalanceProtocolCooperative]; ok {
+		return RebalanceProtocolCooperative, nil
+	}
+	if _, ok := common[RebalanceProtocolEager]; ok {
+		return RebalanceProtocolEager, nil
+	}
```
puellanivis: Hm… my first thought is that this is probably more complicated than it needs to be for just two strategies… however, I guess that depends on how likely there are to be more strategies?

We could maybe YAGNI and write a simpler version that just does direct checks? And then reconsider producing a wider more generic design later if we really get more strategies involved?

dnwe: Yeah I did worry about this - agreed will do a minimal implementation

dnwe: Pushed with something simpler

puellanivis: Looks much simpler to reason about. 👍

puellanivis (approved):

PR #3671, opened 2026-07-15
feat(consumer): add cooperative sticky assignor

dnwe:
Reuse the sticky planner with ownership from subscription metadata. Partitions moving between members are withheld until their previous owner has revoked them.

Advertise both eager and cooperative protocols so existing groups can migrate with two rolling deployments.

### DIFF
--- a/balance_strategy_cooperative_sticky.go
+++ b/balance_strategy_cooperative_sticky.go
@@ -89,11 +89,9 @@ func adjustCooperativeAssignment(plan BalanceStrategyPlan, members map[string]Co
 			continue
 		}
 		topics := plan[newOwner]
-		partitions := slices.DeleteFunc(topics[tp.Topic], func(p int32) bool { return p == tp.Partition })
-		if len(partitions) == 0 {
+		topics[tp.Topic] = slices.DeleteFunc(topics[tp.Topic], func(p int32) bool { return p == tp.Partition })
+		if len(topics[tp.Topic]) == 0 {
 			delete(topics, tp.Topic)
-			continue
 		}
-		topics[tp.Topic] = partitions
 	}
 }
--- a/balance_strategy_cooperative_sticky.go
+++ b/balance_strategy_cooperative_sticky.go
@@ -1,0 +1,99 @@
+package sarama
+
+import (
+	"slices"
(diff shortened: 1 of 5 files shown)


balance_strategy_cooperative_sticky.go
```diff
@@ -0,0 +1,99 @@
...
+		if _, transferring := revoked[tp]; !transferring {
+			continue
+		}
+		topics := plan[newOwner]
+		partitions := slices.DeleteFunc(topics[tp.Topic], func(p int32) bool { return p == tp.Partition })
+		if len(partitions) == 0 {
+			delete(topics, tp.Topic)
+			continue
+		}
+		topics[tp.Topic] = partitions
```
puellanivis: 🤔 I think we might want to just stomp the `slices.Delete` straight back into the map without going through a temporary variable. This is because the deletion is happening in-place, so there’s no point at which this `partition` local variable is not also aliasing the `topics[tp.Topic]`, so it’s kind of technically invalid to be keeping the old slice around at all.
```suggestion
		topics[tp.Topic] = slices.DeleteFunc(topics[tp.Topic], func(p int32) bool { return p == tp.Partition })
		if len(topics[tp.Topic]) == 0 {
			continue
		}
```

A smart enough compiler should be able to elide the lookup for the `len` call anways.

puellanivis (commented):
Narrow technical nitpick. 👍 looks good otherwise.

balance_strategy_cooperative_sticky.go
dnwe: Sure, we can avoid the temporary var, but we still need the delete(…) rather than a bare continue so I can’t use the suggestion as-is 😅

dnwe:
@puellanivis please can you re-review whenever you’re ready?

balance_strategy_cooperative_sticky.go
puellanivis: Yes, yes, still need the delete. I believe an earlier edit had an explanation that the delete still needs to be there. 🤦‍♀️ Somehow, it fell out.

Good catch on it though! (Must have been late in the day.)

puellanivis (approved):

Issue #2820, opened 2024-02-29
`Net.ReadTimeout` is not a network-level property


leonidboykov:
##### Description

The description of the `Net` struct says that inner properties correspond to network-level configuration. However, `Net.ReadTimeout` is affected by some other timeouts, i.e. if `Consumer.Group.Rebalance.Timeout` is greater than `Net.ReadTimeout` the rebalance fails with `i/o timeout` error. Perhaps, Kafka does not reply immediately to the client.

At the moment, the only way to make other timeouts work is to set `Net.ReadTimeout` as the longest timeout. However, this may lead to problems with actual network issues because there is no actual network-level timeout.

##### Versions

| Sarama  | Kafka | Go   |
|---------|-------|------|
| v1.41.3 | 3.1.0 | 1.21 |

##### Configuration

``` go
conf := sarama.NewConfig()
conf.Version = sarama.V3_3_1_0
conf.Consumer.Return.Errors = true
conf.Consumer.Offsets.Initial = sarama.OffsetOldest
conf.Consumer.MaxProcessingTime = 1 * time.Minute
conf.Consumer.Group.Rebalance.Timeout = 10 * time.Minute
conf.Consumer.Group.Session.Timeout = 10 * time.Minute
conf.Consumer.Group.Heartbeat.Interval = 5 * time.Second
conf.Net.ReadTimeout = 11 * time.Minute // Consumer.Group.Rebalance.Timeout with an extra 1 minute.
}
```

##### Additional Context

I've done some research, and it seems, other Sarama users are dealing with the same problem: https://github.com/IBM/sarama/issues/1422#issuecomment-517530147.

I'm not sure, but perhaps, this parameter wasn't tested properly. While default settings for `sarama.Config` set `Consumer.Group.Rebalance.Timeout` (60s) greater than `Net.ReadTimeout` (30s), the value `Consumer.Group.Rebalance.Timeout` is set to only 10 seconds in the corresponding test:

https://github.com/IBM/sarama/blob/3e385a677e5b0aaacc8ba0a56be18a53550275ca/functional_consumer_group_test.go#L396-L405

leonidboykov:
I'm pretty sure, the bug still exists in the main branch.

leonidboykov:
The bug is still present in the master branch.

hugoev:
I'd like to take a look at this — I'll open a PR with a fix.

PR #3672, opened 2026-07-15
fix: anchor Net.ReadTimeout deadline to request send time

hugoev:
## What

`Net.ReadTimeout` was supposed to be a per-request network deadline, but `readFull` called `SetReadDeadline(time.Now().Add(ReadTimeout))` on every call. In `responseReceiver` there are two sequential reads per response (header, then body). Each call reset the clock, so the effective budget was `ReadTimeout` per read — not per response. A broker that dribbled bytes just within each window could stall indefinitely.

## Why it matters

This is the root cause of the pattern documented in #2820: users must set `Net.ReadTimeout` to a value larger than `Consumer.Group.Rebalance.Timeout` to prevent spurious `i/o timeout` errors during rebalances. The timeout should only fire when the network is actually stuck.

## How

Compute the deadline once per request as `requestTime.Add(Net.ReadTimeout)` and pass it to `readFull` as an argument. The deadline now covers the entire read of a response. The same fix is applied to the ApiVersions negotiation, SASL handshake, SASL plain v0, and SCRAM paths which all have a local `requestTime` already.

The `readFull` signature changes from:

```go
func (b *Broker) readFull(buf []byte) (n int, err error)
```

to:

```go
func (b *Broker) readFull(buf []byte, deadline time.Time) (n int, err error)
```

`readFull` is unexported so this is not a public API change.

The `Net.ReadTimeout` doc comment is updated to explain that the value must be at least as large as the longest expected server-side operation.

## Tests

Two new unit tests use a `dripDialer` (a `proxy.Dialer` backed by `net.Pipe`) that writes a valid framed response in two chunks with a configurable pause between them:

- `TestReadTimeoutDeadlineSetOncePerResponse`: pause is within budget → request succeeds
- `TestReadTimeoutExpiresAcrossChunks`: pause exceeds budget → request fails with `net.Error` timeout

Both pass with `-race`.

Relates to #2820

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -1084,10 +1084,16 @@ func (b *Broker) AlterClientQuotas(request *AlterClientQuotasRequest) (*AlterCli
 	return response, nil
 }
 
-// readFull ensures the conn ReadDeadline has been setup before making a
-// call to io.ReadFull
-func (b *Broker) readFull(buf []byte) (n int, err error) {
-	if err := b.conn.SetReadDeadline(time.Now().Add(b.conf.Net.ReadTimeout)); err != nil {
+// readFull sets the read deadline to deadline and then calls io.ReadFull.
+//
+// The deadline is an absolute time.Time computed by the caller as
+// requestTime.Add(Net.ReadTimeout). Anchoring the deadline to the moment
+// the request was sent means a single deadline covers both the header and
+// body reads of one response; resetting it on every call would allow a
+// slow broker to stall indefinitely by returning bytes just before each
+// per-read timeout fires.
+func (b *Broker) readFull(buf []byte, deadline time.Time) (n int, err error) {
+	if err := b.conn.SetReadDeadline(deadline); err != nil {
 		return 0, err
 	}
 
@@ -1337,7 +1343,8 @@ func (b *Broker) responseReceiver() {
 		headerLength := getHeaderLength(promise.response.headerVersion())
 		header := make([]byte, headerLength)
 
-		bytesReadHeader, err := b.readFull(header)
+		deadline := promise.requestTime.Add(b.conf.Net.ReadTimeout)
+		bytesReadHeader, err := b.readFull(header, deadline)
 		requestLatency := time.Since(promise.requestTime)
 		if err != nil {
 			b.updateIncomingCommunicationMetrics(bytesReadHeader, requestLatency)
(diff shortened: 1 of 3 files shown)


broker.go
```diff
@@ -1337,7 +1343,8 @@ func (b *Broker) responseReceiver() {
 		headerLength := getHeaderLength(promise.response.headerVersion())
 		header := make([]byte, headerLength)
 
-		bytesReadHeader, err := b.readFull(header)
+		deadline := promise.requestTime.Add(b.conf.Net.ReadTimeout)
```
puellanivis, line 1346: I’m going to suggest that we set this off in its own paragraph (or in the `header` paragraph above), this will make it more clear that this deadline is used beyond just passing it into the following call…

Otherwise, we would want to just use the `promise.requestTime.Add(b.conf.Net.ReadTimeout)` directly, rather than assigning through a temporary variable. Having it on its own paragraph gives some sort of semantic clue that the value is going to be used for more than just the one following line.

puellanivis (commented):
I’m not sure this actually resolves #2820 ? Their complaint was that the `Net.ReadTimeout` has to be set to the longest timeout, which this PR just further reinforces as, I guess intentional?

I agree this change is probably absolutely appropriate, the timeout really should apply for the whole `readFull` operation, not for individual reads thereof… but I’m just not so sure that it’s what #2820 was questioning.

broker.go
hugoev: Good call — added a blank line below so deadline sits in its own paragraph, making it clear it spans both reads.

hugoev:
You're right — I've updated it to Relates to #2820. The per-read reset is a real correctness issue worth fixing on its own; the idle timeout the reporter is actually after would be a follow-up.

config.go
```diff
@@ -51,8 +51,17 @@ type Config struct {
 		// All three of the below configurations are similar to the
 		// `socket.timeout.ms` setting in JVM kafka. All of them default
```
hindessm, line 52: This PR definitely improves semantics of the `ReadTimeout` as it is already but I wonder if there might be more we should do here.

The consumer configs describe `socket.timeout.ms` as:
> The socket timeout for network requests. The actual timeout set will be max.fetch.wait + socket.timeout.ms.

So it seems like saying these are "similar" is possibly a little misleading and I wonder if we shouldn't make it more similar.  For example, making the deadline for Fetch include the `config.Consumer.Fetch.MaxWaitTime`.

Given the significantly different times different requests might be expected to take, a single max duration of any request (which is what this becomes) possibly isn't the most useful configuration and perhaps we can do better?

dnwe: I did have an old branch somewhere that introduced a “holdTime” to the protocol interface and responseReceiver set the connection read deadline to Net.ReadTimeout + holdTime where holdTime was the appropriate field from each, i.e., FetchRequest (MaxWaitTime), ProduceRequest (Timeout), JoinGroupRequest (RebalanceTimeout), and the admin requests all had their existing Timeout field (CreateTopics, DeleteTopics, CreatePartitions, DeleteRecords, ElectLeaders, Alter/ListPartitionReassignments etc.). I don’t think I ever got around to pushing it up anyway as I wasn’t fully convinced by it

PR #3674, opened 2026-07-16
fix(example): potentional race-condition in consumer-group example

mabrarov:
Fix race-condition in consumer-group example when waiting for consumer readiness. Refer to https://github.com/IBM/sarama/pull/1404#discussion_r3597058952.

### DIFF
--- a/examples/consumergroup/main.go
+++ b/examples/consumergroup/main.go
@@ -84,11 +84,15 @@ func main() {
 		config.Consumer.Offsets.Initial = sarama.OffsetOldest
 	}
 
+	consumerReady := make(chan struct{})
+
 	/**
 	 * Setup a new Sarama consumer group
 	 */
 	consumer := Consumer{
-		ready: make(chan bool),
+		ready: sync.OnceFunc(func() {
+			close(consumerReady)
+		}),
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
@@ -116,11 +120,10 @@ func main() {
 			if ctx.Err() != nil {
 				return
 			}
-			consumer.ready = make(chan bool)
 		}
 	}()
 
-	<-consumer.ready // Await till the consumer has been set up
+	<-consumerReady // Await till the consumer has been set up
 	log.Println("Sarama consumer up and running!...")
 
 	sigusr1 := make(chan os.Signal, 1)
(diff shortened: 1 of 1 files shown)


puellanivis (commented):
As noted, no, there was not a potential race condition in the previous example. No, not even a potential for one.

If we’re going to fix this code, then we should do a full rewrite with `signal.NotifyContext` setup from the start, and we don’t need a whole complicated `sync.OnceXY` pattern because there’s only one place that might ever close the channel, and it’s only ever possible that a single thread could ever be in that code already… but like I explained, _we don’t need that ready condition anyways_.

We can just go straight away to the wait for `NotifyContext` to close. There’s no reason to have to wait for the consumer to be ready. Entirely removing all of this code means everything still works.

mabrarov:
Hi @puellanivis,

I am under impression that waiting for _ready condition_ in this example is used to demonstrate what `sarama.ConsumerGroupHandler.Setup` can be used for.

Regarding lack of race-condition - I replied in https://github.com/IBM/sarama/pull/1404#discussion_r3602532901 - it's more about correctness than about reproducible bug (which is usual for the issues of race-condition nature).

Thank you.

puellanivis:
> I am under impression that waiting for _ready condition_ in this example is used to demonstrate what `sarama.ConsumerGroupHandler.Setup` can be used for.

This is still not a proper reason for why it waits to setup the signal handler. It’s absurd that we will simply crash out on a `SIGTERM` during a partial startup rather than unwinding and gracefully shutdown, which is the whole point of having a signal handler handling `SIGTERM` in the first place.

mabrarov:
Considering https://github.com/IBM/sarama/pull/1404#discussion_r3607048072 - if we still want to keep waiting for `sarama.ConsumerGroupHandler.Setup` to complete (refer to https://github.com/IBM/sarama/pull/3674#issuecomment-5008719001)... which is not the thing this PR makes decision on (is out of scope of this PR) - this PR is still helpful to ensure the wait at line 126 (in the new version of the modified main.go) is unblocked even if _server-side rebalance happen_ and `sarama.ConsumerGroupHandler.Setup` was not called (is it possible?) at the first iteration of loop starting at line 109. Isn't it?

puellanivis:
> `sarama.ConsumerGroupHandler.Setup` was not called (is it possible?) at the first iteration of loop starting at line 109. Isn't it?

Probably. 🤷‍♀️ We should still clean up everything else. Move the signal handler to start, and then `select` on the ready, and signal handler, so we can properly handle the case of getting a SIGTERM while setting up as well, and even without moving the signal handler at all, now the context we setup is entirely unnecessary.

There’s nothing wrong with the code as is, but it’s insufficient for the scope of changes that are being made.

PR #3676, opened 2026-07-17
feat: raise default Kafka version to 2.8.0

dnwe:
Sarama has full protocol coverage up to and including 2.8, so assume that as the baseline rather than 2.6.0 when Version is unset.

### DIFF
--- a/config.go
+++ b/config.go
@@ -529,7 +529,8 @@ type Config struct {
 	// and most 3rdparty ones.
 	ApiVersionsRequest bool
 	// The version of Kafka that Sarama will assume it is running against.
-	// Defaults to the oldest supported stable version. Since Kafka provides
+	// Defaults to 2.8 as Sarama has full protocol coverage for that
+	// version. Since Kafka provides
 	// backwards-compatibility, setting it to a version older than you have
 	// will not break anything, although it may prevent you from using the
 	// latest features. Setting it to a version greater than you are actually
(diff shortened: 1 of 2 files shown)


lizthegrey (commented):
lgtm

dnwe:
@lizthegrey fyi (in case it wasn’t already known) you should be able to set Version to MaxVersion, and Sarama will use the ApiVersionsResponse to gate the apikey versions to use and it’ll use the maximum that both broker and sarama support (which for some apikeys is higher than the 2.8 baseline). At the moment we just keep the Version field around for backwards compatibility

lizthegrey:
> @lizthegrey fyi (in case it wasn’t already known) you should be able to set Version to MaxVersion, and Sarama will use the ApiVersionsResponse to gate the apikey versions to use and it’ll use the maximum that both broker and sarama support (which for some apikeys is higher than the 2.8 baseline). At the moment we just keep the Version field around for backwards compatibility

ooh. fyi @jparsons04

dnwe:
For reference https://github.com/IBM/sarama/releases/tag/v1.46.0 and newer contain this behaviour

 > [!NOTE]  
> This release contains significant changes. Notably Sarama will now use the ApiVersionRequest response from each broker to aid in selecting the protocol version to use. The existing `Version` field in sarama.Config will continue to provide a "pinning" mechanism, but can safely be set to a maximum or higher value than the remote cluster and sarama will sensibly pick compatible versions.

hindessm (approved):
LGTM

PR #3681, opened 2026-07-22
fix(consumer): preserve pauses across reassignment

dnwe:
- prevent reassignment from silently resuming paused consumption
- retain pauses for active partitions and apply them to replacement consumers
- clear retained state through Resume and ResumeAll, with unit test coverage

### DIFF
--- a/consumer.go
+++ b/consumer.go
@@ -106,6 +106,9 @@ type consumer struct {
 	client          Client
 	metricRegistry  metrics.Registry
 	lock            sync.Mutex
+
+	// paused retains pause state while a partition is reassigned, guarded by lock
+	paused map[string]map[int32]none
 }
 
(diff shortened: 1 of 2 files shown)


consumer.go
```diff
@@ -106,6 +106,9 @@ type consumer struct {
 	client          Client
 	metricRegistry  metrics.Registry
 	lock            sync.Mutex
+
+	// paused retains pause state while a partition is reassigned, guarded by lock
```
puellanivis: Perhaps we should restructure the fields to fit the Go style of putting the things protected by `lock` in a paragraph headed with `lock`? That way we don’t have to be so explicit about what’s being protected by lock. (Since for instance, none of the other entries make it clear if they’re protected or not.)

dnwe: Yep good thought, addressed

puellanivis (approved):

PR #3688, opened 2026-07-26
feat(offset): commit revoked partitions on cooperative rebalance

dnwe:
Add removePartitions to commit and stop tracking the offsets for the partitions a cooperative rebalance moves away, without touching the retained ones. A generation lock stops a commit using a stale generation while the assignment is changing.

### DIFF
--- a/offset_manager_test.go
+++ b/offset_manager_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"runtime"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"testing"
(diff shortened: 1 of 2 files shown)


offset_manager_test.go
```diff
@@ -929,3 +930,251 @@ func TestConstructRequestRetentionTime(t *testing.T) {
...
+		var committed []int32
+		for _, req := range capture.requests() {
+			for p := range req.blocks["my_topic"] {
+				committed = append(committed, p)
+			}
+		}
+		require.Contains(t, committed, int32(2))
+		require.Contains(t, committed, int32(3))
+		require.NotContains(t, committed, int32(0))
+		require.NotContains(t, committed, int32(1))
```
hindessm: I may be clearer/stronger to assert:
```suggestion
		require.ElementsMatch(t, []int32{2,3}, committed)
```

hindessm (approved):
Approved with very minor comment.

PR #3690, opened 2026-07-27
fix(producer): wake blocked flush when a muted partition is unmuted

### LINKED ISSUE #3689: Producer permanently stall after upgrading from v1.50.3 to v1.60.0
##### Description

##### Versions

##### Additional Context

- Kafka config:

```
connections.max.idle.ms=1000
```

- Application:

```golang
package main

import (
	"log"
	"time"

	"github.com/IBM/sarama"
)

func main() {

	cfg := sarama.NewConfig()
	cfg.ClientID = "test"

	cfg.Producer.Retry.Max = 1

	if err != nil {
		panic(err)
	}

	for {
		time.Sleep(10 * time.Second)

		deadlockReport(func() {

				{Topic: testTopic},
			})

			if err != nil {
			} else {
			}
		})
	}
}

	done := make(chan struct{})

	go func() {
		defer close(done)
		cb()
	}()

	select {
		<-done
	case <-done:
	}
}
```

khayyamsaleem:
Fixes #3689

`brokerProducer.run()` never wakes up when a muted partition is unmuted. If `tryBuildFlushingBatch` fails because the partition is muted by an in-flight `retryBatch`, the select blocks until the next input/timer/response event. With a sync producer blocked in `SendMessages` and no flush timer, that event never comes and the batch is stranded forever.

`waitForSpace` and `shutdown` already guard this with `muter.awaitUnmuteChan` (#3645); `run()` was the third flush site and was missed. The hazard dates to the muter (v1.48.0, idempotent producers only); #3644 made it reachable for non-idempotent producers in v1.60.0. Full analysis in https://github.com/IBM/sarama/issues/3689#issuecomment-5092089657.

The fix selects on `awaitUnmuteChan(bp.accumulatingBatch)` when the build fails on a non-empty batch, retrying the build immediately if the unmute races in between. The inner loop can't spin: a build only fails on a non-empty batch when every partition is muted, and "not blocked" means none are, so the next build succeeds.

Testing:
- the new regression test fails on main (`timed out waiting for channel`) and passes with the fix, `-race` clean
- the issue's snippet against a live Kafka 3.7.2 with `connections.max.idle.ms=1000` runs indefinitely; on main it stalls on the second batch every run
- `go test -race .` matches main (TestClientBackgroundMetadataUpdater and TestPartitionConsumerComputeBackoff fail identically on clean main under the full suite, unrelated)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -1126,8 +1126,20 @@ func (bp *brokerProducer) run() {
 	Logger.Printf("producer/broker/%d starting up\n", bp.broker.ID())
 
 	for {
+		// if the accumulating batch is ready to flush but every partition is
+		// blocked behind a mute (an in-flight retryBatch), watch for the
+		// unmute signal: no other event may ever arrive to wake us up again (#3689)
+		var unmuteChan <-chan struct{}
 		if bp.flushingBatch == nil && (bp.timerFired || bp.accumulatingBatch.readyToFlush()) {
-			bp.tryBuildFlushingBatch()
+			for !bp.tryBuildFlushingBatch() && !bp.accumulatingBatch.empty() {
+				ch, blocked := bp.parent.muter.awaitUnmuteChan(bp.accumulatingBatch)
+				if blocked {
+					unmuteChan = ch
+					break
+				}
+				// raced with an unmute between the failed build and
+				// awaitUnmuteChan; retry the build immediately
+			}
 		}
 
 		var timerChan <-chan time.Time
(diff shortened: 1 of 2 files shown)


dnwe:
Thanks @khayyamsaleem. The root cause analysis here was spot on, and the live-broker verification was much appreciated. I've fixed the same missed wake-up in #3692 with a slightly different implementation: the unmute signal is snapshotted before the mute attempt (inside tryBuildFlushingBatch), which closes the check/wait race by construction and avoids the need for the retry loop, and the same pattern now covers the waitForSpace and shutdown call sites too, allowing awaitUnmuteChan to be removed entirely. Closing in favour of that PR, but many thanks again for chasing this down so quickly after it was raised.

PR #3691, opened 2026-07-27
fix(producer): prevent deadlock in non-idempotent producer retryBatch

### LINKED ISSUE #3689: Producer permanently stall after upgrading from v1.50.3 to v1.60.0

lbcjbb:
PR #3644 introduced a regression that can cause a `SyncProducer` to block
indefinitely when using non-idempotent mode with connection retries.
    
The deadlock scenario:
    
1. A connection error triggers `handleError()`, which launches a goroutine
   calling `retryBatch()` for the failed batch.
2. `retryBatch()` creates a new `brokerProducer` via `getBrokerProducer()` (refcount=1).
3. `retryBatch()` sends the batch on `bp.output`, then immediately releases its
   reference via defer `unrefBrokerProducer()` (refcount=0).
4. With `refcount=0`, the `brokerProducer` shuts down while the batch is still
   in flight waiting for a response from the broker.
5. The response is lost, `returnSuccesses()` / `returnErrors()` is never called.
6. `SendMessages()` blocks forever waiting on `ProducerMessage.expectation`.
    
The fix stores the `brokerProducer` reference in the `produceSet` itself, so it
is only released after `handleResponse()` has fully processed the batch (success
or retry exhausted). This ensures the `brokerProducer` stays alive until the
batch lifecycle is complete.

Fixes #3689

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -1358,6 +1358,15 @@ func (bp *brokerProducer) handleResponse(response *brokerProducerResponse) {
 		bp.handleSuccess(response.set, response.res)
 	}
 
+	// Release the BP reference if this produceSet was holding one (from retryBatch).
+	// This must be done after handling the response to ensure the BP stays alive
+	// until all processing is complete.
+	if response.set.bpRefBP != nil {
+		bp.parent.unrefBrokerProducer(response.set.bpRefBroker, response.set.bpRefBP)
+		response.set.bpRefBroker = nil
+		response.set.bpRefBP = nil
+	}
+
 	if bp.accumulatingBatch.empty() {
 		bp.rollOver() // this can happen if the response invalidated our buffer
 	}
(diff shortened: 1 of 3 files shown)


dnwe:
Thanks @lbcjbb for the original issue report. Having dug into this, I don't believe the premature shutdown is actually the failure mode: when the retryBatch-created brokerProducer is un-reffed to zero, its shutdown() still drains and handles every pending response after closing output, so the retried batch's response isn't lost and the partition does get unmuted. The deadlock is that this unmute happens on a goroutine the stranded brokerProducer never hears about, because its run() loop had no case to wake on an unmute. That's what the other PR fixes. 

The handleSuccess changes here would also alter non-idempotent retry behaviour beyond the scope of the regression, so it's not obvious what issue they're be fixing or improving?

lbcjbb:
Yes, I came to the same conclusion. I introduced a delay in the execution of `retryBatch()`, but the messages are not lost. The `brokerProducer` is closed/released while a batch is still being processed, but this has no impact now with https://github.com/IBM/sarama/pull/3692.

> The handleSuccess changes here would also alter non-idempotent retry behaviour beyond the scope of the regression, so it's not obvious what issue they're fixing or improving?

I've included these changes in this PR, but they are more complementary to PR https://github.com/IBM/sarama/pull/3644 than directly related to the regression. I'm going to move them to a dedicated PR (https://github.com/IBM/sarama/pull/3693).

Thank you for your reply and for taking the time to look into this :pray:

PR #3692, opened 2026-07-27
fix(producer): wake stranded brokerProducer when partition unmutes

### LINKED ISSUE #3689: Producer permanently stall after upgrading from v1.50.3 to v1.60.0

dnwe:
A brokerProducer whose accumulated batch was blocked behind a partition muted elsewhere (e.g. by an in-flight retryBatch handled on another brokerProducer) never woke up, as its run() loop did not select on the unmute signal, deadlocking non-idempotent producers after connection errors.

- select on the muter unmute signal in run() when the batch cannot be built
- snapshot the signal before any mute attempt so a racing unmute is not missed
- fold awaitUnmuteChan into tryBuildFlushingBatch, returning the signal to wait on
- add an end-to-end regression test for the stranded batch scenario

Fixes #3689

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -225,18 +225,13 @@ func (m *partitionMuter) waitUntilMuted(set *produceSet) bool {
 	return true
 }
 
-func (m *partitionMuter) awaitUnmuteChan(set *produceSet) (<-chan struct{}, bool) {
-	if set == nil || set.empty() {
-		return nil, false
-	}
-
+// nextUnmuteSignal returns the channel that the next unmute (or close) will
+// close.
+func (m *partitionMuter) nextUnmuteSignal() <-chan struct{} {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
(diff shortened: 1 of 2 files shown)


hindessm (approved):
LGTM

PR #3693, opened 2026-07-28
fix(producer): keep partition muted during handleSuccess retries for non-idempotent producers

lbcjbb:
PR #3644 changed `handleError()` to use `retryBatch()` for non-idempotent producers, keeping partitions muted during retries. However, `handleSuccess()` was not updated consistently: it called `retryBatch()` with `alreadyMuted=true`, but did not add the partition to `keepMuted`, causing the partition to be unmuted immediately after launching the retry goroutine.

This inconsistency could lead to message reordering, as a new batch for the same partition could be sent before the retry completes.

This commit aligns `handleSuccess()` with `handleError()` by:

- Always populating `keepMuted` for retriable errors (not just for idempotent)
- Always calling `RefreshMetadata()` before retries (not just for idempotent)

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -1404,12 +1404,11 @@ func (bp *brokerProducer) handleSuccess(sent *produceSet, response *ProduceRespo
 				bp.parent.returnErrors(pSet.msgs, block.Err)
 			} else {
 				retryTopics = append(retryTopics, topic)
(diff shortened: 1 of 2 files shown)


puellanivis (commented):
I don’t see any issues with the code, but I have no idea about the appropriateness of the logic here.

PR #3694, opened 2026-07-28
feat(client): expose ClusterID via Client interface

shashank-reddy-nr:
## Summary

Adds `ClusterID() string` to the `Client` interface, surfacing the cluster ID that brokers already return in `MetadataResponse` (protocol version ≥ 2, Kafka ≥ 0.10.1.0) but that was never stored or accessible to callers.

### Changes

- **`client.go`** — adds `clusterID string` field to the `client` struct; stores the value in `updateMetadata()` under the existing `client.lock` RWMutex (same pattern as `controllerID`); adds `ClusterID() string` to the `Client` interface; implements `(*client).ClusterID()` with `RLock`.
- **`mockresponses.go`** — adds `clusterID` field and `SetClusterID()` builder to `MockMetadataResponse` so tests can configure the mock; populates `MetadataResponse.ClusterID` in `For()`.
- **`client_test.go`** — adds `TestClientClusterID` with two subtests: one at `V0_10_1_0` that asserts the value round-trips correctly, and one at `V0_10_0_0` that asserts an empty string is returned when the broker does not include a cluster ID.
- **`async_producer_test.go`** — adds the `ClusterID() string` stub to `stubLeaderClient` to satisfy the updated interface.

### Design notes

- Returns `string` (not `(string, error)`): the cluster ID is either present or not; returning empty string for pre-0.10.1 brokers is consistent with how `controllerID` behaves before any metadata has been received.
- `nopCloserClient` requires no changes — it embeds the `Client` interface by value and inherits the new method automatically.
- Nil-pointer guard: `MetadataResponse.ClusterID` is `*string`; the store path checks `!= nil && != ""` before overwriting, so a nil pointer from an old-version response never clears a previously cached value.

### DIFF
--- a/client.go
+++ b/client.go
@@ -35,6 +35,11 @@ type Client interface {
 	// and stores it in the local cache. Requires Kafka 0.10 or higher.
 	RefreshController() (*Broker, error)
 
+	// ClusterID returns the cluster ID reported by the broker in the most recent
+	// metadata response. It returns an empty string if the broker has not yet
+	// responded or if the Kafka version predates 0.10.1 (MetadataResponse v2).
+	ClusterID() string
+
 	// Brokers returns the current set of active brokers as retrieved from cluster metadata.
 	Brokers() []*Broker
 
@@ -156,6 +161,7 @@ type client struct {
 	deadSeeds   []*Broker
 
 	controllerID            int32                                   // cluster controller broker id
+	clusterID               string                                  // cluster ID from metadata response (Kafka ≥ 0.10.1)
 	brokers                 map[int32]*Broker                       // maps broker ids to brokers
 	metadata                map[string]map[int32]*PartitionMetadata // maps topics to partition ids to metadata
 	metadataTopics          map[string]none                         // topics that need to collect metadata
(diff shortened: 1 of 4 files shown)


client.go
```diff
@@ -35,6 +35,11 @@ type Client interface {
 	// and stores it in the local cache. Requires Kafka 0.10 or higher.
 	RefreshController() (*Broker, error)
 
+	// ClusterID returns the cluster ID reported by the broker in the most recent
+	// metadata response. It returns an empty string if the broker has not yet
+	// responded or if the Kafka version predates 0.10.1 (MetadataResponse v2).
+	ClusterID() string
```
puellanivis, line 41: This is expanding an exported interface, and thus a breaking change.

client.go
```diff
@@ -1101,6 +1113,9 @@ func (client *client) updateMetadata(data *MetadataResponse, allKnownMetaData bo
 	client.updateBroker(data.Brokers)
 
 	client.controllerID = data.ControllerID
+	if data.ClusterID != nil && *data.ClusterID != "" {
+		client.clusterID = *data.ClusterID
+	}
```
puellanivis, line 1118: The test against `*data.ClusterID != ""` is unnecessary. We can simply assign a zero value as a no-op.

PR #2826, opened 2024-03-05
fix: add locking around broker throttle timer to prevent race condition

chengsha:
The producer can be easily blocked due to a race condition in Broker.throttleTimer, which may result in a panic. Consider changing the type of Broker.throttleTimer to atomic.Value
    
Fixes #2823
    
Signed-off-by: chengsha  <shacheng@tencent.com>

PR #3711, opened 2026-08-14
fix: make broker throttle waits interruptible on close

### LINKED ISSUE #3710: Broker.Close can be blocked for the full broker throttle duration
### Description

A long broker-supplied throttle duration can prevent a producer from closing and recovering after the connection has failed.

We observed a `ProduceResponse` with a throttle duration of `12h42m52.356s`. Sarama honored the response by entering `waitIfThrottled`. Approximately two seconds later, an in-flight request hit a TCP read timeout and the broker producer changed to the closing state, but the broker could not finish closing until the throttle timer expired. Synchronous producers remained blocked during that period, and production did not recover even after the Kafka cluster was healthy again.

Sanitized logs:

```text
broker/242559 *sarama.ProduceResponse throttled 12h42m52.356s
broker/242559 waiting for throttle timer
producer/broker/242559 state change to [closing] because read tcp <client>-><broker>: i/o timeout
```

A goroutine dump showed `(*syncProducer).SendMessage` blocked waiting for its result while the broker producer was unable to complete its close and retry path.

### Versions

| Component | Version |
| --- | --- |
| Sarama | v1.43.1 |
| Go | 1.22 |
| Kafka broker | Unknown |

The same throttle implementation is still present on the current `main` branch and in v1.60.1.

### Root cause

`sendInternal` calls `waitIfThrottled` while its caller holds `Broker.lock`.

`waitIfThrottled` blocks on `<-b.throttleTimer.C` for the complete broker-supplied duration. Since PR #2826, it also holds `throttleTimerLock` for the complete wait.

When an in-flight request fails while another send is waiting for the throttle timer:

1. the broker producer starts its connection error/close path;
2. `Broker.Close` tries to acquire `Broker.lock`;
3. the send that owns `Broker.lock` remains blocked on the throttle timer;
4. close, reconnect, retry, and delivery-result propagation cannot progress until the timer fires.

`Net.ReadTimeout`, `Net.WriteTimeout`, and `Producer.Timeout` do not bound this wait because no network operation is active in `waitIfThrottled`.

### Expected behavior

Sarama should continue to honor a broker throttle while the broker connection is active.

When the broker is closing or the connection has failed, the local throttle wait should be interrupted promptly. The blocked send should return an existing connection/closed error so the normal producer retry, metadata refresh, and reconnect paths can proceed.

### Actual behavior

The broker close and producer recovery paths can remain blocked for the complete broker-supplied throttle duration.

### Proposed scope

This issue does **not** propose ignoring, shortening, or globally capping valid Kafka quota throttles.

A focused fix should:

- make the throttle wait interruptible by broker close or connection failure;
- avoid holding the throttle mutex while blocked on the timer;
- ensure the send path does not prevent `Broker.Close` from making progress;
- preserve the existing throttle behavior while the connection remains active;
- add regression coverage for a long throttle combined with concurrent close/connection failure;
- pass `go test -race` without leaking goroutines.

### Related issues and pull requests

- #2823 reported the race between `setThrottle` and `waitIfThrottled`.
- #2826 fixed that race by adding `throttleTimerLock`, but did not make the wait interruptible.
- #2121 reported producer shutdown remaining blocked for a long time.
- #2133 fixed a different `Broker.Close` deadlock in the broker producer error path and is useful precedent for keeping close paths non-blocking.

liuzengh:
Fixes #3710.

## What changed

- Replace the shared throttle timer with a deadline and a broadcast notification channel.
- Let `Broker.Close` cancel an active throttle wait before it tries to acquire `Broker.lock`.
- Return `ErrNotConnected` from a send interrupted by broker shutdown so the existing producer retry and reconnect paths can proceed.
- Reset throttle state when a new broker connection is opened and ignore throttled responses racing with shutdown.
- Add regression coverage for closing a broker while a send is blocked by a long throttle, and preserve replacement-throttle behavior.

## Why

`sendInternal` waits for a broker-provided throttle while its caller owns `Broker.lock`. Before this change, a long throttle therefore prevented `Broker.Close` from acquiring that lock. If another in-flight request failed during the wait, broker close, reconnect, retries, and delivery-result propagation could all remain blocked for the complete throttle duration.

The throttle wait also held `throttleTimerLock` while receiving from the timer channel. Moving to a deadline plus per-wait timer avoids sharing a timer channel between `setThrottle` and active waiters, while the notification channel lets replacement throttles and shutdown wake all waiters safely.

This change does not cap or ignore broker quota throttles. A healthy active connection continues to honor the complete broker-provided duration.

## Validation

- `GOTOOLCHAIN=auto go test ./...`
- `GOTOOLCHAIN=auto go test ./... -run 'TestBrokerClose|Test_handleThrottledResponse' -count=20`
- `GOTOOLCHAIN=auto go test -race ./... -run 'TestBrokerClose|Test_handleThrottledResponse' -count=10`
- `GOTOOLCHAIN=auto go vet ./...`

A full `GOTOOLCHAIN=auto go test -race ./...` run reaches existing races involving the global `DebugLogger` in `TestClientBackgroundMetadataUpdater` and `TestPartitionConsumerComputeBackoff`. The same races reproduce on an unmodified `origin/main` worktree. No race was reported in the changed throttle and close tests.

### DIFF
--- a/broker.go
+++ b/broker.go
@@ -61,10 +61,10 @@ type Broker struct {
 	kerberosAuthenticator               GSSAPIKerberosAuth
 	clientSessionReauthenticationTimeMs int64
 
+	throttleLock     sync.Mutex
 	throttleDeadline time.Time
 	throttleNotify   chan struct{}
 	throttleCanceled bool
-	throttleLock     sync.Mutex
 }
 
 // SASLMechanism specifies the SASL mechanism the client uses to authenticate with the broker
@@ -2018,16 +2018,9 @@ func (b *Broker) waitIfThrottled() error {
 		b.throttleLock.Unlock()
 
 		DebugLogger.Printf("broker/%d waiting for throttle timer\n", b.ID())
-		timer := time.NewTimer(remaining)
 		select {
-		case <-timer.C:
+		case <-time.After(remaining):
 		case <-notify:
-			if !timer.Stop() {
-				select {
-				case <-timer.C:
-				default:
-				}
-			}
 		}
 	}
 }
--- a/broker.go
+++ b/broker.go
@@ -61,8 +61,10 @@ type Broker struct {
 	kerberosAuthenticator               GSSAPIKerberosAuth
 	clientSessionReauthenticationTimeMs int64
 
-	throttleTimer     *time.Timer
-	throttleTimerLock sync.Mutex
+	throttleDeadline time.Time
+	throttleNotify   chan struct{}
+	throttleCanceled bool
+	throttleLock     sync.Mutex
 }
 
 // SASLMechanism specifies the SASL mechanism the client uses to authenticate with the broker
@@ -201,6 +203,7 @@ func (b *Broker) Open(conf *Config) error {
 	}
 
 	b.lock.Lock()
+	b.resetThrottle()
 
 	if b.metricRegistry == nil {
 		b.metricRegistry = newCleanupRegistry(conf.MetricRegistry)
@@ -372,6 +375,11 @@ func (b *Broker) TLSConnectionState() (state tls.ConnectionState, ok bool) {
 
 // Close closes the broker resources
 func (b *Broker) Close() error {
+	// Wake a send waiting on a broker-provided throttle before acquiring
+	// b.lock. The send owns b.lock while waiting, so acquiring it first would
+	// make Close wait for the complete throttle duration.
+	b.cancelThrottle()
+
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
@@ -394,6 +402,8 @@ func (b *Broker) maybeCloseLocked(err error) bool {
 // closeLocked closes the broker connection and resets state.
 // NOTE: caller must hold b.lock.
 func (b *Broker) closeLocked() error {
+	b.cancelThrottle()
+
 	if b.conn == nil {
 		return ErrNotConnected
 	}
@@ -1182,7 +1192,9 @@ func (b *Broker) sendInternal(rb protocolBody, promise *responsePromise) error {
 	}
 
 	// check and wait if throttled
-	b.waitIfThrottled()
+	if err := b.waitIfThrottled(); err != nil {
+		return err
+	}
 
 	requestTime := time.Now()
 	// Will be decremented in responseReceiver (except error or request with NoResponse)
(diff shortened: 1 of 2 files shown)


broker_test.go
```diff
@@ -1680,17 +1729,24 @@ func Test_handleThrottledResponse(t *testing.T) {
...
 		}
-		startTime := time.Now()
-		broker.waitIfThrottled()
-		if time.Since(startTime) < throttleTime*2 {
-			t.Fatal("expected throttling to use second delay")
+		select {
+		case err := <-waitErrs:
+			require.NoError(t, err)
+		case <-time.After(throttleTime * 2):
+			t.Fatal("throttle wait did not use the replacement delay")
```
liuzengh: This replacement-throttle coverage is now timing-based and can both flake and false-pass.

`Sleep(throttleTime/4)` does not guarantee the waiter is blocked before the second throttle is applied. If the first 100ms deadline expires first (slow scheduler, long Sleep), the waiter returns early and this fatals even when replacement logic is fine.

The two `select`s with `time.After(throttleTime)` are also a weak bound. If replacement is ignored and the wait ends at about 100ms after the second `handleThrottledResponse`, both `waitErrs` and `time.After(throttleTime)` can be ready together; `select` chooses one at random, so the test can pass without proving the 2x delay.

The Close test already waits until the send holds `broker.lock` before calling `Close`. Do the same here: wait until `waitIfThrottled` is actually blocked, apply the replacement, then assert elapsed time from a measured start against `2 * throttleTime` (as the old test did) instead of stacking `time.After` windows.

liuzengh (commented):
The Close interrupt path looks correct. One inline note on the replacement-throttle test: the new timing windows can flake or false-pass.

broker_test.go
liuzengh: Fixed in 4149bac. The waiter now holds broker.lock while calling waitIfThrottled, and the test uses TryLock to confirm that the waiter has entered the throttled section before applying the replacement. I removed the Sleep and overlapping timing selects; the test now measures from the replacement update and asserts that the elapsed time is at least 2 * throttleTime, with only a 5-second watchdog for deadlocks. Verified with 50 repeated targeted runs, 20 race-enabled targeted runs, the full test suite, and go vet.

liuzengh:
I don't have permission to add reviewers from the GitHub UI (external fork contributor). Could @hindessm and @dnwe take a look when you have a chance?

This change makes broker throttle waits interruptible on `Close` (fixes #3710). It sits next to #2536 / #2826 (KIP-219 throttle wait and the timer lock) and #2133 (`Broker.Close` deadlock in the producer path).

broker.go
```diff
@@ -61,8 +61,10 @@ type Broker struct {
 	kerberosAuthenticator               GSSAPIKerberosAuth
 	clientSessionReauthenticationTimeMs int64
 
-	throttleTimer     *time.Timer
-	throttleTimerLock sync.Mutex
+	throttleDeadline time.Time
+	throttleNotify   chan struct{}
+	throttleCanceled bool
+	throttleLock     sync.Mutex
```
puellanivis: Please move this `sync.Mutex` to head the grouping of variables that it protects.

broker.go
```diff
@@ -1974,25 +1986,78 @@ func (b *Broker) handleThrottledResponse(resp protocolBody) {
...
+		select {
+		case <-timer.C:
+		case <-notify:
+			if !timer.Stop() {
+				select {
+				case <-timer.C:
+				default:
+				}
+			}
+		}
```
puellanivis: We are using Go 1.25 now, which means this pattern is no longer necessary. Timers can be recovered by the GC even if they have not expired or stopped, and performing a receive after a `time.Stop` returns `false` is no longer necessary.

The new pattern is significantly more simple:
```suggestion
		select {
		case <-time.After(remaining):
		case <-notify:
		}
```

broker.go
liuzengh: Done in 741f2be. `throttleLock` now heads the throttle state grouping.

broker.go
liuzengh: Done in 741f2be. Switched the wait to `time.After` and dropped the `Stop`/drain path, matching Go 1.25 timer GC behavior.

dnwe:
Thanks for the PR. Interrupting the throttle wait on Close makes sense, we shouldn’t sit holding b.lock for the full throttle duration.

I don’t think we should drop the deadline on close/reopen though. Kafka quotas are keyed on client-id, not connection, so a client that reconnects (e.g. after a request timeout) would skip the remaining penalty. Can we keep throttleDeadline as-is and have cancel just wake the waiter, so a send on the reopened connection waits out whatever remains?

PR #3712, opened 2026-08-15
docs: clarify producer batching configuration

krvladislav:
- Clarify producer batching scope: `Flush.*` is evaluated by each broker producer across all topic/partition batches accumulated for that broker, while `Producer.MaxMessageBytes` independently limits each topic/partition batch. ([per-partition batch limit](https://github.com/IBM/sarama/blob/649156cb0ead112a7ad237a7419ef9320854ba1f/produce_set.go#L326-L329), [produceSet limits and flush thresholds](https://github.com/IBM/sarama/blob/649156cb0ead112a7ad237a7419ef9320854ba1f/produce_set.go#L330-L351))

- Document that `Producer.MaxMessageBytes` not only rejects an oversized individual message but also limits the estimated pre-compression size of each topic/partition batch; Kafka validates the resulting record batch after compression using broker `message.max.bytes` or topic `max.message.bytes`. ([Kafka broker configuration](https://kafka.apache.org/41/configuration/broker-configs/#brokerconfigs_message.max.bytes))

- Remove mappings to legacy JVM `queue.buffering.*` names because they are not one-to-one: `queue.buffering.max.messages` limited the producer's entire unsent queue, while `Flush.MaxMessages` limits one broker request; the time-based mapping is also approximate: the legacy `queue.buffering.max.ms` controlled a producer-wide async queue before messages were grouped by broker and partition; its modern Java successor, `linger.ms`, applies while accumulating each topic/partition batch, whereas Sarama uses a timer for a broker-scoped `produceSet` containing multiple topic/partition batches. I propose documenting Java producer-to-Sarama configuration mappings in a separate table, potentially in the project Wiki, instead of maintaining partial mappings in field comments. ([legacy JVM producer](https://kafka.apache.org/08/configuration/producer-configs/#:~:text=queue.buffering.max.ms,queue.enqueue.timeout.ms), [current Java producer `linger.ms`](https://kafka.apache.org/41/configuration/producer-configs/#producerconfigs_linger.ms))

### DIFF
--- a/config.go
+++ b/config.go
@@ -183,8 +183,15 @@ type Config struct {
 	// Producer is the namespace for configuration related to producing messages,
 	// used by the Producer.
 	Producer struct {
-		// The maximum permitted size of a message (defaults to 1000000). Should be
-		// set equal to or smaller than the broker's `message.max.bytes`.
+		// The maximum permitted size of a message and the pre-compression size estimate
+		// of a batch for a single topic and partition. Messages larger than this are
+		// rejected with a ConfigurationError and are not sent. When adding a message
+		// would reach the batch limit, the current batch is flushed first.
+		// Kafka applies the broker's `message.max.bytes` or topic's `max.message.bytes`
+		// to the whole record batch after compression, so this value may be higher when
+		// compression is enabled, but a compressed batch that exceeds the Kafka limit
+		// will still be rejected.
+		// This value must remain smaller than sarama.MaxRequestSize.
 		MaxMessageBytes int
 		// The level of acknowledgement reliability needed from the broker (defaults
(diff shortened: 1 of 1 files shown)


krvladislav:
Hi @puellanivis, could you please take a look when you have a chance? This is a small documentation-only change. Thanks!

puellanivis (commented):
This contains no code changes. My normal review process for this project is empty.

krvladislav:
@dnwe would you mind reviewing this documentation change? 
If documentation changes should be routed differently please let me know

dnwe (approved):
thanks looks reasonable to me

PR #3722, opened 2026-08-21
fix: correct JoinGroup v8 selection threshold to V3_2_0_0

### LINKED ISSUE #3721: JoinGroupRequest v8 selected for KafkaVersion 3.1.0, but v8 requires 3.2.0
`consumer_group.go` picks JoinGroupRequest v8 as soon as `Config.Version.IsAtLeast(V3_1_0_0)`:

    if c.config.Version.IsAtLeast(V3_1_0_0) {
        req.Version = 8
    } else if c.config.Version.IsAtLeast(V2_5_0_0) {
        req.Version = 7
    ...

But `JoinGroupRequest.requiredVersion()` / `JoinGroupResponse.requiredVersion()` were fixed in #3585 to say v8 needs V3_2_0_0, not V3_1_0_0. The selection switch in consumer_group.go was never updated to match.

`subscriptionVersion()` (same file) is correctly gated at V3_2_0_0:

    case c.config.Version.IsAtLeast(V3_2_0_0):
        return 2 // KIP-792: GenerationID
    case c.config.Version.IsAtLeast(V2_4_0_0):
        return 1 // KIP-429: OwnedPartitions

So for Config.Version in V3_1_0_0, V3_1_1_0, V3_1_2_0 (all < V3_2_0_0), sarama sends a v8 JoinGroupRequest paired with v1 (old) subscription metadata. Against a real broker (Confluent Platform 7.9.1 / Kafka 3.9.x, tested both Zookeeper and KRaft mode), this fails for all three:

    kafka server: The version of API is not supported

V3_0_0_0 and V3_2_0_0 both work fine.

Environment:
- sarama v1.60.1
- Go 1.26
- Kafka 3.9.x (Confluent Platform 7.9.1), tested both Zookeeper and KRaft mode
- Config: `Consumer.Group` defaults, `Net.TLS.Enable = false`

I'll open a PR with a fix and a regression test.

mpicque:
Fixes #3721.

#3585 fixed requiredVersion() for JoinGroupRequest/Response v8: it now says V3_2_0_0, not V3_1_0_0. But the actual selection switch in joinGroupRequest() still checked V3_1_0_0. So Config.Version 3.1.0, 3.1.1 and 3.1.2 sent a v8 request paired with v1 subscription metadata (subscriptionVersion() is already gated at V3_2_0_0). Real brokers reject that with UNSUPPORTED_VERSION.

v8 adds the Reason field from KIP-800, shipped in Kafka 3.2.0. #3585 had the right version, it just didn't fix this switch too.

The first two commits reproduce and fix the issue:
- TestJoinGroupVersionSelection (mock broker) asserts the JoinGroup wire version sent for 3.1.0, 3.1.1, 3.1.2 and 3.2.0 - it fails before the fix because 3.1.x sends v8 instead of v7.
- TestFuncJoinGroupVersionSelection (functional, real broker) joins a consumer group for the same versions - it fails before the fix because the broker actually rejects the v8 request.
- The one-line fix; both tests then pass on all four versions.

A follow-up commit then reworks version selection so this kind of drift can't recur, and fixes an adjacent bug in the block it touches:
- JoinGroup version selection moves into a NewJoinGroupRequest(KafkaVersion) constructor in join_group_request.go, matching NewMetadataRequest and friends. A new subtest walks SupportedVersions and asserts the selected version satisfies requiredVersion(), so a future version bump that only updates one side fails CI.
- The KIP-800 Reason was consumed when the first JoinGroup request was built, so the second KIP-394 request (the one that actually joins) carried no reason. lastSessionCause is kept until a join succeeds, and the version-selection test exercises the two-request flow and asserts both requests repeat the reason.
- Test cleanup: the mock join/sync harness is shared with TestConsumerGroupJoinSync, and the functional test uses the shared defaultConfig() and runs in parallel.

### DIFF
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -540,7 +540,7 @@ func (c *consumerGroup) joinGroupRequest(coordinator *Broker, topics []string, h
 	// send two JoinGroupRequests, once with the empty member id, and then again
 	// with the assigned id from the first response. This is handled via the
 	// ErrMemberIdRequired case.
-	if c.config.Version.IsAtLeast(V3_1_0_0) {
+	if c.config.Version.IsAtLeast(V3_2_0_0) {
 		req.Version = 8
 	} else if c.config.Version.IsAtLeast(V2_5_0_0) {
 		req.Version = 7
--- a/consumer_group.go
+++ b/consumer_group.go
@@ -364,6 +364,9 @@ func (c *consumerGroup) joinSync(ctx context.Context, topics []string, held *hel
 	switch join.Err {
 	case ErrNoError:
 		c.memberID = join.MemberId
+		// clear the session cause only once a join succeeds so the KIP-394
+		// second request repeats the reason
+		c.lastSessionCause = nil
 	case ErrUnknownMemberId, ErrIllegalGeneration:
 		// reset member ID and retry immediately
 		c.memberID = ""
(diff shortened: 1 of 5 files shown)


dnwe (approved):
Thanks! This was a good catch. I added an extra commit to fix a related bug in the join group reason as well as moving the JoinGroup version selection to the protocol file like the rest of Sarama which should also prevent this sort of mismatch in the future

Issue #3703, opened 2026-08-06
[Feature Request] Add a metadata snapshot API


DCjanus:
## Problem

We operate a large number of Kafka clusters through a centralized management system. To populate its UI, a single service node may scan topic and partition metadata across many clusters. Issuing fresh metadata requests for every scan concentrates network and response-processing costs on that node and can make it a bottleneck.

These scans can tolerate slightly stale data; the important requirement is that they can read the client's existing cache without generating Kafka RPC traffic.

Sarama already maintains this metadata internally, but its public APIs do not provide a complete cache-only view:

- `ClusterAdmin.DescribeTopics` sends a `MetadataRequest` on every call.
- The `Client` accessors expose metadata in smaller pieces, and some may refresh metadata on a cache miss.

As a result, there is no way to inspect all currently cached broker and partition metadata with a strict no-refresh guarantee.

## Proposal

Expose a concurrency-safe metadata snapshot with these semantics:

- Reading it never sends a request or triggers a metadata refresh.
- It represents one consistent state of the client cache.
- It includes broker addresses and partition metadata such as leader, replicas, ISR, offline replicas, and errors.
- It is a detached copy, so modifying it cannot affect the client cache.

Callers that need fresh data can explicitly call `RefreshMetadata` before taking a snapshot.

## API compatibility

Our internal fork currently implements this by adding `MetadataSnapshot()` directly to `Client`. It has worked well for this use case, but adding a method to a public Go interface breaks external implementations and mocks.

For upstream, a separate optional interface may be a better fit:

```go
type MetadataSnapshotter interface {
	MetadataSnapshot() *MetadataSnapshot
}
```

Sarama's client implementation could implement this interface without changing the existing `Client` contract.

This is only a suggested API shape. I would be happy to follow maintainer guidance and contribute the implementation.

puellanivis:
Seems reasonable. But maybe a:
```go
type MetadataSnapshotterClient interface {
	Client

	MetadataSnapshot() *MetadataSnapshot
}

PR #3723, opened 2026-08-21
feat(client): add cached metadata snapshots

DCjanus:
## Why

Applications that manage many Kafka clusters may need to inspect a client's existing broker and partition metadata without issuing another Metadata request. The current public APIs expose parts of that state, but may refresh the cache on a miss and do not provide one consistent view.

This implements the optional client-extension direction discussed in #3703.

## What

- Add `MetadataSnapshotterClient`, an optional interface implemented by clients returned from `NewClient`, without changing the existing `Client` interface.
- Add `MetadataSnapshot` as a detached, point-in-time copy of cached controller, broker, and partition metadata. Reading it never sends a request or triggers a refresh; an empty or partially populated cache is returned as-is, while closed clients return `ErrClosedClient`.
- Document cache scope when `Config.Metadata.Full` is disabled, and report `ControllerID == -1` when the controller is unknown or unavailable in Metadata protocol v0.

## Risks

Snapshot creation is O(cache size) and performs the copy while holding the client read lock. This favors a simple, internally consistent view over a more complex immutable-cache or two-phase-copy design.

The internal unknown-controller sentinel changes from the zero value to `-1`. This does not change the existing `Client` API: Metadata protocol v0 already causes `Controller()` to return `ErrUnsupportedVersion`, supported protocol versions populate the real controller ID during metadata refresh, and the field is private.

## Notes

This also updates Staticcheck and golangci-lint so the existing `stable` CI matrix can analyze the project with Go 1.27. Staticcheck `2026.2rc1` is used because it is the first upstream release that supports Go 1.27 export data.

## Validation

On an Apple M1 Pro, snapshots of 10,000 partitions take about 1.59 ms and allocate 2.03 MB; snapshots of 100,000 partitions take about 19.4 ms and allocate 16.84 MB.


metadata_snapshot.go
```diff
@@ -0,0 +1,105 @@
...
+type BrokerSnapshot struct {
+	// Addr is the broker address advertised in cluster metadata.
+	Addr string
+	// Rack is the broker rack advertised in cluster metadata, or nil if the
+	// broker did not advertise a rack.
+	Rack *string
+}
+
+// PartitionSnapshot contains the cluster metadata cached for a partition.
+type PartitionSnapshot struct {
```
puellanivis, line 41: This type doesn’t seem to be different from `PartitionMetadata`? Do we need a separate type? It doesn’t seem like `PartitionMetadata` has any exported receiver methods, after all.

DCjanus: Defining `PartitionSnapshot` as a separate type here is intentional. Although it currently has exactly the same fields as `PartitionMetadata`, the two types represent different API contracts.

`PartitionMetadata` is part of the Kafka Metadata response model and may gain fields or exported methods as protocol support evolves. If the snapshot API reused that type directly, changes to the protocol model would automatically become part of the public snapshot API without necessarily being evaluated against the snapshot semantics.

Keeping a separate type allows the protocol response model and the cached snapshot API to evolve independently. When the protocol model changes, we can explicitly decide whether the corresponding information should also be exposed through snapshots. It also leaves room for the snapshot representation to adopt different fields or semantics in the future.

This does introduce a small amount of duplication in the fields and conversion logic. If the project prefers to minimize the number of types and reuse `PartitionMetadata`, I’m also open to making that change. However, since this is a new public API, I currently think keeping this boundary explicit is worthwhile.

puellanivis: 👍 Sounds good.

In that context then, consider the `Isr` field is an initialism and should be all caps, or we should just change it to the full `InSyncReplicas` as it seems to be the only field whose name is not inherently descriptive?

DCjanus: Changed `Isr` to `InSyncReplicas` in 5691523db4f7a0ffcc1817fa638dfa3273f755c4.

The original name was copied from `PartitionMetadata`, but since `PartitionSnapshot` is a new type, it does not need to follow the existing naming. This is a good opportunity to make the field name more descriptive.

puellanivis (approved):
There is nothing to complain about here.

PR #3729, opened 2026-08-24
fix(producer): drain pending responses immediately once channel closes

navi89mai:
In newBrokerProducer's response-ordering goroutine, once `pending` is closed, <-pending never blocks again (a closed channel is always ready). If `buf` still has buffered responses at that point, the select can keep racing the closed-channel case against `responses <- headRes` instead of making forward progress, since Go picks uniformly among ready cases.

Once closure is observed, there's nothing left to wait for, so drain the remaining buffer directly into `responses` in order instead of looping through select.

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -1072,7 +1072,15 @@ func (p *asyncProducer) newBrokerProducer(broker *Broker) *brokerProducer {
 			select {
 			case res, ok := <-pending:
 				if !ok {
-					continue
+					// pending is closed and never blocks again, so select could keep
+					// racing this case against responses<- instead of progressing.
+					// Once closed, just drain buf directly in order.
+					for buf.Length() > 0 {
+						responses <- buf.Peek()
+						buf.Remove()
+					}
+					close(responses)
+					return
 				}
 				buf.Add(res)
 				continue


navi89mai:
@puellanivis review pls, little enhancement)

async_producer.go
```diff
@@ -1072,6 +1072,10 @@ func (p *asyncProducer) newBrokerProducer(broker *Broker) *brokerProducer {
 			select {
 			case res, ok := <-pending:
 				if !ok {
+					for buf.Length() > 0 {
```
puellanivis, line 1075: Great explanation in the PR body for why this is important. But it’s still not readily clear in the code why this is important. Could you add a concise comment here to explain what you explain there?

async_producer.go
```diff
@@ -1072,6 +1072,10 @@ func (p *asyncProducer) newBrokerProducer(broker *Broker) *brokerProducer {
...
+						responses <- buf.Peek()
+						buf.Remove()
+					}
 					continue
```
puellanivis, line 1079: Fun fact, because we only leave `for buf.Length() > 0` once `buf.Length() == 0` this `continue` leads unconditionally to the `if buf.Length() == 0 { res, ok := <-pending ; if !ok { /* this code runs unconditionally. */ } }` block above.

So, there’s no real reason to control flow into another control flow, when we could also just do the work ourselves: (We are allowed to repeat ourselves a little… as a treat.)
```go
					// We are done forwarding the last pending response
					close(responses)
					return
```

puellanivis (commented):
Looks great. Just polishing recommendations.

async_producer.go
navi89mai: Yep, I've tried not to repeat cleanup logic (at least closing `responses` channel now, but maybe more logic in the future). But you're right, fixed it

async_producer.go
navi89mai: Added some explanation

puellanivis (approved):
Code looks good to me.

dnwe (approved):

PR #3733, opened 2026-08-31
fix: prevent race in asyncProducer retryHandler

dloebl:
There is a potential race in the `asyncProducer` `retryHandler` which is occasionally causing `nil` pointer dereferences on our side:

```
WARNING: DATA RACE
Write at 0x00c0001cd4d0 by goroutine 10:
  github.com/IBM/sarama.TestRetryHandlerHeadersRace.func2()
      /Users/dbl/src/github.com/IBM/sarama/async_producer_race_test.go:25 +0x18c

Previous read at 0x00c0001cd4d0 by goroutine 9:
  github.com/IBM/sarama.(*ProducerMessage).ByteSize()
      /Users/dbl/src/github.com/IBM/sarama/async_producer.go:400 +0x40
  github.com/IBM/sarama.(*asyncProducer).retryHandler()
      /Users/dbl/src/github.com/IBM/sarama/async_producer.go:1614 +0x488
  github.com/IBM/sarama.TestRetryHandlerHeadersRace.func1()
      /Users/dbl/src/github.com/IBM/sarama/async_producer_race_test.go:18 +0x68
```

`retryHandler` publishes a message to `p.input` and then reads it again to update its `ByteSize`:

```go
case p.input <- buf.Peek():
    msgToRemove := buf.Remove()
    currentByteSize -= int64(msgToRemove.ByteSize(version))
```

Once that send completes the message belongs to `dispatcher`, which can write to `Headers` through a `headerInterceptor`.

Only reachable with `Version >= V0_11_0_0` (`version == 2`): below that `ByteSize` skips the `Headers` loop entirely.

**I've added a test to reproduce:**
```
go test -race -run TestRetryHandlerHeadersRace
```

**Output without the fix:**
```
[..]
--- FAIL: TestRetryHandlerHeadersRace (0.00s)
    testing.go:1617: race detected during execution of test
FAIL
exit status 1
FAIL	github.com/IBM/sarama	0.072s
```

**Output with the fix applied:**
```
PASS
ok  	github.com/IBM/sarama	1.082s
```

### DIFF
--- a/async_producer.go
+++ b/async_producer.go
@@ -1580,6 +1580,15 @@ func (bp *brokerProducer) handleError(sent *produceSet, err error) {
 	}
 }
 
+// Message in the retry buffer, paired with the size it contributed to currentByteSize.
+// The size is measured on insert: measuring it again on removal would read a message we have already sent to p.input,
+// which the dispatcher goroutine owns and could modifiy by then:
+// Potential data race through custom headerInterceptor.
+type retryBufEntry struct {
+	msg  *ProducerMessage
+	size int64
+}
+
 // singleton
 // effectively a "bridge" between the flushers and the dispatcher in order to avoid deadlock
 // based on https://godoc.org/github.com/eapache/channels#InfiniteChannel
@@ -1601,11 +1610,7 @@ func (p *asyncProducer) retryHandler() {
 
 	var currentByteSize int64
 	var msg *ProducerMessage
-	var buf queue.Queue[*ProducerMessage]
-	// Each message's size, remembered on insert. Measuring it on removal
-	// instead would read a message we have already sent to p.input, which the
-	// dispatcher goroutine owns and modifies by then (data race).
-	var sizes queue.Queue[int64]
+	var buf queue.Queue[retryBufEntry]
 
 	for {
 		if buf.Length() == 0 {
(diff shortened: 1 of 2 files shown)


async_producer.go
```diff
@@ -1602,6 +1602,10 @@ func (p *asyncProducer) retryHandler() {
 	var currentByteSize int64
 	var msg *ProducerMessage
 	var buf queue.Queue[*ProducerMessage]
+	// Each message's size, remembered on insert. Measuring it on removal
+	// instead would read a message we have already sent to p.input, which the
+	// dispatcher goroutine owns and modifies by then (data race).
+	var sizes queue.Queue[int64]
```
puellanivis: 🤔 I think there’s a separate race condition in the two queues being enqueued and dequeued separately?

dloebl: Good point. There currently is no other race detected by `-race` in the `TestRetryHandlerHeadersRace` test run, but the two queues are still a bit confusing and could get out of sync by accident in a future edit:
```
$ go test -race -run TestRetryHandlerHeadersRace

PASS
ok  	github.com/IBM/sarama	1.082s
```

Let's switch to a single queue with a `struct`:
```Go
type retryBufEntry struct {
	msg  *ProducerMessage
	size int64
}
```
Added with 32f177020b90ea38a490214a537490fae93fff2e

puellanivis (approved):
Code seems fine. As usual, fit for greater design is left for the team.

dnwe (approved):
Thanks, LGTM - added a small tweak to the race_test

PR #3737, opened 2026-09-02
fix(test): capture logs without replacing the global Logger

### LINKED ISSUE #3736: test: race detector reports a race on global Logger
### Description

The race detector reports a race on the package-level `Logger` between two
existing unit tests. This reproduces on the current `main` without any code
changes.

A similar race involving tests replacing the global `Logger` was previously
reported in [#2011](https://github.com/IBM/sarama/issues/2011) and addressed for
a different test in [#2020](https://github.com/IBM/sarama/pull/2020).

### Versions

| Component | Version |
| --- | --- |
| Sarama | 51a916ad6aea00eab802d6c1b10260d9002e67c7 |
| Go | `go1.27.0 darwin/arm64` |
| Kafka | Not applicable; the tests use mock brokers |

### Reproduction

```console
go test -race -run '^(TestConsumerExpiryTicker|TestPartitionConsumerComputeBackoff)$' .
```

This exact command reproduced the race in 10 out of 10 independent runs.

### Race report

```text
WARNING: DATA RACE
Write at ... by goroutine 36:
  github.com/IBM/sarama.TestPartitionConsumerComputeBackoff.func4()
      consumer_test.go:1864

Previous read at ... by goroutine 31:
  github.com/IBM/sarama.(*debugLogger).Printf()
      sarama.go:139
  github.com/IBM/sarama.(*Broker).closeLocked()
      broker.go:417
  github.com/IBM/sarama.(*Broker).Close()
      broker.go:379
  github.com/IBM/sarama.safeAsyncClose.func1()
      utils.go:56

Goroutine 31 (finished) created at:
  github.com/IBM/sarama.safeAsyncClose()
      utils.go:54
  github.com/IBM/sarama.(*client).Close()
      client.go:329
  github.com/IBM/sarama.(*consumer).Close()
      consumer.go:153
  github.com/IBM/sarama.TestConsumerExpiryTicker()
      consumer_test.go:1645

--- FAIL: TestPartitionConsumerComputeBackoff (0.00s)
    --- FAIL: TestPartitionConsumerComputeBackoff/emits_stuck_warning_at_threshold_and_each_multiple (0.00s)
        testing.go:1865: race detected during execution of test
FAIL
FAIL github.com/IBM/sarama 0.184s
```

official-burak:
Sorry this showed up as a `go test -race` failure after a close. `TestPartitionConsumerComputeBackoff` was replacing the package `Logger` while `TestConsumerExpiryTicker` still had a `safeAsyncClose` goroutine reading it through `DebugLogger`.

The stuck-warning subtest now uses `(*log.Logger).SetOutput`, which is documented as safe for concurrent use, so leftover close logs no longer race with the capture. `TestClientBackgroundMetadataUpdater` had the same assignment, so I pointed that at the same helper.

Fixes #3736.

### DIFF
--- a/sarama_test.go
+++ b/sarama_test.go
@@ -4,6 +4,7 @@ package sarama
 
 import (
 	"flag"
+	"io"
 	"log"
 	"os"
 	"testing"
@@ -15,4 +16,21 @@ func TestMain(m *testing.M) {
 		Logger = log.New(os.Stderr, "[DEBUG] ", log.Lmicroseconds|log.Ltime)
 	}
 	os.Exit(m.Run())
+}
+
+// redirectLogger sends Logger output to w for the rest of the test.
+// It prefers (*log.Logger).SetOutput so leftover goroutines from other tests
+// (for example safeAsyncClose) do not race on a write to the Logger variable.
+func redirectLogger(t *testing.T, w io.Writer) {
+	t.Helper()
+	std, ok := Logger.(*log.Logger)
+	if !ok {
+		orig := Logger
+		Logger = log.New(w, "", 0)
+		t.Cleanup(func() { Logger = orig })
+		return
+	}
+	prev := std.Writer()
+	std.SetOutput(w)
+	t.Cleanup(func() { std.SetOutput(prev) })
 }
--- a/client_test.go
+++ b/client_test.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
 	"net"
 	"runtime"
 	"sync"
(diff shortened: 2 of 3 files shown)


sarama_test.go
```diff
@@ -16,3 +17,20 @@ func TestMain(m *testing.M) {
...
+	std, ok := Logger.(*log.Logger)
+	if !ok {
+		orig := Logger
+		Logger = log.New(w, "", 0)
+		t.Cleanup(func() { Logger = orig })
+		return
+	}
+	prev := std.Writer()
+	std.SetOutput(w)
+	t.Cleanup(func() { std.SetOutput(prev) })
```
puellanivis, line 35: 🤔 Maybe a type switch? Then we don’t need the early return, as each type would be a separate `case`?

puellanivis (commented):
:not-bad: A pretty simple and measured solution.

dnwe (approved):

PR #3740, opened 2026-09-06
fix(consumer): stop roundrobin balancer looping on an unsubscribed topic

lenamonj:
`roundRobinBalancer.Plan` never returns when the topics map holds a topic no member subscribes to: the subscriber search walks `members[i%n]` forever. The range and sticky strategies leave such a topic unassigned.

The search is now bounded to one pass over the members and a topic with no subscriber is skipped. Whenever a subscriber exists the same member is chosen and the cursor advances by the same amount, so existing plans are unchanged.

Sarama's own consumer group builds the topics map from the members' subscriptions, so this reaches direct callers of `Plan`.

A row added to `TestBalanceStrategyRoundRobin` covers it; on main that row hangs until the test timeout.


balance_strategy.go
```diff
@@ -450,12 +450,15 @@ func (b *roundRobinBalancer) Plan(memberAndMetadata map[string]ConsumerGroupMemb
...
-		for !m.hasTopic(tp.topic) {
-			i++
-			m = members[i%n]
+		probe := 0
+		for probe < n && !members[(i+probe)%n].hasTopic(tp.topic) {
+			probe++
+		}
+		if probe == n {
+			continue // no member subscribes to this topic
 		}
```
puellanivis: 🤔 We could probably do this by reworking the existing control flow instead:
```go
m := member[i%n]
for range len(members) {
	if m.hasTopic(tp.topic) {
		break
	}
	i++
	m = member[i%n]
}
if !m.hasTopic(tp.topic) {
	continue // no member subscribes to this topic
}
```

lenamonj:
Reworked to your form in the second commit; the same member is chosen and the cursor advances by the same amount, so the added test row and the existing rows still pass.

puellanivis (approved):
Code is good. 👍

dnwe (approved):
Thanks! Great contribution – lgtm

PR #3752, opened 2026-09-19
fix(client): only deregister the broker registered under that ID

### LINKED ISSUE #3145: Function deregisterBroker Maybe Delete Wrong Broker Without Close Broker
##### Description

Because the function parameter is a point. So if the broker has same ID in the map but the memory point was different. that meaning we delete a different broker point. and didn't close it. after that, the deleted point will never be release. it be a ghost.

```
func (client *client) deregisterBroker(broker *Broker) {
	client.lock.Lock()
	defer client.lock.Unlock()

	_, ok := client.brokers[broker.ID()]
	if ok {
		Logger.Printf("client/brokers deregistered broker #%d at %s", broker.ID(), broker.Addr())
		delete(client.brokers, broker.ID())
		return
	}
	if len(client.seedBrokers) > 0 && broker == client.seedBrokers[0] {
		client.deadSeeds = append(client.deadSeeds, broker)
		client.seedBrokers = client.seedBrokers[1:]
	}
}
```

| Sarama | Kafka | Go  |
| ------ | ----- | --- |
|    release    |    NA   | NA     |

hsdfat:
`deregisterBroker` looks the broker up by ID only, so it can remove a different broker from the one it was given.

Every caller gets a broker from `LeastLoadedBroker()`, drops the client lock, sends a request, and on failure closes that broker and calls `deregisterBroker(broker)`. In the meantime a metadata refresh or coordinator lookup can register a new `*Broker` under the same ID, for example:

- `updateBroker` or `registerBroker` replaces it because its address changed. The old instance is closed.
- `checkBrokersHealth` drops it after a socket error, and `updateBroker` registers the instance from the response in its place.

The caller's request fails because the old instance is closed or its socket is gone. Its `deregisterBroker(old)` then finds the ID in the map and deletes the new broker. That broker is never closed (`Close` and `RefreshBrokers` only close what is in the map), and the client has lost a live broker until the next refresh registers another one. This matches the leaks under unstable networks that the issue describes.

As suggested in the issue, `deregisterBroker` now removes the entry only when `client.brokers[broker.ID()] == broker`.

On whether a stale broker should fall through to the seed broker handling: it does, and nothing happens there. Seed brokers are only created by `NewBroker` in `randomizeSeedBrokers` (ID -1), and the brokers map only holds brokers decoded from Metadata and FindCoordinator responses. The two sets never share an instance, and the seed branch already compares `broker == client.seedBrokers[0]` by pointer, so a stale registered broker never matches it. Falling through leaves the rest of the function as it was.

`TestDeregisterBroker` replaces a broker through `updateBroker`, deregisters the old instance, and checks that the new one is still registered and the seed list is unchanged. It then checks that deregistering the registered instance still removes it. On main the first assertion fails because the map entry is gone.

Fixes #3145


dnwe (approved):
LGTM

